PR #2, opened 2017-02-26
Fixes #1 - IO operations on disk and file system

### LINKED ISSUE #1: Add disk read speed check as part of preupgrade checks
Disk read speed should be at least **80 MB/sec** for mongodb, pgsql, pulp directories present under /var/lib.

swapab:
- [x] Speed check for HDD type of disk
- [x] Speed check for Filesystem type of disk
- [x] Unit tests for above

### DIFF
--- a/definitions/checks/disk_speed_minimal.rb
+++ b/definitions/checks/disk_speed_minimal.rb
@@ -1,0 +1,41 @@
+class Checks::DiskSpeedMinimal < ForemanMaintain::Check
+  EXPECTED_IO = 80
+  DEFAULT_UNIT   = 'MB/sec'.freeze
+  DEFAULT_DIRS   = ['/var/lib/pulp', '/var/lib/mongodb', '/var/lib/pgsql'].freeze
+
+  label :disk_io
+  description 'Check for recommended disk speed of pulp, mongodb, pgsql dir.'
+  tags :basic
+
+  confine do
+    execute?('which hdparm') && execute?('which fio')
+  end
+
+  def run
+    success = true
+    io_obj = ForemanMaintain::Utils::Disk::NilDevice.new
+
+    dirs_to_check.each do |dir|
+      io_obj = ForemanMaintain::Utils::Disk::Device.new(dir)
+
+      next if io_obj.read_speed >= EXPECTED_IO
+
+      success = false
+      logger.info "\n Slow disk detected for #{dir} - #{io_obj.read_speed} #{io_obj.unit}."
+      break
+    end
+
+    assert(success, io_obj.slow_disk_error_msg)
+  end
+
+  def check_only_single_device?
+    DEFAULT_DIRS.map do |dir|
+      ForemanMaintain::Utils::Disk::Device.new(dir).name
+    end.uniq.length <= 1
+  end
+
+  def dirs_to_check
+    return DEFAULT_DIRS.first(1) if check_only_single_device?
+    DEFAULT_DIRS
+  end
+end
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -1,3 +1,9 @@
+if RUBY_VERSION <= '1.8.7'
+  require 'rubygems'
+end
+
+require 'json'
+
 module ForemanMaintain
   require 'foreman_maintain/concerns/logger'
   require 'foreman_maintain/concerns/metadata'
@@ -13,6 +19,7 @@ module ForemanMaintain
   require 'foreman_maintain/scenario'
   require 'foreman_maintain/runner'
   require 'foreman_maintain/reporter'
+  require 'foreman_maintain/utils'
 
   class << self
     attr_accessor :config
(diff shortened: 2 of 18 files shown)


lib/foreman_maintain/concerns/disk_utility.rb
```diff
@@ -0,0 +1,41 @@
...
+    module DiskUtility
+      include Logger
+      MIN_SPEED_MBPS = 80
+      DEFAULT_DIRS = ['/var/lib/pulp', '/var/lib/mongodb', '/var/lib/pgsql'].freeze
+
+      def self.included(klass)
+        klass.extend(self)
+      end
+
+      def pass?
```
iNecas: A concern should not define this kind of generic method, that could be easily mistaken/conflicted with some other concern. Perhaps, instead of having this as a concern, the logic that this module defines could go directly to the check.

The the `Concerns::DiskUtility` module, that groups the relevnat classes could be renamed to `Utils::Disk` (while introducing the `Utils` module: I see a benefit of having such a module here and we will utilize that later for sure for other things as well

definitions/features/disk_io.rb
```diff
@@ -0,0 +1,3 @@
+class Features::DiskIO < ForemanMaintain::Feature
```
iNecas: For this particular case, we perhaps don't need a feature at all: not all checks need a feature defined. The check itself can use the disk utility classes to do the job.

definitions/checks/disk_speed_minimal.rb
```diff
@@ -0,0 +1,9 @@
+class Checks::DiskSpeedMinimal < ForemanMaintain::Check
+  for_feature :disk_io
+  description 'Check for recommended disk speed of pulp, mongodb, pgsql dir.'
```
iNecas: On newer versions (checks with sat62), the hdparm command can not be present (needs to be pulled via hdparm rpm). I suggest putting 

```ruby
confine do
  execute?("which hdparm")
end
```

to make it valid only when the tool is in place (and we will ensure the hdparm is there via rpm depednencies.

lib/foreman_maintain.rb
```diff
@@ -1,8 +1,10 @@
+require 'json'
```
iNecas: require 'rubygems` is needed to be able to load `json` with ruby 1.8.7

definitions/checks/disk_speed_minimal.rb
```diff
@@ -0,0 +1,9 @@
...
+  tags :pre_upgrade
+
+  def run
+    assert(feature(:disk_io).pass?, "Disk speed should be atleast #{MIN_SPEED_MBPS} MB/sec")
```
iNecas: The info message sould be more specfic about what subdirs don't match the requirement and what's the current value for the particular disk

iNecas:
Please make sure to run `rake rubocop` (`rake` runs it by defualt when running on ruby > 2.0, as newer rubocop is not compatible on older versions. The travis should start picking up the PRs starting the next commit.

definitions/features/disk_io.rb
swapab: Sure @iNecas, not mandating feature does makes sense.

lib/foreman_maintain/concerns/disk_utility.rb
swapab: > Perhaps, instead of having this as a concern, the logic that this module defines could go directly to the check.

The reason to extract it in a concern(or util) was to be able to unit test it.

If we include everything in check `Checks::DiskSpeedMinimal` then it needs to be functional tested(preferably). Unit testing `Checks::DiskSpeedMinimal` would be feasible? Like initializing checks and testing the instance methods etc.

If it needs to be abstracted out, what would be the best way to do?

swapab: > The the `Concerns::DiskUtility` module, that groups the relevant classes could be renamed to `Utils::Disk`

Yes. This sounds good.

definitions/checks/disk_speed_minimal.rb
swapab: Done.

definitions/checks/disk_speed_minimal.rb
swapab: check confined :+1:

lib/foreman_maintain/concerns/disk_utility.rb
iNecas: This looks good. For unit testing, you should be able to test the check similar way as any other object. Something like `check = Checks::DiskSpeedMinimal.new; assert check.pass`

swapab:
@iNecas I feel this task complete from my side(based on check-list https://github.com/iNecas/foreman_maintain/pull/2#issue-210318603 )
Waiting for your re-review.

lib/foreman_maintain.rb
swapab: required 'rubygems' before json

iNecas:
Look good, thanks @swapab

PR #6, opened 2017-03-01
Fixes #5 - enhanced logger feature

### LINKED ISSUE #5: Enhance logging
Currenttly, the logging is very simple, but we can do better. Things to improve:

1. log into file in debug mode
2. include timestamps in log file names
3. keep history of X log files, delete the older log files

kgaikwad:
1. Deleted existing logger.rb files
2. Now, logs are stored into file under /var/log/foreman-maintain directory in debug mode
3. Number of log files set to 10 i.e
when, there would be 11 files remaining, the oldest one will be deleted
Note - not added timestamp to filename.


kgaikwad:
@iNecas , I have resolved the code conflicts.

lib/foreman_maintain/concerns/logger.rb
```diff
@@ -1,13 +0,0 @@
-require 'logger'
-
-module ForemanMaintain
-  module Concerns
-    module Logger
-      def logger
-        @logger ||= ::Logger.new($stderr).tap do |logger|
```
iNecas: could we still keep this module, just linking to the  `ForemanMaintain.logger` instance? Just for convenience of the code writers: less things to remember :)

iNecas:
Nice trick with the logger arguments: I didn't know about them :), intially I was thinking about having a log file per run, but I guess having one file for multiple runs work ok, especially if we starat logging beginning and end of the command invocations.

lib/foreman_maintain/concerns/logger.rb
kgaikwad: @iNecas ,

Yes, we can keep it.
I have removed it only because every time we need to include this concerns/logger.rb module.

Other way is instead of keeping lib/foreman_maintain/concerns/logger.rb file, we could rename 
lib/foreman_maintain/logger.rb file to lib/foreman_maintain/fm_logger.rb and use this as you mentioned.
So that it will not create conflicts with original Logger class.

What do you think? 
Please let me know your views then I will go with further changes.

iNecas: You should be able to resolve the conflicts just by using `::Logger` when you are referring to the Ruby logger. I prefer `foreman_maintain/concerns/logger.rb` for defining the module to be included where needed, as this mimics the way people are used to organize this kind of modules in rails.

kgaikwad: @iNecas ,
Thank you for your views. I will then go ahead with foreman_maintain/concerns/logger.rb and update the pull-request soon.

kgaikwad: I have updated the pull-request. Please let me know if any changes required.

lib/foreman_maintain.rb
```diff
@@ -57,5 +58,15 @@ def available_checks(*args)
     def available_procedures(*args)
       detector.available_procedures(*args)
     end
+
+    def init_logger
+      # Note - If timestamp added to filename then number of log files i.e second
+      # argument to Logger.new will not work as expected
+      filename = File.expand_path("#{config.log_dir}/fm.log")
```
iNecas: small note: `fm.log` is not very descriptive. Havnig `foreman-maintain.log` would be more suitable

kgaikwad: Yes sure. Updated the pull-request.

lib/foreman_maintain/config.rb
```diff
@@ -1,17 +1,31 @@
...
 module ForemanMaintain
   class Config
-    attr_accessor :definitions_dirs, :log_level
+    attr_accessor :definitions_dirs, :log_level, :log_dir
 
     def initialize(options = {})
       @definitions_dirs = options.fetch(:definitions_dirs,
                                         [File.join(source_path, 'definitions')])
+
       @log_level = options.fetch(:log_level, ::Logger::ERROR)
```
iNecas: Since we are logging in to a file now, I think it's benefitial to have debug mode turned on by default: we can change thsi later, but for this stage, having as much info as possible from the run would be benefitial, unless you think it could bite us somewhere?

kgaikwad: @iNecas , I have modified the default logger level to debug.

iNecas:
Thanks @kgaikwad, merging now

PR #8, opened 2017-03-02
Fixes #7 - Added bin directory to auto include list

### LINKED ISSUE #7: bin directory is missing from installed gem
bin directory is missing from the list of directories to be built into the gem.

gnurag:
* Added bin dir to auto include list


iNecas:
Thanks @gnurag

PR #9, opened 2017-03-02
Added foreman-maintain to executables section in gemspec

gnurag:
* Removed config directory from gem include dirs, since it doesn't exist.
* Added foreman-maintain executable in executables section.


iNecas:
Thanks @gnurag

PR #13, opened 2017-03-05
Fixes #3 - Add option to skip the prompted step

### LINKED ISSUE #1: Add disk read speed check as part of preupgrade checks

### LINKED ISSUE #3: Option to 'skip' or 'continue' the current check

swapab:
* When opted to skip(choice: 'N', 'n' or 'no')
  * Check if any next_steps exists, prepend to `steps_to_run` array
  * Run the next check
* Add 'next' or 'no' option for multiple steps

_Looks like:_
```ruby
sat62 :: ~/foreman_maintain ‹3-option-to-skip-step*› » ./bin/foreman-maintain upgrade check 6.2.z                                                                                                              
Running checks before upgrading to Satellite 6.2.z
Check for recommended disk speed of pulp, mongodb, pgsql dir.:        [OK]
check for paused tasks:                                               [FAIL]
There are currently 5 paused tasks in the system
Continue with step [resume paused tasks]?, [y(yes), n(no), q(quit)]n
check for running tasks:                                              [OK]
```

### DIFF
--- a/lib/foreman_maintain/error.rb
+++ b/lib/foreman_maintain/error.rb
@@ -1,0 +1,6 @@
+module ForemanMaintain
+  module Error
+    class Fail < StandardError
+    end
+  end
+end
--- a/lib/foreman_maintain/reporter/cli_reporter.rb
+++ b/lib/foreman_maintain/reporter/cli_reporter.rb
@@ -4,6 +4,12 @@ require 'highline'
 module ForemanMaintain
   class Reporter
     class CLIReporter < Reporter
+      DECISION_MAPPER = {
+        %w[y yes] => 'add_step',
+        %w[n next no] => 'skip_to_next',
+        %w[q quit] => 'ask_to_quit'
+      }.freeze
+
       # Simple spinner able to keep updating current line
       class Spinner
         def initialize(reporter, interval = 0.1)
@@ -134,14 +140,12 @@ module ForemanMaintain
       def after_scenario_finishes(_scenario); end
 
       def on_next_steps(runner, steps)
-        choice = if steps.size > 1
-                   multiple_steps_selection(steps)
-                 elsif ask_to_confirm("Continue with step [#{steps.first.description}]?")
-                   steps.first
-                 else
-                   :quit
-                 end
-        choice == :quit ? runner.ask_to_quit : runner.add_step(choice)
+        if steps.size > 1
+          runner.ask_to_quit if multiple_steps_selection(steps) == :quit
+        else
+          decision = ask_decision("Continue with step [#{steps.first.description}]?")
+          runner.send(decision, steps.first)
+        end
       end
 
       def multiple_steps_selection(steps)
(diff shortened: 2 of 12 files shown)


iNecas:
I'm thinking if we should keep the skipped step there when the answer is `s`… especially how it would behave when there are are more than one options to go with? What about just skipping the offered step and continuing?

iNecas:
While looking more into the tasks checks, I start thinking about this approach to defining the check
this way instead of going though `next_steps`:

```ruby
def run
  paused_tasks_count = feature(:foreman_tasks).paused_tasks_count
  unless check(paused_tasks_count == 0,
                "There are currently #{paused_tasks_count} paused tasks in the system")
    offer(procedure(Procedures::ForemanTasksResume),
          procedure(Procedures::ForemanTasksUiInvestigate)) do |procedure|
       procedure.run(:state => 'paused')
      run # run the check again after the procedure
    end
  end
end
```

This way, we would be able to define the flow of the fix of failed checks, run the procedures
in parametrized way and rerun the check if needed. One shift would be the fix procedures would
be part run as part of the check step instead of adding separate step to the scenario. On the other hand, what we care most is the result of the check (especially when we re-run it after one selects a procedure). If you're ok with this approach, I would try to implement it tomorrow.

swapab:
> I'm thinking if we should keep the skipped step there when the answer is s…

Yes, we can go with 's' for 'skip'. But then we should get rid of 'N' as it won't have any action.

> especially how it would behave when there are are more than one options to go with? What about just skipping the offered step and continuing?

I had thought about this and had coded it like you said(just skipping and continuing). Showing the status [SKIPPED] would help to understand the flow and history of execution visually. Basic thought was to make it look good.
For multiple options we can concat the options and show like:

```ruby
| check for paused tasks:                                           [FAIL]     |
| There are currently paused tasks in the system                               |
| There are multiple steps to proceed:
| 1) start the present service
| 2) restart present service
Select step to continue, (s) skip (q) quit:
| 1) start the present service
| 2) restart present service                                        [SKIPPED]  |
| check for running tasks:                                          [OK]       |
```

But this looks cluttered... maybe we'll just skip and continue for multiple options?

swapab:
> While looking more into the tasks checks, I start thinking about this approach to defining the check
this way instead of going though next_steps:

Yes! recursive approach looks good and more functional. But doesn't provide a generic approach the way `next_steps` provides. Also, I had thought of using `next_steps` as a way to chain procedures.
We can use both decisively.

iNecas:
Sure, we can keep both

swapab:
@iNecas  All checks passed. Open for review.

lib/foreman_maintain/runner.rb
```diff
@@ -38,5 +43,23 @@ def ask_about_offered_steps(steps)
         @reporter.on_next_steps(self, steps)
       end
     end
+
+    def mock_skip_class(step)
+      Class.new(step.class.superclass) do
```
iNecas: We migth actually not need to show this, at least not for the case, where we offer fix procedure and the user chooses not to run it, this would cause complciations for multi-option case

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -152,26 +156,28 @@ def multiple_steps_selection(steps)
...
+      end
+
       def ask_to_select(message, steps)
-        print "#{message}, (q) for quit"
+        print "#{message}, [n(next), q(quit)]"
         answer = @stdin.gets.chomp
-        case answer
-        when 'q'
+        if %w(n no next).include?(answer)
           :quit
```
iNecas: shouln't here be something else: as I read it, it would just return the symbol, but there is no place that would consume it

lib/foreman_maintain/runner.rb
swapab: Okay, so you are suggesting we do not show `[SKIPPED]` status for the single(binary) procedure?

In case of multi-options, we are not showing `[SKIPPED]` status. So to keep both of them consistent, we can just skip the check without showing anything on the next line and moving on to the next check. Makes sense?

iNecas: ACK

swapab:
Above changes done.
Also fixed new offense introduced in rubocop v0.48.1
```ruby
%w-literals should be delimited by [ and ].
```

lib/foreman_maintain/reporter/cli_reporter.rb
swapab: This has been handled completely now.

lib/foreman_maintain/runner.rb
swapab: Actually, this makes me ask what would be use case of using 'skipped' status?
https://github.com/iNecas/foreman_maintain/blob/master/lib/foreman_maintain/reporter/cli_reporter.rb#L204

iNecas: Probably no use case for now, we can remove it add add it later when the use case is there

swapab: Sure. Removed.

lib/foreman_maintain/error.rb
```diff
@@ -0,0 +1,9 @@
+module ForemanMaintain
+  module Error
+    class Fail < StandardError
+    end
+
+    class Skipped < StandardError
```
iNecas: Since there is not use case for it now, let's remove it, including the related places that deal with it

swapab: Done!

iNecas:
Thanks @swapab !

PR #22, opened 2017-03-10
Add check for tags

Akasurde:
Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>


iNecas:
Just out of curiosity, on which versions it was failing? I haven't seen issue with this on my versions?

Akasurde:
@iNecas I am using github source for reproducing this issue.

iNecas:
I've reproduced this with running 'health check --tags` command, but I was hitting other issues after that. I've opened a PR to address my issues with that, that i believe should cover also your case. Could you have a look at https://github.com/iNecas/foreman_maintain/pull/22 and confirm that addresses issues you are seeing?

Akasurde:
@iNecas OK. I will close this PR as #27 handles this scenario better.

PR #29, opened 2017-03-16
Code usability improvements

iNecas:
While trying to implement the tasks checks, I've hit several issues that were
complicating my life as an author of the check. This is a set of changes I've made
to make the life of others better as well.

**Extract metadata DSL to separate namespace**
The reason for this was to make the work with classes metadata a bit easier,
we could not use MyCheck.label before, becuse the label was already a method
of the DSL. It was needed to be able to do the detections on classes, not instances of the classes (see below)

**Run confine against classes, not objects**

This allows us to have multiple objects of the same procedure in the
system, differing in the constructor parameters. Useful for parametrized
procedures.

**Set next steps with the assert check**

This allows adding more assertions to the check, each adding their own
next steps.

**Make the spinner opt-in, not opt-out**

The default spinner caused issues when one needs to interact with the
user (printing additional messages). As it's needed only for long
running actions, it should not be enforced by default.

### DIFF
--- a/lib/foreman_maintain/check.rb
+++ b/lib/foreman_maintain/check.rb
@@ -10,10 +10,6 @@ module ForemanMaintain
     class Fail < StandardError
     end
 
-    def initialize(associated_feature)
-      @associated_feature = associated_feature
-    end
-
     def assert(condition, error_message)
       raise Fail, error_message unless condition
     end
--- a/lib/foreman_maintain/check.rb
+++ b/lib/foreman_maintain/check.rb
@@ -10,8 +10,23 @@ module ForemanMaintain
     class Fail < StandardError
     end
 
-    def assert(condition, error_message)
-      raise Fail, error_message unless condition
+    # run condition and mark the check as failed when not passing
+    #
+    # ==== Options
+    #
+    # * +:next_steps* - one or more procedures that can be followed to address
+    #                   the failure, will be offered to the user when running
+    #                   in interactive mode
+    def assert(condition, error_message, options = {})
+      unexpected_options = options.keys - [:next_steps]
+      unless unexpected_options.empty?
+        raise ArgumentError, "Unexpected options #{unexpected_options.inspect}"
+      end
+      unless condition
+        next_steps = Array(options.fetch(:next_steps, []))
+        self.next_steps.concat(next_steps)
+        raise Fail, error_message
+      end
     end
 
     # public method to be overriden
--- a/lib/foreman_maintain/concerns/metadata.rb
+++ b/lib/foreman_maintain/concerns/metadata.rb
@@ -42,13 +42,15 @@ module ForemanMaintain
         end
 
         def self.eval_dsl(metadata, &block)
-          self.new(metadata).tap do |dsl|
+          new(metadata).tap do |dsl|
             dsl.instance_eval(&block)
           end.data
         end
       end
 
       module ClassMethods
+        include Finders
+
         def inherited(klass)
           sub_classes << klass
         end
@@ -85,6 +87,18 @@ module ForemanMaintain
           @metadata
         end
 
+        def label
+          metadata[:label] || generate_label
+        end
+
+        def description
+          metadata[:description] || to_s
+        end
+
+        def tags
+          metadata[:tags]
+        end
+
         def initialize_metadata
           { :tags => [],
             :confine_blocks => [] }.tap do |metadata|
@@ -92,6 +106,26 @@ module ForemanMaintain
               metadata[:label] = superclass.metadata[:label]
             end
           end
+        end
+
+        def present?
+          evaluate_confines
+        end
+
+        private
+
+        def evaluate_confines
+          raise 'Recursive confine block call detected' if @confines_evaluation_in_progress
+          @confines_evaluation_in_progress = true
+          metadata[:confine_blocks].all? do |block|
+            instance_exec(&block)
+          end
+        ensure
+          @confines_evaluation_in_progress = false
+        end
+
+        def generate_label
+          name.split('::').last.split(/(?=[A-Z])/).map(&:downcase).join('_').to_sym
         end
       end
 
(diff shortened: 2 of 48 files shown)


iNecas:
@swapab @kgaikwad  here are the chaganges regarding usability of writing the checks: any comments welcome.

lib/foreman_maintain/concerns/metadata.rb
```diff
@@ -93,6 +107,26 @@ def initialize_metadata
...
+          evaluate_confines
+        end
+
+        private
+
+        def evaluate_confines
+          raise 'Recursive confine block call detected' if @confines_evaluation_in_progress
+          @confines_evaluation_in_progress = true
+          metadata[:confine_blocks].all? do |block|
+            instance_exec(&block)
```
swapab: Why use `instance_exec` and not `instance_eval`?

lib/foreman_maintain/concerns/metadata.rb
```diff
@@ -104,35 +138,15 @@ def metadata
       end
 
       def label
-        metadata[:label] || generate_label
+        self.class.label
```
swapab: This is good :+1: 
`class.label` would have produced the same results right? Just clarifying myself. No need to change, I prefer `self.class` chaining for readability.

lib/foreman_maintain/executable.rb
```diff
@@ -40,5 +43,17 @@ def __run__(execution)
...
     end
+
+    # method defined both on object and class to ensure we work always with object
+    # even when the definitions provide us only class
+    def ensure_instance
+      self
+    end
+
+    class << self
+      def ensure_instance
```
swapab: Can you explain this with a scenario/example?

lib/foreman_maintain/check.rb
```diff
@@ -10,8 +10,23 @@ class Check < Executable
...
+    #                   the failure, will be offered to the user when running
+    #                   in interactive mode
+    def assert(condition, error_message, options = {})
+      unexpected_options = options.keys - [:next_steps]
+      unless unexpected_options.empty?
+        raise ArgumentError, "Unexpected options #{unexpected_options.inspect}"
+      end
+      unless condition
+        next_steps = Array(options.fetch(:next_steps, []))
+        self.next_steps.concat(next_steps)
```
swapab: Can we use [`add_steps`](https://github.com/iNecas/foreman_maintain/pull/29/commits/0895f48516306b3ea225a9141b3a01e179f065c1#diff-b47081cbc34a956f34fd187f253c1bc4R50) here?

test/lib/reporter_test.rb
```diff
@@ -40,10 +40,10 @@ module ForemanMaintain
       assert_equal <<STR, captured_out
 Running present_service upgrade scenario
 --------------------------------------------------------------------------------
-| present service run check:                                        [OK]       |
```
swapab: So we have got rid of the bars. :+1:

lib/foreman_maintain/concerns/metadata.rb
iNecas: Afaik `class` is one of the things you can't use with implicit receiver, as it's also a keyword, so the interpreter considers it beginning of a class definition

```
class.to_s
SyntaxError: (irb):3: syntax error, unexpected '.'
```

lib/foreman_maintain/executable.rb
iNecas: The this allows to use both class and instance of class in `next_steps` definition.

With this, you can do both:

```
:next_steps => [Producedures::ForemanTasksResume]
# as well as
:next_steps => [Producedures::ForemanTasksResume.new]
```

What do you think? I guess we could neforce only using instances in this cases, to not make to much confusion over how it actually works

test/lib/reporter_test.rb
iNecas: yes :)

lib/foreman_maintain/concerns/metadata.rb
swapab: Oh yes, I remember this.

lib/foreman_maintain/check.rb
iNecas: No, as we don't have reference to the scenario from check here: I would like to keep the logic how the scenario/running handles the next steps outside of the check itself. Here, it's just collecting the steps that can be done, and leaving the decision on what will actually be done for later/other parts of code.

iNecas:
@swapab anything I should address before merging?

lib/foreman_maintain/executable.rb
iNecas: While I was looking at removing this, I started hitting more issues, as we return classes some places (such as `find_checks`), and we need to turn them into instances at some point. I'm for keeping the current model for now, and instanciating the classes if needed (+ use the `manual_detection` metadata for classes that can't be initialized automatically)

iNecas:
After talking to @kgaikwad today, I'm merging the PR as is and leave the code for improvement in futher work, as it blocks some other activity

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -73,19 +84,51 @@ def before_scenario_starts(scenario)
...
+      end
+
+      def print(string)
+        new_line_if_needed
+        @stdout.print(string)
+        @stdout.flush
+        record_last_line(string)
       end
 
-      def on_execution_update(execution, update)
```
swapab: @iNecas I think I missed this in review.
But [`Execution#update`](https://github.com/iNecas/foreman_maintain/blob/master/lib/foreman_maintain/runner/execution.rb#L55) is still calling `on_execution_update`. Origin seems to be `say` method. It looks like if you are **not** opting for the spinner and calling say then it raises an exception
```undefined method `on_execution_update' for #<ForemanMaintain::Reporter::CLIReporter:0x00000001eb5900>```

iNecas: Good catch: after recent updates of the spinner interface and adding `puts` and `print` methods that go though the reporter. I'm leaning towards just removing the `say` keyword in favor of the others alternatives. What do you think?

swapab: Sure. No problem at all. Unless and until the purpose is served. We can remove `say` keyword.

swapab: I was also going to ask this, what's different between `puts` and `print`? (In `foreman_maintain` of course, and not ruby :smile:)

iNecas: :D The difference is the very same as in ruby. The reason we override the methods here is to be able to capture that in the reporter and decide how to work with it. This way, we can still put
the status label at the same line, even when people use puts. The non-cli reporter would save the stings to send it somewhere else if needed.

PR #30, opened 2017-03-17
Fixes #19 - Handled sync plans before & after upgrade

### LINKED ISSUE #19: Offer to disable sync plans before upgrade

kgaikwad:
With this commit, below changes included -
1. Two new checks
   - Disable sync plans before upgrade
   - Enable sync plans after upgrade which are disabled previously.

2. Created directory lib/foreman_maintain/storage in which
   using singleton objects we can store data in yaml.

3. Added new config attribute :storage_file to mention file name
   to store data.

### DIFF
--- a/config/foreman_maintain.yml.test
+++ b/config/foreman_maintain.yml.test
@@ -9,3 +9,6 @@
 :definitions_dirs:
   - 'test/lib/support/definitions'
   - 'test/lib/support/additional_definitions'
+
+# Mention file path to store data
+:storage_file: 'test/data_test.yml'
--- a/definitions/features/sync_plans.rb
+++ b/definitions/features/sync_plans.rb
@@ -1,0 +1,74 @@
+class Features::SyncPlans < ForemanMaintain::Feature
+  metadata do
+    label :sync_plans
+  end
+
+  def active_sync_plans_count
+    feature(:foreman_database).query(
+      <<-SQL
+        SELECT count(*) AS count FROM katello_sync_plans WHERE enabled ='t'
+      SQL
+    ).first['count'].to_i
+  end
+
+  def ids_by_status(enabled = true)
+    enabled = enabled ? 't' : 'f'
+    feature(:foreman_database).query(
+      <<-SQL
+        SELECT id FROM katello_sync_plans WHERE enabled ='#{enabled}'
+      SQL
+    ).map { |r| r['id'].to_i }
+  end
+
+  def disabled_plans_count
+    data[:disabled].length
+  end
+
+  def make_disable(ids)
+    update_records(ids, false)
+  end
+
+  def make_enable
+    update_records(data[:disabled], true)
+  end
+
+  private
+
+  def update_records(ids, enabled)
+    updated_record_ids = []
+    ids.each do |sp_id|
+      result = hammer("sync-plan update --id #{sp_id} --enabled #{enabled}")
+      if result.include?('Sync plan updated')
+        updated_record_ids << sp_id
+      else
+        raise result
+      end
+    end
+  ensure
+    new_data = sync_plan_data(enabled, updated_record_ids)
+    save_state(new_data)
+  end
+
+  def data
+    upgrade_storage = ForemanMaintain.storage(:upgrade)
+    @data ||= upgrade_storage.data.fetch(:sync_plans, :enabled => [], :disabled => [])
+    @data
+  end
+
+  def sync_plan_data(enabled, new_ids)
+    sync_plan_hash = data
+    if enabled
+      sync_plan_hash[:disabled] -= new_ids
+      sync_plan_hash[:enabled] = new_ids
+    else
+      sync_plan_hash[:disabled].concat(new_ids)
+    end
+    sync_plan_hash
+  end
+
+  def save_state(sync_plan_hash = {})
+    storage = ForemanMaintain.storage(:upgrade)
+    storage[:sync_plans] = sync_plan_hash
+    storage.save
+  end
+end
--- a/definitions/procedures/sync_plans/enable.rb
+++ b/definitions/procedures/sync_plans/enable.rb
@@ -1,0 +1,18 @@
+module Procedures::SyncPlans
+  class Enable < ForemanMaintain::Procedure
+    metadata do
+      for_feature :sync_plans
+      description 're-enable active sync plans'
+    end
+
+    def run
+      enabled_sync_plans
+    end
+
+    private
+
+    def enabled_sync_plans
+      feature(:sync_plans).make_enable
+    end
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 Gemfile.lock
 log/
 config/foreman_maintain.yml
+data.yml
--- a/config/foreman_maintain.yml.example
+++ b/config/foreman_maintain.yml.example
@@ -7,3 +7,6 @@
 
 # Mention definitions directories. Default
 # :definitions_dirs:
+
+# Mention file path to store data
+# :storage_file: 'data.yml'
--- a/definitions/checks/sync_plans/with_disabled_status.rb
+++ b/definitions/checks/sync_plans/with_disabled_status.rb
@@ -1,0 +1,16 @@
+module Checks::SyncPlans
+  class WithDisabledStatus < ForemanMaintain::Check
+    metadata do
+      for_feature :sync_plans
+      description 'check for disabled sync plans'
+      tags :post_upgrade
+    end
+
+    def run
+      disabled_plans_count = feature(:sync_plans).disabled_plans_count
+      assert(disabled_plans_count == 0,
+             "There are #{disabled_plans_count} disabled sync plans which needs to be enabled",
+             :next_steps => Procedures::SyncPlans::Enable.new)
+    end
+  end
+end
--- a/definitions/checks/sync_plans/with_enabled_status.rb
+++ b/definitions/checks/sync_plans/with_enabled_status.rb
@@ -1,0 +1,16 @@
+module Checks::SyncPlans
+  class WithEnabledStatus < ForemanMaintain::Check
+    metadata do
+      for_feature :sync_plans
+      description 'check for enabled sync plans'
+      tags :pre_upgrade
+    end
+
+    def run
+      active_sync_plans_count = feature(:sync_plans).active_sync_plans_count
+      assert(active_sync_plans_count == 0,
+             "There are total #{active_sync_plans_count} active sync plans in the system",
+             :next_steps => Procedures::SyncPlans::Disable.new)
+    end
+  end
+end
--- a/definitions/procedures/sync_plans/disable.rb
+++ b/definitions/procedures/sync_plans/disable.rb
@@ -1,0 +1,22 @@
+module Procedures::SyncPlans
+  class Disable < ForemanMaintain::Procedure
+    metadata do
+      for_feature :sync_plans
+      description 'disable active sync plans'
+    end
+
+    def run
+      disable_all_enabled_sync_plans
+    end
+
+    private
+
+    def disable_all_enabled_sync_plans
+      with_spinner('disabling sync plans') do |spinner|
+        ids = feature(:sync_plans).ids_by_status(true)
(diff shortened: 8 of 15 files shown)


kgaikwad:
@iNecas , 
I am getting error "Prefer using YAML.safe_load over YAML.load." for latest ruby versions.
Using gem safe_yaml, we can use YAML.load with safe argument.

Please provide me your suggestions.

iNecas:
Since we're not loading yaml from untrusted sources, I don't think the safe_yaml suggestion is valid for us, and suggest turning this rubocop check off: the safe_yaml gem would be unnecessary dependency for us.

kgaikwad:
Thank you @iNecas.
I will make this check off.

swapab:
@kgaikwad 
I think this should do:
```yaml
Security/YAMLLoad:
    Enabled: false
```

kgaikwad:
@swapab , yes added with same key with enabled flag. 
I will update the pull-request shortly as I was struggling with rvm config on sat machine :-)

Please review other changes of pull-request and feel free to add comments.

lib/foreman_maintain/hammer.rb
```diff
@@ -0,0 +1,37 @@
+module ForemanMaintain
+  class Hammer
```
iNecas: What about turning the hammer into a feature? I think it would fit a bit better to the whole concept

lib/foreman_maintain/utils/sync_plans.rb
```diff
@@ -0,0 +1,112 @@
...
+        private
+
+        def update_records(ids, enabled, sync_plans_action = MAKE_SYNC_PLANS_DISABLE)
+          updated_record_ids = []
+          ids.each do |sp_id|
+            updated_record_ids = execute_update_cmd(
+              sp_id, enabled, updated_record_ids, sync_plans_action
+            )
+          end
+          store_to_yaml(updated_record_ids, sync_plans_action)
```
iNecas: I guess this should go an ensure block, so that you don't have to deal with storing in `execute_update_cmd` method

lib/foreman_maintain/utils/sync_plans.rb
```diff
@@ -0,0 +1,112 @@
...
+            updated_record_ids << id
+          else
+            store_to_yaml(updated_record_ids, sync_plans_action)
+            raise result
+          end
+          updated_record_ids
+        end
+
+        def store_to_yaml(ids, action_name)
+          file_path = File.expand_path FILE_TO_STORE_MANIPULATED_DATA
```
iNecas: I expect we will need more data to be stored on per-upgrade basis. What about introducing a separate singleton object for stornig the data. Something like

`ForemanMaintain.storage(:upgrade)`, which could have the mehods around storing and retrieving useful data, such as info about enabled/disabled sync plans.

lib/foreman_maintain/utils/sync_plans.rb
```diff
@@ -0,0 +1,112 @@
+module ForemanMaintain
+  module Utils
+    module SyncPlans
```
iNecas: I suggest extracting the code regarding the yaml serialization into a 'persistence' object, that will handle the manipulation of yaml (loading/saving), and trying to extract the hammer-specific code into hammer feature, leaving there only the code specific to the  sync plans.

lib/foreman_maintain/utils/sync_plans.rb
```diff
@@ -0,0 +1,112 @@
...
+      FILE_TO_STORE_MANIPULATED_DATA = 'data.yml'.freeze
+      MAKE_SYNC_PLANS_DISABLE = 'add'.freeze
+      MAKE_SYNC_PLANS_ENABLE = 'remove'.freeze
+
+      class << self
+        def ids(db_feature, enabled = true)
+          enabled = enabled ? 't' : 'f'
+          db_feature.query(
+            <<-SQL
+              SELECT id FROM katello_sync_plans WHERE enabled ='#{enabled}'
```
iNecas: The `katello` feature should know how to return the list of ids of sync plans.

definitions/procedures/make_sync_plans_enable.rb
```diff
@@ -0,0 +1,12 @@
+class Procedures::MakeSyncPlansEnable < ForemanMaintain::Procedure
+  for_feature :sync_plans
+  description 'enable sync plans which are disabled before upgrade'
+
+  def run
```
iNecas: After `https://github.com/iNecas/foreman_maintain/pull/29` we could have just one precodure (`UpgradeSyncPlansDisable`), which could have a parameter (such as `rollback`) to disable (and enable) the sync plans at the end

lib/foreman_maintain/hammer.rb
iNecas: Oh, I see the feature is there: I think it should also provide the methods that we now define here (unless some methods are too generic that could be used from other places). The idea is the lib will contain only very generic code, that would be version independent.

lib/foreman_maintain/concerns/hammer_helpers.rb
```diff
@@ -0,0 +1,17 @@
...
+    module HammerHelpers
+      def hammer_settings
+        @hammer_settings ||= ForemanMaintain.hammer_settings
+      end
+
+      def self.included(klass)
+        klass.extend(self)
+      end
+
+      def hammer_command
```
iNecas: I would expect this rather being a method taking arguments and performing the commands on it's own, returning the results in some unified format

lib/foreman_maintain/hammer.rb
kgaikwad: Yes, I have added hammer as a separate feature using which we can have 3 methods as below

1. password_available?
2. username_available?
3. main_command which will return hammer command string with -u & -p  options.

Should I move logic of retrieving of username & password from this file into feature itself?

lib/foreman_maintain/utils/sync_plans.rb
kgaikwad: Yeah, sorry I have missed this. I will add these changes and update the pull-request.

definitions/features/sync_plans.rb
```diff
@@ -0,0 +1,11 @@
+class Features::SyncPlans < ForemanMaintain::Feature
```
iNecas: As suggested in SyncPlans, I would suggest putting this into Katello feature

iNecas: Scratch that for now

lib/foreman_maintain/utils/sync_plans.rb
iNecas: it can more to sync_plans feature as it's already there

definitions/checks/disabled_sync_plans.rb
```diff
@@ -0,0 +1,14 @@
+class Checks::DisableddSyncPlans < ForemanMaintain::Check
```
swapab: @kgaikwad I think is should be `DisabledSyncPlans`?

lib/foreman_maintain/utils/sync_plans.rb
```diff
@@ -0,0 +1,83 @@
...
+        def perform_yaml_action(old_ids, new_ids, action_name)
+          old_ids ||= []
+          case action_name
+          when MAKE_SYNC_PLANS_DISABLE
+            old_ids.concat(new_ids)
+          when MAKE_SYNC_PLANS_ENABLE
+            old_ids -= new_ids
+          end
+          old_ids.uniq!
+          old_ids
```
swapab: I guess you can remove the returning line of code and replace
`.uniq! with .uniq` unless you want to return `nil` for already uniq collection. 
 https://ruby-doc.org/core-2.2.0/Array.html#method-i-uniq-21

lib/foreman_maintain/utils/sync_plans.rb
kgaikwad: @iNecas, 
I have other query about argument for ForemanMaintain.storage.
Here argument would be a label like in metadata we have label or it will represent class name.

iNecas: For now, I would suggest using symbols for distinguishing between the storages, might change in the future ,but for now, I would keep it simple.

kgaikwad:
@iNecas and @swapab , 
I have made code refactoring changes as suggested.
Please let me know if any additional changes required.

definitions/procedures/upgrade_sync_plans_disable.rb
```diff
@@ -0,0 +1,34 @@
+class Procedures::UpgradeSyncPlansDisable < ForemanMaintain::Procedure
```
iNecas: The checks and procedures should follow the same namespacing strategy, if we have `Checks::SyncPlan`, we should have `Procedures::SyncPlans` where we would contain the code regarding this

kgaikwad: Yes, I will modify name of procedure.

kgaikwad:
Sure @iNecas ,
I am doing modifications to storage code changes as per you mentioned. Once done I will add tests for YamlStorage.

kgaikwad:
@iNecas,
I modified the changes as per your comments. Please let me know if any other changes required.
Once done, I will add tests for YamlStorage.

iNecas:
The code looks good now, I would like to give it a try in real world. Please add some unit tests for the storage, as well as for the definitions

kgaikwad:
@iNecas &  @swapab , 
I have added few tests for yaml_storage and checks.
Please let me know if any changes required.

iNecas:
I still need to check against real system, but works for me code-wise

config/foreman_maintain.yml.test
```diff
@@ -9,3 +9,6 @@
 :definitions_dirs:
   - 'test/lib/support/definitions'
   - 'test/lib/support/additional_definitions'
+
+# Mention file path to store data
+:storage_file: 'test/data_test.yml'
```
iNecas: Please updated foreamn_maintan.yml.example as well so that people see all options there

definitions/procedures/sync_plans.rb
```diff
@@ -0,0 +1,34 @@
+class Procedures::SyncPlans < ForemanMaintain::Procedure
```
iNecas: At the current state, it would probably make sense to have `SyncPlans::Enable` and `SyncPlans::Disable` procedures,
and have better message than `[enable/disable sync plans before/after upgrade]?, [y(yes), n(no), q(quit)]`.

definitions/procedures/sync_plans.rb
```diff
@@ -0,0 +1,34 @@
...
+      enabled_sync_plans
+    else
+      disable_all_enabled_sync_plans
+    end
+  end
+
+  private
+
+  def disable_all_enabled_sync_plans
+    with_spinner('disabling all sync plans') do |spinner|
```
iNecas: Any reason for using spinner for disable but not for enable?

kgaikwad: In disable scenario, ids of enabled sync_plans are fetched through sql query which may takes sometime. That's why I added spinner there. 
In enable scenario, ids are read from file only. 
This is the only reason, I added spinner for disable but not for enable.

Please let me know if any changes required in it.

iNecas: Since even for the enable case, there is one request per sync plan, having that behind a spinner might be a good idea

definitions/procedures/sync_plans/enable.rb
```diff
@@ -0,0 +1,18 @@
+module Procedures::SyncPlans
+  class Enable < ForemanMaintain::Procedure
+    metadata do
+      for_feature :sync_plans
+      description 'enable sync plans those are disabled through tool'
```
iNecas: `re-enable active sync plans` would sound a bit better to me

iNecas:
Thanks @kgaikwad, works nicely, merging

definitions/procedures/sync_plans.rb
kgaikwad: @iNecas 
Sorry, I missed this comment. 
I have done these changes, I will opened new PR :-).

Issue #16, opened 2017-03-06
Offer to delete the tasks that are not valid any more


swapab:

swapab:
@iNecas 
* Delete all tasks in planning state
  Q. Shall we delete **all** tasks those are in planning state or should restrict to certain date/type of tasks?

 * Delete all tasks in pending state that are safe to delete
  Q1. Same as above
  Q2. How can be best define 'safe to delete' tasks?

iNecas:
This is more a question for @mccun934, I would recommand, I would rather got with a smaller set, that can happen often, such as `Actions::Katello::Repository::Sync` and perhaps `Actions::Candlepin::ListenOnCandlepinEvents` and widen the range as we go.

swapab:
Closed #33

PR #33, opened 2017-03-18
RFE #16 - Delete invalid tasks

swapab:
**Sub checks and relevant procedures:**
- Check for old or aged foreman tasks(condition: _Paused_ or _stopped_ tasks beyond **30** days)
  - Procedure invoked: `Procedures::ForemanTasksDelete::Aged`
    - Backup _pausesd_ tasks before cleanup
    - Delete aged tasks

- Check for all tasks in planning state
  - Procedure invoked: `Procedures::ForemanTasksDelete::Planning`
  - Backup tasks in _planning_ state
  - Delete tasks in _planning_ state

---
- refactor device.rb did it as I saw it(sorry to mix-up refactoring in this)

### DIFF
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -1,4 +1,16 @@
 class Features::ForemanTasks < ForemanMaintain::Feature
+  MIN_AGE = 30
+
+  SAFE_TO_DELETE = %w[
+    Actions::Katello::Host::GenerateApplicability
+    Actions::Katello::RepositorySet::ScanCdn
+    Actions::Katello::Host::Hypervisors
+    Actions::Katello::Host::HypervisorsUpdate
+    Actions::Foreman::Host::ImportFacts
+    Actions::Candlepin::ListenOnCandlepinEvents
+    Actions::Katello::EventQueue::Monitor
+  ].freeze
+
   metadata do
     label :foreman_tasks
 
@@ -6,6 +18,17 @@ class Features::ForemanTasks < ForemanMaintain::Feature
       check_min_version('ruby193-rubygem-foreman-tasks', '0.6') ||
         check_min_version('tfm-rubygem-foreman-tasks', '0.7')
     end
+  end
+
+  def backup_tasks(state)
+    backup_table('dynflow_execution_plans', state, 'uuid') { |status| yield(status) }
+    backup_table('dynflow_steps', state) { |status| yield(status) }
+    backup_table('dynflow_actions', state) { |status| yield(status) }
+
+    yield('Backup Tasks [running]')
+    export_csv("SELECT * FROM foreman_tasks_tasks WHERE #{condition(state)}",
+               'foreman_tasks_tasks.csv', state)
+    yield('Backup Tasks [DONE]')
   end
 
   def running_tasks_count
@@ -22,8 +45,86 @@ class Features::ForemanTasks < ForemanMaintain::Feature
         WHERE state IN ('paused')
     SQL
     unless ignored_tasks.empty?
-      sql << "AND label NOT IN (#{ignored_tasks.map { |task| "'#{task}'" }.join(',')})"
+      sql << "AND label NOT IN (#{quotize(ignored_tasks)})"
     end
     feature(:foreman_database).query(sql).first['count'].to_i
+  end
+
+  def count(state)
+    feature(:foreman_database).query(<<-SQL).first['count'].to_i
+     SELECT count(*) AS count FROM foreman_tasks_tasks WHERE #{condition(state)}
+    SQL
+  end
+
+  def delete(state)
+    tasks_condition = condition(state)
+
+    feature(:foreman_database).psql(<<-SQL)
+     BEGIN;
+       DELETE FROM dynflow_steps USING foreman_tasks_tasks WHERE (foreman_tasks_tasks.external_id = dynflow_steps.execution_plan_uuid) AND #{tasks_condition};
+       DELETE FROM dynflow_actions USING foreman_tasks_tasks WHERE (foreman_tasks_tasks.external_id = dynflow_actions.execution_plan_uuid) AND #{tasks_condition};
+       DELETE FROM dynflow_execution_plans USING foreman_tasks_tasks WHERE (foreman_tasks_tasks.external_id = dynflow_execution_plans.uuid) AND #{tasks_condition};
+       DELETE FROM foreman_tasks_tasks WHERE #{tasks_condition};
+     COMMIT;
+    SQL
+
+    count(state)
+  end
+
+  def condition(state)
+    raise 'Invalid State' unless valid(state)
+
+    if state == :old
+      old_tasks_condition
+    else
+      tasks_condition(state)
+    end
+  end
+
+  private
+
+  def backup_dir(state)
+    "/var/lib/foreman/backup-tasks/#{state}/#{Time.now.strftime('%Y-%m-%d_%H-%M-%S')}"
+  end
+
+  def backup_table(table, state, fkey = 'execution_plan_uuid')
+    yield("Backup #{table} [running]")
+    sql = "SELECT #{table}.* FROM foreman_tasks_tasks JOIN #{table} ON\
+       (foreman_tasks_tasks.external_id = #{table}.#{fkey})"
+    export_csv(sql, "#{table}.csv", state)
+    yield("Backup #{table} [DONE]")
+  end
+
+  def export_csv(sql, file_name, state)
+    dir = prepare_for_backup(state)
+    filepath = "#{dir}/#{file_name}"
+    execute("echo \"COPY (#{sql}) TO STDOUT WITH CSV;\" \
+      | su - postgres -c '/usr/bin/psql -d foreman' > #{filepath}")
+    execute("bzip2 -9 #{filepath}")
+  end
+
+  def old_tasks_condition(state = "'stopped', 'paused'")
+    "foreman_tasks_tasks.state IN (#{state}) AND " \
+       "foreman_tasks_tasks.started_at < CURRENT_DATE - INTERVAL '#{MIN_AGE} days'"
+  end
+
+  def prepare_for_backup(state)
+    dir = backup_dir(state)
+    execute("mkdir -p #{dir}")
+    dir
+  end
+
+  def quotize(array)
+    array.map { |el| "'#{el}'" }.join(',')
+  end
+
+  def tasks_condition(state)
+    safe_to_delete_tasks = quotize(SAFE_TO_DELETE)
+    "foreman_tasks_tasks.state = '#{state}' AND " \
+    "foreman_tasks_tasks.label IN (#{safe_to_delete_tasks})"
+  end
+
+  def valid(state)
+    [:old, :planning, :pending].include?(state)
   end
 end
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -29,6 +29,7 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     export_csv("SELECT * FROM foreman_tasks_tasks WHERE #{condition(state)}",
                'foreman_tasks_tasks.csv', state)
     yield('Backup Tasks [DONE]')
+    @backup_dir = nil
   end
 
   def running_tasks_count
@@ -84,7 +85,8 @@ class Features::ForemanTasks < ForemanMaintain::Feature
   private
 
   def backup_dir(state)
-    "/var/lib/foreman/backup-tasks/#{state}/#{Time.now.strftime('%Y-%m-%d_%H-%M-%S')}"
+    @backup_dir ||=
+      "/var/lib/foreman/backup-tasks/#{state}/#{Time.now.strftime('%Y-%m-%d_%H-%M-%S')}"
   end
 
   def backup_table(table, state, fkey = 'execution_plan_uuid')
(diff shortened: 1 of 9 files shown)


definitions/features/foreman_tasks.rb
```diff
@@ -19,4 +21,58 @@ def paused_tasks_count
...
+  end
+
+  def backup_tasks
+    backup_table('dynflow_execution_plans', 'uuid') { || yield(status) }
+    backup_table('dynflow_steps') { |status| yield(status) }
+    backup_table('dynflow_actions') { |status| yield(status) }
+
+    yield('Backup Tasks [running]')
+    export_csv("SELECT * FROM foreman_tasks_tasks WHERE #{old_paused_tasks_condition}",
+               '$DIR/foreman_tasks_tasks.csv')
```
iNecas: Where does the $DIR varaible come from?

definitions/procedures/foreman_tasks_delete/aged.rb
```diff
@@ -0,0 +1,18 @@
+module Procedures::ForemanTasksDelete
+  class Aged < ForemanMaintain::Procedure
+    label :delete_old_foreman_tasks
+    description 'Delete Old Tasks'
+
+    def run
+      old_tasks_count = feature(:foreman_tasks).old_tasks_count
```
iNecas: Once we get https://github.com/iNecas/foreman_maintain/pull/29 in, it should be possible to parametrize the procedure, where we could specify the parameters of which kind of tasks we want to delete.

What would be a reason to delete one kind of tasks?

definitions/features/foreman_tasks.rb
swapab: It actually goes as a `file` argument to `export_csv` which then executes the string. So $DIR will be environment variable if any is set. Else the files will be saved to /
Which is not a good idea. What would be the best place to write the backup? One I can think of is user home(~/<some-folder>) or `/var/local/`

iNecas: `/var/lib/foreman` might be the best place for now. Also I suggest putting there a subdirectory including timestamp of the operation to better cover multiple runs of this

swapab:
@iNecas I have improvised the agenda of this PR from `Check and delete old tasks` to a much higher level: `Check and delete invalid tasks`. Which would have multiple asserts(viz. check and delete old tasks, tasks in planning state, tasks safe to delete).

But as I try, I am not able to run multiple `asserts` within one `run` of the check.
https://github.com/swapab/foreman_maintain/commit/9dbd7115e64aca21c68d73b6a70b404ccac8bd3b#diff-e6f212952523f23e427e5b4a6e810537R18
The second `assert` (assert for tasks in planning state) only runs if previous assert passes and next one fails. If both fails then, only first one executes and next gets skipped.

Am I doing it incorrectly?

swapab:
@iNecas I have made the respective updates per this PR description.
Can you please review and let me know how it looks?

---
I have used the knowledge shared by **Mike McCune**(@mccun934) in the email with subject "Define Invalid Task".

```
# Planning state tasks restricted to:
Actions::Candlepin::CandlepinListeningService
Actions::Katello::EventQueue::Monitor

# Pending and safe to delete tasks restricted to:
Actions::Katello::Host::GenerateApplicability
Actions::Katello::RepositorySet::ScanCdn
Actions::Katello::Host::Hypervisors
Actions::Katello::Host::HypervisorsUpdate
Actions::Foreman::Host::ImportFacts
```

iNecas:
Sound like sane list of such tasks.

swapab:
@iNecas This PR is ready for review.

definitions/features/foreman_tasks.rb
```diff
@@ -22,8 +48,92 @@ def paused_tasks_count(ignored_tasks = [])
...
+  def safe_to_delete_tasks_pending
+    PENDING
+  end
+
+  def safe_to_delete_tasks_planning
+    PLANNING
+  end
+
+  def tasks_condition(query)
+    safe_to_delete_tasks = quotize(send("safe_to_delete_tasks_#{query}"))
```
iNecas: The query feels like it should contain some kind of query, condition, but in facts it's more like state. We probably need a better naming for this. Also I thing we tasks that are safe to delete should not be distinquished by planning/pening.  It's more like a set of tasks that are safe to delete no matter in what state they are.

Another suggestion is avoiding the send and calling methods dynamically: it makes it pretty hard to read and navigate thought the code later: imagine one would like so find where is the `safe_to_delete_task_planning` method called from…

definitions/procedures/foreman_tasks/delete/old.rb
```diff
@@ -0,0 +1,32 @@
+module Procedures::ForemanTasks
+  module Delete
+    class Old < ForemanMaintain::Procedure
```
iNecas: It seems we could still unify the `delete/old.rb` and `delete/pending_state.rb` just by having a `delete.rb`, that would take the properties of what task to delete, based on arguments

definitions/features/foreman_tasks.rb
swapab: I was getting that notion of naming. Changed it to `state`.

Agreed on avoiding using `send`. Removed.

swapab:
@iNecas Refactoring done. Can you please re-review?

test/definitions/checks/disk_speed_minimal_test.rb
```diff
@@ -1,38 +0,0 @@
-require 'test_helper'
```
iNecas: Was it intentional to delete this?

swapab: @iNecas Yes. Deleted intentionally.
Hard time getting around error message.

```ruby
  1) Error:
Checks::DiskSpeedMinimal#test_0002_executes successfully for disk with minimal speed:
RuntimeError: Trying to get execution information before the run started happened
    /root/foreman_maintain/lib/foreman_maintain/executable.rb:30:in `execution'
    /root/foreman_maintain/definitions/checks/disk_speed_minimal.rb:17:in `run'
    /root/foreman_maintain/test/definitions/checks/disk_speed_minimal_test.rb:22:in `block (2 levels) in <top (required)>'


  2) Failure:
Checks::DiskSpeedMinimal#test_0003_raise error if disk speed does not meet minimal requirement [/root/foreman_maintain/test/definitions/checks/disk_speed_minimal_test.rb:35]:
[ForemanMaintain::Check::Fail] exception expected, not
Class: <RuntimeError>
Message: <"Trying to get execution information before the run started happened">
---Backtrace---
/root/foreman_maintain/lib/foreman_maintain/executable.rb:30:in `execution'
/root/foreman_maintain/definitions/checks/disk_speed_minimal.rb:17:in `run'
/root/foreman_maintain/test/definitions/checks/disk_speed_minimal_test.rb:35:in `block (3 levels) in <top (required)>'
```

iNecas: you should be able to use https://github.com/iNecas/foreman_maintain/blob/master/test/definitions/test_helper.rb#L30 for preparing the execution properly, with the dependencies it needs.

swapab: @iNecas Tons of thanks for the pointer. **It worked** :smile:
_Test case restored_

extra/backup-tasks.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
iNecas: Are this files use by the code somewhere else in the checks? If not, I would rather remove them for now

swapab: **YAGNI** _removed_

definitions/features/foreman_tasks.rb
```diff
@@ -1,4 +1,16 @@
 class Features::ForemanTasks < ForemanMaintain::Feature
+  MIN_AGE = 30
+
+  SAFE_TO_DELTE = %w[
+    Actions::Katello::Host::GenerateApplicability
+    Actions::Katello::RepositorySet::ScanCdn
+    Actions::Katello::Host::Hypervisors
+    Actions::Katello::Host::HypervisorsUpdate
+    Actions::Foreman::Host::ImportFacts
+    Actions::Candlepin::CandlepinListeningService
```
iNecas: It should be `Actions::Candlepin::ListenOnCandlepinEvents`

definitions/features/foreman_tasks.rb
```diff
@@ -1,4 +1,16 @@
 class Features::ForemanTasks < ForemanMaintain::Feature
+  MIN_AGE = 30
+
+  SAFE_TO_DELTE = %w[
```
iNecas: Typo

definitions/features/foreman_tasks.rb
```diff
@@ -22,8 +45,84 @@ def paused_tasks_count(ignored_tasks = [])
...
+  end
+
+  def export_csv(sql, file_name, state)
+    dir = prepare_for_backup(state)
+    execute("echo \"COPY (#{sql}) TO STDOUT WITH CSV;\" \
+      | su - postgres -c '/usr/bin/psql -d foreman' > #{dir}/#{file_name}")
+  end
+
+  def old_tasks_condition(state = "'stopped', 'paused'")
+    "foreman_tasks_tasks.query IN (#{state}) AND " \
```
iNecas: `s/query/state`

definitions/features/foreman_tasks.rb
```diff
@@ -22,8 +45,84 @@ def paused_tasks_count(ignored_tasks = [])
...
+    yield("Backup #{table} [running]")
+    sql = "SELECT #{table}.* FROM foreman_tasks_tasks JOIN #{table} ON\
+       (foreman_tasks_tasks.external_id = #{table}.#{fkey})"
+    export_csv(sql, "#{table}.csv", state)
+    yield("Backup #{table} [DONE]")
+  end
+
+  def export_csv(sql, file_name, state)
+    dir = prepare_for_backup(state)
+    execute("echo \"COPY (#{sql}) TO STDOUT WITH CSV;\" \
```
iNecas: There can be quite a lot of data there: I recommend running the output though bzip ` bzip 2 -9` with `csv.bz2` suffix

swapab: You mean compressing the `csv` file(to `.bz2`)  after copying the data?

definitions/features/foreman_tasks.rb
swapab: Oops. Corrected.

definitions/features/foreman_tasks.rb
swapab: Corrected.

definitions/features/foreman_tasks.rb
swapab: Done. `execute("bzip2 -9 #{filepath}")`

iNecas: I would recommend rather going though `|` there, to avoid storing the original file (that already can be quite big) and saving just the compressed one.

definitions/features/foreman_tasks.rb
```diff
@@ -22,8 +45,86 @@ def paused_tasks_count(ignored_tasks = [])
...
+    export_csv(sql, "#{table}.csv", state)
+    yield("Backup #{table} [DONE]")
+  end
+
+  def export_csv(sql, file_name, state)
+    dir = prepare_for_backup(state)
+    filepath = "#{dir}/#{file_name}"
+    execute("echo \"COPY (#{sql}) TO STDOUT WITH CSV;\" \
+      | su - postgres -c '/usr/bin/psql -d foreman' > #{filepath}")
+    execute("bzip2 -9 #{filepath}")
```
iNecas: Don't forget to add `.bz2` suffix when compressing

swapab: Yeah, I did that. but it wasn't compressing in this case. When I removed the `.bz2` suffix it created the `.bz2` files implicitly. Probably it expects the exact file to be compressed.

swapab: And yes it removes the original file, once done compressing. That's what I saw on my machine, could there be a possibility of difference in behavior due to version change?
Mine is `1.0.6`

```
sat62 :: ~ » bzip2 --version
bzip2, a block-sorting file compressor.  Version 1.0.6, 6-Sept-2010.
   
   Copyright (C) 1996-2010 by Julian Seward.
   
   This program is free software; you can redistribute it and/or modify
   it under the terms set out in the LICENSE file, which is included
   in the bzip2-1.0.6 source distribution.
   
   This program is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
   LICENSE file for more details.
   
bzip2: I won't write compressed data to a terminal.
bzip2: For help, type: `bzip2 --help'.
```

iNecas: have you tried redirecting the output to a file?

swapab: Yes, `redirect` would definitely work. But I am not able to understand the point in using it as normal `bzip2 -9 filename` is giving the expected result.

iNecas: The thing is that you need first the disk space to hold the uncompressed file + additional space for compressed (until the uncompressed gets deleted), while with redirect, you only need the space for compressed. Hopefully, users should have more space on the disk, but still… if we can choose, we should go for less demanding solution

swapab: Done. @iNecas Could you please re-review?
Thanks!

iNecas:
Thanks @swapab

PR #35, opened 2017-03-21
Hammer initial support

iNecas:
@kgaikwad I've tried to turn the hammer code into a feature (see a commit history) but then it seemed that the hammer acts more like an infrastructure for other features/checks/procedures, than actual feature, so I moved it back to utility class. I've removed the code about loading credentials from hammer config files, as we actually don't need them, if they are already present.

The code should be extendable for being able to ask the user for credentials and store it somewhere, when some checks/procedures would require it.


iNecas:
Asking for hammer password could be at similar stage as we would install missing packages https://github.com/iNecas/foreman_maintain/issues/15

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -63,31 +64,51 @@ def parse_csv(data)
...
+
+      def parse_execute_options(options)
+        options = options.dup
+        stdin = options.delete(:stdin)
+        hidden_strings = Array(options.delete(:hidden_patterns))
+        raise ArgumentError, "Unexpected options: #{options.keys.inspect}" unless options.empty?
+        [stdin, hidden_strings]
+      end
+
+      def shellescape(string)
```
kgaikwad: This method is not used anywhere. Please correct me if wrong.

iNecas: It is not: I had it there for escaping the password on command line, but I left it there for future cases, where we will need to put user input on command line, which I expect to happen sooner or alter

lib/foreman_maintain/utils/hammer.rb
```diff
@@ -0,0 +1,42 @@
...
+      # run_command('host list')
+      def run_command(args)
+        output = execute("#{command_base} #{args}")
+        if output =~ /Invalid username or password/
+          # TODO: change the error message after we are able to ask the user
+          # for credentials and store it somewhere
+          raise CredentialsError, 'Invalid hammer credentials: '\
+            'we expect the hammer username/password to be stored'\
+            'in ~/.hammer/cli.modules.d/foreman.yml'
+        end
```
kgaikwad: Return the output of command from this method.

iNecas: Good catch @kgaikwad ! Updated and convered in tests

kgaikwad: @iNecas, 
Done with review. Now, we can merge this changes.

iNecas:
Thanks @kgaikwad for review

PR #36, opened 2017-03-22
Fixes #18 - introduce configuration file

### LINKED ISSUE #18: Introduce configuration file for foreman-maintain
1. put it in `config/foremain_maintain.yml` (should be in `.gitignore)
2. have `config/foreman_maintain.yml.example` in the repository
3. expose all configuration options (as stated in https://github.com/iNecas/foreman_maintain/blob/ce490caa16cfe2525a8406ddff73fdaf869ec4de/lib/foreman_maintain/config.rb) there (especially configuration of log directory).

kgaikwad:


lib/foreman_maintain/settings.rb
```diff
@@ -0,0 +1,14 @@
+require 'yaml'
+module ForemanMaintain
+  class Settings
+    def self.load_from_file(file_dir)
+      file_path = File.join(file_dir, 'foreman_maintain.yml')
```
iNecas: For this code, it's perhaps not needed for having outside of the runner class, + it would be nice to have info about what file was used for configuration in the logger (which is quite hard doing form here, as the setup happens afterwards). What about a simpler version, directly in side the executable file

```ruby
config_file = File.expand_path('../config/foreman_maintain.yml', __FILE__)
 if File.exists?(config_file)
  config = YAML.load_file(config_file)
else
  config = {}
  default_config = true
end
ForemanMaintain.setup(config)
if default_config
  ForemanMaintain.info("Config file #{config_file} not found, using default configuration")
end
```

iNecas: Alternatively, we could let the `ForemanMaintain.setup` to accept a strnig, in which case it would assume it was a patch to yaml and perform the logic there, which might look a bit nicer than in executable

kgaikwad: Yes, second way is good to implement. I will go with second comment.

config/foreman_maintain.yml.example
```diff
@@ -0,0 +1,10 @@
+# Directory where the logs are stored.
+# The default file is named as /log/foreman_maintain.log
+:log_dir: 'log'
```
kgaikwad: @iNecas ,
No any specific reason. 
Generally, user can only allow to provide directory for logs not a log file. 

Assuming that if user is frequently changing directory path of log files due to his environment setup then at least we should know the log file name. Using which it will become easy to locate and find directory path.

After allowing users to set log file, it will create difficulties for us during troubleshooting.
In the future, what If these logs are included in foreman-debug then file name will work.

Please let me know your opinion.

iNecas: Works for me

lib/foreman_maintain.rb
```diff
@@ -67,9 +79,22 @@ def init_logger
...
       end
     end
+
+    def load_custom_configs(file_path)
+      file_path ||= ''
+      custom_configs = {}
+      if File.exist?(file_path)
+        custom_configs = YAML.load(File.open(file_path)) || {}
+      else
+        $stderr.puts "Config file #{file_path} not found, using default configuration"
```
iNecas: Could you think of changing the code in a way that this messages would be produced after the logger is initialized and producing it via a logger?

kgaikwad: Yes. Excepting the exception messages we can store other messages in a queue. 
Once a logger initialized, we can produce these messages via logger.

iNecas: Makes sense

kgaikwad:
@iNecas ,
Done with the rebasing, branch is now up to date with the master.

lib/foreman_maintain.rb
```diff
@@ -28,8 +29,19 @@ module ForemanMaintain
...
+      'debug' => ::Logger::DEBUG,
+      'info' => ::Logger::INFO,
+      'warn' => ::Logger::WARN,
+      'error' => ::Logger::ERROR,
+      'fatal' => ::Logger::FATAL,
+      'unknown' => ::Logger::UNKNOWN
+    }.freeze
+
+    def setup(options = {})
+      @msg_queue = []
```
iNecas: The `@msg_queue` could have a more descriptive name. What about `@pre_setup_log_messages`, and + a small comment why we are doing so.

lib/foreman_maintain.rb
```diff
@@ -68,9 +81,28 @@ def init_logger
...
       filename = File.expand_path("#{config.log_dir}/foreman-maintain.log")
       @logger = Logger.new(filename, 10, 10_240_000).tap do |logger|
-        logger.level = config.log_level || Logger::DEBUG
+        logger.level = LOGGER_LEVEL_MAPPING[config.log_level] || Logger::DEBUG
         logger.datetime_format = '%Y-%m-%d %H:%M:%S%z '
       end
+      popup_msg_queue
+    end
+
+    def popup_msg_queue
```
iNecas: The method should perhaps be called `pickup_log_messages`, as `popup` seems more like the messages should popup somewhere in the user interface

lib/foreman_maintain.rb
```diff
@@ -68,9 +81,28 @@ def init_logger
...
+      return if @msg_queue.empty?
+      @msg_queue.each { |msg| logger.info msg }
```
iNecas: We should clean the queue once finished

kgaikwad: @iNecas ,
Corrected the changes.

lib/foreman_maintain.rb
```diff
@@ -68,9 +82,29 @@ def init_logger
...
+      file_path ||= ''
+      custom_configs = {}
+      if File.exist?(file_path)
+        custom_configs = YAML.load(File.open(file_path)) || {}
+      else
+        @pre_setup_log_messages << "Config file #{file_path} not found, using default configuration"
+      end
+      custom_configs
+    rescue => e
+      $stderr.puts "Couldn't load configuration file: #{e.message}"
```
iNecas: We probably should raise an error here: if there is an error in the config file, we should not even continue

iNecas:
Thanks @kgaikwad

PR #40, opened 2017-03-23
Foreman tasks paused tasks check and resume procedures

iNecas:
This PR introduces checking for paused tasks in the system and offers either resuming them via
hammer or investigating them via UI (pointing the user to the right url)

As part of working on that, I've added:

* ability to organize the checks and procedures into module namespaces (while keeping the labels
working as before, `Check::FormanTasks::Resume` class will have autogenerated label 'foreman-tasks-resume' as it had before (@swapab would this layout work for you also for tasks cleanup?)
* ability to ask question from procedures (when reporter supports it)
* more helpers to test definitions (`run_step` with procedures `run_procedure` and `run_check`) that allows to simulate the check/procure run and look at the result/outputs produced

Includes commits from (https://github.com/iNecas/foreman_maintain/pull/35 - will remove them once it's merged)


iNecas:
We will need to run the tasks cleanup BEFORE the paused->resume check, as some paused tasks might not be relevant for resuming anymore and we need to clean them beforehand

swapab:
> @swapab would this layout work for you also for tasks cleanup?

Absolutely it will work.
Comparing current state of #33 and  your commits(#40). Checks namespace looks fine.

Procedures would look something like
```bash
definitions/procedures
├── foreman_tasks
│   └── delete
│       ├── aged.rb
│       ├── pending_state.rb
│       └── planning_state.rb
```
I think this should be good. :+1:

iNecas:
Rebased, ready for re-review

iNecas:
I've rebased against latest master. @swapab I've noticed the multipe-steps selection was not working and as part of the fixing I did some refactoring towards decoupling reporter and runner.

Before this commit, we required the reporter to call back runner
methods, which complicated the things. Also the multiple-steps select
was not working properly.

After this change, the reporter just returns a decision as result of
`on_next_steps` method, and the runner is responsible for deciding
what to do next with it.

iNecas:
Also, I've added some conditions to ignore some tasks that are known to be paused but not to cause issues

swapab:
>  decoupling reporter and runner.

This is a life saver :smiley: 
It was getting difficult to clutter around `runner` object.

iNecas:
Squashed and merged 019ca085c0327d185bd3b0f61889e612736845b0

PR #45, opened 2017-03-31
Compatiblity with rubocop v0.48.0

swapab:
1. Disabled: Use %i or %I for an array of symbols
2. Fixed: Use 2 spaces for indentation in a heredoc by using some library(e.g. ActiveSupport's String#strip_heredoc). Created `foreman_maintain/core_ext.rb` extend String class.
3. Fixed: Parenthesize the param find to make sure that the block will be associated with the assert/refute method call.
4. Fixed: Use >= instead of inverting <.

Following offense fixed by disabling IverseMethods check:

```ruby
lib/foreman_maintain/concerns/metadata.rb:62:11: C: Use >= instead of inverting <.
          !(superclass < Metadata)
```

Disabled `InverseMethod` rubocop check:
```yaml
Style/InverseMethods:
  Enabled: false
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,3 +53,11 @@ Security/YAMLLoad:
 # some proper logging
 Metrics/MethodLength:
   Max: 15
+
+Style/SymbolArray:
+  Description: 'Use %i or %I for arrays of symbols.'
+  StyleGuide: '#percent-i'
+  Enabled: false
+
+Style/InverseMethods:
+  Enabled: false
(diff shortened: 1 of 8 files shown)


iNecas:
+1 for disabling the InverseMethod check

swapab:
Thanks @iNecas. We can rely on peer review to catch the above offense.

lib/foreman_maintain/customize_core.rb
```diff
@@ -0,0 +1,6 @@
+class String
```
iNecas: I've seen multiple gems having `core_ext.rb` file for this kind of things. We can follow this here as well

swapab: Yes. I was trying to recollect that term :smile: Thanks!

iNecas:
Thanks @swapab !

Issue #23, opened 2017-03-13
Check: Verify self registration


lzap:
It would be good to create a check that searches for `katello-ca-consumer-*.rpm` installed and compares the hostname in the RPM package with hostname of the system.

If they are same, I'd issue WARNING: This system is self-registered.

Server can be sometimes self-registered, therefore this is not an ERROR but WARNING. We've encountered an issue when Capsule was self-registered and we did not notice that, it took lot of time to figure out. A warning could should this immediately.

swapab:
@lzap This check has been implemented in PR https://github.com/iNecas/foreman_maintain/pull/53
Can you please try it out and let us know if it works as you expected?

Also, do you think it would be a good idea to convey the user about workaround(or resolution steps) for the WARNING?

lzap:
Sorry for late response, I think the warning is only relevant on Capsule. On Server, they are sometimes self-registered and it should not be a warning, that's [OK]. Apologize for confusion:

Server:
* CDN or self registered - OK
* WARNIN otherwise

Capsule:
* CDN or self registered - WARNING: Capsule should have been registered to Satellite Server
* OK otherwise

iNecas:
Tracking as http://projects.theforeman.org/issues/19715, closing the issue as we are moving away from github issue tracker to Redmine http://projects.theforeman.org/projects/foreman-maintain.

PR #53, opened 2017-04-19
RFE #23 - Add new status [WARNING] - Check: is system self registered

swapab:
- Add check `Checks::SystemRegistration`
- Add new status and label named `warning`
- Add context block `test/minitest_spec_context.rb`

### DIFF
--- a/definitions/checks/system_registration.rb
+++ b/definitions/checks/system_registration.rb
@@ -1,0 +1,31 @@
+class Checks::SystemRegistration < ForemanMaintain::Check
+  metadata do
+    label :verify_self_registration
+    description 'Check whether system is self-registered or not'
+    tags :default
+
+    confine do
+      file_exists?('/etc/rhsm/rhsm.conf')
+    end
+  end
+
+  def run
+    if system_is_self_registerd?
+      raise ForemanMaintain::Error::Warn, 'System is self registered'
+    else
+      puts 'System is not self registered'
+    end
+  end
+
+  def system_is_self_registerd?
+    rhsm_hostname.casecmp(hostname).zero?
+  end
+
+  def rhsm_hostname
+    execute("grep '\\bhostname\\b' < #{rhsm_conf_file}  | grep -v '^#'").sub(/.*?=/, '').strip
+  end
+
+  def rhsm_conf_file
+    '/etc/rhsm/rhsm.conf'
+  end
+end
--- a/lib/foreman_maintain/check.rb
+++ b/lib/foreman_maintain/check.rb
@@ -7,9 +7,6 @@ module ForemanMaintain
 
     attr_accessor :associated_feature
 
-    class Fail < StandardError
-    end
-
     # run condition and mark the check as failed when not passing
     #
     # ==== Options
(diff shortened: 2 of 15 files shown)


lib/minitest_spec_context.rb
```diff
@@ -0,0 +1,9 @@
+require 'minitest/spec'
+
+module MiniTest
```
iNecas: Please keep this in `test/test_helper.rb` or some file in that directory. It looks strange having it in `lib` which should not contain test-related code

iNecas:
One comment, but looks good otherwise

lib/minitest_spec_context.rb
swapab: :+1:
Moved it to `test/helpers/minitest_spec_context.rb`

definitions/checks/system_registration.rb
```diff
@@ -0,0 +1,29 @@
...
+      raise ForemanMaintain::Error::Warn, 'System is self registered'
+    else
+      puts 'System is not self registered'
+    end
+  end
+
+  private
+
+  def system_is_self_registerd?
+    rhsm_hostname.include?(hostname)
```
iNecas: We should match only the last word from the line and match on equality, rather than include: you can have `sat.example.com' registered to `another-sat.example.com` and this check would fail: a corner case, but still…

test/helpers/minitest_spec_context.rb
```diff
@@ -0,0 +1,9 @@
+require 'minitest/spec'
```
iNecas: we use `support` directory in `test/lib`, we can name this directory `support` as well, for consistency

swapab: Yes I had thought of that. I thought it would be an `ambiguity` :stuck_out_tongue:

definitions/checks/system_registration.rb
swapab: Aah. Good catch. I did a split on ` = ` later checked for `hostname` in the derived array.

definitions/checks/system_registration.rb
```diff
@@ -0,0 +1,29 @@
...
+      raise ForemanMaintain::Error::Warn, 'System is self registered'
+    else
+      puts 'System is not self registered'
+    end
+  end
+
+  private
+
+  def system_is_self_registerd?
+    rhsm_hostname.split(' = ').include?(hostname)
```
iNecas: No need to match for the string before ' = ', also, there could be 'hostname=myhost.example.com' that would not get though this. This (or similar variant) should be better `rhsm_hostname.sub(/.*?=/,'').strip`

swapab: Better :+1: Thanks!

definitions/checks/system_registration.rb
```diff
@@ -0,0 +1,27 @@
...
+      puts 'System is not self registered'
+    end
+  end
+
+  def system_is_self_registerd?
+    rhsm_hostname.casecmp(hostname).zero?
+  end
+
+  def rhsm_hostname
+    execute("grep '\\bhostname\\b' < /etc/rhsm/rhsm.conf").sub(/.*?=/, '').strip
```
iNecas: While testing, I hit sill one issue, when I had a line commented out with this value. `'^\\s*\\bhostname\\b'` to match online uncommented line.

swapab: Handled this case as well. Thanks!

iNecas:
Thanks @swapab

PR #55, opened 2017-04-20
Added spinner to enable sync_plans

kgaikwad:


definitions/features/sync_plans.rb
```diff
@@ -47,6 +47,7 @@ def update_records(ids, enabled)
   ensure
     new_data = sync_plan_data(enabled, updated_record_ids)
     save_state(new_data)
+    updated_record_ids
```
iNecas: As we already talked in a past: ensure block will not influence what the method returns. Please move it before the ensure block

kgaikwad: @iNecas , 
Again my bad, sorry!!
Updated the PR, I forgot method returns ids from each loop not updated_record_ids and thats why it is executed successfully.

iNecas:
Thanks @kgaikwad

PR #56, opened 2017-04-24
Fixes #39 - raise warning on unread mails

### LINKED ISSUE #39: Check: If there are unread mail for root issue warning
We have many crons, they end up in root mail folder (or file technically).

kgaikwad:


definitions/checks/unread_mail.rb
```diff
@@ -0,0 +1,40 @@
...
+  metadata do
+    label :unread_mail
+    description 'Check for unread mail for root user.'
+    tags :default
+  end
+
+  attr_reader :file_path
+
+  def initialize
+    @file_path = '/var/spool/mail/root'
```
gnurag: Would it be feasible to retrieve mail spool path from site-specific configuration file `/etc/login.defs`? This file defines the default path for `MAIL_DIR`, and its value differs based on distro.

On Fedora*
```shell
[anurag@coruscant ~]$ cat /etc/redhat-release 
Fedora release 25 (Twenty Five)
[anurag@coruscant ~]$ grep ^MAIL_DIR /etc/login.defs 
MAIL_DIR	/var/spool/mail
[anurag@coruscant ~]$ 
```

On Debian*
```shell
anurag@ellen:~$ cat /etc/debian_version 
wheezy/sid
anurag@ellen:~$ grep ^MAIL_DIR /etc/login.defs 
MAIL_DIR        /var/mail
anurag@ellen:~$ 
```
You may want to retain the default value of `/var/spool/mail/root` and fallback to it if the above queries fail.

definitions/checks/unread_mail.rb
```diff
@@ -0,0 +1,40 @@
...
+    fail_msg += ' Please check before proceeding further.'
+    raise ForemanMaintain::Error::Fail, fail_msg
+  end
+
+  def file_exists?
+    File.exist?(file_path)
+  end
+
+  def number_of_mails_present
+    cmd = "egrep '^Message-Id' #{file_path} | wc -l"
```
gnurag: `egrep --count` can provide a count of matching lines, without passing the output to wc for counting.

definitions/checks/unread_mail.rb
kgaikwad: @gnurag,
Thank you for providing information. 
I will go through it and fix the code as per your suggestions.

kgaikwad: @gnurag , 
Done with the suggested changes. 
Instead of hard-coded path, value of `MAIL_DIR` is set as a mailbox path.
This value of `MAIL_DIR` is retrieved from config file `/etc/login.defs` .

gnurag: LGTM.

definitions/checks/unread_mail.rb
```diff
@@ -0,0 +1,62 @@
...
+      output = execute(cmd).strip
+      dir = output unless output.empty?
+    end
+    dir
+  end
+
+  def guidline_steps
+    <<-EOF
+    Please follow steps manually to clear root mailbox i.e #{file_path}:
+    1) To read mails use vim or cat for mailbox file
```
iNecas: What about defining this steps into `procedures/system_mailbox` and linking to them from the check
in assertion: this would be really nice user experience, if we listed the file for them, as well as allowed them
to delete that file.

kgaikwad:
@iNecas, @gnurag, @swapab,
Done with the suggested changes. Ready for review.

- Converted manual text steps to view & delete unread mail list to procedures
- Added `error_type` key in options for method `assert`. So that instead of fail, we can pass and print other error types too.

swapab:
I think, there should be a `mail` feature. And check should be for `any new mail?`.
It would be a structural change with all the check logic moving to mail feature and call `feature(:mail).new_mail?` from the check.

kgaikwad:
@swapab,
I don't think to create `mail` as feature would be needed here.
There will be very less chance where future tasks will be related to system mails.
Due to this reason only I have not created `mail` as feature separately.

If you still think, this change should be needed, please let me know. 
I will do modifications and update PR.

definitions/checks/unread_mail.rb
```diff
@@ -0,0 +1,46 @@
...
+  attr_reader :file_path, :mailbox_dir
+
+  def initialize
+    @mailbox_dir = find_mailbox_directory
+    @file_path = root_mail_file_path
+  end
+
+  def run
+    if file_exists?(file_path)
+      mails_count = find_mails_count.to_i
```
swapab: can move `to_i` to the method itself.
Method name can be `mails_count`.

swapab:
can keep the check as is. we'll introduce `mail` when firm use-case arrives.

kgaikwad:
Ready for review. 
As I am not using any mail client package and there is no any info about any mail which says it is unread,  
I am showing all mails count from file `/var/spool/mail/root `.

lib/foreman_maintain/check.rb
```diff
@@ -15,18 +15,14 @@ class Check < Executable
     #                   the failure, will be offered to the user when running
     #                   in interactive mode
     #
-    # * +:warn* - issue warning instead of failure: this is less strict check,
-    #      that could be considered as non-critical for continuing with the scenario
+    # * +:error_type* - error type if wants to override failure
```
kgaikwad: @iNecas , @swapab 
Here, I have replaced `:warn` option to `:error_type`.
Using this option, we can raise warning or any other error_type of exceptions.

kgaikwad:
@iNecas ,
Resolved conflicts.

swapab:
@kgaikwad  I tried this on my local setup.

```
Check for verifying syntax for ISP DHCP configurations:         
Not Found 404. Response: ""
              [FAIL]
Please check and verify DHCP configurations.
check for paused tasks:                                               [OK]
Check for mails in root's mailbox.:                                   [WARNING]
WARNING: Found 20 mail(s) in root's mailbox.
1) list mails from root's mailbox ({:file_path=>#<ForemanMaintain::Param:0x0000000194d8c8 @name=:file_path, @description="", @options={}, @required=false, @flag=false, @block=nil, @array=false>})
2) clear root's mailbox ({:file_path=>#<ForemanMaintain::Param:0x0000000194fbc8 @name=:file_path, @description="", @options={}, @required=false, @flag=false, @block=nil, @array=false>})
Select step to continue, [n(next), q(quit)] q
```

I think the key `:file_path` should have a `String` as value instead of `ForemanMaintain::Param` object.

kgaikwad:
@swapab ,
I guess something is broken in my code due to new changes, I will check on this.

kgaikwad:
@swapab ,

Sorry for confusion nothing is broken so far :-)
This param list shown along with procedure's description due to method [runtime_message](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/concerns/metadata.rb#L255)

I have overridden this method in both procedures. Now, it will show only procedure's description for this check.

kgaikwad:
@swapab ,

Reverted the previous changes as fixed in [PR-97](https://github.com/theforeman/foreman_maintain/pull/97).

kgaikwad:
Pulled new changes from master and updated the PR.

ntkathole:
@kgaikwad Testing against satellite 6.3 on RHEL 7 gives some error on my setup

```
./bin/foreman-maintain advanced procedure run system-mailbox-clear-mailbox
clear root's mailbox:
/ clearing all existing mailssh: -c: line 0: syntax error near unexpected token `2'
sh: -c: line 0: `cat /dev/null >  2>&1'
                                         [OK]

```
and also running system-mailbox-list-mailbox stucked (never ends)

```
./bin/foreman-maintain advanced procedure run system-mailbox-list-mailbox
list mails from root's mailbox:
```

sean797:
Whats the motivation behind this? Can we at least add it to the commit message please.

kgaikwad:
@sean797, updated the commit message.

@ntkathole , added `advanced_run` metadata to above procedures and updated the pull-request.

definitions/checks/unread_mail.rb
```diff
@@ -0,0 +1,46 @@
+class Checks::UnreadMail < ForemanMaintain::Check
+  metadata do
+    label :unread_mail
+    description "Check for mails in root's mailbox."
+    tags :unread_mail
```
ntkathole: @kgaikwad We are tagging this check with unread_mail...Currently this  check won't run before upgrade unless user run it explicitly ( foreman-maintain health check --label unread-mail ). So I think to this need to add in scenario definitions https://github.com/theforeman/foreman_maintain/blob/master/definitions/scenarios/upgrade_to_satellite_6_3.rb#L23 

add_step(Checks:::UnreadMail.new)

kgaikwad: @ntkathole ,
It's not intended to be a pre/post upgrade checks. That's why tagged as `unread_mail` so that any user can run it explicitly whenever required.

definitions/checks/unread_mail.rb
```diff
@@ -0,0 +1,46 @@
...
+      mcount = mails_count
+      assert(mcount == 0, "WARNING: Found #{mcount} mail(s) in root's mailbox.",
+             :next_steps => [Procedures::SystemMailbox::List.new(:file_path => file_path),
+                             Procedures::SystemMailbox::Clear.new(:file_path => file_path)],
+             :error_type => ForemanMaintain::Error::Warn)
+    end
+  end
+
+  private
+
```
mbacovsky: @kgaikwad, would it make sense to make `file_path` and `mailbox_dir` lazy so that they are evaluated during the `run` instead of initialization? I see two advantages there:
 - it takes time during scenario execution when there is already something printed out and scenario can start quicker
 - if (hypothetically) if file_path is subject of change in some of the previous steps we won't notice

kgaikwad: Yes, make sense! And advantages mentioned are legit.
I will do modifications as suggested.

kgaikwad:
@mbacovsky, @ntkathole,

ntkathole:
Looks good!

sean797:
Whats the motivation behind this? I did ask already but it seems the commit message was updated telling me what this commit does (which is useful :+1: ) but the motivation behind why we are doing it would be nice.

Right now, I'm thinking why the hell would we want to warn users about unread mails in root's mailbox!!?

from `man cron` "When executing commands, any output is mailed to the owner" this means foreman-maintain is going to warn users that they have cronjobs ran that produced output, I'm wondering why that is useful?

kgaikwad:
@sean797,
Initially, this check was suggested by @lzap so he can provide better insights.

lzap:
I think I was brainstorming but the original ticket is no longer accessible probably because folks disabled issues in this repo - don't know the details. I guess it won't hurt to warn user, but the amount of code written to do one-line check is just worrisome :-) If anyone told me in advance we are gonna deliver 100 lines of code for this, I'd say: please no, just close the idea.

upadhyeammit:
As per the recent comment of @lzap and @sean797 I am closing this pr.

PR #57, opened 2017-05-04
Preparation steps - setups steps before the main scenario runs

iNecas:
This PR introduces the following:

1. possibility to define preparation steps for checks: it allows installing additional dependencies or asking for hammer credentials. If multiple checks require the same preparation step, we ensure to run it only once
2. installing hdparm and fio when disk io check is run
3. setting up hammer credentials: we now use custom hammer config file, we first look at the default location and search for stored credentials, and ask the user for the credentials if we don't save them. At the end, we save the credentials into custom file (config/foreman-maintain-hammer.yml)
4. definition of procedure params, currently it's used just to streamline the procedure initialization, but later, it would be reused when running the procedures directly (it was needed to add to be able to unify
unique preparation steps)
5. added --assumeyes (-y) option to minimize the required interactions


iNecas:
I've also added the --assumeyes option

swapab:
Looks good to me.

kgaikwad:
I have reviewed code. Looks good!! 
I would like to check --assumeyes option on my setup.

kgaikwad:
Done.

iNecas:
Thanks. Merging now

PR #60, opened 2017-05-22
Fixes #19564 define order for checks

swapab:
- [x] Implement ruby `TSort` for `Checks`
- [x] Introduce new metadata `before` and `after`
- [x] Set initial order for checks.
- [x] Caching mechanism for `label => class`
- [x] Add test cases for same.

### DIFF
--- a/definitions/checks/foreman_tasks/not_paused.rb
+++ b/definitions/checks/foreman_tasks/not_paused.rb
@@ -5,7 +5,6 @@ module Checks::ForemanTasks
       for_feature :foreman_tasks
       description 'check for paused tasks'
       tags :default
-      after :disk_io, :verify_self_registration
     end
 
     def run
--- a/definitions/checks/foreman_tasks/not_paused.rb
+++ b/definitions/checks/foreman_tasks/not_paused.rb
@@ -5,6 +5,7 @@ module Checks::ForemanTasks
       for_feature :foreman_tasks
       description 'check for paused tasks'
       tags :default
+      after :disk_io, :verify_self_registration
     end
 
     def run
--- a/lib/foreman_maintain/dependency_graph.rb
+++ b/lib/foreman_maintain/dependency_graph.rb
@@ -4,7 +4,7 @@ module ForemanMaintain
   class DependencyGraph
     include TSort
 
-    attr_reader :graph, :collection
+    attr_reader :graph, :collection, :labels
 
     def self.sort(collection)
       new(collection).tsort.map(&:ensure_instance)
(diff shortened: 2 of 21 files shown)


swapab:
@iNecas @kgaikwad PR ready for review.

definitions/checks/foreman_tasks/not_paused.rb
```diff
@@ -5,6 +5,7 @@ class NotPaused < ForemanMaintain::Check
       for_feature :foreman_tasks
       description 'check for paused tasks'
       tags :default
+      after :disk_io, :verify_self_registration
```
iNecas: Why the dependency on this checks? I would recommend to reduce the amount of rules to just those that are justified by the behaviour (such as check on running tasks happening after resuming the paused ones or invalidating the tasks before resuming)

lib/foreman_maintain/dependency_graph.rb
```diff
@@ -0,0 +1,76 @@
+require 'tsort'
+
+module ForemanMaintain
+  class DependencyGraph
+    include TSort
+
+    attr_reader :graph, :collection
+
+    def self.sort(collection)
+      new(collection).tsort.map(&:ensure_instance)
```
iNecas: If I read this properly, it would mean we would loose the original objects and re-create them again. This goes against
the intention of being able to add parametrized steps in the scenarios. Think of this example:

```ruby
# in scenario
add_step(Procedures::Packages::Install.new(:packages => 'hdparm'))
```

After the sorting, we would loose the `:packages` parameter

lib/foreman_maintain/object_cache.rb
```diff
@@ -0,0 +1,34 @@
+require 'singleton'
+
+module ForemanMaintain
+  class ObjectCache
+    include Singleton
```
iNecas: Should the ObjectCache be singleton? Different `DependencyGarphs` could have different collections of steps, and having a cache, that would interfere with those might be not desired

swapab: But every run would have only one set of checks. `Procedure` are in pipeline and would come eventually, but not considered here.

definitions/checks/foreman_tasks/not_paused.rb
swapab: I thought of keeping the checks for paused and running tasks together. These would collectively run before task-cleanup checks.
This was just to keep it understandable while execution.
`before :foreman_tasks_not_running` would do? or should get rid of it altogether?

lib/foreman_maintain/dependency_graph.rb
swapab: [Update] TSort operations would be done on objects as received.

iNecas:
Nice! Tested and works well. Thanks @swapab and sorry for the delay on review

PR #61, opened 2017-05-22
Fixes #59 - missing error for hammer-config file

### LINKED ISSUE #59: Cannot setup foreman_maintain, complains about foreman-maintain-hammer.yml not found
Looking at the https://github.com/iNecas/foreman_maintain/tree/master/config , I can see just two config files  foreman_maintain.yml.example,hammer.yml.example.
When i run the foreman-maintain have the setting configured
```

./foreman-maintain health check
 
Running checks with tags [default]
check for paused tasks:                                               [FAIL]
There are currently 2 paused tasks in the system
1) resume paused tasks
2) investigate the tasks via UI
Select step to continue, [n(next), q(quit)] 1
resume paused tasks:                                                  [OK]
Error: Custom configuration file /root/foreman_maintain/config/foreman-maintain-hammer.yml does not exist.
check for paused tasks:                                               [FAIL]
There are currently 2 paused tasks in the system
```

Am I missing something?

kgaikwad:
As we already have foreman-tasks feature, moved hammer code from resume procedure to feature file.
Which will ask hammer credentials instead of showing error message for missing config file.

### DIFF
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -82,6 +82,10 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     end
   end
 
+  def resume_task_using_hammer
+    hammer('task resume')
+  end
+
   private
 
   def backup_dir(state)
--- a/definitions/checks/foreman_tasks/not_paused.rb
+++ b/definitions/checks/foreman_tasks/not_paused.rb
@@ -1,5 +1,6 @@
 module Checks::ForemanTasks
   class NotPaused < ForemanMaintain::Check
+    include ForemanMaintain::Concerns::Hammer
     metadata do
       for_feature :foreman_tasks
       description 'check for paused tasks'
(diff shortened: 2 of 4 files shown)


definitions/features/foreman_tasks.rb
```diff
@@ -82,6 +82,10 @@ def condition(state)
     end
   end
 
+  def resume_task_using_hammer
+    hammer('task resume')
```
iNecas: Don't we need to include hammer module here as well?

kgaikwad: @iNecas
Hammer module is already included in `lib/foreman_maintain/feature.rb`. That's why we don't need to include hammer module here.

iNecas: Oh, you're right. Thanks for explanation.

iNecas:
Thanks @kgaikwad

PR #67, opened 2017-05-29
Full upgrade scenario

iNecas:
This commit intoroduces more structured upgrade scenario consisting of
this scenarios:

  * pre_upgrade_check - non-modifying checks: is the system ready?
  * pre_migrations - modifying procedures before source update
  * migrations - source update and main migrations
  * post_migrations - additional steps after migration
  * post_upgrade_check - non-modifying checks: was the upgrade successful

If there were some failures or warnings in the checks, the reporter
informs the user about it and asks for confirmation.

Things to finish:

  - [x] - `--whitelist` option
  - [x] - full upgrade experience: move some checks to pre/post migrations + add setting of repositories, yum update and insaller run
  - [ ] - add status command
  - [ ] - extract target version and handle situation when some upgrade is already in progress
  - [x] - use `run --phase` instead of `advanced run --phase`

PR #68, opened 2017-05-30
Fixes #41 - Add capsule clear

### LINKED ISSUE #41: Command: Wipe capsule and start over
Pulp does not handle broken repositories well, it just reports errors like "Will not create a symlink to a non-existent source" but a repo will never start working again. We need a command that will wipe all capsule package data (and mongo) and starts again:

https://bugzilla.redhat.com/show_bug.cgi?id=1276911#c7

chris1984:


chris1984:
@iNecas other than the style errors since this is my first pr, is there other stuff that is needed I assume?

iNecas:
I would have some suggestions for refactorings, also related to other procedures, that need similar helpers. For example, stopping/strating services assumes that the services are running at the beginning. In case they were not (such as as part of upgrade scenario), we would unintentionally change the state of the system. I will come to it later this week to suggest some steps for improvement. Stay tuned :)

chris1984:
Sounds good ty.

chris1984:
@iNecas When you get a chance, can you do another review of this for the refactor etc.

iNecas:
Sry for the delays @chris1984, so where we are right now is https://github.com/theforeman/foreman_maintain/pull/52 contains a feature that would be very beneficial here (`feature(:katello_service).make_stop() {}`), that handles correct stopping and staring of the services.
It's waiting on https://github.com/theforeman/foreman_maintain/pull/52/files#r123739959 + https://github.com/theforeman/foreman_maintain/pull/67. Once those are in, we can refactor this one to use those features. I'm sorry it takes a bit more time, but that happens when too many things happen at once. Patience please :)

chris1984:
@iNecas hey was this still on the roadmap for foreman-maintain or can I close it out and reopen it at a different time? I have ocd with open things that I cant move forward.

iNecas:
The required change was merged in https://github.com/theforeman/foreman_maintain/pull/83 that makes it easier to stop and start the services.

Please have a look at https://github.com/theforeman/foreman_maintain/pull/52/files#diff-d960f2313918737e5fdda6323248d300R1 and update the procedure to:

1. define `confine` to be applicable only when appropriate. Would this work for the main server or this is capsule only? You will probably need first to add `features/pulp` and use that for checking on when the procedure is appropriate

2. use the `make_stop` method with block https://github.com/theforeman/foreman_maintain/pull/52/files#diff-d960f2313918737e5fdda6323248d300R9 so that you don't need to stop/start on your own

3. we moved to redmine in the meantime, please link it to the corresponding redmine issue: I suppose you also have a BZ for it?

4. another question is how safe this procedure is? Wouldn't this make my capsule unusable? Is there a KCS this would help with?

PR #69, opened 2017-05-31
Fixes #19715 - [Check] WARN if foreman-proxy is self-registered

swapab:
Raise WARNING for smart-proxy if self-registered.
REF: https://github.com/theforeman/foreman_maintain/issues/23#issuecomment-300745839

### DIFF
--- a/definitions/checks/system_registration.rb
+++ b/definitions/checks/system_registration.rb
@@ -5,20 +5,16 @@ class Checks::SystemRegistration < ForemanMaintain::Check
     tags :default
 
     confine do
-      file_exists?('/etc/rhsm/rhsm.conf')
+      file_exists?('/etc/rhsm/rhsm.conf') &&
+        !feature(:foreman_server) &&
+        feature(:foreman_proxy)
     end
   end
 
   def run
-    if system_is_self_registerd?
+    if rhsm_hostname_eql_hostname?
       warn! 'System is self registered'
-    else
-      puts 'System is not self registered'
     end
-  end
-
-  def system_is_self_registerd?
-    rhsm_hostname.casecmp(hostname).zero?
   end
 
   def rhsm_hostname
@@ -27,5 +23,9 @@ class Checks::SystemRegistration < ForemanMaintain::Check
 
   def rhsm_conf_file
     '/etc/rhsm/rhsm.conf'
+  end
+
+  def rhsm_hostname_eql_hostname?
+    @result ||= rhsm_hostname.casecmp(hostname).zero?
   end
 end
--- a/definitions/features/foreman_proxy.rb
+++ b/definitions/features/foreman_proxy.rb
@@ -1,6 +1,9 @@
 class Features::ForemanProxy < ForemanMaintain::Feature
   metadata do
     label :foreman_proxy
+    confine do
+      find_package('foreman-proxy')
+    end
   end
 
   attr_reader :dhcpd_conf_file, :cert_path
@@ -36,7 +39,7 @@ class Features::ForemanProxy < ForemanMaintain::Feature
     array_output = curl_resp.split(/\r\n/)
     result_array = array_output.last.split(/\n/)
     curl_http_status = result_array.delete_at(result_array.length - 1).strip.to_i
-    curl_http_resp = json_parse(result_array.join(''))
+    curl_http_resp = parse_json(result_array.join(''))
     ForemanMaintain::Utils::CurlResponse.new(
       curl_http_resp,
       curl_http_status,
--- a/definitions/features/katello.rb
+++ b/definitions/features/katello.rb
@@ -1,0 +1,17 @@
+class Features::Katello < ForemanMaintain::Feature
+  metadata do
+    label :katello
+
+    confine do
+      find_package('katello')
+    end
+  end
+
+  def data_dirs
+    @dirs ||= ['/var/lib/pulp', '/var/lib/mongodb', '/var/lib/pgsql']
+  end
+
+  def current_version
+    @current_version ||= rpm_version('katello')
+  end
+end
--- a/lib/foreman_maintain/concerns/system_helpers.rb
+++ b/lib/foreman_maintain/concerns/system_helpers.rb
@@ -26,30 +26,6 @@ module ForemanMaintain
         end
       end
 
-      def hostname
-        execute('hostname -f')
-      end
-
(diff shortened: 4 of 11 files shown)


swapab:
I have piggybacked `katello` feature. Which would confine disk_io check to katello's presence.

definitions/features/katello.rb
```diff
@@ -0,0 +1,17 @@
+class Features::Katello < ForemanMaintain::Feature
+  metadata do
+    label :katello
+  end
+
+  def dirs
```
iNecas: The `dirs` is too generic. Can you thing of something more specific? `data_dirs` perhaps?

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -89,9 +50,60 @@ def execute(command, options = {})
...
+
+      def install_packages(packages, options = {})
+        options.validate_options!(:assumeyes)
+        yum_options = []
+        yum_options << '-y' if options[:assumeyes]
+        execute!("yum #{yum_options.join(' ')} install #{packages.join(' ')}", :interactive => true)
+      end
+
+      def server?
+        find_package('foreman')
```
iNecas: shouldn't this be `foreman` and `foreman-proxy` features?

definitions/features/katello.rb
swapab: Yes makes sense. Done :+1:

lib/foreman_maintain/concerns/system_helpers.rb
swapab: Absolutely!
Added new features. `foreman-server` and `foreman-proxy` to be more specific and `label :foreman` was already taken.

definitions/checks/system_registration.rb
```diff
@@ -10,15 +10,23 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
 
-  def system_is_self_registerd?
-    rhsm_hostname.casecmp(hostname).zero?
+  def system_self_registerd?
+    @result ||= rhsm_hostname.casecmp(hostname).zero?
+  end
+
+  def warn?
+    if feature(:foreman_server).present?
+      !system_self_registerd?
```
iNecas: If I read it properly, we would warn if run on main server and the system was not self-registered. Despite the fact, that the error message would be `System is self registered`, is this what we actually want?

swapab: Per Lukas's comment, a warning should be raised if the system is a **smart-proxy and self-registered**.
https://github.com/theforeman/foreman_maintain/issues/23#issuecomment-300745839

swapab: I have updated the check `Checks::SystemRegistration` to check self-registration if **only smart-proxy present** and foreman-server is absent

swapab:
@iNecas I have updated the PR with following changes:
1. Confine `feature(:foreman_server)` with package presence.
2. Confine `check(:verify_self_registration)` with condition `Checks::SystemRegistration#system_self_registered?`

definitions/features/foreman_proxy.rb
```diff
@@ -5,6 +5,10 @@ class Features::ForemanProxy < ForemanMaintain::Feature
 
   attr_reader :dhcpd_conf_file, :cert_path
 
+  def present?
+    !feature(:foreman_server).present? && find_package('foreman-proxy')
```
iNecas: You can have foreman_proxy feature present also at the same machine as the foreman. This condition probably belongs to the confine block of the self-registered check itself

definitions/checks/system_registration.rb
```diff
@@ -5,27 +5,32 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
+    def rhsm_hostname_eql_hostname?
+      @result ||= rhsm_hostname.casecmp(hostname).zero?
+    end
 
-  def rhsm_conf_file
-    '/etc/rhsm/rhsm.conf'
+    def system_self_registered?
+      file_exists?('/etc/rhsm/rhsm.conf') &&
+        !feature(:foreman_server) &&
+        feature(:foreman_proxy).present? &&
```
iNecas: Similarly as the `:foreman_server` also the `:foreman_proxy` and `:katello` should rather use confine blocks than the `present?` method, to match with the other features

iNecas: Also, this condition should go to the confine block of this check

definitions/features/foreman_proxy.rb
swapab: Ah, right. That's correct `present?` is redundant here.

definitions/checks/system_registration.rb
swapab: You mean the whole condition of `system_self_registered` should belong to confine block? OR just
```
confine do
  file_exists?('/etc/rhsm/rhsm.conf') &&
    !feature(:foreman_server) &&
    feature(:foreman_proxy)
end
```

swapab:
I have updated the checks accordingly but, the tests have become pretty slow now.
I think it's because of the package presence checks in the respective confine blocks.

```
Finished in 55.495118s, 0.5946 runs/s, 1.3154 assertions/s.

33 runs, 73 assertions, 0 failures, 0 errors, 0 skips
```
It used to execute in ~20 seconds before. Need to find the slow tests and stub confine blocks for them.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -26,16 +26,66 @@ def build
...
       end
 
       def file_exists?(filename)
         File.exist?(filename)
       end
 
-      def version(value)
-        Version.new(value)
+      def find_package(name)
+        execute(%(rpm -qa '#{name}'))
```
iNecas: the `qa` variant is too slow. `rpm -q ` should be used instead. Also make sure it returns nil/false if the pacakge is not found: currently it returns '' instead: therefore the feature looks like it's present even though it isn't

swapab:
Requested changes were done. Failing tests fixed.

lib/foreman_maintain/concerns/system_helpers.rb
swapab: Tests running in `< 6 seconds` now.
Thanks! @iNecas

iNecas:
Tested and works well now. Thanks @swapab

PR #70, opened 2017-06-01
Fixes #27945 - modify description of --tags option

### LINKED ISSUE #65: allow to enter multiple tags for check filtering
`foreman-maintain health check --help`  command gives two options: 

 ```
--label label                 Limit only for a specific label. (Use "list" command to see available labels)
 --tags tags                   Limit only for specific set of labels. (Use list-tags command to see available 
```
Lets say, `tags = [default, pre-upgrade, post-upgrade] `
I have tried below commands to apply `default` & `pre-upgrade` tags for filtering checks which is not working.

`foreman-maintain health list --tags default --tags pre-upgrade` -> it is overriding default tag value with pre-upgrade.
`foreman-maintain health list --tags default,pre-upgrade` -> no any output
`foreman-maintain health list --tags default, pre-upgrade` -> ERROR too many arguments.

Is there any way to apply multiple tags to filter checks

### PR DESCRIPTION
**Pull request by kgaikwad**
With this commit, it will allow user to enter multiple values separated with commas for tags filter.

### DIFF
--- a/lib/foreman_maintain/cli/base.rb
+++ b/lib/foreman_maintain/cli/base.rb
@@ -135,7 +135,7 @@ module ForemanMaintain
 
       def self.label_option
         option '--label', 'label',
-               'Limit only for a specific label. ' \
+               'Run only a specific check with a label. ' \
                  '(Use "list" command to see available labels)' do |label|
           raise ArgumentError, 'value not specified' if label.nil? || label.empty?
           underscorize(label).to_sym
@@ -144,7 +144,7 @@ module ForemanMaintain
 
       def self.tags_option
         option('--tags', 'tags',
-               'Limit only for specific set of labels. ' \
+               'Run only those with all specific set of tags. ' \
                  '(Use list-tags command to see available tags)',
                :multivalued => true) do |tags|
           raise ArgumentError, 'value not specified' if tags.nil? || tags.empty?
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ Subcommands:
       list                          List the checks based on criteria
       list-tags                     List the tags to use for filtering checks
       check                         Run the health checks against the system
-        --label label               Run only a specific check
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by swapab** [comment]
@kgaikwad 
1. Can you please also update the help description with `Separate multiple tags with comma` and an example probably?
Also the current one is incorrectly saying it as `labels` which should be `tags`
```
--tags tags                   Limit only for specific set of labels. (Use list-tags command to see available tags)
```

2. A command line test would be a bonus.

Apart from that, it looks good to me.

**Comment by iNecas** [comment]
I'm not sure about this change: originally, the behavior was to do the selection using the `and` condition, in order to be able to filter checks for a specific version and specific phase. After this change, it would not be possible to do easily. If people want the `or` functionality, they could just run the commands one after the other.

**Comment by swapab** [comment]
It actually feels confusing, when you use comma separated values for any options and expect the results be a union of those values.
It happened with me while demoing `foreman-maintain` to a peer. :smile: 

May be we can do one or more of following:
1. remove the multiple tags functionality
2. Add more description for `--tags` regarding `AND`ing of the tags
3. Wait for it to reach a larger crowd and act based on their feedback.

**Comment by iNecas** [comment]
I'm for (2) for now.

File: lib/foreman_maintain/cli/base.rb [hunk 1]
```diff
@@ -144,7 +144,7 @@ def self.label_option
 
       def self.tags_option
         option('--tags', 'tags',
-               'Limit only for specific set of labels. ' \
+               'Limit only where all specific set of tags exist. ' \
```
mbacovsky: I still had to read the other BZ to understand the description.
Wouldn't be `Run only checks with all these tags` more explanatory?

To be consistent we could change the description for `--label` to 'Run only check with this label'

@kgaikwad, what do you think?

kgaikwad: @mbacovsky,  modified description for tags & label.

PR #72, opened 2017-06-13
Newly added config constants for packaging.

kgaikwad:
Sorry for confusion as [log file size](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain.rb) is already set to `10mb`. 
So I didn't make any changes to it.

I forgot that 3rd argument of logger.new is in bytes not in KB. 
` Logger.new('foo.log', 10, 1024000)` -> 3rd argument is in bytes.

### DIFF
--- a/bin/foreman-maintain
+++ b/bin/foreman-maintain
@@ -4,6 +4,8 @@ $LOAD_PATH.unshift(File.expand_path('../../lib', __FILE__))
 
 require 'foreman_maintain'
 
+CONFIG_FILE = '/etc/foreman-maintain/config/foreman_maintain.yml'.freeze
+
 ForemanMaintain.setup
 
 require 'foreman_maintain/cli'
--- a/config/foreman_maintain.yml.packaging
+++ b/config/foreman_maintain.yml.packaging
@@ -1,15 +1,16 @@
 # Directory where the logs are stored.
 # The default file is named as /log/foreman_maintain.log
-:log_dir: 'log'
+:log_dir: '/var/log/foreman-maintain/'
 
 # Logger levels: mention one of debug, info, warning, error, fatal
-:log_level: 'error'
+# :log_level: 'error'
 
 # Mention definitions directories. Default
 # :definitions_dirs:
 
 # Mention file path to store data
-# :storage_file: 'data.yml'
+:storage_file: '/lib/foreman-maintain/data.yml'
 
 # Mention directory to store whole backup data
-# :backup_dir: '/lib/foreman-maintain'
+:backup_dir: '/lib/foreman-maintain'
+
--- a/test/definitions/test_helper.rb
+++ b/test/definitions/test_helper.rb
@@ -84,6 +84,6 @@ module DefinitionsTestHelper
 end
 
 TEST_DIR = File.dirname(__FILE__)
-ForemanMaintain.setup(
-  :config_file => File.join(TEST_DIR, 'config/foreman_maintain.yml.test')
-)
+CONFIG_FILE = File.join(TEST_DIR, 'config/foreman_maintain.yml.test').freeze
+
+ForemanMaintain.setup
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -88,9 +88,13 @@ class Features::ForemanTasks < ForemanMaintain::Feature
 
   private
 
+  def parent_backup_dir
+    File.expand_path(ForemanMaintain.config.backup_dir)
+  end
+
   def backup_dir(state)
     @backup_dir ||=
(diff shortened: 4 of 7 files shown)


test/definitions/test_helper.rb
```diff
@@ -84,6 +84,9 @@ def refute_scenario(filter)
 end
 
 TEST_DIR = File.dirname(__FILE__)
-ForemanMaintain.setup(
-  :config_file => File.join(TEST_DIR, 'config/foreman_maintain.yml.test')
-)
+CONFIG_FILE = File.join(TEST_DIR, 'config/foreman_maintain.yml.test').freeze
+STORAGE_DATA_FILE = '/lib/foreman-maintain/data_test.yml'.freeze
```
kgaikwad: I declared constants twice for tests.

I have tried to move these constants into separate file under test folder & that file I have required into `test/definitions/test_helper.rb`  & `test/lib/test_helper.rb`
which results in test case failures. 
Please let me know if any other way around.

swapab: That's strange. `require` on a constants file for example `foreman_maintain/constants.rb` should ideally work. But the tests are failing, looks like it's not able to recognize definitions under test folder.
Few of the errors are:
```
NameError: uninitialized constant Scenarios::PresentUpgrade
NameError: uninitialized constant Procedures::Setup
NameError: uninitialized constant Checks::PresentServiceIsRunning
```
And failures:
```
 18) Failure:
ForemanMaintain::Cli::HealthCommand::list-tags#test_0001_lists the defined tags [/root/foreman_maintain/test/lib/cli/health_command_test.rb:50]:
--- expected
+++ actual
@@ -1,3 +1,4 @@
 "[default]
-[pre-upgrade-check]
+[post-upgrade]
+[pre-upgrade]
```

kgaikwad: yes @swapab . These are test failures that I was getting when I shifted code into separate file.

swapab: I'll think about it once again. Maybe @iNecas might have a clue about what's going wrong here.

bin/foreman-maintain
```diff
@@ -4,6 +4,11 @@ $LOAD_PATH.unshift(File.expand_path('../../lib', __FILE__))
 
 require 'foreman_maintain'
 
+CONFIG_FILE = '/etc/foreman-maintain/config/foreman_maintain.yml'.freeze
+STORAGE_DATA_FILE = '/lib/foreman-maintain/data.yml'.freeze
```
iNecas: Isn't needed just the CONFIG_FILE, and we could have different default config files for git and pacakge-based distribution?

test/definitions/test_helper.rb
iNecas: I think having them serparately in `test_helper` and the executable is acceptable, especially when we eliminate the amount of the constants to one, and the rest would be configured in settings.

bin/foreman-maintain
kgaikwad: @iNecas,
Yes, we have config files for git. 
Did you mean we need to declare separate config files for packaging like test environment, right?
For example, foreman_maintain.yml.packaging
Please correct me.

iNecas: yes, that was the suggestion, but I'm open to other ideas

kgaikwad: @iNecas , 
I will go with `foreman_maintain.yml.packaging` option. 
Previously I thought that user have to create foreman_maintain.yml from .yml.example file that's why I added these constants additionally. 
I will modify code and update the PR for review.

kgaikwad:
@iNecas, @swapab & @gnurag, 

Sorry for delay in updating PR as facing issues on my laptop. 
Using temporary laptop I was able to update the code as per review comments.
Removed constants except CONFIG_FILE and defined a new config file for packaging.

Ready for review. Please let me know if any further changes needed in PR.

config/foreman_maintain.yml.packaging
```diff
@@ -0,0 +1,16 @@
+# Directory where the logs are stored.
+# The default file is named as /log/foreman_maintain.log
+:log_dir: '/var/log/foreman-maintain/'
+
+# Logger levels: mention one of debug, info, warning, error, fatal
+:log_level: 'error'
```
iNecas: I would suggest keeping it on debug for several first versions at users: it has a lot of useful information for us

iNecas: Or perhaps just  leave it commented out so that we can control the default log level from the code, and set when it's ok to switch from debug to error by default.

iNecas:
Tested and works well. One request about the log level, but otherwise good to go

kgaikwad:
@iNecas , 
Done with changes. I have commented config line for `:log_level`.

iNecas:
Thanks @kgaikwad, merging now

PR #74, opened 2017-07-04
Fixed - Config related to packaging & updated version.

kgaikwad:
1. Required absolute path instead of relative path for default CONFIG_FILE.
2. Changed data directory to /var/lib instead of /lib
3. Bump version to 0.0.5

PR #76, opened 2017-07-05
Fixed - Packaging config changes

kgaikwad:
1. Required absolute path instead of relative path for default CONFIG_FILE.
2. Changed data directory to /var/lib instead of /lib


kgaikwad:
Code changes are already reviewed in [PR-74](https://github.com/theforeman/foreman_maintain/pull/74).
Removed version updated code. I will add it separately.

PR #77, opened 2017-07-07
Fixes #20267 - Report Command

swapab:
```
Subcommands:
    disk-usage                    Report disk consumption and availability of directories
    table-counts                  Get table counts
    table-sizes                   Get table sizes
    all                           Generate all reports
```

- [x] All commands working
- [x] Namespace `DiskSpeedMininal` check as `Disk::Performance`
- [x] Functional `json` and yaml output.
       `-o, --output [json|plain-text|yaml] Specify output format. (default: "plain-text")`
- [x] Test cases
- [x] Params validations (http://projects.theforeman.org/issues/20729)
- [x] Merge Procedure-CLI

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,11 @@ Subcommands:
       check                         Run the health checks against the system
         --label label               Run only a specific check
         --tags tags                 Limit only for specific set of tags
-
+    report                        Generate reports
+      disk-usage                    Report disk consumption and availability of directories
+      table-counts                  Get table counts
+      table-sizes                   Get table sizes
+      all                           Generate all reports
     upgrade                       Upgrade related commands
       list-versions                 List versions this system is upgradable to
       check --target-version TARGET_VERSION   Run pre-upgrade checks for upgrading to specified version
--- a/definitions/features/foreman_database.rb
+++ b/definitions/features/foreman_database.rb
@@ -7,7 +7,7 @@ class Features::ForemanDatabase < ForemanMaintain::Feature
     label :foreman_database
 
     confine do
-      file_exists?(FOREMAN_DB_CONFIG)
+      file_exist?(FOREMAN_DB_CONFIG)
     end
   end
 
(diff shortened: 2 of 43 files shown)


swapab:
@iNecas @kgaikwad This PR is ready for review.

definitions/features/foreman_database.rb
```diff
@@ -7,6 +7,12 @@ class Features::ForemanDatabase < ForemanMaintain::Feature
     end
   end
 
+  def count(table, condition = '1=1')
+    query("SELECT count(*) AS count FROM #{table} WHERE #{condition}").first['count'].to_i
+  rescue CSV::MalformedCSVError
```
kgaikwad: would it be good to add logger statement here in rescue block?

README.md
```diff
@@ -15,7 +15,11 @@ Subcommands:
       check                         Run the health checks against the system
         --label label               Run only a specific check
         --tags tags                 Limit only for specific set of tags
-
+    report                        Generate reports
+      disk-usage                    Report disk consumption and availability of directories
+      table-counts                  Get table counts
+      table-sizes                   Get table sizes
+      all                           Generate all re
```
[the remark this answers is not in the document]
swapab: @kgaikwad Thanks for the review. Changes were done.

definitions/features/foreman_database.rb
swapab: Yes, makes sense.

lib/foreman_maintain/cli/system_report.rb
```diff
@@ -0,0 +1,62 @@
+module ForemanMaintain
+  module Cli
+    class SystemReport < Base
+      class << self
+        def limit_table_size_option
+          option ['-s', '--size'], '[>=1mb|<=1MB]',
+                 'Show tables with size specified operator(>= or <= or > or <).'\
+                 'Defaults to >=1MB.' do |size|
+            size = Report::Params::Size.new(size)
+            size.validate!
```
kgaikwad: @swapab 
I have thought that instead of calling `validate!` method separately here why not to put it in `initialize` method only.

For example - 

```
def initialize(size)
     @size = disintegrate_size(size)
     validate!
end
```

lib/foreman_maintain/cli/system_report.rb
```diff
@@ -0,0 +1,62 @@
...
+            size.to_params
+          end
+        end
+
+        def output_format_option
+          option ['-o', '--output'], '[json|plain-text|yaml]',
+                 'Specify output format.',
+                 :default => 'plain-text' do |output|
+            output = Report::Params::Output.new(output)
+            output.validate!
```
kgaikwad: ditto

lib/foreman_maintain/cli/system_report.rb
swapab: That's a good point. But I feel `Object#initialize` method should only care about initializing objects.  Public methods should be called as needed.

Also, in this case, it provides a good readability.

swapab:
Rebased.

kgaikwad:
LGTM

swapab:
@iNecas Per Postgres docs https://wiki.postgresql.org/wiki/Disk_Usage function `pg_indexes_size` is only available for version `>= 9.0`.

Shall we keep the `report table-sizes` command confined to above postgres version?

iNecas:
Yes please

swapab:
Confined `report table-sizes` for postgres version `>= 9.0`
The build is passing. PR ready for re-review.

ntkathole:
@swapab I am trying to list report according to table-sizes <=1mb, it gives 

```
# ./bin/foreman-maintain report table-sizes -s <=1MB
undefined method `gsub' for nil:NilClass
```

swapab:
@ntkathole I think bash is ignoring the last parameter(<=1MB) because of special character(<=).
Using quotes `'<=1MB'` should work fine. I have updated the help document accordingly.
```
-s, --size ['>=1mb'|'<=1MB']  Show tables with size specified operator(>= or <= or > or <). Defaults to '>=1MB'
```

ntkathole:
@swapab Yes, it works for me. But can we handle nil value passed to `report table-sizes -s` (just for usability of tool)?
`raise Error::UsageError, "--size not specified"`

kgaikwad:
Closing in favor of #173

PR #78, opened 2017-07-11
Fixes #19496 - added passenger recycler

lzap:
Extra script to recycle passenger. Read the YAML config which explains most of it.

In the first phase we want to have the config file in /etc and cron file in /etc/cron.d and this will do the work. I can submit packaging changes if necessary.

### DIFF
--- a/bin/passenger-recycler
+++ b/bin/passenger-recycler
@@ -1,0 +1,68 @@
+#!/usr/bin/env ruby
+#
+# Trivial Passenger memory monitor and recycler. See the configuration file
+# /etc/passenger-recycler.yaml for options. Execute via SCL.
+#
+require "yaml"
+
+$conf = {}
+CONFIG = '/etc/passenger-recycler.yaml'
+$conf = YAML.load_file(CONFIG) if File.readable?(CONFIG)
+exit 0 unless $conf[:ENABLED]
+
+def running?(pid)
+  return Process.getpgid(pid) != -1
+rescue Errno::ESRCH
+  return false
+end
+
+def debug msg
+  puts(msg) if $conf[:DEBUG]
+end
+
+def verbose msg
+  puts(msg) if $conf[:VERBOSE]
+end
+
+require 'phusion_passenger'
+require 'phusion_passenger/platform_info'
+require 'phusion_passenger/platform_info/ruby'
+require 'phusion_passenger/admin_tools/memory_stats'
+PhusionPassenger.locate_directories
+stats = PhusionPassenger::AdminTools::MemoryStats.new
+unless stats.platform_provides_private_dirty_rss_information?
+  puts "Please run as root or platform unsupported"
+  exit 1
+end
+killed = 0
+stats.passenger_processes.each do |p|
+  pid = p.pid.to_i
+  debug "Checking #{pid} with RSS of #{p.private_dirty_rss}"
+  if p.private_dirty_rss > $conf[:MAX_PRIV_RSS_MEMORY]
+    started = `ps -p#{pid} -o start=`.strip rescue '?'
+    status_ps = `ps -p#{pid} -u`
+    status_all = `passenger-status 2>/dev/null`
+    status_backtraces = `passenger-status --show=backtraces 2>/dev/null`
+    verbose "Terminating #{pid} (started #{started}) with private dirty RSS size of #{p.private_dirty_rss} MB"
+    Process.kill "SIGUSR1", pid
+    sleep $conf[:GRACEFUL_SHUTDOWN_SLEEP]
+    if running?(pid) && $conf[:KILL_BUSY]
+      verbose "Process #{pid} is still running, sending SIGKILL"
+      Process.kill "KILL", pid
+      sleep 5
+    end
+    if running?(pid)
+      verbose "Process #{pid} still terminating, moving on..."
+    else
+      verbose "Process successfully #{pid} terminated"
+    end
+    if $conf[:SEND_STATUS]
+      verbose status_ps
+      verbose status_all
+      verbose status_backtraces
+    end
+    killed += 1
+    exit(1) if killed >= $conf[:MAX_TERMINATION]
+  end
+end
+exit 0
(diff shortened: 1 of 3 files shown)


lzap:
To test execute with:

    scl enable tfm -- ruby passenger-recycler

Lower down the limit by deploying /etc/passenger-recycler.yaml to test how it works.

lzap:
I tested this myself on Satellite 6.3 and 6.2.

lzap:
Ping?

swapab:
Should this be a new command? Or can be a `health check` step?

I see 2 steps _todo_ here:
1. Add a cron job
2. Add new check to `healt check` command, that would do step 1 and also run the script explicitly once during the run.

@lzap can you please ellaborate on:

> Note this implementation is very simple, it could be improved (e.g. waits for termination, it could check the PID in a loop).

lzap:
Separate command for now, that's what I propose. There is high demand downstream to have this in foreman-maintain to workaround issues with memory. Long term, Ivan told me he would like to build some generic set of commands around this.

What you mean by cron job, it's present in the PR already.

swapab:
@lzap, sure a new command it is.
Considering the importance of the task the script performs, a user readable command would be good?
Example: `foreman-maintain free-memory`

The cron job (`config/passenger-recycler.cron`) needs be added to `/etc/crontab`?
Please correct me if I am wrong.

lzap:
I was thinking just adding the binary on the path (which ruby gems should do by default) so you can use it directly as "passenger-recycler". That is how I reference it in the cron job.

The script needs to be installed via package manager (RPM) into /etc/cron.hourly making sure it has executable permissions.

swapab:
I thought of creating a utility named 'recycle' (independent of app server - which can be changed later if required) WIP : https://github.com/swapab/foreman_maintain/commit/505835606caa4e20905d7837084ebc8903c690d7
This way it feels more modular, OO and readable.

IMHO we should keep commands under foreman-maintain.

---
I tried to run it on my sat-6.3 setup. Facing errors:

```
sat63 :: ~/foreman_maintain ‹19496-passenger-recycler*› » ./bin/passenger-recycler                 
/usr/local/rvm/rubies/ruby-2.4.0/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- phusion_passenger (LoadError)
	from /usr/local/rvm/rubies/ruby-2.4.0/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from ./bin/passenger-recycler:36:in `<main>'
```

* After installing `gem install passenger --no-ri --no-rdoc`

```
sat63 :: ~/foreman_maintain ‹19496-passenger-recycler› » ./bin/passenger-recycler                                              
/usr/local/rvm/rubies/ruby-2.4.0/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `require': cannot load such file -- /phusion_passenger/utils/tmpio (LoadError)
	from /usr/local/rvm/rubies/ruby-2.4.0/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `rescue in require'
	from /usr/local/rvm/rubies/ruby-2.4.0/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:40:in `require'
	from /usr/local/rvm/gems/ruby-2.4.0@foreman_maintain/gems/passenger-5.1.7/src/ruby_supportlib/phusion_passenger.rb:240:in `require_passenger_lib'
	from /usr/local/rvm/gems/ruby-2.4.0@foreman_maintain/gems/passenger-5.1.7/src/ruby_supportlib/phusion_passenger/platform_info.rb:26:in `<top (required)>'
	from /usr/local/rvm/rubies/ruby-2.4.0/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from /usr/local/rvm/rubies/ruby-2.4.0/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from ./bin/passenger-recycler:37:in `<main>'
```

Am I missing something?

lzap:
Yes, the problem with that approach is that foreman maintain CLI does not run in SCL, but we need the script to run in SCL in order to load passenger native libraries. I think separate script is better in this case.

lzap:
Rebased, cleaned codebase a bit and added MAX_TERMINATION to prevent from killing more than 4 processes at once. I really believe this should not be in maintain binary but a separate command. You need to execute the script via SCL:

    /usr/bin/scl enable tfm -- ruby /usr/bin/passenger-recycler

swapab:
@lzap 
Agreed. We will package it alongside foreman-maintain binary.

Can the script be named as `foreman-maintain-passenger-recycler`. This would be helpful in differentiating passenger shipped script and foreman-scripts.

lzap:
Hmmm that's a bit slow, any better ideas? Also the script recycles ALL passenger processes, so it's not foreman only. I kinda like it's name to be honest :-)

swapab:
The prefix(foreman-maintain) is just to avoid any confusion with users that the `script/command` comes with foreman-maintain and not the passenger.

Also, we will be using the script as a procedure under `foreman-maintain advanced` command.
There it would definitely be passenger-recycler.
Such as:
`./bin/foreman-maintain advanced procedure run passenger-recycler`

lzap:
I still don't like it, this would mean that I need to rename configuration files e.g. `/etc/crond.d/foreman-maintain-passenger-recycler` or `config/foreman-maintain-passenger-recycler.yaml` that sound terrible. Not matching the binary is confusing for users.

I really don't understand why you push so hard for including into `foreman-maintain` subcommand, this is a separate utility which will run from cron every few minutes, you don't want to boot complex programs from there.

@iNecas is really this the right place for the tool? I start to believe it is not. All I want to leverage is your RPM packaging, I have no interest in deep integrations with the tool for now. I understand there might be some use cases in the future, but why to bother doing it now just to get it done? That's time wasted.

swapab:
@lzap I think you missed my comment here https://github.com/theforeman/foreman_maintain/pull/89#issue-252220795

I have done some changes in the script to comply with Robocop and moved `passenger-recycler.cron` to `foreman_maintain/extras` directory.

Everything else is kept **as is**. Installation of the script will be taken care in packaging.

Would it be fine with you to carry out further discussion over #89 and close this PR?

lzap:
Sure.

PR #83, opened 2017-08-15
Fixes #20604 - katello-service feature to make services start/stop

kgaikwad:


swapab:
I think we already have a `katello` feature here https://github.com/theforeman/foreman_maintain/pull/69
I'll rebase the PR and update accordingly.

@kgaikwad You may have a look and see if it meets the purpose?

kgaikwad:
@swapab ,

I have already added `katello-service` feature in qpidd PR.
Could you please check PR [#52](https://github.com/theforeman/foreman_maintain/pull/52).

We have some queries related to PR #52 that's why I marked it as `On-Hold`.
To unblock other PRs, I have separated out the code into this new PR.

kgaikwad:
@swapab , 

One more thing I forgot to mention here is that `katello` & `katello-service` are two different things.
`katello` is plugin and `katello-service` comes under katello-packaging using which we can perform actions like `start/stop/status` for few services. So we can't mix these two.

swapab:
@kgaikwad Got it.

I think it would be sensible to add `katello-service` under `definition/services/`

iNecas:
Based on scrum call discussion, I'm ok with leaving this one as is and not introducing another namespace for now. Thanks @kgaikwad

PR #85, opened 2017-08-17
Fixes #20565 - Add license to the README.rst file

rahulbajaj0509:


swapab:
@rahulbajaj0509 
Thanks for noticing this.

It would be great if we tell people
 **How they can contribute?**[1]
 **Where to report issues?**[2]

Also, since we are updating the README with above changes, please update commit message and PR message accordingly.

[1] a new section would be helpful.
[2] Use Redmine link: http://projects.theforeman.org/projects/foreman-maintain/issues/new

iNecas:
Let's address the other comments in separate PR. I'm ok with pulling this small change as it is :)

iNecas:
Thanks @rahulbajaj0509

PR #91, opened 2017-08-29
Fixes #20798 - proper repo-id and VERSION from RPM

mccun934:


definitions/features/downstream.rb
```diff
@@ -34,14 +34,14 @@ def setup_repositories(version)
...
-     "rhel-#{rh_version.major}-rhscl-#{rh_version.major}-rpms",
+     "rhel-server-rhscl-#{rh_version.major}-rpms",
      "rhel-#{rh_version.major}-server-satellite-#{sat_version.major}.#{sat_version.minor}-rpms"]
   end
 
   def rh_version
     return @rh_version if defined? @rh_version
     release_package = execute!('rpm -qf /etc/redhat-release')
-    @rh_version = rpm_version(release_package, 'RELEASE')
+    @rh_version = rpm_version(release_package, 'VERSION')
```
iNecas: Interesting, I was testing this with rhel6: need to check if it was just my mistaker there is something different between this two

iNecas: So, from the testing on rhel6:

```
rpm -q --queryformat='%{VERSION}' $(rpm -qf /etc/redhat-release)
6Server
rpm -q --queryformat='%{RELEASE}' $(rpm -qf /etc/redhat-release)
6.9.0.4.el6
```

We probably need to match first the `release_package` against inclusion of ".el6" vs. ".el7" to decide what attribute to match against

definitions/features/downstream.rb
```diff
@@ -34,14 +34,14 @@ def setup_repositories(version)
   def rh_repos(sat_version)
     sat_version = version(sat_version)
     ["rhel-#{rh_version.major}-server-rpms",
-     "rhel-#{rh_version.major}-rhscl-#{rh_version.major}-rpms",
+     "rhel-server-rhscl-#{rh_version.major}-rpms",
```
iNecas: Good catch

iNecas:
Thakns @mccun934, we need to make the rhel6 vs. rhel7 a bit more sophisticated. Good to go after that

mccun934:
will re-test this on rhel6 to get it working on both cases

iNecas:
Thanks @mccun934, this looks good.

PR #93, opened 2017-09-05
Fixes #20848 - Fix error messages indentation

iNecas:
Also fixes #20847 - unify steps description

### DIFF
--- a/test/lib/support/definitions/checks/dummy.rb
+++ b/test/lib/support/definitions/checks/dummy.rb
@@ -19,6 +19,17 @@ module Checks::Dummy
     end
   end
 
+  class Fail2 < ForemanMaintain::Check
+    metadata do
+      label :dummy_check_fail2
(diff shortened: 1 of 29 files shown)


iNecas:
Fixed rubocop

test/lib/support/definitions/checks/dummy.rb
```diff
@@ -19,6 +19,17 @@ def run
     end
   end
 
+  class Fail2 < ForemanMaintain::Check
+    metadata do
+      label :dummy_check_fail2
+      description 'check that ends up with fail'
```
swapab: Can we follow the `uppercase` in tests as well?

iNecas:
Updated + made the test less dependent on the description in the future (by checking against label instead of description in LogReporter)

swapab:
This PR is good to merge.

PR #95, opened 2017-09-07
Fixes #20597 - store the current target version of upgrade

iNecas:
This allows to prevent running one upgrade before finishing another one.
It also fixes an issue, where after updating the packages, one would not
be able to continue with the current upgrade when something goes wrong.


iNecas:
Updated the storing of the target version when failure happens on pre-upgrade checks

swapab:
LGTM :100:

iNecas:
Thanks @swapab for review

PR #96, opened 2017-09-12
Fixes #20623 - Failed scenario should exit with code 1

swapab:


swapab:
@iNecas https://travis-ci.org/theforeman/foreman_maintain/jobs/274487488
This is failing on my local as well. Do you know what might be causing the tags to show twice?

lib/foreman_maintain/runner.rb
```diff
@@ -84,6 +86,14 @@ def storage
 
     private
 
+    def execute_scenario_steps(scenario)
+      scenario.before_scenarios.flatten.each { |before_scenario| run_scenario(before_scenario) }
+      yield
```
iNecas: I think we can also move `return if quit? # the before scenarios caused the stop of the execution` here and not use the `yield`, for readability purposes

iNecas:
Don't know. Please try to find the cause: we can discuss on status call today.

lib/foreman_maintain/runner.rb
swapab: Yes, I thought returning from extracted-method might cause some unexpected behaviour.

swapab:
It was `require 'checks/dummy'` in `scenarios/dummy.rb`. :smile:
Removed.

swapab:
Added `exit!` to exit with code `1` when encountered with `UsageError`.

lib/foreman_maintain/cli.rb
```diff
@@ -19,19 +19,42 @@ def run(*arguments)
...
+        end
+
+        puts error.message
+        logger.error(error)
+
+        @exit_code = 1
+      end
+
+      def process_usage_error(error)
+        log_exit_code_info(exit_code)
```
iNecas: I'm getting `undefined local variable or method `exit_code' for` when running `bin/foreman-maintain upgrade run --target-version nonexisting`

swapab: That's a test case missed. Fixed!

iNecas:
This branch has conflicts now, please rebsae

swapab:
Conflicts resolved. Rebased.

iNecas:
Looks good, thanks @swapab

PR #101, opened 2017-09-15
Fixes #20870 - hammer ping check with default tag

kgaikwad:


iNecas:
rubocop is complaining

iNecas:
Need to test but looks good so far

kgaikwad:
@iNecas ,
Yes, I will look into rubocop errors and one test case is also failing for ruby 1.8.7.

kgaikwad:
@iNecas , @swapab 

Updated the PR with timeout changes. Ready for review.

definitions/features/katello_service.rb
```diff
@@ -32,8 +32,62 @@ def make_start(spinner, options = {})
...
+      end
+    rescue Timeout::Error => e
+      logger.error e.message
+    end
+  end
+
   private
 
+  def retry_hammer_ping(spinner)
+    5.times do |retry_count|
```
swapab: Probably should be a constant or private method(RETRIES/retries).

swapab:
Procedure step question asked twice.
```
Continue with step [katello-service restart]?, [y(yes), n(no), q(quit)] n
Continue with step [katello-service restart]?, [y(yes), n(no), q(quit)] n
```
Reproducible always: When you chose `y` for first attempt and `n` for next attempt.

hammer ping multiple ask issue

First attempt choose `n`: Works fine

hammer-first-time

kgaikwad:
@swapab,

Updated the PR with suggested changes. 
I tried reproducing an issue that you mentioned above related to `Continue with step` 
and was unable to.
If you are still facing the same issue, please share your inputs to reproduce it.

definitions/features/katello_service.rb
```diff
@@ -27,13 +29,67 @@ def make_start(spinner, options = {})
...
+    else
+      execute!('katello-service restart')
+    end
+  end
+
+  def run_with_hammer_ping_retry(spinner, action = '', services_options = {})
+    meth_call_by_action(action, spinner, services_options)
+    spinner.update 'Sometimes, services will take time to be ready'
+    begin
+      Timeout.timeout(30) do
```
iNecas: Is the Timeout necessary here? I believe the `hammer ping` will finish eventually: I would like to avoid us spawning more requests against the server until the the previous ones are processed.

definitions/features/katello_service.rb
```diff
@@ -27,13 +29,67 @@ def make_start(spinner, options = {})
...
+  def retry_hammer_ping(spinner)
+    RETRIES_FOR_SERVICES_RESTART.times do |retry_count|
+      spinner.update "Retry #{retry_count + 1}: checking status by hammer ping"
+      result = feature(:hammer).hammer_ping_cmd
+      if result[:success]
+        spinner.update 'All services are running.'
+        break
+      elsif retry_count < (RETRIES_FOR_SERVICES_RESTART - 1)
+        spinner.update 'Lets try checking status once again'
+        sleep 2
```
iNecas: the 2 should be extracted to a constant. Also, perhaps something like 30 s would be more suitable interval

iNecas:
Some comments regarding the timeouts, will test after that's resolved

definitions/features/katello_service.rb
kgaikwad: @iNecas , 
We don't need `Timeout` once we apply `30 s sleep`. I have added it so that it will only wait max up to 30 s.
I will remove it and add 30 s sleep.

kgaikwad:
@iNecas,

Done with suggested changes. Updated the PR.

definitions/procedures/katello_service/restart.rb
```diff
@@ -0,0 +1,17 @@
+module Procedures::KatelloService
+  class Restart < ForemanMaintain::Procedure
+    metadata do
+      description 'katello-service restart'
+      param :only, 'A comma-separated list of services to include', :array => true
+      param :exclude, 'A comma-separated list of services to skip', :array => true
+    end
+
+    def run
```
iNecas: Currently, the next step describes as:

```
Continue with step [katello-service restart]
```

By overriding the `runtime_description` method, we should be able to get also info about `--only` and `--exclude`
options

lib/foreman_maintain.rb
```diff
@@ -6,6 +6,7 @@
 require 'json'
 require 'logger'
 require 'yaml'
+require 'timeout'
```
iNecas: This is no longer used

kgaikwad: Oops, sorry!! I will do this.

definitions/features/katello_service.rb
```diff
@@ -27,13 +29,65 @@ def make_start(spinner, options = {})
...
 
+  def retry_hammer_ping(spinner)
+    RETRIES_FOR_SERVICES_RESTART.times do |retry_count|
+      spinner.update "Retry #{retry_count + 1}: checking status by hammer ping"
+      result = feature(:hammer).hammer_ping_cmd
+      if result[:success]
+        spinner.update 'All services are running.'
+        break
+      elsif retry_count < (RETRIES_FOR_SERVICES_RESTART - 1)
+        spinner.update 'Lets try checking status once again'
```
iNecas: What about some more descriptive message, such as (note the "\n" at the first line is needed for properly getting the message to the new line):

```
        puts "\n#{result[:message]}"
        spinner.update "Waiting #{PING_RETRY_INTERVAL} seconds before retry."
        sleep PING_RETRY_INTERVAL
```

definitions/features/katello_service.rb
```diff
@@ -27,13 +29,65 @@ def make_start(spinner, options = {})
...
+    rescue StandardError => e
+      logger.error e.message
+    end
+  end
+
   private
 
+  def retry_hammer_ping(spinner)
+    RETRIES_FOR_SERVICES_RESTART.times do |retry_count|
+      spinner.update "Retry #{retry_count + 1}: checking status by hammer ping"
```
iNecas: Please include the number max attempts as well and use `Try` instead of `Retry` as the first attempt is not effectively a retry :)

```ruby
      spinner.update "Try #{retry_count + 1}/#{RETRIES_FOR_SERVICES_RESTART}: checking status by hammer ping"
```

definitions/features/hammer.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def exec_hammer_cmd(cmd, required_json = false)
+    response = ForemanMaintain::Utils::Hammer.instance.run_command(cmd)
+    json_str = json_parse(response) if required_json
+    json_str ? json_str : response
+  end
+
+  def msg_for_failed_resources(resources)
+    "#{resources.join(', ')} resource(s) are failing. Please check by restarting services."
```
iNecas: The sentence `Please check by restarting services.` dosn't fit here, especially if we put the info on what went wrong in into the `retry_hammer_ping` method.

definitions/features/katello_service.rb
```diff
@@ -27,13 +29,65 @@ def make_start(spinner, options = {})
...
+      execute!('katello-service restart')
+    end
+  end
+
+  def run_with_hammer_ping_retry(spinner, action = '', services_options = {})
+    meth_call_by_action(action, spinner, services_options)
+    spinner.update 'Sometimes, services will take time to be ready'
+    begin
+      retry_hammer_ping(spinner)
+    rescue StandardError => e
```
iNecas: What is the error we are expecting here? Sholudn't it to to the retry_hammer_ping?

definitions/features/katello_service.rb
```diff
@@ -27,13 +29,65 @@ def make_start(spinner, options = {})
...
     else
       filters = "--only #{services.join(',')}"
-      spinner.update 'Starting the katello services..'
+      spinner.update 'Starting katello services..'
       execute!("katello-service start #{filters}")
     end
   end
 
+  def restart(spinner, options = {})
+    spinner.update 'Restarting katello services..'
```
iNecas: Given spinner is only being used here, what about moving this call to the procedure itself, so that the restart method doesn't have to know about the spinner at all?

definitions/procedures/katello_service/restart.rb
```diff
@@ -0,0 +1,17 @@
...
+      with_spinner('restarting katello service(s)') do |spinner|
```
iNecas: Are `meth_call_by_action` and `run_with_hammer_ping_retry` necessary. What if we did something like:

```
spinner.update("Restarting services")
feature(:katello_service).restart(:only => @only, :exclude => @exclude)
feature(:katello_service).hammer_ping_retry(spinner)
```

kgaikwad:
@iNecas ,

Thank you for review suggestion. 
Done with the modifications and updated PR.
PR is ready for review.

iNecas:
Tested and works well. Thanks @kgaikwad !

PR #105, opened 2017-10-04
Fixes #21186 - Add description to --version

swapab:


definitions/procedures/repositories/setup.rb
```diff
@@ -5,7 +5,9 @@ class Setup < ForemanMaintain::Procedure
       confine do
         feature(:downstream) || feature(:upstream)
       end
-      param :version, :required => true
+      param :version,
+            'Satellite/Foreman version for which repositories needs to be setup',
```
kgaikwad: Instead of showing both, can we show either satellite or foreman in this message depending upon which feature available.

swapab: Yeah, I had thought about that. But was thinking to make a generic solution based on this discussion: https://bugzilla.redhat.com/show_bug.cgi?id=1487403

swapab: @kgaikwad I have kept it as Satellite for now.

swapab: As discussed, kept just `Version for which...` instead of branding.

PR #106, opened 2017-10-05
Fixes #20974 - Procedures for Katello-Service stop and start command

kgaikwad:


iNecas:
As per face 2 face discussion, the solution to the ordering should be by explicitly mentioning the migration steps instead of relying on the searching by tags.

kgaikwad:
@iNecas ,
I am trying to check these procedures on real sat_instance but I am not able to create box using `satellite-6.1.12-rhel7.box`.
It is giving me error on file unpacking. 
I will try again to download file freshly and will notify the result once it is done.

definitions/features/katello_service.rb
```diff
@@ -7,16 +7,15 @@ def make_stop(spinner, options = {})
     services = find_services_for_only_filter(running_services, options)
     if services.empty?
       spinner.update 'No katello service running'
-      yield
+      yield if block_given?
     else
       begin
         filters = "--only #{services.join(',')}"
         spinner.update 'Stopping katello running services..'
```
iNecas: `Stopping katello services` would sound better.

iNecas:
Tested and works well, needs rebase as it's conflicting with other PR + one comment on spinner message. Otherwise good to go

kgaikwad:
@iNecas , 
 Done! Updated the PR.

iNecas:
Thanks @kgaikwad

PR #107, opened 2017-10-05
Fixes #21208 - upgrade runner should comply with -y option

swapab:
Pass on the `assumeyes` flag value from `UpgradeRunner` to `upgrade-scenario` which then invokes procedures with `assumeyes` option passed as args.

### DIFF
--- a/definitions/scenarios/upgrade_to_satellite_6_2.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_2.rb
@@ -43,7 +43,7 @@ module Scenarios::Satellite_6_2
 
     def compose
       add_step(Procedures::Repositories::Setup.new(:version => '6.2'))
-      add_step(Procedures::Packages::Update.new(:assumeyes => false))
+      add_step(Procedures::Packages::Update.new(:assumeyes => assumeyes?))
       add_step(Procedures::Installer::Upgrade.new)
     end
   end
--- a/definitions/scenarios/upgrade_to_satellite_6_3.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_3.rb
@@ -43,7 +43,7 @@ module Scenarios::Satellite_6_3
 
     def compose
       add_step(Procedures::Repositories::Setup.new(:version => '6.3'))
-      add_step(Procedures::Packages::Update.new(:assumeyes => false))
+      add_step(Procedures::Packages::Update.new(:assumeyes => assumeyes?))
       add_step(Procedures::Installer::Upgrade.new)
     end
   end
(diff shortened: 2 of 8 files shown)


swapab:
@iNecas @kgaikwad 

This is an approach to let every scenario know about `--assumeyes` flag. Let me know if this looks good and I'll add relevant tests.

iNecas:
+1 for this approach

swapab:
@iNecas @kgaikwad  Ready for review.

definitions/scenarios/upgrade_to_satellite_6_2.rb
```diff
@@ -43,7 +43,7 @@ class Migrations < Abstract
 
     def compose
       add_step(Procedures::Repositories::Setup.new(:version => '6.2'))
-      add_step(Procedures::Packages::Update.new(:assumeyes => false))
+      add_step(Procedures::Packages::Update.new(:assumeyes => assumeyes?))
```
kgaikwad: If you don't pass param `assumeyes` to `Procedures::Packages::Update` procedure then nil case is already handled in file [update.rb](https://github.com/theforeman/foreman_maintain/blob/master/definitions/procedures/packages/update.rb#L9)  where it takes value from `assumeyes?` method.
So no need to pass param `assumeyes` here.

definitions/scenarios/upgrade_to_satellite_6_3.rb
```diff
@@ -43,7 +43,7 @@ class Migrations < Abstract
 
     def compose
       add_step(Procedures::Repositories::Setup.new(:version => '6.3'))
-      add_step(Procedures::Packages::Update.new(:assumeyes => false))
+      add_step(Procedures::Packages::Update.new(:assumeyes => assumeyes?))
```
kgaikwad: Ditto

definitions/scenarios/upgrade_to_satellite_6_2.rb
swapab: @kgaikwad  Thanks for reviewing.
Right, but the instance variable in [update.rb](https://github.com/theforeman/foreman_maintain/blob/master/definitions/procedures/packages/update.rb#L9) is different than the one used here.

Arg `:assumeyes => assumeyes?` in this case refers to [Scenario#assumeyes?](https://github.com/swapab/foreman_maintain/blob/21208-package-update-comply-assumeyes/lib/foreman_maintain/scenario.rb#L10).
So, the argument is not optional it needs to be passed.

iNecas: Thinking more about this, I would probably not even ask the user and put always true here: they already confirmed they want to update before and there is a little chance hey would ever answer no here.

swapab:
Closing in favor of #115

PR #109, opened 2017-10-12
Fixes #21293 - Refactor check: DiskIO - show speed in MB/s

swapab:


definitions/checks/disk/performance.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      def compute_disk_speed(spinner)
+        success = true
+        io_obj = ForemanMaintain::Utils::Disk::NilDevice.new
+
+        dirs_to_check.each do |dir|
+          io_obj = ForemanMaintain::Utils::Disk::Device.new(dir)
+
+          spinner.update("[Speed check In-Progress] device: #{io_obj.name}")
+          push_stats(io_obj)
```
kgaikwad: Currently, I don't see any other place where we have used method `push_stats`.
Instead of creating a method, can we please place code here itself?

swapab: I feel this to be a good coding practice. For readability.

swapab: As I write this I think the method name can be improved with `stats <<`

definitions/checks/disk/performance.rb
```diff
@@ -0,0 +1,63 @@
+module Checks
+  module Disk
+    class Performance < ForemanMaintain::Check
+      metadata do
+        label :disk_performance
+        description 'Check for recommended disk speed of pulp, mongodb, pgsql dir.'
+        tags :pre_upgrade
+        preparation_steps { Procedures::Packages::Install.new(:packages => %w[hdparm fio]) }
```
iNecas: Shouldn't this confine to `:pulp`?

swapab: Yes, this can now be confined to `pulp`

swapab: The requested change was done.

iNecas:
Tested and works well, thanks @swapab

PR #110, opened 2017-10-16
Fixes #21320 - repeated --assumeyes option in packages

kgaikwad:
screenshot-before-after-change

We are already have an interactive option `-y, --assumeyes`.
To maintain consistency between `foreman-maintain advanced procedure run {procedure-name} -h` command, with this commit suppressed an option ` --assumeyes` added from procedure.
So that it won't display repetitive ` --assumeyes` option.

### DIFF
--- a/lib/foreman_maintain/cli/base.rb
+++ b/lib/foreman_maintain/cli/base.rb
@@ -98,6 +98,8 @@ module ForemanMaintain
       end
 
       def self.interactive_option
+        delete_duplicate_assumeyes_if_any
+
         option ['-y', '--assumeyes'], :flag,
                'Automatically answer yes for all questions'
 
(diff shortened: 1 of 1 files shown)


lib/foreman_maintain/cli/base.rb
```diff
@@ -98,6 +98,8 @@ def self.tags_option
       end
 
       def self.interactive_option
+        delete_duplicate_assumeyes_if_any
+
         option ['-y', '--assumeyes'], :flag,
```
swapab: I think we can check here if options already have `--assumeyes` flag and opt to add accordingly.

kgaikwad: Previously, I have made same changes as you mentioned. 
But description of --assumeyes option is different in package update/install  procedure.
To make the consistency between interactive options and their descriptions throughout procedures I did modifications.

swapab:
Thanks @kgaikwad

PR #111, opened 2017-10-16
Fixes #21346 - use the beta Satellite repositories until 6.3 GA

mccun934:


definitions/features/downstream.rb
```diff
@@ -34,9 +34,17 @@ def setup_repositories(version)
   def rh_repos(sat_version)
     sat_version = version(sat_version)
     rh_version_major = execute!('facter operatingsystemmajrelease')
+    # Use Beta repositories for 6.3 until GA
+    if sat_version.to_s.strip == '6.3'
```
iNecas: Rubocop complains on:

```
definitions/features/downstream.rb:38:5: C: Use the return of the conditional for variable assignment and comparison.

    if sat_version.to_s.strip == '6.3' ...

    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So it wants

```ruby
    sat_repo_id = if sat_version.to_s.strip == '6.3'
      "rhel-server-#{rh_version_major}-satellite-6-beta-rpms"
    else
      "rhel-#{rh_version_major}-server-satellite-#{sat_version.major}.#{sat_version.minor}-rpms"
    end
```

iNecas: Is the change of `server` word in the naming intentional?

iNecas:
Thanks @mccun934

PR #112, opened 2017-10-17
Fixes #21351 - foreman-rake katello:upgrade_check in pre_upgrade

kgaikwad:
With this commit, `foreman-rake katello:upgrade_check` will get run in pre_upgrade checks.

If there is no active task present then it will show `ok` result and in case of failure whole output of rake task will get shown. 
As this check is again related to tasks, provided two steps for user:
- Resume tasks 
- Investigate URL with search query `state = paused`

As mentioned in [document](https://access.redhat.com/solutions/2089951), currently tasks cleanup script is present but specific to sat 6.2 so I have not included it in steps.

### DIFF
--- a/definitions/checks/foreman_rake/katello_upgrade_check.rb
+++ b/definitions/checks/foreman_rake/katello_upgrade_check.rb
@@ -1,0 +1,27 @@
+module Checks::ForemanRake
+  class KatelloUpgradeCheck < ForemanMaintain::Check
+    metadata do
+      label :katello_upgrade_check
+      tags :pre_upgrade
+      description 'Check for active tasks using katello:upgrade_check'
+      confine do
+        feature(:katello) && command_exists?('foreman-rake')
+      end
+    end
+
+    def run
+      output = execute('foreman-rake katello:upgrade_check')
+      assert(ready_to_upgrade?(output),
+             output,
+             :next_steps =>
+                [Procedures::ForemanTasks::Resume.new,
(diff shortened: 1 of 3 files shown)


iNecas:
Oh, I have not realized that the upgrade check is just checking for the curently running tasks: in that case, let's fix https://bugzilla.redhat.com/show_bug.cgi?id=1496384 instead. I was first thinking about moving the check for running tasks into the pre_migrations (http://projects.theforeman.org/issues/21118#note-2), but I believe that even in pre-upgrade checks should be covering most of the cases.

kgaikwad:
@iNecas ,
Yes, when I check description I got know it is related to running/active tasks.
Ok, I will look into [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1496384).

In the document, they have suggested to run this `katello:upgrade_check` twice:
- In pre_upgrade checks and 
-  Before `katello-service stop` & `yum update`
We may need to check this in `pre_migrations` also.

First, I will start on the BZ. Let's decide it later after discussion.

iNecas:
Yes, probably having it on both places would make sense, but we can add it one by one

kgaikwad:
@iNecas ,
Created [PR](https://github.com/theforeman/foreman_maintain/pull/114) for an [issue](http://projects.theforeman.org/issues/21118).

iNecas:
Closing in favor of #114

PR #114, opened 2017-10-17
Fixes #21118 - Verify if there are no any running tasks

kgaikwad:


definitions/checks/foreman_tasks/not_running.rb
```diff
@@ -9,8 +9,17 @@ class NotRunning < ForemanMaintain::Check
...
 
     def run
-      assert(feature(:foreman_tasks).running_tasks_count == 0,
-             'There are actively running tasks in the system')
+      task_count = feature(:foreman_tasks).running_tasks_count
+      assert(task_count == 0,
+             failure_message(task_count),
+             :next_steps =>
+               [Procedures::ForemanTasks::Resume.new,
+                Procedures::ForemanTasks::UiInvestigate.new('search_query' => 'state = paused')])
```
iNecas: The search query should be `state = running`, shouln't it?

definitions/checks/foreman_tasks/not_running.rb
```diff
@@ -9,8 +9,17 @@ class NotRunning < ForemanMaintain::Check
     end
 
     def run
-      assert(feature(:foreman_tasks).running_tasks_count == 0,
-             'There are actively running tasks in the system')
+      task_count = feature(:foreman_tasks).running_tasks_count
+      assert(task_count == 0,
+             failure_message(task_count),
+             :next_steps =>
+               [Procedures::ForemanTasks::Resume.new,
```
iNecas: The resume is not valid for `running` tasks. We could have a procedure, that polls and waits for the tasks to finish.

definitions/checks/foreman_tasks/not_running.rb
kgaikwad: Yes, actually!!

definitions/checks/foreman_tasks/not_running.rb
kgaikwad: In the running_tasks_count, I have mentioned  filter state field  should be either running or paused. That's why I mentioned this step.

Should I remove paused from state filter values?

kgaikwad: Also, please correct me if I am misunderstood.
Here, polls and wait for tasks to finish means we need to ask user to enter timeframe to wait.
After timeframe elapsed, we need to exit from procedure.

kgaikwad: @iNecas ,

I have queries regarding this:
- Do we need to set any timeout for this procedure to exit? Or we really need to wait till all tasks to finish.
- What interval should we set before retrying status of tasks?

iNecas: We have the `ForemanTasks::NotPaused` check for the paused, so this should be for running only.

Having a timeout is a good idea. I would suggest something around 5 minutes for start.

For the retry interval, I would go for every 10 seconds, as it's not very expensive operation.

kgaikwad: @iNecas ,
Is it okay if I change order for NotRunning check as it is showing after NotPaused check?

I know that it is a rare situation where any running task which will get paused somehow in NotRunning check. But in that scenario it will proceed with upgrade.

I have pushed the changes but they are incomplete. I just saw your comments.
I will update PR.

kgaikwad:
@iNecas ,

Updated the PR with `timeout` of `5 mins` and interval of `10 sec`.
Ready for review!

definitions/procedures/foreman_tasks/fetch_tasks_status.rb
```diff
@@ -0,0 +1,15 @@
+module Procedures::ForemanTasks
+  class FetchTasksStatus < ForemanMaintain::Procedure
+    metadata do
+      for_feature :foreman_tasks
+      description 'Fetch tasks status and wait till they finish'
+      param :state, :required => true
+    end
+
+    def run
+      with_spinner("wait till #{@state} tasks get finish") do |spinner|
```
iNecas: `till` is quite informal, let's use `until` instead. I would also recommend using `ing` from, as we are describing, what's happening at the moment, and instead of `get finish`, it would sound better to me to `waiting for 2 tasks to finish` or `waiting until tasks get finished`.

I would recommend `waiting for #{@state} tasks to finish`

definitions/checks/foreman_tasks/not_running.rb
```diff
@@ -9,8 +9,17 @@ class NotRunning < ForemanMaintain::Check
...
 
     def run
-      assert(feature(:foreman_tasks).running_tasks_count == 0,
-             'There are actively running tasks in the system')
+      task_count = feature(:foreman_tasks).running_tasks_count
+      assert(task_count == 0,
+             failure_message(task_count),
+             :next_steps =>
+               [Procedures::ForemanTasks::FetchTasksStatus.new(:state => 'running'),
+                Procedures::ForemanTasks::UiInvestigate.new('search_query' => 'state = running')])
```
iNecas: The search query should be `state = running AND label !^(Features::ForemanTasks::EXCLUDE_ACTIONS_FOR_RUNNING_TASKS.join(" "))`

iNecas: The `UiInvestigate` procedure is using `press ENTER after the paused tasks are resolved` wording, which is not precise for this case. Using `press ENTER after the tasks are resolved` would be matching both running and paused tasks.

kgaikwad: Good Catch!! :+1:

kgaikwad:
@iNecas ,

Sorry for small mistakes.
Updated PR as per suggested changes. Ready for review.

iNecas:
Thanks @kgaikwad, tested and works well. Merging and will release a new version. Enjoy the holiday!

kgaikwad:
Thank you !! 😊
Yes, now we can release new version 👍

iNecas:
0.0.10 is out now

PR #119, opened 2017-10-30
Fixes #21452 - Add yum clean all before packages update

kgaikwad:
`yum clean all` doesn't delete `/var/cache/yum`.
According to below BZ issue and one knowledge based article: 
- https://bugzilla.redhat.com/show_bug.cgi?id=1214203
- https://access.redhat.com/solutions/1989973 

Added `rm -rf /var/cache/yum` and `rm -rf /var/cache/dnf` after `yum clean all` command for safer side. Due to removal of cache folder for `dnf`, it will take some time at next run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ class Procedures::InstallPackage < ForemanMaintain::Procedure
   end
 
   def run
-    install_packages(@packages)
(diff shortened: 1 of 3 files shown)


iNecas:
Thanks @kgaikwad

PR #120, opened 2017-10-30
Fixes #21507 - Added check for curl status code 000

gnurag:
In certain scenarios, curl seems to be exiting with $http_code of 000. When this happens, find_http_error_msg function does not handle the status code and encounters a nil exception.

The [issue](http://projects.theforeman.org/issues/21507) was originally reported by manually stopping foreman-proxy service which closes the port curl queries, but the same behavior is shown when the hostname is (temporarily) unresolvable too.

example:
```
$ curl -w '%{http_code}' https://nonexistant.example.com/
curl: (6) Could not resolve host: nonexistant.example.com
000$
```

### DIFF
--- a/definitions/features/foreman_proxy.rb
+++ b/definitions/features/foreman_proxy.rb
@@ -26,12 +26,16 @@ class Features::ForemanProxy < ForemanMaintain::Feature
   end
 
   def find_http_error_msg(array_output, curl_http_status)
-    http_line = ''
-    array_output.each do |str|
-      next unless str.include?('HTTP')
-      http_line = str
+    if curl_http_status == 0
+      'No valid HTTP response (Connection failed)'
+    else
+      http_line = ''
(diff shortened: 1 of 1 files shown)


definitions/features/foreman_proxy.rb
```diff
@@ -26,12 +26,16 @@ def dhcp_curl_cmd
   end
 
   def find_http_error_msg(array_output, curl_http_status)
-    http_line = ''
-    array_output.each do |str|
-      next unless str.include?('HTTP')
-      http_line = str
+    if curl_http_status == 0
+      'Curl error'
```
kgaikwad: Instead of giving this message, can we give a precise message for user so that he can debug further.
for example - `No valid HTTP response code (either network issues or Connection failed for service)`.

iNecas:
Thakns @gnurag and @kgaikwad

PR #122, opened 2017-11-01
Fixes #21506 - scenario error on upgrade run

kgaikwad:


lib/foreman_maintain/detector.rb
```diff
@@ -73,6 +73,10 @@ def ensure_features_detected
       @all_features_scanned = true
     end
 
+    def scenarios_without_evaluating_confines(filter_conditions = nil)
```
iNecas: what about the `all_scenarios`?

lib/foreman_maintain/scenario.rb
```diff
@@ -158,9 +158,13 @@ def to_hash
     end
 
     def self.new_from_hash(hash)
-      scenarios = find_scenarios(:label => hash[:label])
+      condition = { :label => hash[:label] }
+      scenarios = find_scenarios(condition)
```
iNecas: What is the reason for not using non-confine variant first?

kgaikwad: Yeah, totally agree!! :+1: 
There is no use having confine evaluation here as we are again checking for non-confine variant. 
I was thinking too much unnecessary about this I guess :-)
I will update PR according to review comments.

kgaikwad:
@iNecas ,
I have one query which may or may not be relevant here.
In `scenario` method from [upgrade_runner.rb](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/upgrade_runner.rb#L70), we have used `find_scenarios` with tags condition. 
This will get execute only in case where there is no phase in `scenario_cache`.
Here we don't need to change it with `find_all_scenario`.

Please correct me if I misunderstood anything.

iNecas:
I think we should also change it to `find_all_scenari`

iNecas:
Tested and works well so far, waiting for the update in upgrade_runner.rb, based on our last comments.

kgaikwad:
@iNecas,
Thank you.
Done with the changes.

iNecas:
Tested with the new changes and works well.

PR #123, opened 2017-11-09
Fixes #21588 - cofine hammer ping check to katello feature

swapab:
hammer ping command is not available for standalone foreman instance


iNecas:
Thanks @swapab, makes sence

PR #126, opened 2017-11-23
Fixes #21721 - Don't downcase stdin if its hammer credentials

gnurag:
Added condition to `ask()` where it skips `.downcase` method if the console input is hammer username and password.

Alternatively, an easier way could be to simply skip downcasing of input. But I'm not sure if doing this would have any side-effects elsewhere.
`answer.to_s.chomp if answer`

### DIFF
--- a/lib/foreman_maintain/reporter/cli_reporter.rb
+++ b/lib/foreman_maintain/reporter/cli_reporter.rb
@@ -117,7 +117,7 @@ module ForemanMaintain
         # add space at the end as otherwise highline would add new line there :/
         message = "#{message} " unless message =~ /\s\Z/
         answer = @hl.ask(message) { |q| q.echo = false if options[:password] }
-        answer.to_s.chomp.downcase if answer
+        answer.to_s.chomp if answer
       end
 
       def new_line_if_needed
@@ -195,6 +195,7 @@ module ForemanMaintain
 
       def filter_decision(answer)
         decision = nil
+        answer   = answer.downcase
         DECISION_MAPPER.each do |options, decision_label|
           decision = decision_label if options.include?(answer)
         end


lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -117,7 +117,11 @@ def ask(message, options = {})
         # add space at the end as otherwise highline would add new line there :/
         message = "#{message} " unless message =~ /\s\Z/
         answer = @hl.ask(message) { |q| q.echo = false if options[:password] }
-        answer.to_s.chomp.downcase if answer
+        if answer
+          answer.to_s.chomp!
+          answer.downcase! unless options[:username] or options[:password]
```
kgaikwad: `or` & `and`  are used as `control flow operators`, not as boolean operators.
So, use `||` operator here.
Please modify syntax like  `options[:username] || options[:password]` 
instead of `options[:username] or options[:password]`

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -117,7 +117,11 @@ def ask(message, options = {})
         # add space at the end as otherwise highline would add new line there :/
         message = "#{message} " unless message =~ /\s\Z/
         answer = @hl.ask(message) { |q| q.echo = false if options[:password] }
-        answer.to_s.chomp.downcase if answer
+        if answer
+          answer.to_s.chomp!
+          answer.downcase! unless options[:username] || options[:password]
```
iNecas: How does the `:username` option even get here? Notice, that we validate what options we are accepting on line 113.

I think we should remove the downcase here, and use it in `filter_decision` instead (+ check other places we use `ask` to make sure we have not missed something.

kgaikwad: Yes, right! I have missed it.
Only for password, passed options `password` to `ask` method.

iNecas:
ping

kgaikwad:
@iNecas & @swapab,

I have tested this change and other checks. Works well so far.
Please let me know if we can go and merge this PR.

PR #128, opened 2017-12-04
Fixes #21713 - remove bad content definitions from cp_env_content

kgaikwad:


definitions/checks/candlepin/cp_env_content/verify_bad_content_definitions.rb
```diff
@@ -0,0 +1,21 @@
+module Checks::Candlepin
+  module CpEnvContent
```
swapab: `Cp` can be dropped here? since we are already scoping this checks with `Candlepin`

definitions/features/candlepin_database.rb
```diff
@@ -0,0 +1,31 @@
+class Features::CandlepinDatabase < ForemanMaintain::Feature
+  CANDLEPIN_DB_CONFIG = '/etc/candlepin/candlepin.conf'.freeze
+
+  include ForemanMaintain::Concerns::BaseDatabase
+  metadata do
```
swapab: A new line here.

definitions/features/foreman_database.rb
```diff
@@ -1,21 +1,23 @@
 class Features::ForemanDatabase < ForemanMaintain::Feature
+  FOREMAN_DB_CONFIG = '/etc/foreman/database.yml'.freeze
+
+  include ForemanMaintain::Concerns::BaseDatabase
```
swapab: And a new line.

definitions/features/foreman_database.rb
```diff
@@ -1,21 +1,23 @@
...
   metadata do
     label :foreman_database
 
     confine do
-      File.exist?('/etc/foreman/database.yml')
+      File.exist?(FOREMAN_DB_CONFIG)
```
swapab: Can use system helper method `file_exists?`.

definitions/checks/candlepin/cp_env_content/verify_bad_content_definitions.rb
kgaikwad: @swapab,
There is table having name cp_env_content thats why I mentioned it here.
Please confirm then I will go and remove it.

swapab: Oh okay makes sense. Can keep it as is.

iNecas: Do we even need a separate namespace for this?

iNecas:
Tested and works well, one suggestion about removing the module

definitions/checks/candlepin/cp_env_content/verify_bad_content_definitions.rb
kgaikwad: @iNecas,
I wanted to give some context of table in candlepin DB + action for that table.
So, added namespace for procedure to avoid long class name.
That's why in checks also I have placed the namespace.
What I can do is remove namespace from check and modify name of check to get a context of table.

Please let me know your suggestions on it so that I can maintain context of `table + action` on it.

kgaikwad:
@iNecas,
Ok, I will remove a module from check and update the PR.

kgaikwad:
@iNecas ,

Done with the modifications as per the suggestions.
Removed extra namespaces from `checks/candlepin` & `procedures/candlepin`.
Updated PR.

kgaikwad:
@iNecas ,

Please review modified changes so that we can merge this PR as for `cpdb check` I will use same `candlepin` feature.

iNecas:
As per https://bugzilla.redhat.com/show_bug.cgi?id=1503177, I wonder if we actually need the check/fix procedure in foreman-maintain. as the issue itself should be addressed in the candlepin already, Or am I missing something?

Anyway, I've tested the code and it works in general, so if you need some parts in other PRs, I'm ok with merging the feature part itself, maybe as part of the other PR?

kgaikwad:
This is main BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1503176.
And above ticket is cloned from this.

Using cpdb command, it is validating the DB for bad product/content/activation-key. So closing this PR as it is already addressed into cpdb.

PR #129, opened 2017-12-04
Fixes #20156 - EPEL repo check

upadhyeammit:
This adds check for EPEL repository, if repository is enabled then check fails.

### DIFF
--- a/definitions/checks/check_epel_repository.rb
+++ b/definitions/checks/check_epel_repository.rb
@@ -1,0 +1,21 @@
+class Checks::CheckEpelRepository < ForemanMaintain::Check
+  metadata do
+    label :check_epel_repository
+    description 'Check if EPEL repository enabled on system'
+    tags :pre_upgrade
+    confine do
+      feature(:downstream)
+    end
+  end
+
+  def run
+    with_spinner('Checking for presence of EPEL repository') do
+      assert(!epel_enabled?, 'System is subscribed to EPEL repository')
+    end
+  end
+
+  def epel_enabled?
+    system_repos = execute("yum repolist enabled -d 6 -e 0| grep -E 'Repo-baseurl|Repo-id'")
+    system_repos.to_s.match(/\bepel\b/i)
+  end
+end


definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,17 @@
+class Checks::CheckRepositories < ForemanMaintain::Check
```
swapab: The class name should be same as file name.
`Checks::CheckEpelRepository`

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,17 @@
...
+  metadata do
+    label :check_epel_repository
+    description 'Check if EPEL repoistory enabled on system'
+    tags :default
+  end
+  def run
```
swapab: Add new line here.

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,17 @@
...
+    epel_enabled?
+  end
+
+  def epel_enabled?
+    system_repos = execute("yum repolist enabled -d 6 -e 0| grep -E 'Repo-baseurl|Repo-id'")
+    if system_repos.is_a?(String) && system_repos.include?('epel')
```
swapab: This can be concisely written as `if system_repos.to_s.include?('epel')`

upadhyeammit:
@swapab done all changes as suggested, request to check. Thank You!

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,18 @@
+class Checks::CheckEpelRepository < ForemanMaintain::Check
+  metadata do
+    label :check_epel_repository
+    description 'Check if EPEL repoistory enabled on system'
+    tags :default
```
iNecas: I think this should be limited to `:pre_upgrade` tag

iNecas: Also, there should be limited to downstream only deployments

```ruby
confine do
  feature(:downstream)
end
```

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def run
+    epel_enabled?
```
iNecas: Nitpick: I would expect the `epel_enabled?` only to return `true`/`false`, and raising the exception based on the result here.

upadhyeammit:
@iNecas changes done as you have mentioned. Thank You!

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+  end
+
+  def run
+    assert(!epel_enabled?, 'System is subsribed to EPEL repository')
+  end
+
+  def epel_enabled?
+    system_repos = execute("yum repolist enabled -d 6 -e 0| grep -E 'Repo-baseurl|Repo-id'")
+    return true if system_repos.to_s.include?('epel')
```
kgaikwad: @upadhyeammit ,
No need to return `true` explicitly. 
As `include?` method returns a `boolean` value, please mention only `system_repos.to_s.include?('epel')`.

upadhyeammit: @kgaikwad changes done as requested. Thank You!

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,19 @@
...
+    label :check_epel_repository
+    description 'Check if EPEL repoistory enabled on system'
+    tags :pre_upgrade
+    confine do
+      feature(:downstream)
+    end
+  end
+
+  def run
+    assert(!epel_enabled?, 'System is subsribed to EPEL repository')
```
iNecas: Since the `yum repolist` can take a bit of time, especially when registered with subscription manager, I would recommend using spinner while waiting:

```ruby
with_spinner("Checking for presence of EPEL repository") do
  assert(!epel_enabled?, 'System is subsribed to EPEL repository')
end
```

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def epel_enabled?
+    system_repos = execute("yum repolist enabled -d 6 -e 0| grep -E 'Repo-baseurl|Repo-id'")
+    system_repos.to_s.include?('epel')
```
iNecas: I think better would be matching on whole words (to avoid false positives, that would have epel as a subname, not that it would be very likely) + case insensitive, just in case:

```ruby
system_repos.to_s.match(/\bepel\b/i)
```

upadhyeammit:
@iNecas added spinner and modified to use match method.
Thank You!

swapab:
@upadhyeammit Can you please rebase with the latest master? This should fix Travis build.

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,21 @@
+class Checks::CheckEpelRepository < ForemanMaintain::Check
+  metadata do
+    label :check_epel_repository
+    description 'Check if EPEL repoistory enabled on system'
```
ntkathole: Typo `repoistory`

definitions/checks/check_epel_repository.rb
```diff
@@ -0,0 +1,21 @@
...
+    tags :pre_upgrade
+    confine do
+      feature(:downstream)
+    end
+  end
+
+  def run
+    with_spinner('Checking for presence of EPEL repository') do
+      assert(!epel_enabled?, 'System is subsribed to EPEL repository')
```
ntkathole: Typo `subsribed`

ntkathole:
Looks good! 
```
Check if EPEL repository enabled on system:
| Checking for presence of EPEL repository                            [OK]
```
if EPEL repos enabled

```
Check if EPEL repository enabled on system:
/ Checking for presence of EPEL repository                            [FAIL]
System is subscribed to EPEL repository

```

kgaikwad:
@iNecas, can we merge this PR?

iNecas:
Thanks @upadhyeammit, @kgaikwad and @ntkathole

PR #130, opened 2017-12-11
Fixes #21730 - remove empty puppet cacert request files

kgaikwad:


definitions/checks/puppet/verify_no_empty_cacert_requests.rb
```diff
@@ -0,0 +1,20 @@
+module Checks::Puppet
+  class VerifyNoEmptyCacertRequests < ForemanMaintain::Check
+    metadata do
+      label :check_no_empty_cacert_requests_exist
```
iNecas: I suggest having `puppet` in the prefix of the label, something like `label :puppet_check_no_empty_cert_requests`

iNecas:
Tested and works well: one comment regarding the label naming

ntkathole:
@kgaikwad While testing I removed directory `rmdir /var/lib/puppet/ssl/ca/requests/` . It showed `undefined method empty? for nil:NilClass` if directory not found.

Can we make it better?
```
Check to verify no empty CA cert requests exist: /var/lib/puppet/ssl/ca/requests directory not found
                     [FAIL]
undefined method `empty?' for nil:NilClass

  [puppet-check-no-empty-cert-requests]

use --whitelist="puppet-check-no-empty-cert-requests"

```

kgaikwad:
@ntkathole,
Handled above condition which checks for directory.
Updated the PR.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -85,6 +85,10 @@ def file_exists?(filename)
         File.exist?(filename)
       end
 
+      def directory_exists(dirname)
```
mbacovsky: I'd suggest to avoid extending this helper concern unless inevitable. It extends all executable defs and a few other places and adding too much stuff to their scope does not seem like a good practice. This increases risk of conflits and leads to too much methods in the objects. What added value this helper has? Could it be avoided at all?

kgaikwad: Yes, there is no value by adding this method to helper file instead I can use `File.directory?` method directly. 
I thought similar to `file_exists?` method I should move this method here so moved to helper file.

Updated PR by reverting these changes and using `File.directory?` .

definitions/features/puppet_master.rb
```diff
@@ -0,0 +1,36 @@
+class Features::PuppetMaster < ForemanMaintain::Feature
+  metadata do
+    label :puppet_master
+
+    confine do
+      find_package('puppet-server')
```
mbacovsky: On my Sat 6.4 the package is called `puppetserver-5.1.0-2.el7sat.noarch` -->
```ruby 
find_package('puppet-server') || find_package('puppetserver')
```

definitions/features/puppet_master.rb
```diff
@@ -0,0 +1,36 @@
+class Features::PuppetMaster < ForemanMaintain::Feature
```
mbacovsky: I found out, there is also `PuppetServer` feature. Shouldn't both be merged together?
`PuppetServer` is currently just the place for the `services` method that would need to be changed to
```ruby
  def services
    # We only check puppetserver and not puppet-server, as puppet-server
    # is a part of httpd and relies on httpd service to restart, therefore
    # not requiring a separate service to restart
    find_package('puppetserver') ? { 'puppetserver' => 30 } ? {}
  end

File: definitions/features/puppet_server.rb [hunk 5]
```diff
@@ -2,15 +2,42 @@ class Features::PuppetServer < ForemanMaintain::Feature
   metadata do
     label :puppet_server
 
-    # We only check puppetserver and not puppet-server, as puppet-server
-    # is a part of httpd and relies on httpd service to restart, therefore
-    # not requiring a separate service to restart
     confine do
-      find_package('puppetserver')
+      find_package('puppet-server') || find_package('puppetserver')
```
kgaikwad: @mbacovsky, @iNecas,

I can see there is one more feature called [`puppet` ](https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/puppet.rb) and in which there is similar confine block i.e

```
confine do
   find_package('puppetserver')
end
```
In that feature we have method `config_files`. Should I move this method to `puppet_server` feature?
and instead of checking `find_package('puppetserver')` may I add `find_package('puppet')` for `puppet` feature?

please confirm.

iNecas: I'm okw with merging, but it should be called `:puppet_server` - as might want to add the `puppet_agent` in the future.

kgaikwad: Thank you @mbacovsky and @iNecas. I will update PR with above changes.

johnpmitsch: +1 to what @iNecas said

kgaikwad:
@mbacovsky, @iNecas,

Removed `puppet` feature and moved `config_files` in `puppet_server` feature with extra condition in confine block for `puppet` package.
Updated the PR.

kgaikwad:
@ntkathole, done with the rebase. Please have a look.

ntkathole:
@kgaikwad  foreman-maintain health check on 6.4 showed
```
Check to verify no empty CA cert requests exist:                      [SKIPPED]
Warning: No section specified; defaulting to 'main'.
Set the config section by using the `--section` flag.
For example, `puppet config --section user print foo`.
For more information, see https://puppet.com/docs/puppet/latest/configuration.html
Resolving settings from section 'main' in environment 'production'
/etc/puppetlabs/puppet/ssl/ca/requests directory not found
```

upadhyeammit:
@kgaikwad and @ntkathole the 'puppet config print ssldir' command is showing that warning message, if I redirect its error on bash to null then it does not show this warning message,

# puppet config print ssldir 2> /dev/null
/etc/puppetlabs/puppet/ssl

Same thing I tried but it did not help to fix this issue, so there is something in execute! method which we need to check to know why redirect is not working.

definitions/features/puppet_server.rb
```diff
@@ -2,19 +2,58 @@ class Features::PuppetServer < ForemanMaintain::Feature
...
+  end
+
+  private
+
+  def ca_directory_path
+    "#{puppet_ssldir_path}/ca"
+  end
+
+  def puppet_ssldir_path
+    execute!('puppet config print ssldir')
```
upadhyeammit: This execute throws warning and redirect 2> /dev/null is not working here, request to check this one.

mbacovsky: @kgaikwad, @upadhyeammit, @iNecas I found out we are not the only ones suffering from this [issue](https://tickets.puppetlabs.com/browse/PUP-8566) and we can as an alternative use 
```
 puppet master --configprint ssldir
``` 
I positively tested it with puppet 4,5 but have no version 3. It would be great if someone could confirm.

mbacovsky:
Thanks for the updates @kgaikwad. It is good to go for me when the 'puppet print' issue is resolved.

kgaikwad:
@mbacovsky, @upadhyeammit,

Thank you for your inputs. :+1: 
As per @mbacovsky mentioned above, using `puppet master --configprint ssldir` command we can resolve the `puppet print` issue. Updated the PR.

@ntkathole, please confirm.

mbacovsky:
:+1: from me, thanks @kgaikwad. Let's wait on @ntkathole approval.

kgaikwad:
@ntkathole,

Done with rebase and resolved conflicts for puppet feature as we have recently updated puppet config files in puppet feature.
Please verify this PR again.

mbacovsky:
@ntkathole could you please re-check this is good to go?

jameerpathan111:
@kgaikwad @mbacovsky 
Tested pr with Satellite 6.2, 6.3, 6.4 and 6.5
```
# bin/foreman-maintain health check --label puppet-check-no-empty-cert-requests
Check to verify no empty CA cert requests exist:                      [FAIL]
Found 1 empty file(s) under /var/lib/puppet/ssl/ca/requests
Continue with step [Delete empty CA cert request files]?, [y(yes), n(no), q(quit)] y
Delete empty CA cert request files:                                             
Empty CA cert file(s) under /var/lib/puppet/ssl/ca/requests:
1. /var/lib/puppet/ssl/ca/requests/jp.pem
- Done with deleting empty cert request files                         [OK]      
```
```
# rmdir /var/lib/puppet/ssl/ca/requests/
[root@smqa-x3650-01-vm03 foreman_maintain]# bin/foreman-maintain health check --label puppet-check-no-empty-cert-requests
Check to verify no empty CA cert requests exist:                      [SKIPPED]
/var/lib/puppet/ssl/ca/requests directory not found
```
PR looks good to me. :+1:

kgaikwad:
@mbacovsky & @upadhyeammit,
Could you please check whether this PR good to merge?

mbacovsky:
:+1: If it gets approval from @upadhyeammit I can merge this.

upadhyeammit:
Approved from my side, checked all things can see all things are good to go ! Thank You @kgaikwad

PR #132, opened 2017-12-14
Fixes #21974 - make --state mandatory in tasks delete

kgaikwad:


definitions/procedures/foreman_tasks/delete.rb
```diff
@@ -1,7 +1,19 @@
...
+    ALLOWED_TASK_STATES_TO_DELETE = %w[old planning pending].freeze
+
     metadata do
-      param :state, 'In what state should the task be deleted'
+      param :state,
+            "In what state should the task be deleted.\
+             Possible values are #{ALLOWED_TASK_STATES_TO_DELETE.join(', ')}",
+            :required => true do |val|
+              unless ALLOWED_TASK_STATES_TO_DELETE.include?(val.to_s)
+                raise ArgumentError,
```
kgaikwad: One query I have that for `:required => true`, it is giving proper user friendly message.
But  in case of `block execution`, it raises exception and error stack with message.
For example -

```
#  ./bin/foreman-maintain advanced procedure run foreman-tasks-delete --state xyz
/root/foreman_maintain/definitions/procedures/foreman_tasks/delete.rb:11:in `block (2 levels) in <class:Delete>': State value 'xyz' is not allowed. Possible values are old, planning, pending (ArgumentError)
	from /root/foreman_maintain/lib/foreman_maintain/param.rb:31:in `call'
	from /root/foreman_maintain/lib/foreman_maintain/param.rb:31:in `process'
	from /root/foreman_maintain/lib/foreman_maintain/executable.rb:35:in `block in setup_params'
	from /root/foreman_maintain/lib/foreman_maintain/executable.rb:34:in `each_value'
	from /root/foreman_maintain/lib/foreman_maintain/executable.rb:34:in `setup_params'
	
```
When I tried to call clamp methods here it failed for those methods.
Please suggest me some pointers.

swapab: Can you try raising UsageError?

swapab: This can be best handled with `allowed_values` or `in` option for params.

kgaikwad: @swapab , Thanks.
`ForemanMaintain::Error::UsageError` works well. Updated the PR.

kgaikwad: @swapab ,
Yes, I will modify code to use `allowed_values` or `in` option for params.

swapab: We can do that in a different task(unless params already provide the above options).

swapab:
@kgaikwad Thanks for fixing this. :+1:

PR #135, opened 2017-12-27
Fixes #22082 - new conventions & warnings for rubocop 0.52.0

kgaikwad:


lib/foreman_maintain/detector.rb
```diff
@@ -122,6 +122,7 @@ def detect_feature(label)
       end
       present_feature
     end
+    # rubocop:enable Metrics/AbcSize
```
iNecas: Does this mean that with new rubocop, the cop is disabled until it's explicitly enable again?

kgaikwad: As Inline config is only applied to the given file, rubocop gives warnings to select block of code where we need to disable cop. 
If we ignore those warnings then disabling cop will be applied to current file.

For example - 
```
lib/foreman_maintain/detector.rb:107:1: W: Lint/MissingCopEnableDirective: Re-enable Metrics/AbcSize cop with # rubocop:enable after disabling it.
```

PR #139, opened 2018-01-18
Fixes #21927 - check to validate candlepin db using cpdb

kgaikwad:


definitions/checks/candlepin/validate_candlepin_db.rb
```diff
@@ -0,0 +1,20 @@
+module Checks::Candlepin
+  class ValidateCandlepinDb < ForemanMaintain::Check
+    metadata do
+      description 'Check to validate candlepin database'
+      tags :pre_upgrade
+      label :validate_candlepin_db
+
+      confine do
+        feature(:candlepin) &&
```
iNecas: Why not just confine on `feature(:candlepin_database) && feature(:candlepin_database).check_option_using_cpdb_help('validate')` I would recommend defining the `candlepin` feature to the times that actually has some other reasons. It seems we don't need it at this point.

definitions/checks/candlepin/validate_candlepin_db.rb
```diff
@@ -0,0 +1,20 @@
...
+          feature(:candlepin_database).check_option_using_cpdb_help('validate') != 0
+      end
+    end
+
+    def run
+      feature(:candlepin_database).execute_cpdb_validate_cmd
+      output = execute('echo $?').strip.to_i
```
iNecas: As you are using `execute!` in `execute_cpdb_validate_cmd`, isn't it by itself already raising an exception on non-zero exit code?

definitions/checks/candlepin/validate_candlepin_db.rb
```diff
@@ -0,0 +1,20 @@
+module Checks::Candlepin
+  class ValidateCandlepinDb < ForemanMaintain::Check
+    metadata do
+      description 'Check to validate candlepin database'
+      tags :pre_upgrade
+      label :validate_candlepin_db
+
+      confine do
+        feature(:candlepin) &&
+          feature(:candlepin_database).check_option_using_cpdb_help('validate') != 0
```
iNecas: I would recommend hiding the implementation detail, that we're checking help for this output, by calling something like `feature(:candlpin_database).validate_available?` and letting this method decide how we determine that (which in this case would be calling the `check_option_using_cpdb_help('validate')` method there.

definitions/features/candlepin_database.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def cpdb_help_output
+    @cpdb_help_output || run_cpdb_help_cmd
+  end
+
+  def check_option_using_cpdb_help(option_name, parent_cmd = '')
+    parent_cmd = '/usr/share/candlepin/cpdb' if parent_cmd.empty?
+    help_cmd = "#{parent_cmd} --help |  grep -c '\\-\\-#{option_name} '"
+    execute(help_cmd).strip.to_i
```
iNecas: you can use `execute?` and return boolean directly

definitions/features/candlepin_database.rb
```diff
@@ -0,0 +1,60 @@
...
+      file_exists?(CANDLEPIN_DB_CONFIG)
+    end
+  end
+
+  def configuration
+    @configuration || load_configuration
+  end
+
+  def cpdb_help_output
+    @cpdb_help_output || run_cpdb_help_cmd
```
iNecas: Did you mean?

```ruby
@cpdb_help_output ||= run_cpdb_help_cmd
```

definitions/features/candlepin_database.rb
```diff
@@ -0,0 +1,60 @@
...
+  def check_option_using_cpdb_help(option_name, parent_cmd = '')
+    parent_cmd = '/usr/share/candlepin/cpdb' if parent_cmd.empty?
+    help_cmd = "#{parent_cmd} --help |  grep -c '\\-\\-#{option_name} '"
+    execute(help_cmd).strip.to_i
+  end
+
+  def execute_cpdb_validate_cmd
+    main_cmd = cpdb_validate_cmd
+    unless main_cmd.empty?
+      main_cmd += " -u #{configuration['username']} -p '#{configuration[%(password)]}'"
```
iNecas: I would recommend using `shellescape` helper here (that we have defined in the `SystemHelpers`, and also wrapping the `username` argument into `''`. Perhaps additional helper in the system helpers, that would take the args like this and turn them into proper command would be useful for other cases as well:

```ruby
main_cmd += format_shell_args('-u' => configuration['username'], '-p' => configuration['password'])
```

definitions/checks/candlepin/validate_candlepin_db.rb
kgaikwad: Yes, right.
`execute! ` already raising an exception on non-zero exit code.

kgaikwad:
@iNecas ,
Done with the modification as per your suggestions.
Tested it on sat 6.2.14.

definitions/checks/candlepin/validate_candlepin_db.rb
```diff
@@ -0,0 +1,17 @@
+module Checks::Candlepin
+  class ValidateCandlepinDb < ForemanMaintain::Check
```
iNecas: Since it's in `Candlepin` namespace, the `Candlepin` name here is unnecessary.

definitions/checks/candlepin/validate_candlepin_db.rb
```diff
@@ -0,0 +1,17 @@
...
+    metadata do
+      description 'Check to validate candlepin database'
+      tags :pre_upgrade
+      label :validate_candlepin_db
```
iNecas: I'm for the lavel reflecting the naming of the class, so `:candlepin_validate_db` would be better for me: it helps seeing the checks grouped by component, when calling `foreman-maintain health list`

iNecas:
I want to give this a try in production: @kgaikwad could you give me some example of an sql command to introduce some inconsinstency?

kgaikwad:
@iNecas,

You can create dummy records either in cp_pool table or cp_env_content table.
For example:

1. `contentid` should be invalid in below sql query
```
insert into cp_env_content(id, contentid, environment_id) VALUES ('ff8080815a25a54b3', '8763863586325', 1);

# Run below sql to check -

SELECT e.id, ec.contentid
    FROM cp_environment e
     JOIN cp_env_content ec ON e.id = ec.environment_id
    LEFT JOIN cp_content c ON c.id = ec.contentid
    WHERE c.id IS NULL;  
```

2. `productid` should be invalid in below sql query
Note that - Please check column name because in latest version column name changed to `product_uuid`

```
insert into cp_pool (id, productid, productname,  quantity, version, owner_id, enddate, startdate) VALUES ('ff8080815a25a54b015a2695d3530d2b', 'UN1502908878099', 'test_pp', -1, 0, '8aca0c8257429f86015742a2ec4b0001', '2047-08-09 14:41:18.344-04', '2017-08-16 14:41:18.344-04');
```

kgaikwad:
@iNecas, 
Done with suggested changes. 
Pull request ready for review.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -138,6 +138,14 @@ def shellescape(string)
       def version(value)
         Version.new(value)
       end
+
+      def format_shell_args(options = {})
+        cmd = ''
```
iNecas: What about using inject for that:

```ruby
options.inject('') do |cmd, (shell_optn, val)|
  cmd += " #{shell_optn} '#{shellescape(val)}'"
end
```

or `map` + `join`:

```ruby
options.map { |shell_optn, val| "#{shell_optn} '#{shellescape(val)}'" }.join(' ')
```

iNecas:
Tested and works well. One nitpick

definitions/features/candlepin_database.rb
```diff
@@ -0,0 +1,62 @@
...
+  def check_option_using_cpdb_help(option_name, parent_cmd = '')
+    parent_cmd = '/usr/share/candlepin/cpdb' if parent_cmd.empty?
+    help_cmd = "#{parent_cmd} --help |  grep -c '\\-\\-#{option_name} '"
+    execute?(help_cmd)
+  end
+
+  def execute_cpdb_validate_cmd
+    main_cmd = cpdb_validate_cmd
+    unless main_cmd.empty?
+      main_cmd += format_shell_args(
```
iNecas: Missing space here, leads to `no such option: --verbose-u`

kgaikwad: @iNecas,

Yes, sorry my bad. I have tested code on console only.
Corrected and updated the pull-request.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -138,6 +138,10 @@ def shellescape(string)
       def version(value)
         Version.new(value)
       end
+
+      def format_shell_args(options = {})
+        options.map { |shell_optn, val| " #{shell_optn} '#{shellescape(val)}'" }.join(' ')
```
iNecas: nitpick: in this case, we don't need `' '` in `join` anymore: so it should be 

```ruby
options.map { |shell_optn, val| " #{shell_optn} '#{shellescape(val)}'" }.join
```

iNecas:
Tested and works great, thanks @kgaikwad

PR #141, opened 2018-01-29
Fixes #22341 - no more special case for 6.3, use the standard repos

mccun934:


mccun934:
Once we get to 6.4 we may want to revisit having a standard procedure for handling a 'beta' version

iNecas:
We were to talking with @swapab about doing this actually as part of move to GA. Swapnil: any progress on checking the available repositories and deciding what repositories to enable based on what repositories are available?

mccun934:
@iNecas the beta repos will always be available, even after GA, they will *most likely* be empty but we can't necessarily count on that

iNecas:
I meant the other way arround: once the GA repos are there, not using the beta.

swapab:
@iNecas This is an attempt to generalize enabling of necessary repositories.
https://github.com/swapab/foreman_maintain/commit/a8415a3455d886982afeddb5c749723278f4c55e

---
Tested with command
`./bin/foreman-maintain advanced procedure run repositories-setup --version 6.2`
and
`./bin/foreman-maintain advanced procedure run repositories-setup --version 6.3`

Previous one enables relevant 6.2 repos for satellite and satellite-tools and later one enables beta repos since 6.3 is not yet GA.
Maintenance repo being `rhel-7-server-satellite-maintenance-6-rpms`(always as of now).

iNecas:
@swapab the only possible caveat of this one is that case, when the redhat.repo was not refreshed yet, so it would not know about the existing repositories. I'm afraid your code would not handle this situation. I guess we would need to call `subscription-manager repos` always and then cache the value for later use. I guess we will need more time for tune this code out. And we will need to merge this one for now.

iNecas:
Thanks @mccun934

swapab:
@iNecas Yes, that does make sense. We'll revisit the generic way later. Thanks!

PR #142, opened 2018-02-15
Fixes #22578 - force fail self registered servers

swapab:


definitions/checks/system_registration.rb
```diff
@@ -5,15 +5,19 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
 
     confine do
-      file_exists?('/etc/rhsm/rhsm.conf') &&
-        !feature(:foreman_server) &&
-        feature(:foreman_proxy)
+      file_exists?('/etc/rhsm/rhsm.conf')
     end
+
+    message <<-MESSAGE
+      https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3-beta/html/installation_guide/preparing_your_environment_for_installation#system_requirements
```
kgaikwad: Since these are Beta docs, those links will not be valid after GA release. 
@swapab, you need to change those links.

swapab: Yes. Above links won't work since now 6.3 is GA

definitions/checks/system_registration.rb
```diff
@@ -5,15 +5,19 @@ class Checks::SystemRegistration < ForemanMaintain::Check
     tags :default
 
     confine do
-      file_exists?('/etc/rhsm/rhsm.conf') &&
-        !feature(:foreman_server) &&
-        feature(:foreman_proxy)
+      file_exists?('/etc/rhsm/rhsm.conf')
     end
+
+    message <<-MESSAGE
```
swapab: @iNecas @kgaikwad  Let me know what you think of this new metadata. If it sounds good I'll create a new PR for `message` metadata and not piggyback it with this PR(unless piggybacking is Okay)

lib/foreman_maintain/reporter/resolution_reporter.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      def type
+        if status == :error
+          'failed'
+        else
+          'warned'
+        end
+      end
+
+      def messages?
```
kgaikwad: By convention, methods ending in ? return a boolean. 
Instead of calling `empty?` method on `messages?` below, could you please modify method `messages?`
so that `empty?` condition will be handled by this method.

definitions/checks/system_registration.rb
kgaikwad: @swapab,
In my opinion it is good to have but `message` key-name will be very generalized one. 
Naming of this metadata i.e `message` may create a confusion while understanding a purpose of it.

lib/foreman_maintain/reporter/resolution_reporter.rb
swapab: Ah. right, fixed!

definitions/checks/system_registration.rb
swapab: @kgaikwad 
Yes, that's right. I thought of it. But couldn't find a suitable name for it.
`failure_message` but the text can be shown for a warning as well as a failure check.
`notification_message` doesn't sound relevant either
`resolution_message` close enough(but not sure if we are showing a resolution or just wanted to tell the administrator something related to failed/warned check)

Any name you have in mind?

ntkathole:
@swapab While testing this PR, I failed the check for verify-self-registration. But still it gives me option to whitelist and perform the upgrade. I was able to perform the upgrade as well.

Looking into BZ https://bugzilla.redhat.com/show_bug.cgi?id=1545136, it states that it should not allow upgrades of self-registered Satellites.  Am I missing something here?

```
Check for verifying syntax for ISP DHCP configurations:               [FAIL]
undefined method `strip' for nil:NilClass
Check whether system is self-registered or not:                       [FAIL]
System is self registered

  [foreman-proxy-verify-dhcp-config-syntax]
  [verify-self-registration]

use --whitelist="foreman-proxy-verify-dhcp-config-syntax,verify-self-registration"

========== Please follow steps to resolve failed check(s) ==========
  [verify-self-registration]

      https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/installation_guide/preparing_your_environment_for_installation#syste
m_requirements

      https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite
#migrating_a_self_registered_satellite

====================================================================


```

lib/foreman_maintain/reporter/resolution_reporter.rb
iNecas: Was this really fixed?

lib/foreman_maintain/reporter/resolution_reporter.rb
```diff
@@ -0,0 +1,66 @@
...
+        @messaged ||= steps.map(&:message).join
```
iNecas: Typo

definitions/checks/system_registration.rb
iNecas: I'm not sure it's a good place fore metadata. Besides the name, there is also issue when we wanted to update the content on upstream vs. downstream. I would probably like this more passing as argument to `fail!` or `assert` methods

lib/foreman_maintain/reporter/resolution_reporter.rb
swapab: My original VM got crashed. Changes overwrote because of force push on new VM(which was not updated). Fixed for real now :smile:

definitions/checks/system_registration.rb
swapab: A `yaml` file(2 actually one for downstream and one for upstream)? to map failure and/or warning messages per definition.

iNecas: It would mean decoupling the messages from the checks, which would be confusing (at least for me). I would like to keep that in one code.

swapab: Okay. I have dropped the thought of a new `metadata`. And used existing `fail!` to show the documentation notification message.

definitions/checks/system_registration.rb
```diff
@@ -5,15 +5,26 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
+        notification_message = <<-MESSAGE
+  Satellite is self registered
+
+  Self registered satellite won't be supported as of Satellite version 6.3. Please follow following documentation to resolve.
+
+  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/installation_guide/preparing_your_environment_for_installation#system_requirements
+
+  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite#migrating_a_self_registered_satellite
+    MESSAGE
+        fail!(notification_message)
```
ekohl: I think it's cleaner to move this `fail!()` out of the if/else statement. Just set notification_message in both branches (but different content). That way it's clear you always fail, regardless of upstream/downstream.

definitions/checks/system_registration.rb
```diff
@@ -5,15 +5,26 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
   end
 
   def run
     if rhsm_hostname_eql_hostname?
-      warn! 'System is self registered'
+      if feature(:downstream)
+        notification_message = <<-MESSAGE
+  Satellite is self registered
+
+  Self registered satellite won't be supported as of Satellite version 6.3. Please follow following documentation to resolve.
```
sean797: s/won't be/are not/

definitions/checks/system_registration.rb
```diff
@@ -5,15 +5,26 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
+
+  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/installation_guide/preparing_your_environment_for_installation#system_requirements
```
sean797: Is this link really needed? Giving the user 2 links to could be confusing IMO. We have already told them it's not supported, let's just give them the link below that tells them how to migrate?

definitions/checks/system_registration.rb
sean797: s/Please follow following/Please follow **the** following/

sean797: s/following documentation to resolve/following documentation to migrate/

Not required, but I think migrate reads better.

kgaikwad:
Closing in favor of #192

PR #145, opened 2018-02-22
Fixes #22663 - handled nil.strip for DHCP check

kgaikwad:


kgaikwad:
@swapab & @iNecas,

Fixed the DHCP issue for nil.strip. 
During testing I found that it was failing for positive scenario as well. 
Did few modifications in methods.

definitions/features/foreman_proxy.rb
```diff
@@ -34,16 +34,17 @@ def find_http_error_msg(array_output, curl_http_status)
...
   end
 
   def run_dhcp_curl
     curl_resp = execute(dhcp_curl_cmd)
-    array_output = curl_resp.split(/\r\n/)
+    array_output = curl_resp.split(/\r?\n/)
     result_array = array_output.last.split(/\n/)
     curl_http_status = result_array.delete_at(result_array.length - 1).strip.to_i
-    curl_http_resp = parse_json(result_array.join(''))
+    curl_http_resp = parse_json(array_output[0])
```
iNecas: Just in case, shouoln't we also check that `array_output` is not empty?

kgaikwad: @iNecas, yes we can.
But array_output contains result and http status code. Result can be empty but array_output will always have http status code.
For empty result + status code 200, we are raising failure.

definitions/features/foreman_proxy.rb
```diff
@@ -34,16 +34,17 @@ def find_http_error_msg(array_output, curl_http_status)
...
+      return http_line if http_line.empty?
       http_line.split(curl_http_status.to_s).last.strip
     end
   end
 
   def run_dhcp_curl
     curl_resp = execute(dhcp_curl_cmd)
-    array_output = curl_resp.split(/\r\n/)
+    array_output = curl_resp.split(/\r?\n/)
     result_array = array_output.last.split(/\n/)
```
iNecas: I don't understand this code: why are we doing split on `/\n/` for something that was just split based on that on the previous line. 

```ruby
"one\ntwo\nthree".split(/\r?\n/)
# => ["one", "two", "three"]
```
In which case the `array_output.last` can actually contain `\n`?

definitions/features/foreman_proxy.rb
iNecas: ok

definitions/features/foreman_proxy.rb
kgaikwad: yes, I will modify this.

kgaikwad: @iNecas ,
Corrected the code. Ready for review.

kgaikwad:
@iNecas,

Updated PR as per suggested changes. Ready for review.

ntkathole:
LGTM

For positive scenario,
```
./bin/foreman-maintain health check
```
If katello-service stopped,
```
Check for verifying syntax for ISP DHCP configurations:
No valid HTTP response (Connection failed) 0. Response: ""
              [FAIL]
Please check and verify DHCP configurations.
Check for paused tasks:                                               [FAIL]
Illegal quoting in line 2.
Check whether all services are running using hammer ping:             [FAIL]
Error: Failed to open TCP connection to <satellite-hostname>:443 (Connection refused - connect(2) for "<satellite-hostname>" port 443)

```

iNecas:

PR #146, opened 2018-02-22
Add FreeBSD support to passenger-recycler

karli-sjoberg:
Tested and working on FreeBSD-11.1-RELEASE

### DIFF
--- a/bin/passenger-recycler
+++ b/bin/passenger-recycler
@@ -46,71 +46,44 @@ require 'phusion_passenger/platform_info'
 require 'phusion_passenger/platform_info/ruby'
 require 'phusion_passenger/admin_tools/memory_stats'
 stats = PhusionPassenger::AdminTools::MemoryStats.new
-
-case PhusionPassenger::PlatformInfo.os_name_simple
-when "linux"
-  unless stats.platform_provides_private_dirty_rss_information?
-    puts 'Please run as root or platform unsupported'
-    exit 1
-  end
-  killed = 0
-  stats.passenger_processes.each do |p|
-    pid = p.pid.to_i
-    debug "Checking #{pid} with RSS of #{p.private_dirty_rss}"
-    next unless p.private_dirty_rss > CONFIG[:MAX_PRIV_RSS_MEMORY]
-    started = begin
+killed = 0
+stats.passenger_processes.each do |p|
+  pid = p.pid.to_i
+  started = begin
               `ps -p#{pid} -o start=`.strip
             rescue StandardError => e
               verbose "Error: #{e.message} \nReturning '?'"
               '?'
             end
-    status_ps = `ps -p#{pid} -u`
-    status_all = `passenger-status 2> /dev/null`
-    status_backtraces = `passenger-status --show=backtraces 2>/dev/null`
-    verbose "Terminating #{pid} (started #{started}) with private dirty RSS" \
-          " size of #{p.private_dirty_rss} MB"
-    Process.kill 'SIGUSR1', pid
-    sleep CONFIG[:GRACEFUL_SHUTDOWN_SLEEP]
-    kill(pid)
-    process_status?(pid)
-    if CONFIG[:SEND_STATUS]
-      verbose status_ps
-      verbose status_all
-      verbose status_backtraces
-    end
-    killed += 1
-    exit(1) if killed >= CONFIG[:MAX_TERMINATION]
+  if not p.private_dirty_rss.nil?
+    get_pid_mem_kb = p.private_dirty_rss
+    get_pid_mem_mb = p.private_dirty_rss_in_mb
+    debug_message = "Checking #{pid} with RSS of #{get_pid_mem_kb}"
+    verbose_message = "Terminating #{pid} (started #{started}) with private" +
+        " dirty RSS size of #{get_pid_mem_mb} MB"
+  else
+    get_pid_mem_kb = p.rss
+    get_pid_mem_mb = p.rss_in_mb
+    debug_message = "Checking #{pid} with RSS of #{get_pid_mem_kb}"
+    verbose_message = "Terminating #{pid} (started #{started}) RSS size of" +
+        " #{get_pid_mem_mb} MB"
   end
-when "freebsd"
-  killed = 0
-  stats.passenger_processes.each do |p|
-    pid = p.pid.to_i
-    pid_rss = `procstat -r #{pid} | awk '/maximum RSS/{print$5}'`.strip.to_i
-    pid_rss_mb = sprintf("%.1f", pid_rss / 1024.0)
-    debug "Checking #{pid} with RSS of #{pid_rss}"
-    next unless pid_rss > CONFIG[:MAX_PRIV_RSS_MEMORY]
-    started = begin
-              `ps -p#{pid} -o start=`.strip
-            rescue StandardError => e
-              verbose "Error: #{e.message} \nReturning '?'"
-              '?'
-            end
-    status_ps = `ps -p#{pid} -u`
-    status_all = `passenger-status 2> /dev/null`
-    status_backtraces = `passenger-status --show=backtraces 2>/dev/null`
-    verbose "Terminating #{pid} (started #{started}) with RSS size of" \
-            " #{pid_rss_mb} MB"
-    Process.kill 'SIGUSR1', pid
-    sleep CONFIG[:GRACEFUL_SHUTDOWN_SLEEP]
-    kill(pid)
-    process_status?(pid)
-    if CONFIG[:SEND_STATUS]
-      verbose status_ps
-      verbose status_all
-      verbose status_backtraces
-    end
-    killed += 1
-    exit(1) if killed >= CONFIG[:MAX_TERMINATION]
+  debug debug_message
+  next unless get_pid_mem_kb > CONFIG[:MAX_PRIV_RSS_MEMORY]
+  status_ps = `ps -p#{pid} -u`
+  status_all = `passenger-status 2> /dev/null`
+  status_backtraces = `passenger-status --show=backtraces 2>/dev/null`
+  verbose verbose_message
+  Process.kill 'SIGUSR1', pid
+  sleep CONFIG[:GRACEFUL_SHUTDOWN_SLEEP]
+  kill(pid)
+  process_status?(pid)
+  if CONFIG[:SEND_STATUS]
+    verbose status_ps
+    verbose status_all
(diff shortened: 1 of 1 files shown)


bin/passenger-recycler
```diff
@@ -41,41 +41,76 @@ def process_status?(pid)
...
+      verbose status_backtraces
+    end
+    killed += 1
+    exit(1) if killed >= CONFIG[:MAX_TERMINATION]
+  end
+when "freebsd"
+  killed = 0
+  stats.passenger_processes.each do |p|
+    pid = p.pid.to_i
+    pid_rss = `procstat -r #{pid} | awk '/maximum RSS/{print$5}'`.strip.to_i
```
ekohl: Could this be extracted to a ```get_pid_mem_mb``` so we can reduce the amount of duplication?

karli-sjoberg: Absolutely! I´ve also noticed that there are functions I overlooked in phusion_passenger that negates the the need to use specific shell commands, e.g: 

    pid_rss = p.rss.to_i
    pid_rss_mb = p.rss_in_mb

Which where the only variables I really needed. Beware though, that the debug and verbose messages are also different, and needs to change, since on FreeBSD, I´m not using "private dirty RSS", only RSS!

karli-sjoberg: Would you like me to make a new pull request like that instead?

iNecas:
Rubocop is complaining about some style issues, please have a look at https://travis-ci.org/theforeman/foreman_maintain/jobs/345129122

iNecas:
Also, please follow the hint @theforeman-bot gives about the commit message: I've created a redmine issue for you here http://projects.theforeman.org/issues/22667

karli-sjoberg:
@iNecas Thank you for creating a Redmine issue for me!

About the rubocop style issues, let me address them one by one.
1)
    bin/passenger-recycler:50:1: C: Block has too many lines. [37/25]
    stats.passenger_processes.each do |p| ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
There´s little to do about that, since you wanted code deduplication, it had to live inside the block to fetch the variables from the Passenger objects.

2)
    bin/passenger-recycler:58:6: C: Use ! instead of not.
      if not p.private_dirty_rss.nil?
         ^^^
Easy fix, I can commit that in a second.

3,4,5,6)

    bin/passenger-recycler:62:78: C: Use \ instead of + or << to concatenate those strings.
        verbose_message = "Terminating #{pid} (started #{started}) with private" +
                                                                             ^
    bin/passenger-recycler:63:9: C: Align the operands of an expression in an assignment spanning multiple lines.
            " dirty RSS size of #{get_pid_mem_mb} MB"
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    bin/passenger-recycler:68:77: C: Use \ instead of + or << to concatenate those strings.
        verbose_message = "Terminating #{pid} (started #{started}) RSS size of" +
                                                                            ^
    bin/passenger-recycler:69:9: C: Align the operands of an expression in an assignment spanning multiple lines.
            " #{get_pid_mem_mb} MB"
            ^^^^^^^^^^^^^^^^^^^^^^^

Hm, wierd! I tried to have it just like that but then the script crashed, but now it works. Nevermind, I fixed it!

bin/passenger-recycler
```diff
@@ -41,31 +41,39 @@ def process_status?(pid)
...
+    debug_message = "Checking #{pid} with RSS of #{get_pid_mem_kb}"
+    verbose_message = "Terminating #{pid} (started #{started}) with private" \
+                      " dirty RSS size of #{get_pid_mem_mb} MB"
+  else
+    get_pid_mem_kb = p.rss
+    get_pid_mem_mb = p.rss_in_mb
+    debug_message = "Checking #{pid} with RSS of #{get_pid_mem_kb}"
+    verbose_message = "Terminating #{pid} (started #{started}) RSS size of" \
+                      " #{get_pid_mem_mb} MB"
+  end
```
lzap: This could be easily written as

    get_pid_mem_kb = p.try(:private_dirty_rss) || p.rss

and then the whole if statement would not be necessary.

bin/passenger-recycler
```diff
@@ -41,31 +41,39 @@ def process_status?(pid)
...
 stats.passenger_processes.each do |p|
   pid = p.pid.to_i
-  debug "Checking #{pid} with RSS of #{p.private_dirty_rss}"
-  next unless p.private_dirty_rss > CONFIG[:MAX_PRIV_RSS_MEMORY]
   started = begin
               `ps -p#{pid} -o start=`.strip
             rescue StandardError => e
               verbose "Error: #{e.message} \nReturning '?'"
               '?'
             end
```
lzap: Move this `ps` call back after `next unless` so it does not need to be call every loop.

bin/passenger-recycler
iNecas: No, `try` is coming from rails, so it's not available here.

bin/passenger-recycler
```diff
@@ -41,31 +41,39 @@ def process_status?(pid)
 end
 
 require 'phusion_passenger'
+PhusionPassenger.locate_directories
 require 'phusion_passenger/platform_info'
 require 'phusion_passenger/platform_info/ruby'
 require 'phusion_passenger/admin_tools/memory_stats'
-PhusionPassenger.locate_directories
```
lzap: This change needs to be tested on SCL, I remember this was pain to get it working on both SCL and non-SCL.

iNecas:
The length of the block could be reduced by extracting some code into separate methods, something like this (haven't tested):

```patch
diff --git a/bin/passenger-recycler b/bin/passenger-recycler
index cd464b1..a584c1b 100755
--- a/bin/passenger-recycler
+++ b/bin/passenger-recycler
@@ -47,27 +47,32 @@ require 'phusion_passenger/platform_info/ruby'
 require 'phusion_passenger/admin_tools/memory_stats'
 stats = PhusionPassenger::AdminTools::MemoryStats.new
 killed = 0
-stats.passenger_processes.each do |p|
-  pid = p.pid.to_i
-  started = begin
-              `ps -p#{pid} -o start=`.strip
-            rescue StandardError => e
-              verbose "Error: #{e.message} \nReturning '?'"
-              '?'
-            end
-  if ! p.private_dirty_rss.nil?
-    get_pid_mem_kb = p.private_dirty_rss
-    get_pid_mem_mb = p.private_dirty_rss_in_mb
-    debug_message = "Checking #{pid} with RSS of #{get_pid_mem_kb}"
-    verbose_message = "Terminating #{pid} (started #{started}) with private" \
-                      " dirty RSS size of #{get_pid_mem_mb} MB"
+
+def get_process_start(process)
+  `ps -p#{process.pid.to_i} -o start=`.strip
+rescue StandardError => e
+  verbose "Error: #{e.message} \nReturning '?'"
+  '?'
+end
+
+def get_rss_info(process)
+  if !p.private_dirty_rss.nil?
+    get_pid_mem_kb = process.private_dirty_rss
+    get_pid_mem_mb = process.private_dirty_rss_in_mb
   else
     get_pid_mem_kb = p.rss
     get_pid_mem_mb = p.rss_in_mb
-    debug_message = "Checking #{pid} with RSS of #{get_pid_mem_kb}"
-    verbose_message = "Terminating #{pid} (started #{started}) RSS size of" \
-                      " #{get_pid_mem_mb} MB"
   end
+  [get_pid_mem_kb, get_pid_mem_mb]
+end
+
+stats.passenger_processes.each do |p|
+  pid = p.pid.to_i
+  started = get_process_start(p)
+  get_pid_mem_kb, get_pid_mem_mb = get_rss_info(p)
+  debug_message = "Checking #{pid} with RSS of #{get_pid_mem_kb}"
+  verbose_message = "Terminating #{pid} (started #{started}) with private" \
+                    " RSS size of #{get_pid_mem_mb} MB"
   debug debug_message
   next unless get_pid_mem_kb > CONFIG[:MAX_PRIV_RSS_MEMORY]
   status_ps = `ps -p#{pid} -u`
```

It might also help with readability a bit (that's why rubocop has the rule on the length of the block in the first place)

bin/passenger-recycler
karli-sjoberg: Well, I know I need it for the script to work at all. I get strange issues without this, like you can't load Passenger libraries because it can't find it :)

'bin/passenger-recycler'

    require 'phusion_passenger/platform_info'

'phusion_passenger/platform_info.rb'

    PhusionPassenger.require_passenger_lib 'utils/tmpio'

= fail, without it...

karli-sjoberg:
@iNecas That looks great, I'll test that and commit on success :)

bin/passenger-recycler
lzap: No problem, this was a message for myself to test this with and without SCL.

bin/passenger-recycler
lzap: Well

    get_pid_mem_kb = p.private_dirty_rss || p.rss

then, it is set to `nil` according to the if statement there, should work. I still think this is too complicated code, four lines copied...

karli-sjoberg: It actually gets set to `?` instead of `nil`, so that doesn´t work, I tested it just now. Your ruby-foo is obviously better than mine, but unless you have another equally elegant solution up your sleaves, that´s that way it´s got to be.

karli-sjoberg: I got it working, but I had to write `get_pid_mem_mb` like this:

    get_pid_mem_mb = sprintf("%.0f", get_pid_mem_kb / 1024)

Because `process.rss_in_mb` for some reason returned `?` instead of `nil`

karli-sjoberg: `phusion_passenger/admin_tools/memory_stats.rb`

        def private_dirty_rss_in_mb
          if private_dirty_rss.is_a?(Numeric)
            return sprintf("%.1f MB", private_dirty_rss / 1024.0)
          else
            return "?"
          end
        end

Funky :grinning:

karli-sjoberg:
Hey fellas!

So seeing as there were no more objections, how about getting a merge on this sucker?

/K

iNecas:
Tested and work well for me: the only thing is the rubocop, that started complaining again. Please have a look at the failures: I'm ready to merge after that is resolved.

iNecas:
Thanks @karli-sjoberg

karli-sjoberg:
@iNecas No, thank you! :grinning:

lzap:
Thanks!

PR #148, opened 2018-02-26
Fixes #22686 - subscription-manager refresh to get new repos

kgaikwad:


definitions/features/downstream.rb
```diff
@@ -23,6 +23,7 @@ def setup_repositories(version)
       execute!(%(subscription-manager register #{org_options}\
                   --activationkey #{shellescape(activation_key)} --force))
     else
+      execute!(%(subscription-manager refresh))
```
kgaikwad: @iNecas,

While sat 6.3 upgrading, users are facing issue to get new repositories.
This is the common method we have used for repository setup.

Please confirm - is it ok to add this step without a condition `version` .
or should I add condition of sat 6.3 for this step.

evgeni: I think it does not hurt to always run it.

PR #149, opened 2018-02-27
Add context into scenario

mbacovsky:
Context adds to scenario an ability to distribute own parameters
to included executables. Scenario accepts parameters and keeps mapping
of the params to params of the executables. The mapping is opt-in
so each executable gets just the correct params.

For the usage see the scenario_test

### DIFF
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -20,6 +20,7 @@ module ForemanMaintain
   require 'foreman_maintain/top_level_modules'
   require 'foreman_maintain/yaml_storage'
   require 'foreman_maintain/config'
+  require 'foreman_maintain/context'
   require 'foreman_maintain/detector'
   require 'foreman_maintain/dependency_graph'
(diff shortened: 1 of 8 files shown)


mbacovsky:
I didn't add the context to the `PreparationScenario` and `FilteredScenario` as it was not needed yet. I can add if it helps.

For another usage example you can see my branch with backup command https://github.com/mbacovsky/foreman_maintain/blob/backup_command/definitions/scenarios/backup.rb
 and https://github.com/mbacovsky/foreman_maintain/blob/backup_command/lib/foreman_maintain/cli/backup_command.rb#L27-L39

johnpmitsch:
The context passing is working fine in my development using this PR :+1:

iNecas:
I have not found any comment I could ask to improve (and boy I tried:). Thanks @mbacovsky and @johnpmitsch for testing. Since this is not influencing any of the current functionality, I have no issues with merging this right away and see that grow in further PRs

PR #150, opened 2018-03-01
Fixes #22687 - Add openscap report check and procedure

xprazak2:
wip as I need to test in production

### DIFF
--- a/definitions/procedures/foreman_openscap/invalid_report_associations.rb
+++ b/definitions/procedures/foreman_openscap/invalid_report_associations.rb
@@ -1,0 +1,15 @@
+module Procedures::ForemanOpenscap
+  class InvalidReportAssociations < ForemanMaintain::Procedure
+    metadata do
+      param :ids_to_remove, 'Ids of reports to remove', :required => true
+      for_feature :foreman_openscap
+      tags :pre_migration, :foreman_openscap, :openscap_report_associations
+      advanced_run false
+      description 'Delete reports with association issues'
+    end
+
+    def run
+      feature(:foreman_openscap).delete_reports(@ids_to_remove)
+    end
+  end
+end
--- a/definitions/checks/foreman_openscap/invalid_report_associations.rb
+++ b/definitions/checks/foreman_openscap/invalid_report_associations.rb
@@ -1,0 +1,27 @@
+module Checks
+  module ForemanOpenscap
+    class InvalidReportAssociations < ForemanMaintain::Check
+      metadata do
+        label :openscap_report_associations
+        for_feature :foreman_openscap
+        description 'Check whether reports have correct associations'
+        tags :pre_upgrade, :foreman_openscap, :openscap_report_associations
+      end
+
+      def run
+        ids_to_remove = to_remove
+        assert(ids_to_remove.empty?,
+               "There are #{ids_to_remove.count} reports with issues that will be removed",
+               :next_steps => Procedures::ForemanOpenscap::InvalidReportAssociations.new(
+                 :ids_to_remove => ids_to_remove
+               ))
+      end
+
+      def to_remove
+        (feature(:foreman_openscap).report_ids_without_policy +
+         feature(:foreman_openscap).report_ids_without_proxy +
+         feature(:foreman_openscap).report_ids_without_host).uniq
+      end
+    end
+  end
+end
--- a/definitions/features/foreman_openscap.rb
+++ b/definitions/features/foreman_openscap.rb
@@ -1,0 +1,57 @@
+class Features::ForemanOpenscap < ForemanMaintain::Feature
+  metadata do
+    label :foreman_openscap
+
+    confine do
+      check_min_version('tfm-rubygem-foreman_openscap', '0.5.3')
+    end
+  end
+
+  def report_ids_without_host
+    reports_without_attribute('host_id')
+  end
+
+  def report_ids_without_proxy
+    reports_without_attribute('openscap_proxy_id')
+  end
+
+  def report_ids_without_policy
+    sql = <<-SQL
+      SELECT id
+      FROM reports
+      WHERE id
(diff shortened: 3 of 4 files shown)


definitions/checks/foreman_openscap/report_associations.rb
```diff
@@ -0,0 +1,29 @@
+module Checks
+  module ForemanOpenscap
+    class ReportAssociations < ForemanMaintain::Check
+      metadata do
+        label :report_associations
```
iNecas: I would suggest using `:openscap_` prefix in the label: it makes it easier when listing the checks in `fm health list` command

definitions/checks/foreman_openscap/report_associations.rb
```diff
@@ -0,0 +1,29 @@
...
+        for_feature :foreman_openscap
+        description 'Check whether reports have correct associations'
+        tags :pre_upgrade, :foreman_openscap, :report_associations
+
+        confine do
```
iNecas: When using `for_feature` the `confine` is not needed

definitions/checks/foreman_openscap/report_associations.rb
```diff
@@ -0,0 +1,29 @@
+module Checks
+  module ForemanOpenscap
+    class ReportAssociations < ForemanMaintain::Check
```
iNecas: The check should describe what's wrong about the state, if it fails: what about something like `InvalidReportAssociations`

definitions/procedures/foreman_openscap/report_associations.rb
```diff
@@ -0,0 +1,15 @@
+module Procedures::ForemanOpenscap
+  class ReportAssociations < ForemanMaintain::Procedure
```
iNecas: Similarly to the checks naming, I would suggest `ClearReportAssociations`

ntkathole:
```
 ./bin/foreman-maintain advanced procedure run foreman-openscap-report-associations
Delete reports with association issues:                               [FAIL]
undefined method `join' for nil:NilClass

  [foreman-openscap-report-associations]

use --whitelist="foreman-openscap-report-associations"
```

xprazak2:
I made changes as suggested. I am not entirely sure what causes the nil issue as I cannot reproduce, but I made an extra precaution for that as well.

ntkathole:
@xprazak2 I still see the same issue o sat62 and sat63. Let me know if you need setup.
```
E, [2018-03-18 03:39:41-0400 #25968] ERROR -- : undefined method `join' for nil:NilClass (NoMethodError)
/root/foreman_maintain/definitions/features/foreman_openscap.rb:37:in `delete_reports'
/root/foreman_maintain/definitions/procedures/foreman_openscap/invalid_report_associations.rb:12:in `run'
/root/foreman_maintain/lib/foreman_maintain/executable.rb:119:in `__run__'
/root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:76:in `block (2 levels) in run'
/root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:99:in `capture_errors'
/root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:75:in `block in run'
/root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:93:in `with_metadata_calculation'
/root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:74:in `run'
/root/foreman_maintain/lib/foreman_maintain/runner.rb:109:in `run_step'
/root/foreman_maintain/lib/foreman_maintain/runner.rb:98:in `run_steps'
/root/foreman_maintain/lib/foreman_maintain/runner.rb:91:in `execute_scenario_steps'
/root/foreman_maintain/lib/foreman_maintain/runner.rb:45:in `run_scenario'
```

definitions/procedures/foreman_openscap/invalid_report_associations.rb
```diff
@@ -0,0 +1,15 @@
+module Procedures::ForemanOpenscap
+  class InvalidReportAssociations < ForemanMaintain::Procedure
+    metadata do
+      param :ids_to_remove, 'Ids of reports to remove'
```
kgaikwad: @xprazak2 ,

param ` :ids_to_remove` is required to run this procedure so you could mention `:required => true` to param declaration.

xprazak2:
The nil issue was caused by running the procedure directly through `advanced procedure run`, but it requires a list of ids which are supplied by check. I added `advanced_run false` to disable running it directly. Thanks @kgaikwad!

ntkathole:
Tested on sat 6.2 : 
Steps:
1. Created scap report with policy
2. deleted policy

Before performing upgrade, it asks to delete orphaned reports. But my concern is if we are handling the same in installer as well https://bugzilla.redhat.com/show_bug.cgi?id=1547607#c16 , @xprazak2 @iNecas shall we still have this check in foreman-maintain ?

xprazak2:
Partially, the migration in bz just removes reports without policy. Here we check for other associations as well. The migration runs as a part of upgrade and only once, having this in foreman_maintain provides value to users outside of upgrade scenario.

xprazak2:
@iNecas, any additional comments?

iNecas:
Last question: should this apply to any upgrade with 'tfm-rubygem-foreman_openscap' >= '0.5.3', or is this specific let's say for sat6.3 -> sat6.4 but would not be applicable for sat6.4 -> sat6.5?

xprazak2:
People were having problems on 6.2, so this is mostly relevant for 6.2 -> 6.3. I do not expect this problem in later versions, steps were taken to prevent this from happening again. So not too relevant for the 6.3 -> 6.4 and later, but the check and procedure can be run manually as well if the users need it.

iNecas:
Ok, let's pull it in then. Thanks @xprazak2

PR #151, opened 2018-03-01
Fixes #22458 - Add Katello service command

johnpmitsch:
Todo
- [x] write tests
- [x] test on standalone foreman and foreman proxy w/ content


johnpmitsch:
@iNecas can you move this ticket?

definitions/features/katello.rb
```diff
@@ -14,4 +14,29 @@ def data_dirs
...
+      'pulp_resource_manager'    => 20,
+      'pulp_streamer'            => 20,
+      'foreman-proxy'            => 20,
+      'smart_proxy_dynflow_core' => 20,
+      'httpd'                    => 30,
+      #'puppetserver'             => 30,
+      'foreman-tasks'            => 30,
+      'goferd'                   => 30
+    }
+  end
```
johnpmitsch: @iNecas @mbacovsky Should I split up these services into different features? They are being selected based if they [exist on the sytem](https://git.io/vADZI) We could keep one master list of services and it shouldn't make a difference if its standalone foreman/foreman-proxy/etc. Curious to hear your thoughts.

lib/foreman_maintain/cli/service_command.rb
```diff
@@ -0,0 +1,125 @@
...
+  module Cli
+    class ServiceCommand < Base
+      option '--exclude', "EXCLUDE", 'A comma-separated list of services to skip'
+      option '--only', "ONLY", 'A comma-separated list of services to include'
+
+      subcommand 'start', 'Start applicable services' do
+        # TODO: DRY up labels
+        label_option
+        tags_option
+        interactive_option
```
johnpmitsch: what is the best way to not repeat these option methods each subcommand?

definitions/procedures/services/list.rb
```diff
@@ -0,0 +1,21 @@
+#TODO: figure out why I need this?
+require_relative 'base.rb'
```
johnpmitsch: @iNecas @mbacovsky Is there a better way to do this? I would have thought the base class would be required already here, but it wasn't. Maybe I am missing something obvious  :man_shrugging:

definitions/features/katello.rb
johnpmitsch: If we do go with one list, I also am wondering where the best home for it would be :smile:

lib/foreman_maintain/cli/service_command.rb
johnpmitsch: Also do I need all of these? If I didn't include them I got an error (something about "assumeyes" not found iirc), but not sure if I  need all of them

definitions/procedures/services/list.rb
kgaikwad: @johnpmitsch,

Yes, you don't need to require base class here. Instead of Base class, you just need to inherit List class from ForemanMaintain::Procedure. Example - 

```
module Procedures::Services
  class List < ForemanMaintain::Procedure
       # code
  end
```

lib/foreman_maintain/cli/service_command.rb
kgaikwad: Currently, we don't have any way to avoid this repetition. If we want to achieve then we will need to add wrapper on top of it but I am not sure how easy it will be.

It is giving error because when we called method `run_scenario` from [base class](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/cli/base.rb#L44) we have used assumeyes? method. 
Note - For flag options, Clamp provides "#{flag-option-name}?" method.

iNecas:
@ehelms was faster than me in moving the ticked. I however put you to developers of foreman-maintain in redmine, so you should be able to do the change on your own.

definitions/features/katello.rb
iNecas: What about every relevant feature (foreman, katello, pulp, tasks…) including some module (something like FeatureWithServices) and providing a method (e.g. `services_info`) that would provide the partial hash for it's part.  Alternatively to including some module, we could mark those features with some metadata tag.

We could then get the list of relevant services with something like `features.find_all { |f| f.is_a? FeatureWithServices }.map(&:service_info).reduce(&:merge)`.

WIth this approach, I believe we can achieve better modularity of the scripts

mbacovsky: I like the idea of moving the services into the features. I used the same approach with config files in backup (each feature with configs has method config_files returning the list). 
I also like the the idea of tagging the features and putting the master list together dynamically. For the config files the `HasConfigFiles` tag would equal `respond_to?(:config_files)` but for services it may be different, so +1 for tagging.

The question is how it would look like for features with more than one service such as Pulp. Do we want to create feature for each service? 

I agree it may make sense to include some module FeatureWithServices making sure the feature comply with some service related interface that we can relay on.

iNecas: Having a feature per service would be probably an overkill: the pulp feature could just return something like this:

```
def services
  { 'pulp_workers'             => 20,
    'pulp_celerybeat'          => 20,
    'pulp_resource_manager'    => 20,
    'pulp_streamer'            => 20
  }
end
```

lib/foreman_maintain/cli/service_command.rb
mbacovsky: @johnpmitsch do you need label and tags? It worked for for me when removed. The interactive_option(s) are needed by runner but I'm not sure if --assumeyes and --whitelist makes sense in the scripts. I'd suggest to fix the runner to not depend on these and use it only when available.

Back to your question - if not removed it could be included in common predecessor for the service subcommands. For such case there is another notation for subcommand declaration https://github.com/mdub/clamp#declaring-subcommands.

One thing I noticed and am not sure about is the declaration of --exclude and --only options. The way it is done it allows you to call `foreman-maintain service --only httpd status` and also `foreman-maintain service status --only httpd`. Both ways work both make sense but I'd stick with one to avoid confusion. While I like the first notation (reminds me sysv init scripts notation) it is the second one our CLI is using quite uniformly.

lib/foreman_maintain/cli/service_command.rb
```diff
@@ -0,0 +1,125 @@
...
+        end
+      end
+
+      subcommand 'restart', 'Restart applicable services' do
+        label_option
+        tags_option
+        interactive_option
+
+        def execute
+          run_scenario([stop_scenario, start_scenario])
```
mbacovsky: It maybe useful to support the restart action even in the procedure or have separate procedure for it. Restart of the service is quite common task that may be used on other scenarios

definitions/procedures/services/list.rb
johnpmitsch: @kgaikwad 

The [base class](https://github.com/theforeman/foreman_maintain/pull/151/files/7c5cf13a2a0de4fd7c4de0cfb8c96bc9acf0fc1a#diff-003670ad8a995e0ca76650538024b9cb) is for the services  procedures to share methods, and it inherits from `ForemanMaintain::Procedure`. I'm fine with the require_relative here but I am just wondering if there is a better way such as adding something in `lib/foreman_maintain.rb

definitions/features/katello.rb
johnpmitsch: Thanks for the suggestions, I'll add services per feature in a module with methods to return them

Is there a way to say "what features does this machine have?" I imagine that would be helpful not only here but in the other katello scripts as well

lib/foreman_maintain/cli/service_command.rb
johnpmitsch: @mbacovsky thanks, I can try to remove the label and tags option. I agree that `foreman-maintain service status --only httpd` would be good syntax and match our CLI usage already, I'll update to that :100:

lib/foreman_maintain/cli/service_command.rb
johnpmitsch: @mbacovsky I'll play around with it, but I do worry about repeating code with another procedure. This may be a good topic for us to discuss next week :smiley:

definitions/features/katello.rb
mbacovsky: `ForemanMaintain.available_features` should return that. You can also pass filtering params and filter based on `class`, `label` or `tag`. So `ForemanMaintain.available_features(:tag => 'HasServices')` should do the trick.

mbacovsky: I tried with the config files and this worked:
```ruby
   features = ForemanMaintain.available_features(:tags => :has_config_files)
```
when the features are tagged with 
```ruby
metadata do
  tags :has_config_files
end
```

johnpmitsch: @mbacovsky why a tag? Why not just have a method with a blank hash in `ForemanMaintain::Feature` and override it in the features with services? 

Then I could just do `ForemanMaintain.available_features.map(&:services).reduce(&:merge)`
A simliar strategy could be done for config files too

johnpmitsch: I updated this part in my latest commit

definitions/features/katello.rb
```diff
@@ -14,4 +14,17 @@ def data_dirs
...
+    # TODO: remove foreman services, where should they go?
+    {
+      'postgresql'               => 5,
+      'qpidd'                    => 10,
+      'qdrouterd'                => 10,
+      'httpd'                    => 30,
+      # 'puppetserver'             => 30,
+      'foreman-tasks'            => 30,
+      'goferd'                   => 30
+    }
```
johnpmitsch: @iNecas @mbacovsky what feature should base foreman services be stored in? (postgresql, httpd)

iNecas:
Btw. just for awareness, we already have a wrapper feature around katello_service here https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/katello_service.rb (we use it for some upgrade steps). Perhaps we could move the relevant parts there.

definitions/features/katello.rb
iNecas: ForemanServer for httpd and ForemanDatabase for postgresql should be good enough IMO

johnpmitsch:
@iNecas @mbacovsky Should we be handling root permissions for the systemctl commands? Right now I have to run the commands like this or the commands fail:
`sudo ./bin/foreman-maintain service start --only httpd`

sean797:
If were going to move katello-service to foreman-maintain can we please ensure to port the bash_completion script :)

https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello/katello-service-bash_completion.sh

iNecas:
We have an assumption foreman-maintain running as a root: I think it would be safer just to make this as a check (not even as ForemanMaintain::Check, rather than first thing to try before run), to require that.

johnpmitsch:
@sean797 I definitely agree we shouldn't regress in the bash completion katello-service currently has. I do think that it wouldn't make sense to just add it for one command, rather we should add to all of foreman-maintain. Discussing with @mbacovsky off-thread, he does have some ideas around this, i.e. a rake task that uses either a static or dynamically generated dictionary to create a bash completion script. I think this is something to do as a separate task

johnpmitsch:
@iNecas @mbacovsky this is ready for review now, here are a few notes on the changes:

- The katello_service feature has been removed as the logic is now in the service procedures, all places that use this feature have been updated.
- The "hammer ping retry" functionality has moved to the `Procedures::Services::Restart` class as that seemed to be its intention (check hammer ping after restarting). This allows us to also expose this to the user `foreman-maintain service restart -p`. The only place I saw this utilized was in the [hammer ping check](https://github.com/theforeman/foreman_maintain/blob/master/definitions/checks/hammer_ping.rb) which I refactored to continue to use this option.
- A lot of the underlying functionality is the same between these subcommands as they are all basically systemctl wrappers. I tried to keep all of the functionality in the `Procedures::Services::Base` class. Then there is a Procedure and a Scenario for each command. Any suggestions on DRYing this up is appreciated.
- I added a root check that is used before running some of the service commands

That should cover the major changes, let me know if you have any questions. I'll squash my commits after reviews

ntkathole:
- Might be service-base is not required under advanced procedure run.
```
# ./bin/foreman-maintain advanced procedure run service-base
Procedures::Service::Base:                                            [RUNNING]
/root/foreman_maintain/lib/foreman_maintain/executable.rb:89:in `run': NotImplementedError (NotImplementedError)
        from /root/foreman_maintain/lib/foreman_maintain/executable.rb:119:in `__run__'
        from /root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:76:in `block (2 levels) in run'
        from /root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:99:in `capture_errors'
        from /root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:75:in `block in run'
        from /root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:93:in `with_metadata_calculation'
        from /root/foreman_maintain/lib/foreman_maintain/runner/execution.rb:74:in `run'
        from /root/foreman_maintain/lib/foreman_maintain/runner.rb:109:in `run_step'
        from /root/foreman_maintain/lib/foreman_maintain/runner.rb:98:in `run_steps'
```

- For Satellite 6.1 upgrade to satellite 6.2, katello-service stop/start  elasticsearch also. Can we include it here ? I am not sure, it will affect anything or not.
```
[root@qe-sat61-rhel7 foreman_maintain]# katello-service start
Redirecting to /bin/systemctl start mongod.service

Redirecting to /bin/systemctl start qpidd.service

Redirecting to /bin/systemctl start qdrouterd.service

Redirecting to /bin/systemctl start tomcat.service

Starting elasticsearch (via systemctl):                    [  OK  ]
Redirecting to /bin/systemctl start foreman-proxy.service

Redirecting to /bin/systemctl start pulp_celerybeat.service

Redirecting to /bin/systemctl start pulp_resource_manager.service

Redirecting to /bin/systemctl start pulp_workers.service

Redirecting to /bin/systemctl start httpd.service

Redirecting to /bin/systemctl start foreman-tasks.service

Success!
```

johnpmitsch:
@ntkathole We've discussed this before and have talked about "gating" these commands for 6.1 and 6.2. The executables like katello-service already exist on those Satellite versions and we don't want to introduce new functionality in older versions, so we have talked about redirecting `'foreman-maintain service status` to `katello-service` or showing a message saying `foreman-maintain service` is not supported for 6.2 and earlier. I'll discuss with @mbacovsky and @iNecas and see what they think

iNecas:
One think I'm not sure about is the move from the feature -> procedure. The example where the procedure approach doesn't work well is the example we have in https://github.com/theforeman/foreman_maintain/pull/52/files#diff-d960f2313918737e5fdda6323248d300R9, where we need to stop a service, do something and then start it again if it's was stopped by us.

I would like to see the code from procedures to actually be provided by features, and the procedures just using that code.

sean797:
@johnpmitsch I totally agree that is the best approach, if fact that is how 'oc' (open shift client tool) does it. If you want to go down that route I would like to see that PR merged first, or otherwise it would be a regression ;)

That bash competition script was actually my first contribution to open source, when I was a Satellite customer. I created and contributed it as it made my life as a sysadmin better.

Side note, if we could create a separate gem to create a bash competition script (I don't believe one already exists) we could use it with hammer also, while I understand this isn't a "quick fix" I hope I have expressed the value I use to as a sysadmin (and I'm sure others) put on that completion script.

iNecas:
While I agree about the importance of the completion script, I would not block this PR on that, especially as we have other tasks blocked by this one.

johnpmitsch:
@iNecas Update to use a `service` feature for most of the logic. This will allow us to do things like `features(:service).handle_service(spinner, "stop", {:only => ["httpd"]}`, allowing that to be called on its own outside of a procedure.

sean797:
> I would not block this PR on that, especially as we have other tasks blocked by this one.

Okay, then can we block a release? I would hate to release with a regression we know about.

iNecas:
Yes, we can

ntkathole:
@johnpmitsch  unsupported arguments: enable/disable for katello-service
 
```
Redirecting to:
katello-service enable --only tomcat
                                 [FAIL]
Failed executing katello-service enable --only tomcat, exit status 1:
 katello-service: Received unsupported arguments: enable
```

ntkathole:
--only and --exclude won't work with katello-service command 
```
# ./bin/foreman-maintain service stop --only tomcat
Running Stop Services
Stop applicable services:
Services are handled by katello-service in Satellite versions 6.2 or earlier.
Redirecting to:
katello-service stop --only tomcat
                                   [FAIL]
Failed executing katello-service stop --only tomcat, exit status 1:
 /usr/bin/katello-service:31:in `block in <main>': invalid option: --only (OptionParser::InvalidOption)
        from /usr/bin/katello-service:24:in `new'
        from /usr/bin/katello-service:24:in `<main>'
Scenario [Stop Services] failed.
```

johnpmitsch:
@iNecas @mbacovsky is there a way to DRY up the [metadata for each procedure?](https://git.io/vxLQp)

johnpmitsch:
@ntkathole which version of Satellite is this?

ntkathole:
@johnpmitsch Satellite 6.1.12

johnpmitsch:
updating to support 6.1, plz hold off on reviews for now :pray:

iNecas:
@johnpmitsch you could define a method in `base.rb` that would define the params, and in the metadata, you could call it:

```ruby
# in base.rb
def self.common_params
  param :exclude
  param :include
end

# in the child procedures

class Status < #...
  metadata do
    common_params
  end
end
```

johnpmitsch:
@iNecas I can't seem to do that, it looks like I can't reference anything from the Service::Base class inside the metadata block (notice `ForemanMaintain::Concerns::Metadata`)
```
/home/vagrant/foreman_maintain/definitions/procedures/service/enable.rb:7:in `block in <class:Enable>': 
undefined method `common_params' for #<ForemanMaintain::Concerns::Metadata::DSL:0x00000000028e3198> (NoMethodError)
```

iNecas:
This would be one approach:

```patch
diff --git a/definitions/procedures/service/base.rb b/definitions/procedures/service/base.rb
index a78a3a1..15005b3 100644
--- a/definitions/procedures/service/base.rb
+++ b/definitions/procedures/service/base.rb
@@ -1,6 +1,12 @@
 module Procedures
   module Service
     class Base < ForemanMaintain::Procedure
+      def self.common_params(context)
+        context.instance_eval do
+          param :only, 'A comma-separated list of services to include', :array => true
+          param :exclude, 'A comma-separated list of services to skip', :array => true
+        end
+      end
       def run_service_action(action, options)
         action_noun = feature(:service).action_noun(action).capitalize
         puts "#{action_noun} the following service(s):\n"
diff --git a/definitions/procedures/service/enable.rb b/definitions/procedures/service/enable.rb
index 8ccf4a6..5ceccd3 100644
--- a/definitions/procedures/service/enable.rb
+++ b/definitions/procedures/service/enable.rb
@@ -4,8 +4,7 @@ module Procedures::Service
   class Enable < Base
     metadata do
       description 'Enable applicable services'
-      param :only, 'A comma-separated list of services to include', :array => true
-      param :exclude, 'A comma-separated list of services to skip', :array => true
+      Base.common_params(self)
     end
 
     def run
```

johnpmitsch:
I updated to support satellite 6.1 and 6.2,

- any satellite <=6.2 will use katello-deploy for all service commands.
- and `service enable` and `service disable` will display a message that this command is not supported <=6.2 for those versions.
- All commands that are supported on katello-deploy are supported by `foreman-maintain status`
- RHEL6 versions work as well

@iNecas @mbacovsky @ntkathole please re-review when you have a chance

johnpmitsch:
@ntkathole how can I hide `service-base` from `advanced procedure run` options?

kgaikwad:
@johnpmitsch ,
To hide procedure from` advanced procedure run `command, you can use newly added metadata option i.e `advanced_run false`. Default value is true.

For example -

```
class SampleProcedure < #...
    metadata do
       advanced_run false
    end
end

```

iNecas:
Deferring full review to @mbacovsky, but :+1: on the current direction.

iNecas:
And of course to @kgaikwad, that has also been involved a lot with this part in the past.

ntkathole:
Looks good from testing side ACK!

johnpmitsch:
@ntkathole thanks for the review! To confirm, were you able to test out on both 6.1 and 6.2? Also testing RHEL6? I did my best to test this out locally but would like someone to confirm that they tested this w/o issue

ntkathole:
@johnpmitsch Tested on sat 6.1, 6.2 and 6.3. For 6.2, when service are stopped, I faced an issue http://projects.theforeman.org/issues/22945 but, it is not related to this PR. RHEL 6 is not tested as afaik foreman-maintain don't support it (we use clone tool for migration). Please correct me here/ let me know if rhel6 testing is needed/

iNecas:
Since so far, we've been targeting with one release against multiple versions, and 6.1 and 6.2 are supported on rhel6, we should also test this scenario on rhel6

iNecas:
And yes, the regressions you mentioned is not related

definitions/features/foreman_server.rb
```diff
@@ -7,6 +7,13 @@ class ForemanServer < ForemanMaintain::Feature
           server?
         end
       end
+
+      def services
+        {
+          'postgresql'               => 5,
```
mbacovsky: Shouldn't postgresql  be in foreman_database and candlepin_database features?

mbacovsky:
@johnpmitsch so far it looks good, I've tested it on recent Foreman + Katello and Sat 6.1 on RHEL 6. I' d like to give it a try on plain Foreman.
The only unexpected thing so far was that on Sat 6.1 + RHEL6 the following restarted `tomcat6` while the same command on Foreman + Katello restarted only `httpd` 
```
# ./bin/foreman-maintain service restart --only httpd
Running Restart Services
Restart applicable services: 
httpd

/ Services are handled by katello-service in Satellite versions 6.2 and earlier. Redirecting to: 
katello-service stop --exclude mongod,postgresql,qpidd,qdrouterd,pulp_resource_manager,pulp_workers,foreman-proxy,pulp_celerybeat,elasticsearch,foreman-tasks
- Stopping httpd: [  OK  ]                                                      
Stopping tomcat6: waiting for processes 28529 to exit
[  OK  ]
Success!

httpd

| Services are handled by katello-service in Satellite versions 6.2 and earlier. Redirecting to: 
katello-service start --exclude mongod,postgresql,qpidd,qdrouterd,pulp_resource_manager,pulp_workers,foreman-proxy,pulp_celerybeat,elasticsearch,foreman-tasks
/ [Mon Mar 26 12:00:41 2018] [warn] module passenger_module is already loaded, skipping
Starting tomcat6: [  OK  ]
Starting httpd: [  OK  ]
Success!
                                                                    [OK]

```

mbacovsky:
@johnpmitsch on Foreman without katello the -p option is not applicable. It may make sense to condition it by presence of feature(:katello).

mbacovsky:
@johnpmitsch I've investigated the issue with `restart --only httpd` restarting also the tomcat. The reason is  when the `--only` flag is simulated with `--exclude` we omit the tomcat in the list. The reason is that on Sat 6.1 the tomcat service is called `tomcat6`. Adding `tomcat6` to the list of services solved the issue.

johnpmitsch:
@mbacovsky I've updated the hammer ping part to be restricted by `feature(:katello)` and was able to use the `option_wrapper` method I added for this as well, it seems to be hidden properly on foreman and show on katello just fine, but please double check me :smiley_cat: 

@mbacovsky I added tomcat6 to the `candlepin_database` feature for now, but I am thinking we should make a candlepin feature that is confined by the package `candlepin` and add the services there?

johnpmitsch:
(sorry changes aren't in a separate commit, I have a bad amend + force push habit) :smoking:

johnpmitsch:
actually hold on reviewing plz, for some reason hammer ping is running all the time now

mbacovsky:
@johnpmitsch +1 on adding the candlepin feature

johnpmitsch:
@mbacovsky updated the option_wrapper according to our discussion off thread and also moved the candlepin services to a new candlepin feature

johnpmitsch:
@mbacovsky one update on the option_wrapper, I had to move away from passing in a default argument like `option_wrapper("whitelist", [])`, as this led to whitelist being nil for the commands that already had it. 

i.e. for `health check`, the options would be
`{:assumeyes=>nil, :whitelist=>nil, :force=>nil}`

which is why we need something like `option_wrapper("whitelist") || []`, which will actually pass the empty array when the param is responded to.

I kept the argument there in case its needed, but it defaults to `false`

mbacovsky:
@johnpmitsch, I've tested against plain Foreman and Sat 6.1 and all the issues were fixed. Thanks!

One last thing I'd prefer to get fixed is moving postgres service to candlepin_database and foreman_database features as they are bound closer to the service than the :foreman_server is.

We will need to do this anyway for adding support for remote DBs, so I'm okay to postpone till then if you prefer to no longer block this PR.

mbacovsky:
ACK, given we address the postgres service later. Thanks @johnpmitsch, :fireworks: :+1:

mbacovsky:
@iNecas, If there are no objections could you merge, please?

iNecas:
Thanks @johnpmitsch, @mbacovsky and @ntkathole

PR #152, opened 2018-03-07
Fixes #22762 - Check availability of repositories before upgrade

swapab:


definitions/features/downstream.rb
```diff
@@ -30,6 +30,16 @@ def setup_repositories(version)
     end
   end
 
+  def absent_repos(version)
+    all_repos = execute!(
```
iNecas: I guess we should call `subscription-manager refresh` first, similarly as in https://github.com/theforeman/foreman_maintain/pull/148/files

definitions/procedures/repositories/validate.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      param :version,
+            'Version for which repositories needs to be validated',
+            :required => true
+    end
+
+    def run
+      with_spinner("Validating repositories for #{@version}") do
+        absent_repos = (feature(:downstream) || feature(:upstream)).absent_repos(@version)
```
iNecas: Since we don't define this for upstream, I think this should stay downstream only.

definitions/features/downstream.rb
```diff
@@ -30,6 +30,16 @@ def setup_repositories(version)
...
+      "subscription-manager repos --list | grep 'Repo ID' | " \
+         "awk -F':' '/:/{gsub(/ /, \"\", $2); print $2}'"
```
sean797: No need to use `grep` ;) It can be written like:

```
subscription-manager repos --list | awk -F':' '/Repo ID/{gsub(/ /, "", $2); print $2}

Comment (swapab) [reply definitions/features/downstream.rb]: Ah. Nice! thanks @sean797

File: definitions/features/downstream.rb (hunk 1 above)
Comment (swapab) [reply definitions/features/downstream.rb]: Right. It's better to refresh first before listing repos. Done!

File: definitions/procedures/repositories/validate.rb (hunk 2 above)
Comment (swapab) [reply definitions/procedures/repositories/validate.rb]: Noted and done.

**Comment by ntkathole** [comment]
Looks good :+1: 
@iNecas here we are validating availability for satellite-tools repo as well though I think it is not required either for installation or upgrade of satellite. any comments?

```
Validate repositories:
| Validating repositories for 6.3                                     [FAIL]
Following repositories are not available on your system: rhel-7-server-rpms, rhel-server-rhscl-7-rpms, rhel-7-server-satellite-maintenance-6-rpms, rhel-7-server-satellite-tools-6.3-rpms, rhel-7-server-satellite-6.3-rpms
```

**Comment by iNecas** [comment]
@ntkathole  There are reason for that https://bugzilla.redhat.com/show_bug.cgi?id=1512959 we actually added that to resolve the particular BZ

File: definitions/features/downstream.rb [hunk 4]
```diff
@@ -30,6 +30,17 @@ def setup_repositories(version)
     end
   end
 
+  def absent_repos(version)
+    execute!(%(subscription-manager refresh))
+    all_repos = execute!(
+      'subscription-manager repos --list | ' \
```
iNecas: Just of safety reasons, we should ensure we are running in english locale `LANG=en_US.utf-8`

swapab: Sure @iNecas 
Is the following change relevant to what you meant above?
Or setting english locale `LANG=en_US.utf-8` is actually something different.
```
-      'subscription-manager repos --list | ' \
-        "awk -F':' '/Repo ID/{gsub(/ /, \"\", $2); print $2}'"
+      %('subscription-manager repos --list | ') \
+        %("awk -F':' '/Repo ID/{gsub(/ /, \"\", $2); print $2}'")
```

iNecas: The reason why I'm suggesting `LANG=en_US.utf-8` is, that in case folks have different localization on the sytem, the subscription manager can include something else, than `Repo ID` in the output, in which case it could fail. I'm not sure it would happen, but it's better to be on safe side.

See what we do with hammer `https://github.com/theforeman/foreman_maintain/blob/531e504360858b0df97bf49c24087bb4d43c94d9/lib/foreman_maintain/utils/hammer.rb#L70`

ntkathole:
@swapab 
```
Running Migration scripts to Satellite 6.2
Validate repositories:
/ Validating repositories for 6.2                                     [FAIL]
Failed executing 'LANG=en_US.utf-8 subscription-manager repos --list | ', exit status 127:
 sh: LANG=en_US.utf-8 subscription-manager repos --list | : command not found
```
```
E, [2018-03-13 08:10:30-0400 #7743] ERROR -- : Failed executing 'LANG=en_US.utf-8 subscription-manager repos --list | ', exit status 127:
 sh: LANG=en_US.utf-8 subscription-manager repos --list | : command not found (ForemanMaintain::Error::ExecutionError)
/root/foreman_maintain/lib/foreman_maintain/utils/command_runner.rb:50:in `execution_error'
```

iNecas:
Closing in favor of https://github.com/theforeman/foreman_maintain/pull/158

PR #154, opened 2018-03-07
Freeze rubocop version on 0.50.0

iNecas:
Two reasons:

1. I'm tired of unexpected failres on PRs caused by new comps being
added to Rubocop

2. new rubocop is broken at some checks:

```
An error occurred while Layout/SpaceInsideArrayLiteralBrackets
cop was inspecting
foreman_maintain/test/lib/cli/upgrade_command_test.rb:89:44.
foreman_maintain/test/lib/cli/upgrade_command_test.rb:101:44.
```


swapab:
:+1:

swapab:
[FYI] Restarted this travis-ci job: https://travis-ci.org/theforeman/foreman_maintain/jobs/350283391

Gemfile
```diff
@@ -9,7 +9,7 @@ if RUBY_VERSION <= '1.8.7'
 end
 
 if RUBY_VERSION >= '2.0'
-  gem 'rubocop'
+  gem 'rubocop', '0.52.1'
```
swapab: https://travis-ci.org/theforeman/foreman_maintain/jobs/350283391
I think we can lock it to `<= 0.52.1` ?

iNecas: While in general I agree with `<=`, rubocop is fairly specific as once you configure some cop that has not been defined in previous versions, it starts failing with older versions. Therefore I preffer matching against a specific version

iNecas:
I needed to lock the version to 0.50.0, but should be good to go now.

.rubocop.yml
```diff
@@ -74,7 +74,5 @@ Style/FormatStringToken:
 Naming/HeredocDelimiterNaming:
   Enabled: false
 
-# Enable check once issue is resolved:
-# https://github.com/bbatsov/rubocop/issues/4940
-Lint/UnneededRequireStatement:
+Style/RedundantReturn:
```
swapab: Can we also mute cop: `FrozenStringLiteralComment`
On my new setup it complains about it in each and every class/module.
```
test/test_helper.rb:1:1: C: Style/FrozenStringLiteralComment: Missing magic comment # frozen_string_literal: true.
```

iNecas: I have not seen that myself, but updated.

iNecas:
Addressed all additional comments. Merging to unblock other PRs to get in.

PR #155, opened 2018-03-11
Fixes #22759 - enhance cpdb validate check

kgaikwad:
@iNecas, 
could you please suggest me which is the ideal location to store candlepin dump files?
so that I will modify `value for candlepin_db_backup_dir` in config files.

Currently, I have set default value to `/tmp` which needs to be change.
For value `'/var/lib/foreman-maintain/db-backups'` to `candlepin_db_backup_dir`, it is giving me permission error while creating dump. I have tried by giving different `chmod` but still failing. 
Do I need to modify `chown` for this directory?

### DIFF
--- a/definitions/features/candlepin_database.rb
+++ b/definitions/features/candlepin_database.rb
@@ -27,12 +27,20 @@ class Features::CandlepinDatabase < ForemanMaintain::Feature
 
   def execute_cpdb_validate_cmd
     main_cmd = cpdb_validate_cmd
-    unless main_cmd.empty?
-      main_cmd += format_shell_args(
-        '-u' => configuration['username'], '-p' => configuration[%(password)]
-      )
-      execute!(main_cmd, :hidden_patterns => [configuration['password']])
-    end
+    return [true, nil] if main_cmd.empty?
+    main_cmd += format_shell_args(
+      '-u' => configuration['username'], '-p' => configuration[%(password)]
+    )
+    execute_with_status(main_cmd, :hidden_patterns => [configuration['password']])
+  end
+
+  def env_content_ids_with_null_content
+    sql = <<-SQL
+      SELECT ec.id
+      FROM cp_env_content ec
+      LEFT JOIN cp_content c ON ec.contentid = c.id WHERE c.id IS NULL
+    SQL
+    query(sql).map { |r| r['id'] }
   end
 
   private
@@ -40,15 +48,15 @@ class Features::CandlepinDatabase < ForemanMaintain::Feature
   def load_configuration
     raw_config = File.read(CANDLEPIN_DB_CONFIG)
     full_config = Hash[raw_config.scan(/(^[^#\n][^=]*)=(.*)/)]
-    uri = %r{://(([^/:]*):?([^/]*))/(.*)}.match(full_config['org.quartz.dataSource.myDS.URL'])
+    uri = %r{://(([^/:]*):?([^/]*))/(.*)}.match(full_config['jpa.config.hibernate.connection.url'])
     @configuration = {
-      'username' => full_config['org.quartz.dataSource.myDS.user'],
-      'password' => full_config['org.quartz.dataSource.myDS.password'],
+      'username' => full_config['jpa.config.hibernate.connection.username'],
+      'password' => full_config['jpa.config.hibernate.connection.password'],
       'database' => uri[4],
       'host' => uri[2],
       'port' => uri[3] || '5432',
-      'driver' => full_config['org.quartz.dataSource.myDS.driver'],
-      'url' => full_config['org.quartz.dataSource.myDS.URL']
+      'driver_class' => full_config['jpa.config.hibernate.connection.driver_class'],
+      'url' => full_config['jpa.config.hibernate.connection.url']
     }
   end
 
--- a/definitions/procedures/candlepin/delete_orphaned_records_from_env_content.rb
+++ b/definitions/procedures/candlepin/delete_orphaned_records_from_env_content.rb
@@ -1,0 +1,34 @@
+module Procedures::Candlepin
+  class DeleteOrphanedRecordsFromEnvContent < ForemanMaintain::Procedure
+    metadata do
+      description 'Delete orphaned record(s) from cp_env_content with unresolvable content'
+      label :candlepin_delete_orphaned_records_from_env_content
+
+      confine do
+        feature(:candlepin_database) &&
+          feature(:downstream) &&
+          feature(:downstream).current_minor_version == '6.2' &&
+          feature(:candlepin_database).table_exist?('cp_env_content')
+      end
+    end
+
+    def run
+      with_spinner('Deleting cp_env_content record(s) with unresolvable content') do |spinner|
+        spinner.update 'Finding cp_env_content records with unresolvable content'
+        env_content_ids = feature(:candlepin_database).env_content_ids_with_null_content
+        if env_content_ids.empty?
+          spinner.update 'No orphaned records found'
+        else
+          spinner.update 'Taking a backup of the candlepin database'
+          feature(:candlepin_database).perform_backup
+
+          puts "Total #{env_content_ids.length} records with unresolvable content"
+          spinner.update 'Deleting record(s) from cp_env_content table'
+          feature(:candlepin_database).delete_records_by_ids(
+            'cp_env_content', env_content_ids
+          )
+        end
+      end
+    end
+  end
+end
--- a/definitions/procedures/knowledge_base_article.rb
+++ b/definitions/procedures/knowledge_base_article.rb
@@ -1,0 +1,29 @@
+class Procedures::KnowledgeBaseArticle < ForemanMaintain::Procedure
+  metadata do
+    description 'Show knowledge base article for troubleshooting'
+
+    confine do
+      feature(:downstream)
+    end
+    param :doc,
+          'Document name required to select a correct article',
+          :required => true
+    advanced_run false
+  end
+
+  def run
+    ask(<<-MESSAGE.strip_heredoc)
+      Go to #{kcs_documents[@doc]}
+      please follow steps from above article to resolve this issue
+      press ENTER once done.
+    MESSAGE
+  end
+
+  private
+
+  def kcs_documents
+    {
+      'fix_cpdb_validate_failure' => 'https://access.redhat.com/solutions/3362821'
+    }
+  end
+end
--- a/lib/foreman_maintain/concerns/base_database.rb
+++ b/lib/foreman_maintain/concerns/base_database.rb
@@ -22,6 +22,49 @@ module ForemanMaintain
         psql('SELECT 1 as ping', config)
       end
 
+      def backup_file_path(config = configuration)
+        dump_file_name = "#{config['database']}_#{Time.now.strftime('%Y-%m-%d_%H-%M-%S')}.dump"
+        "#{backup_dir}/#{dump_file_name}.bz2"
+      end
+
+      def backup_db_command(file_path, config = configuration)
+        pg_dump_cmd = "pg_dump -Fc #{config['database']}"
+        "runuser - postgres -c '#{pg_dump_cmd}' | bzip2 -9 > #{file_path}"
+      end
+
+      def backup_dir
+        @backup_dir ||= File.expand_path(ForemanMaintain.config.db_backup_dir)
+      end
+
+      def perform_backup(config = configuration)
+        file_path = backup_file_path(config)
+        backup_cmd = backup_db_command(file_path, config)
+        execute!(backup_cmd)
+        puts "\n Note: Database backup file path - #{file_path}"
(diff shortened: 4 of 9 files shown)


lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -22,6 +22,40 @@ def ping(config = configuration)
         psql('SELECT 1 as ping', config)
       end
 
+      def file_path_to_backup_db(config = configuration)
+        dump_file_name = "#{config['database']}_#{Time.now.strftime('%Y-%m-%d_%H-%M-%S')}.dump"
+        "#{db_backup_dir}/#{dump_file_name}"
+      end
+
+      def backup_db_command(backup_file_path, config = configuration)
+        "runuser - postgres -c 'pg_dump -Fc #{config['database']} > #{backup_file_path}'"
```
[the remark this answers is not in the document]
iNecas: Also, I guess it's not a bad idea to filter it though some compression `bzip2 -9 > #{backup_file_path}.bz2`

definitions/procedures/candlepin/delete_orphaned_records_from_env_content.rb
```diff
@@ -0,0 +1,34 @@
+module Procedures::Candlepin
+  class DeleteOrphanedRecordsFromEnvContent < ForemanMaintain::Procedure
+    metadata do
+      label :cp_env_content_delete_orphaned_records
```
iNecas: Since we have `candlepin` prefix in the candlepin checks, I would recommend following this convention also here. Ideally so that the label matches the class name `:candlepin_delete_orphaned_records_from_env_content`

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -61,6 +61,12 @@ def execute(command, options = {})
         command_runner.output
       end
 
+      def response_on_execute(command, options = {})
+        result_msg = execute(command, options)
+        result = $CHILD_STATUS.success?
```
iNecas: Having the exit code, rather than true/false might be more universal, especially if we cared about the exit code

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -61,6 +61,12 @@ def execute(command, options = {})
         command_runner.output
       end
 
+      def response_on_execute(command, options = {})
```
iNecas: I would not know from the name what to expect from the name. What about `execute_with_status`?

lib/foreman_maintain/config.rb
```diff
@@ -27,6 +25,19 @@ def load_log_configs
...
     end
 
+    def load_backup_dir_paths
+      @storage_file = @options.fetch(:storage_file, 'data.yml')
+      @backup_dir = find_dir_path(
+        @options.fetch(:backup_dir, '/var/lib/foreman-maintain')
+      )
+      @db_backup_directories = {}
+      @db_backup_directories = {
+        :candlepin => find_dir_path(
```
iNecas: As I mentioned in the settings, I think we would be ok with having just one backup dir for now, not specific to just candlepin. It also might change as we move the backup/restore scripts.

iNecas:
Looks good in general, Few comments to address before we move to testing it.

lib/foreman_maintain/concerns/base_database.rb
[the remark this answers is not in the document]
kgaikwad: @iNecas,
`-Fc` option also provide some compression.
```
Description of -Fc:
Output a custom-format archive suitable for input into pg_restore. Together with the directory output format, this is the most flexible output format in that it allows manual selection and reordering of archived items during restore. This format is also compressed by default.
```

Please confirm so that I will go & add `bzip2 -9` to command.

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -22,6 +22,41 @@ def ping(config = configuration)
...
       end
 
+      def file_path_to_backup_db(config = configuration)
+        dump_file_name = "#{config['database']}_#{Time.now.strftime('%Y-%m-%d_%H-%M-%S')}.dump"
+        "#{db_backup_dir}/#{dump_file_name}.bz2"
+      end
+
+      def backup_db_command(backup_file_path, config = configuration)
+        pg_dump_cmd = "pg_dump -Fc #{config['database']}"
+        "runuser - postgres -c '#{pg_dump_cmd}' | bzip2 -9 > #{backup_file_path}"
```
kgaikwad: @iNecas,
I would like to know that what difference it makes after extracting out backup file from inner pg_dump command to `runuser`.
I can see that previously, it tried to create dump file with `postgres` user and that's why it was failing with permission error.  After I modification, dump file gets created using `root` user.
Any other differences?

kgaikwad:
@iNecas ,
Thank you for reviewing the pull request.
Updated PR with suggested changes. PR ready for review.

lib/foreman_maintain/concerns/base_database.rb
iNecas: The difference is under which user the file gets created: the postgres user doesn't have rights to write to the destination, the root user has.

definitions/procedures/candlepin/delete_orphaned_records_from_env_content.rb
```diff
@@ -0,0 +1,34 @@
...
+          feature(:candlepin_database).table_exist?('cp_env_content')
+      end
+    end
+
+    def run
+      with_spinner('Deleting cp_env_content record(s) with unresolvable content') do |spinner|
+        spinner.update 'Finding cp_env_content records with unresolvable content'
+        env_content_ids = feature(:candlepin_database).env_content_ids_with_null_content
+        if env_content_ids.empty?
+          spinner.update 'No any orphaned record(s) found'
```
iNecas: Nitpick: `No orphaned records found` would sound more naturally to me.

definitions/procedures/knowledge_base_article.rb
```diff
@@ -0,0 +1,28 @@
...
+      please follow steps from above article to resolve this issue
+      press ENTER once done.
+    MESSAGE
+  end
+
+  private
+
+  def kcs_documents
+    {
+      'fix_cpdb_validate_failure' => 'https://access.redhat.com/solutions/3362821'
```
iNecas: I like this approach

definitions/features/candlepin_database.rb
```diff
@@ -27,12 +27,32 @@ def check_option_using_cpdb_help(option_name, parent_cmd = '')
...
-      execute!(main_cmd, :hidden_patterns => [configuration['password']])
-    end
+    return [true, nil] if main_cmd.empty?
+    main_cmd += format_shell_args(
+      '-u' => configuration['username'], '-p' => configuration[%(password)]
+    )
+    execute_with_status(main_cmd, :hidden_patterns => [configuration['password']])
+  end
+
+  def db_backup_dir
```
iNecas: Since we are in the context of database class, the `db_` prefix is probably unnecessar

iNecas:
Some additional comments before we move the testing.

kgaikwad:
@iNecas ,

Updated the pull-request. Ready for review as well as for testing.

iNecas:
ACK codewise - waiting for @ntkathole feedback from testing

ntkathole:
- Will need some time to test and work on this as it looks critical change. @kgaikwad As of now, facing issue with satellite 6.1 
```
./bin/foreman-maintain health check
/root/foreman_maintain/definitions/features/candlepin_database.rb:55:in `load_configuration': undefined method `[]' for nil:NilClass (NoMethodError)
        from /root/foreman_maintain/definitions/features/candlepin_database.rb:15:in `configuration'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/base_database.rb:8:in `query'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/base_database.rb:52:in `table_exist?'
        from /root/foreman_maintain/definitions/procedures/candlepin/delete_orphaned_records_from_env_content.rb:11:in `block (2 levels) in <class:DeleteOrphanedRecordsFromEnvContent>'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:215:in `instance_exec'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:215:in `block in evaluate_confines'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:214:in `each'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:214:in `all?'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:214:in `evaluate_confines'
        from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:191:in `present?'
        from /root/foreman_maintain/lib/foreman_maintain/detector.rb:53:in `block in find_present_classes'
        from /root/foreman_maintain/lib/foreman_maintain/detector.rb:52:in `each'

```

- Just to clearify, @iNecas we are  just resolving bad cp_env_content here. Is their any plan to deal with bad products(delete orphaned records from cp_pool)?

iNecas:
@ntkathole we are keeping that strictly in scope of https://access.redhat.com/solutions/3362821 right now.

kgaikwad:
@iNecas , @ntkathole ,

I have made changes in candlepin_database feature while loading `candlepin` DB configs from `/etc/candlepin/candlepin.conf` file.

Previously, we are loading configurations using` ' org.quartz.dataSource.myDS.*' `.
Now, modified to `'jpa.config.hibernate.connection.*' `as quartz related configurations not there in config file on sat 6.1.

As per my observation,` 'jpa.config.hibernate.connection.*' ` options are always be there but we need to make sure from` candlepin team` whether is it safe to use those setting options for DB connection.

kgaikwad:
@iNecas,

I have two queries related to new naming changes latest candlepin package.

1. In sat 6.3, I have `candlepin` version i.e `candlepin-2.1.14-1.el7 ` having different table_names and column_names. For example here -
Tables from `cp_env_content  to cp2_environment_content` ,  `cp_content to cp2_content` and 
column names such as `contentid to content_uuid` and at few places `id to uuid`.
 
Is it ok if I add changes based on candlepin version check? that way I can handle this.
or any suggestion on this.
My next concern is that now if I add changes based on candlepin version even while building SQL but in future there might be changes in naming again.

2. In sat 6.1, there is table called `cp_env_content` but no `--validate` option to cpdb command but procedure visibility is there. So is that ok if user can run procedure using `advanced procedure run`.

iNecas:
1. what about making this check available to sat 6.2 with the confine block, as that's the only release currently with the issue? It would sound as the most simple solution to the question

2. similarly to (1), this is valid only for sat 6.2 and we could just hide that from others via the confine block.

kgaikwad:
@iNecas,

```
1. what about making this check available to sat 6.2 with the confine block, as that's the only release currently with the issue? It would sound as the most simple solution to the question
2. similarly to (1), this is valid only for sat 6.2 and we could just hide that from others via the confine block.
```
As you mentioned, for both queries, adding a condition that checks for sat 6.2 in confine block will work.
Instead of check, I have added condition to confine block of procedure and while adding steps in `next_steps` as we may need to run `cpdb --validate` command for other sat releases.
Updated PR. 

@ntkathole , made changes as per above discussion.

ntkathole:
@kgaikwad when cpdb --validate check failed on sat 6.2 gives error

```
   raise Exception("Error running command")
Exception: Error running command
E, [2018-03-16 02:48:30-0400 #74166] ERROR -- : undefined method `ensure_instance' for #<Array:0x00000001f816b8> (NoMethodError)
/root/foreman_maintain/lib/foreman_maintain/runner.rb:133:in `map'
/root/foreman_maintain/lib/foreman_maintain/runner.rb:133:in `ask_about_offered_steps'
/root/foreman_maintain/lib/foreman_maintain/runner.rb:117:in `post_step_decisions'
```

kgaikwad:
@ntkathole,
My bad, I did mistake on line where I mentioned the [next-steps](https://github.com/theforeman/foreman_maintain/pull/155/files#diff-1788e0473e8ca0a6cccf2e333e486ce2R16). Fixed.

iNecas:
ACK code-wise, waiting for final results from testing

ntkathole:
Looks good except advanced procedure run knowledge-base-article command. I would like to have it `advanced_run false` as this does not seem fully functional, if you agree @kgaikwad and @iNecas ? 
```
# ./bin/foreman-maintain advanced procedure run knowledge-base-article --doc 2
Show knowledge base article for troubleshooting:
Go to
please follow steps from above article to resolve this issue
press ENTER once done.

                                                                      [OK]

```

kgaikwad:
@iNecas  & @ntkathole ,
I guess, for now we can apply `advanced_run false` for procedure knowledge_base_article.
Updated the PR.

ntkathole:
ACK

iNecas:

PR #157, opened 2018-03-15
Fixes #22738 - check for verifying symlink under setting.d dir

kgaikwad:
With this commit, it verifies symlink exist under `/etc/smart_proxy_dynflow_core/settings.d`
If yes, it gives step to remove `settings.d` directory.

### DIFF
--- a/definitions/checks/remote_execution/verify_settings_file_already_exists.rb
+++ b/definitions/checks/remote_execution/verify_settings_file_already_exists.rb
@@ -1,0 +1,42 @@
+module Checks::RemoteExecution
+  class VerifySettingsFileAlreadyExists < ForemanMaintain::Check
+    metadata do
+      description 'Check to verify remote_execution_ssh settings already exist'
+
+      confine do
+        feature(:downstream) &&
+          feature(:downstream).current_minor_version == '6.2' &&
+          find_package('tfm-rubygem-smart_proxy_dynflow_core') &&
+          file_exists?('/etc/smart_proxy_dynflow_core')
+      end
+    end
+
+    def run
+      if file_exists?('/etc/smart_proxy_dynflow_core/settings.d')
+        symlinks = find_symlinks(settingd_dir_path)
+        assert(
+          symlinks.empty?,
+          failure_msg(settingd_dir_path, symlinks),
+          :next_steps => [
+            Procedures::RemoteExecution::RemoveExistingSettingsd.new(
+              :dirpath => settingd_dir_path
+            )
+          ]
+        )
+      end
+    end
+
+    private
+
+    def failure_msg(dir_path, symlinks)
+      'Settings related to remote_execution_ssh are already present'\
+      " under #{dir_path} and " \
+      "\nit would conflict with the installer from the next version." \
+      "\nsymlinks available - #{symlinks.join(', ')}"
+    end
+
+    def settingd_dir_path
+      '/etc/smart_proxy_dynflow_core/settings.d'
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


definitions/checks/remote_execution/verify_settings_file_already_exists.rb
```diff
@@ -0,0 +1,42 @@
+module Checks::RemoteExecution
+  class VerifySettingsFileAlreadyExists < ForemanMaintain::Check
+    metadata do
+      description 'Check to verify remote_execution_ssh settings already exist'
+      tags :pre_upgrade
+
+      confine do
+        feature(:downstream) &&
```
kgaikwad: @iNecas,

Could you please suggest what conditions should be needed in confine block?
Is this check applicable for upstream scenario as well?

iNecas: I believe this is downstream only, so I'm quite happy with the confine block I see right now.

definitions/procedures/remote_execution/remove_existing_settingsd.rb
```diff
@@ -0,0 +1,17 @@
+module Procedures::RemoteExecution
+  class RemoveExistingSettingsd < ForemanMaintain::Procedure
+    metadata do
+      param :dirpath,
+            'Directory path for settings.d folder',
+            :required => true
+      description 'Remove existing settings.d directory before installer run'
```
iNecas: I would add also, that `The next run of the installer will re-create the directory properly`

definitions/checks/remote_execution/verify_settings_file_already_exists.rb
```diff
@@ -0,0 +1,42 @@
...
+        )
+      end
+    end
+
+    private
+
+    def failure_msg(dir_path, symlinks)
+      'Settings related to remote_execution_ssh are already present'\
+      " under #{dir_path}"\
+      "\nsymlinks available - #{symlinks.join(', ')}"
```
iNecas: It would be also good to mention something like: `and it would conflict with the installer from the next version`.

definitions/checks/remote_execution/verify_settings_file_already_exists.rb
```diff
@@ -0,0 +1,42 @@
+module Checks::RemoteExecution
+  class VerifySettingsFileAlreadyExists < ForemanMaintain::Check
+    metadata do
+      description 'Check to verify remote_execution_ssh settings already exist'
+      tags :pre_upgrade
```
iNecas: I think we should add this manually to the 6.3 upgrade, as otherwise, this would apply also for `6.2.z` upgrades, where we don't want to do that.

kgaikwad:
@iNecas ,
Done with the modifications as per suggestions. Ready for review!

iNecas:
Ack code wise, @ntkathole ready to test.

ntkathole:
ACK :+1:

definitions/checks/remote_execution/verify_settings_file_already_exists.rb
```diff
@@ -0,0 +1,42 @@
...
+      confine do
+        feature(:downstream) &&
+          feature(:downstream).current_minor_version == '6.2' &&
+          find_package('tfm-rubygem-smart_proxy_dynflow_core') &&
+          file_exists?('/etc/smart_proxy_dynflow_core')
+      end
+    end
+
+    def run
+      if file_exists?('/etc/smart_proxy_dynflow_core/settings.d')
```
iNecas: There was one more thing I've noticed, where this string should be replaced by `settingd_dir_path`, since it was very straight forward thing and I've tested this before/after the change, I've just updated that and merged the rest.

iNecas:
Merged as da73bad92e55b9d7afde5991a4efe6eea78b46a7

PR #158, opened 2018-03-16
Fixes #22762 - Check availability of repositories before upgrade (updated)

iNecas:
I've taken the original PR https://github.com/theforeman/foreman_maintain/pull/152,
fixed running the subscription-manager command and moved the check to pre-upgrade checks.

### DIFF
--- a/definitions/features/downstream.rb
+++ b/definitions/features/downstream.rb
@@ -30,7 +30,7 @@ class Features::Downstream < ForemanMaintain::Feature
   end
 
   def absent_repos(version)
-    all_repo_lines = execute!(%(LANG=en_US.utf-8 subscription-manager repos --list | ) +
+    all_repo_lines = execute(%(LANG=en_US.utf-8 subscription-manager repos --list | ) +
                               %(grep '^Repo ID:')).split("\n")
     all_repos = all_repo_lines.map { |line| line.split(/\s+/).last }
     repos_required = rh_repos(version)
--- a/definitions/features/downstream.rb
+++ b/definitions/features/downstream.rb
@@ -30,6 +30,17 @@ class Features::Downstream < ForemanMaintain::Feature
     end
   end
 
+  def absent_repos(version)
+    execute!(%(subscription-manager refresh))
+    all_repos = execute!(
+      %('LANG=en_US.utf-8 subscription-manager repos --list | ') \
+        %("awk -F':' '/Repo ID/{gsub(/ /, \"\", $2); print $2}'")
+    ).split("\n")
+    repos_required = rh_repos(version)
+    repos_found = repos_required & all_repos
+    repos_required - repos_found
+  end
+
   private
 
   def rh_repos(sat_version)
--- a/definitions/features/downstream.rb
+++ b/definitions/features/downstream.rb
@@ -32,10 +32,9 @@ class Features::Downstream < ForemanMaintain::Feature
 
   def absent_repos(version)
     execute!(%(subscription-manager refresh))
-    all_repos = execute!(
-      %('LANG=en_US.utf-8 subscription-manager repos --list | ') \
-        %("awk -F':' '/Repo ID/{gsub(/ /, \"\", $2); print $2}'")
-    ).split("\n")
+    all_repo_lines = execute!(%(LANG=en_US.utf-8 subscription-manager repos --list | ) +
+                              %(grep '^Repo ID:')).split("\n")
+    all_repos = all_repo_lines.map { |line| line.split(/\s+/).last }
     repos_required = rh_repos(version)
     repos_found = repos_required & all_repos
     repos_required - repos_found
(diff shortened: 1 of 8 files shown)


iNecas:
@ntkathole ready for final testing

definitions/features/downstream.rb
```diff
@@ -30,6 +30,16 @@ def setup_repositories(version)
     end
   end
 
+  def absent_repos(version)
+    execute!(%(subscription-manager refresh))
```
sean797: I this is not necessary, sub-man gets a fresh list when you run `subscription-manager repos --list`.

https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/managercli.py#L2040

ntkathole:
@iNecas If we are moving this check to pre-upgrade checks, I expect foreman-maintain health list command should show this check with pre-upgrade tag which I could not find.

definitions/features/downstream.rb
iNecas: I don't see the `repos --list` calling this code: https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/managercli.py#L632

Calling the `refresh` was actually suggested by our CEE, when working on the check, and I would prefer keeping it there to stay on safe side here.

iNecas:
@ntkathole we can't do that at the moment, as the check requries argument (version number) that we don't allow to run via `health check` command. I would like to keep addressing this observation out of this PR. The preferred way to run the pre-upgrade checks should be `foreman-maintain upgrade check`.

ntkathole:
@iNecas  ACK :+1:

definitions/features/downstream.rb
sean797: - This sets `repos` https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/managercli.py#L2026
- which calls `RepoUpdateActionCommand` https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/repolib.py#L202
- That class gets a refresh list from the server. https://github.com/candlepin/subscription-manager/blob/8b2c0c60e7694694ca78ce7f9b8596f314871f57/src/subscription_manager/repolib.py#L326

Even if you don't follow the code, try enabling a repo on the server for a client, and then running `sub-man repo --list` without doing a `sub-man refresh` and you'll see your newly enabled repo.

iNecas: Still is doesn't trigger the refreshing the entitlements on the server side that way the `rhsm refresh` does https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/managercli.py#L629.

The point is that there is an evidence from the field that calling the `rhsm refresh` call helped and I don't want to remove it just because in some cases, it seems the `rhsm repos` can do fine on it's own. And there is no indication that I could see that would mean that `repos --list` is calling exactly the same code as the the `refresh`: it seems that the `refresh` actually is doing something else and we sometimes need it.

What I can do is checking the `rhsm repos` first, and in case the repos is not there, calling `rhsm refresh` and trying agian. However, I would probabely prefer the way it's written right now.

sean797: `sub-man refresh` I think like you say does do more than `sub-man repos --list`, my point is that `sub-man --list` will always do enough for our use case (get an up-to-date list of enabled repos).

My initial comment was due to the slowness of `sub-man refresh`, our installer and checks are IMO already quite slow and this only makes it worse, but leave it in if you want.

sean797: And if there's an edge case where `sub-man repos --list` doesn't get an up-to-date list of enabled repos, then that is a `sub-man` bug, we shouldn't need to workaround a potential bug here.

iNecas: Ok, what I did was making the `rhsm refresh` conditional in case the repos don't show up after `rhsm list`, I've also removed the call for `rhsm refresh` that we called later (that was done as fix for https://bugzilla.redhat.com/show_bug.cgi?id=1548368), and after this we don't need to do it twice.

In case the `rhsm list` is enough, we will not call the `refresh` at all, and in case it would not help, I still think it's better to call the `rhsm refresh`, even if that would be the bug on rhsm side, as the users for foreman-maintain don't care that much about where the bug is coming from, but if there is a way on our side to fix the issue for them automatically, we should at least try :)

definitions/features/downstream.rb
```diff
@@ -23,13 +23,25 @@ def setup_repositories(version)
...
     else
-      execute!(%(subscription-manager refresh))
       execute!(%(subscription-manager repos --disable '*'))
       enable_options = rh_repos(version).map { |r| "--enable=#{r}" }.join(' ')
       execute!(%(subscription-manager repos #{enable_options}))
     end
   end
 
+  def absent_repos(version)
+    all_repo_lines = execute!(%(LANG=en_US.utf-8 subscription-manager repos --list | ) +
```
ntkathole: @iNecas 
```
- Validating availability of repositories for 6.3                     [FAIL]
Failed executing LANG=en_US.utf-8 subscription-manager repos --list | grep '^Repo ID:', exit status 1
Scenario [Checks before upgrading to Satellite 6.3.z] failed.
```

iNecas: Thanks @ntkathole for catching this, updated the code

iNecas:
Merging now, thanks @ntkathole and @sean797 for help

PR #159, opened 2018-03-19
Add installer feature

mbacovsky:
The installer feature can
 - parse the answer files regardless of installer version
 - detect installer type (scenario, non-scenario based)
 - detect last used scenario
 - list related config files

### DIFF
--- a/definitions/features/installer.rb
+++ b/definitions/features/installer.rb
@@ -1,0 +1,75 @@
+class Features::Installer < ForemanMaintain::Feature
+  metadata do
+    label :installer
+
+    confine do
+      find_package('foreman-installer') ||
+        find_package('katello-installer') ||
+        find_package('capsule-installer')
+    end
+  end
+
+  def initialize
+    @installer_type = if find_package('foreman-installer')
+                        :scenarios
+                      elsif find_package('katello-installer')
+                        :katello
+                      elsif find_package('capsule-installer')
+                        :capsule
+                      end
+  end
+
+  def answers
+    case @installer_type
+    when :scenarios
+      last_scenario_answers
+    when :katello
+      load_answers(File.join(config_directory, 'katello-installer.yaml'))
+    when :capsule
+      load_answers(File.join(config_directory, 'capsule-installer.yaml'))
+    end
+  end
+
+  def with_scenarios?
+    @installer_type == :scenarios
+  end
+
+  def config_directory
+    case @installer_type
+    when :scenarios
+      '/etc/foreman-installer'
+    when :katello
+      '/etc/katello-installer'
+    when :capsule
+      '/etc/capsule-installer'
+    end
+  end
+
+  def config_files
+    Dir.glob(File.join(config_directory, '**/*'))
+  end
+
+  def last_scenario
+    return nil unless with_scenarios?
+    File.basename(last_scenario_config).split('.')[0]
+  end
+
+  private
+
+  def load_answers(config_file)
+    config = YAML.load_file(config_file)
+    YAML.load_file(config[:answer_file])
+  end
+
+  def scenario_answers(scenario)
+    load_answers(File.join(config_directory, "scenarios.d/#{scenario}.yaml"))
+  end
+
+  def last_scenario_answers
+    scenario_answers(last_scenario)
+  end
+
+  def last_scenario_config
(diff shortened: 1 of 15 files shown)


mbacovsky:
I isolated the installer feature into separate PR. This feature will be needed in backup script and in the hammer feature refactoring.

test/data/installer/capsule-installer/config_header.txt
```diff
@@ -0,0 +1,8 @@
+# Format:
```
iNecas: Nitpick: having `test/definitions/data` would be a better place for definitions-specific mock data IMO, but leaving the decision up to you

iNecas:
The code works for me, @kgaikwad / @johnpmitsch any comments before merging?

kgaikwad:
@iNecas, 
For me looks good.

test/data/installer/capsule-installer/config_header.txt
mbacovsky: @iNecas I had it there originally. Then I added some tests for the hammer utils and while it called the feature internaly I had to mock the configs there so I moved the data to common space. 

Is it allowed practice to use definitions from the `lib`? For Hammer utils it made sense as a short term solution but could be avoided. Testing the other parts e.g. `cli` might also benefit from common data for mock ups. Saying that these data aren't that definition-specific so it makes sense to keep it shared among lib and def's tests.

definitions/features/installer.rb
```diff
@@ -0,0 +1,75 @@
...
+        find_package('capsule-installer')
+    end
+  end
+
+  def initialize
+    @installer_type = if find_package('foreman-installer')
+                        :scenarios
+                      elsif find_package('katello-installer')
+                        :katello
+                      elsif find_package('capsule-installer')
```
johnpmitsch: Upstream foreman-proxy-with-contents will use a `foreman-proxy-content` package, shouldn't this be accounted for? i.e.
```
[vagrant@bacon-proxy ~]$ rpm -qa | grep proxy
foreman-proxy-content-3.7.0-1.nightly.el7.noarch
bacon-proxy.jomitsch.example.com-apache-1.0-1.noarch
bacon-proxy.jomitsch.example.com-qpid-client-cert-1.0-1.noarch
gssproxy-0.7.0-4.el7.x86_64
bacon-proxy.jomitsch.example.com-puppet-client-1.0-1.noarch
rubygem-smart_proxy_pulp-1.3.0-1.el7.noarch
bacon-proxy.jomitsch.example.com-foreman-proxy-client-1.0-1.noarch
bacon-proxy.jomitsch.example.com-qpid-router-server-1.0-1.noarch
bacon-proxy.jomitsch.example.com-qpid-broker-1.0-1.noarch
foreman-proxy-1.18.0-0.develop.201802282028gitaf41016.el7.noarch
bacon-proxy.jomitsch.example.com-foreman-proxy-1.0-1.noarch
bacon-proxy.jomitsch.example.com-qpid-router-client-1.0-1.noarch
```

test/data/installer/capsule-installer/config_header.txt
iNecas: :+1: then

iNecas:
Could we leverage this code also in https://github.com/theforeman/foreman_maintain/blob/master/definitions/procedures/installer/upgrade.rb#L17, so that we could simplify the upgrade procedure?

definitions/features/installer.rb
johnpmitsch: We resolved this outside of this thread, misunderstanding on my part

mbacovsky: @johnpmitsch, would it improve the readability if I renamed the installer_type symbols to `scenario`, `:legacy_katello` and `:legacy_capsule`. Or `:non_scenario_katello` and `:non_scenario_capsule`?

johnpmitsch: @mbacovsky I think that would be a good idea. My vote goes to the `:legacy_` options

mbacovsky:
@johnpmitsch, @iNecas, rebased, renamed the installer_type values, refactored the upgrade procedure. Changes were added in separate commit, feel free to squash before merge
Please test upgrade.

iNecas:
Thanks @mbacovsky and @ntkathole

PR #160, opened 2018-03-21
Fixes #22945 - show failure if DB service stopped

kgaikwad:
@iNecas , 
Doing testing on my end. 
I will update you once done.
If you have any suggestions on code please let me know.


definitions/checks/candlepin/validate_db.rb
```diff
@@ -11,7 +11,7 @@ class ValidateDb < ForemanMaintain::Check
 
     def run
       result, result_msg = feature(:candlepin_database).execute_cpdb_validate_cmd
-      next_steps = []
+      next_steps = [Procedures::StartService.new(:service => 'postgresql')]
```
kgaikwad: @iNecas,
Instead of hard-coding service name, should I need to handle this based on DB type, right?

kgaikwad: I guess, we can add this step later on after [PR-151](https://github.com/theforeman/foreman_maintain/pull/151) . This step i.e procedure helps to start respective DB service.

iNecas:
Needs testing but :+1: on the approach

definitions/checks/candlepin/validate_db.rb
iNecas: :+1: for deferring this decision for now.

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -14,8 +14,18 @@ def query_csv(sql, config = configuration)
...
+          execute("PGPASSWORD='#{config[%(password)]}' #{psql_db_connection_str(config)}",
+                  :stdin => query)
+        else
+          raise_service_error
+        end
+      end
+
+      def up?(config = configuration)
+        query = 'SELECT 1 as ping'
+        execute?("PGPASSWORD='#{config[%(password)]}' #{psql_db_connection_str(config)}",
```
iNecas: The duplicate code is IMO worth extracting to some private method such as`run_psql`, `psql_command` or something like that, and use it bot from here and from the `psql` method

iNecas: Also, what's the difference between this method and the `ping` one? Should the `ping` method be just replaced with this one or the other way around?

kgaikwad: Yes, I agree.
I can see that `ping` method is not used anywhere so worth
modifying existing method & remove duplicate code. I will update PR.

kgaikwad:
@iNecas, updated PR as per suggested changes.

ntkathole:
Looks good...I see issue no longer exists :+1:

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -14,12 +14,15 @@ def query_csv(sql, config = configuration)
...
+        if ping(config)
+          execute(psql_connection_string(config), :stdin => query)
+        else
+          raise_service_error
+        end
       end
 
       def ping(config = configuration)
-        psql('SELECT 1 as ping', config)
+        execute?(psql_connection_string(config), :stdin => 'SELECT 1 as ping')
```
iNecas: One last nitpick: `connection_string` phrase is connected to a special case (or at least for me, and I can imaging others having similar connotation) of passing to some driver to connect to db. In this case, I would rather go for `psql_command_string` or simply just `psql_command`

iNecas:
One last comment and good to go

PR #161, opened 2018-03-21
Fixes #22975 - satellite-installer requires LANG=en_US.utf-8

sean797:
Upgrading a Satellite from 6.2 -> 6.3 today I got the following error that this resolves:

```
The LANG envrionment variable should not be set to C
Your system does not meet configuration criteria
```

See https://github.com/Katello/katello-installer/blob/master/checks/lang.rb

### DIFF
--- a/definitions/procedures/installer/upgrade.rb
+++ b/definitions/procedures/installer/upgrade.rb
@@ -1,7 +1,7 @@
 module Procedures::Installer
   class Upgrade < ForemanMaintain::Procedure
     def run
-      execute!("#{installer_command} --upgrade", :interactive => true)
+      execute!("LANG=en_US.utf-8 #{installer_command} --upgrade", :interactive => true)
     end
 
     private


iNecas:
Interesting, would you mind creating an issue in http://projects.theforeman.org/projects/foreman-maintain/issues/new and assign this commit to it. It helps us keeping the changelog organized a bit, which is useful, especially with issues that other folks might be hitting. I know it's burden for one-liners like this, but still :)

sean797:
Updated, thanks @iNecas - I was being a bit cheeky :smiley:

ntkathole:
ACK :+1: 
```
D, [2018-03-22 06:43:37-0400 #119495] DEBUG -- : Running command LANG=en_US.utf-8 satellite-installer --upgrade with stdin nil
D, [2018-03-22 06:47:32-0400 #119495] DEBUG -- : output of the command:
 Script started on Thu 22 Mar 2018 06:43:37 AM EDT
Upgrading...
Upgrade Step: stop_services...
Redirecting to /bin/systemctl stop foreman-tasks.service
```

iNecas:
Thanks @sean797 and @ntkathole

PR #162, opened 2018-03-21
Fixes #22976, #22977 - foreman-installer doesnt take --upgrade

sean797:
and katello-installer doesn't exist anymore

### DIFF
--- a/definitions/procedures/installer/upgrade.rb
+++ b/definitions/procedures/installer/upgrade.rb
@@ -1,16 +1,24 @@
 module Procedures::Installer
   class Upgrade < ForemanMaintain::Procedure
     def run
-      execute!("LANG=en_US.utf-8 #{installer_command} --upgrade", :interactive => true)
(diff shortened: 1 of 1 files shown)


iNecas:
Don't forget, that this is meant to be used with katello upstream, where katello-installer is still pretty common and active https://github.com/Katello/katello-installer. A redmine issue against http://projects.theforeman.org/projects/foreman-maintain/issues/new is also appreciated.

iNecas:
Scratch that, I was too fast with the comment: on the other hand, since we still target satellite 6.1 (at least in theory), we should still keep `katello-installer` option there.

sean797:
Hows that @iNecas ? Thanks!

iNecas:
Much better, ready for testing @ntkathole

iNecas:
Well, ready for testing after rebase, as there is a conflict at the moment

iNecas:
Thanks @sean797  and @ntkathole

PR #163, opened 2018-03-29
Fixes #23057 - fix mapping between procedure's option to param

kgaikwad:


iNecas:
Approved code-wise, waiting for @ntkathole to approve from testing

PR #165, opened 2018-04-04
Fixes #22739, #21740 - Maintain hammer configuration

mbacovsky:
Updated hammer feature for review. List of changes:
 - hammer commands are run using `feature(:hammer).run('os list')
 - the run makes sure hammer is configured properly
 - stored hammer config is tested before use and reconfigured in case of failure
 - it can recover when admin password has changed
 - hammer configs are loaded at feature start, we load them as Hammer does. We end up with the same setup
 - we look up credentials in hammer configs and in installer answer file. When not valid we ask the user
 - we check the hostname we use is the right hostname so we can recover from localhost and ssl issues
 - hammer related logic is on one place. Hammer utils are not necessary
 - hammer concern adding hammer setup as preparation step to every executable remains (but could be removed when we decide on adding this only when needed)
 - global reporter was added to `ForemanMaintain.reporter`. Some refactoring of executable may follow 
 - updated docs


definitions/features/hammer.rb
```diff
@@ -1,6 +1,14 @@
+require 'uri'
+require 'socket'
```
mbacovsky: remove - not used

lib/foreman_maintain/utils/hammer.rb
```diff
@@ -1,79 +1 @@
-module ForemanMaintain
```
mbacovsky: remove from repo

test/lib/utils/hammer_test.rb
```diff
@@ -1,32 +1 @@
-require 'test_helper'
```
mbacovsky: remove from repo

lib/foreman_maintain/utils/hammer.rb
mbacovsky: done

test/lib/utils/hammer_test.rb
mbacovsky: done

kgaikwad:
Code wise looks good to me!

definitions/procedures/hammer_setup.rb
```diff
@@ -1,50 +1,15 @@
...
     description 'Setup hammer'
+    for_feature :hammer
   end
 
   def run
-    setup_from_default || setup_from_answers
-    puts "New settings saved into #{hammer.config_file}"
-    hammer.run_command('architecture list') # if not setup properly, an error will be risen
+    result = feature(:hammer).setup_admin_access
+    logger.info 'Hammer was configured seccessfully.' if result
```
ntkathole: Typo `seccessfully`

lib/foreman_maintain/cli/base.rb
```diff
@@ -101,7 +103,9 @@ def self.interactive_option
         delete_duplicate_assumeyes_if_any
 
         option ['-y', '--assumeyes'], :flag,
-               'Automatically answer yes for all questions'
+               'Automatically answer yes for all questions' do |assume|
+          ForemanMaintain.reporter.assumeyes(assume)
```
ntkathole:
```
# ./bin/foreman-maintain advanced procedure run hammer-setup -y
undefined method `assumeyes' for #<ForemanMaintain::Reporter::CLIReporter:0x000000024ef3e8>
```

```
# ./bin/foreman-maintain health check -y
undefined method `assumeyes' for #<ForemanMaintain::Reporter::CLIReporter:0x00000001a283d8>
```

kgaikwad: @ntkathole, 
Thanks for testing!
Error that you specified is due to this line. 
Instead of method call, assignment for `assumeyes`  needed here.

mbacovsky: Fixed, this was last minute change suggested by rubocop :)

mbacovsky:
I've updated the README to reflect the changes.
@iNecas, If there are no other objections could you please squash and merge?

lib/foreman_maintain/cli/base.rb
```diff
@@ -33,10 +33,12 @@ def print_check_info(check)
         puts "#{desc} #{tags}".strip
       end
 
+      def parse(arguments)
```
iNecas: Just out of curiosity, why do we need this method here?

iNecas:
One small question, but ready to merge after that's resolved

lib/foreman_maintain/cli/base.rb
mbacovsky: Good catch, this is leftover from some of my testing. I will remove it. I wanted to see when the clamp options are parsed to see if we can use the value of `assumeyes` during feature detection. 

BTW I found out the features are instantiated before Clamp parses the CLI options. I we will need the `assumeyes` during a feature init we will need to add parsing this option for ourselves by regexp or something.

mbacovsky:
@iNecas, method was removed.

lib/foreman_maintain/cli/base.rb
mbacovsky: Initializing the features after clamp is not possible due to `advanced procedure run` (and perhaps other commands in the future) which creates procedure-commands/options dynamically. At the point when the Command is declared the Features are detected/initialized. Then Clamp builds the options and only then it is able to parse CLI according to that options.

Solutions I see:
 - move the --assumeyes to the ForemanMaintain main command (which would move their possition on CLI: `foreman-maintain --assumeyes advanced procedure run`
 - drop dynamic commands/options
 - parse ARGV ourself when needed 

I prefer the last option as it keeps the dynamic CLI and better UX.

Anyway it does not block us now as we don't use assume yes in a feature init.

iNecas:
Thanks @mbacovsky

PR #166, opened 2018-04-05
[WIP] Fixes #22771 - make generic configurations for foreman_proxy

kgaikwad:


definitions/features/foreman_proxy.rb
```diff
@@ -6,30 +6,34 @@ class Features::ForemanProxy < ForemanMaintain::Feature
...
+  DHCPD_CONFIG_PATHS = ['/etc/dhcp3/dhcpd.conf',
+                        '/etc/dhcp/dhcpd.conf',
+                        '/usr/local/etc/dhcpd.conf'].freeze
+
+  attr_reader :dhcpd_conf_file, :settings_file, :proxy_settings
 
   def initialize
-    @dhcpd_conf_file = '/etc/dhcp/dhcpd.conf'
-    @cert_path = ForemanMaintain.config.foreman_proxy_cert_path
+    @dhcpd_conf_file = lookup_into(DHCPD_CONFIG_PATHS)
```
ekohl: A more advanced path would be to read `:config` from `dhcp_isc.yml`. The benefit is that you're reading what the proxy would read. `dhcp_isc.yml` can be found in `:settings_directory` which is a setting in `settings.yml`. I do realize you're starting to replicate the proxy configuration class, but it can catch misconfigurations within the proxy.

You may also need to verify when `dhcp_isc.yml` was introduced. IIRC this was 1.11 so older installs will read `:dhcp_config` from `dhcp.yml`. I think this would affect satellite 6.1 but not 6.2.

There's also the `dhcp_remote_isc` plugin which uses `dhcp_remote_isc.yml` instead of `dhcp_isc.yml` but is otherwise similar. You can find out which provider is used in `dhcp.yml` (`:use_provider`).

definitions/features/foreman_proxy.rb
```diff
@@ -6,30 +6,34 @@ class Features::ForemanProxy < ForemanMaintain::Feature
...
+    !dhcpd_conf_file.nil?
   end
 
   private
 
   def dhcp_curl_cmd
-    "curl -w '\n%{http_code}' --silent -ks --cert #{cert_path}/client_cert.pem \
-      --key #{cert_path}/client_key.pem \
-      --cacert #{cert_path}/proxy_ca.pem https://$(hostname):9090/dhcp"
+    "curl -w '\n%{http_code}' -ks --cert #{proxy_settings[:foreman_ssl_cert]} \
```
ekohl: If you provide the right ```--cacert``` then ```-k``` shouldn't be needed.

definitions/features/foreman_proxy.rb
```diff
@@ -84,4 +88,19 @@ def syntax_error_exists?
...
+  def load_proxy_settings
+    if settings_file
+      @proxy_settings = YAML.load_file(settings_file) || {}
+    else
+      raise "Couldn't find settings file at locations #{FOREMAN_PROXY_SETTINGS_PATHS.join(', ')}"
+    end
+  end
+
+  def lookup_into(file_paths)
+    file_paths.each do |file_path|
```
ekohl: This is equivalent to `file_paths.find { |file_path| file_exists?(file_path) }`

definitions/features/foreman_proxy.rb
```diff
@@ -6,30 +6,34 @@ class Features::ForemanProxy < ForemanMaintain::Feature
...
     dhcp_req_pass? && !syntax_error_exists?
   end
 
-  def services
-    {
-      'foreman-proxy'            => 20,
-      'smart_proxy_dynflow_core' => 20
-    }
+  def dhcpd_conf_file?
+    !dhcpd_conf_file.nil?
```
ekohl: You could use `dhcp.yml` (from `settings.d`) to check if the dhcp module is enabled (`:enabled`). Then if `:use_provider` is `dhcp_isc` then the ISC config should be validated. We also have the `dhcp_remote_isc` plugin which reads `dhcp_remote_isc.yml` and is mostly similar to the `dhcp_isc` provider.

ekohl:
@witlessbird could you have a look as well? I'd be interested in your opinion as a smart proxy maintainer.

kgaikwad:
@ekohl,
Thank you for pointers that you provided. I will go through it.
And if there is way, where I can handle this check in smart-proxy, I will be happy to add it to smart-proxy itself.

@lzap, could you please go through the above discussion and code changes from this PR.
As originally issue created by you for this check so it would be great and helpful if you provide your opinions as well before I go ahead & add more changes to it.

definitions/features/foreman_proxy.rb
```diff
@@ -27,9 +38,9 @@ def services
   private
 
   def dhcp_curl_cmd
-    "curl -w '\n%{http_code}' --silent -ks --cert #{cert_path}/client_cert.pem \
-      --key #{cert_path}/client_key.pem \
-      --cacert #{cert_path}/proxy_ca.pem https://$(hostname):9090/dhcp"
+    "curl -w '\n%{http_code}' -s --cert #{proxy_settings[:foreman_ssl_cert]} \
+      --key #{proxy_settings[:foreman_ssl_key]} \
```
iNecas: This should default to `proxy_settings[:ssl_{certificate,private_key,ca_file}]` if ithe `foreman_*` one are not specified.

ekohl: I actually wonder if it shouldn't always default to the non-foreman ones. They should be signed by the same CA as it's listening on so they should always be accepted. Then it is important that the host itself is in trusted hosts, but IIRC the installer does that by default.

Foreman can have a custom CA (say letsencrypt) so then the CA verification would fail. cert/key could also be different but that's less likely.

iNecas: If we didn't used primarily the `foreman_*` one, we would actually break the custom-cert case. The fact that in many cases, this two are the identical is just a lucky coincidence (connected to the fact, that we are using the client certs also as server certs, but we should not make this assumption IMO)

kgaikwad:
@ekohl , @iNecas ,

Updated pull-request with suggested changes. I have kept `DHCPD_CONFIG_PATHS` as it is. 
Please let me know whether I should consider `DHCPD_CONFIG_PATHS` or remove it completely.

mbacovsky:
@kgaikwad, during rebase It will be needed to move your ssl certs work into the new `curl_cmd` method. There is also the new `config_files` method that is there to collect all config files related to the feature. It would be great if you could update it with your new `*_PATH` constants. Let me know if it needs some clarification.

kgaikwad:
@iNecas, @mbacovsky, @ntkathole,

- Modified code so that `ssl certs` work with new `curl_cmd` method
- Updated `config_files` method with `settings.yml` path as well as `dhcp` directory according to `dhcpd.conf` location 
- Modified `confine` block to check feature `dhcp` instead of whole file path condition

definitions/features/foreman_proxy.rb
```diff
@@ -77,9 +93,12 @@ def backup_features(for_features)
...
 
   def curl_cmd
-    "curl -w '\n%{http_code}' --silent -ks --cert #{cert_path}/client_cert.pem \
-      --key #{cert_path}/client_key.pem \
-      --cacert #{cert_path}/proxy_ca.pem https://$(hostname):9090"
+    ssl_cert = proxy_settings[:foreman_ssl_cert] || proxy_settings[:ssl_certificate]
+    ssl_key = proxy_settings[:foreman_ssl_key] || proxy_settings[:ssl_private_key]
+    ssl_ca = proxy_settings[:foreman_ssl_ca] || proxy_settings[:ssl_ca_file]
+
+    "curl -w '\n%{http_code}' -s --cert #{ssl_cert} --key #{ssl_key} --cacert #{ssl_ca} \
```
iNecas: Please use `''` around the arguments ` --cert '#{ssl_cert}' --key '#{ssl_key}' --cacert '#{ssl_ca}' `, or even better, use `format_shell_args` from https://github.com/theforeman/foreman_maintain/blob/77e624b77b3c197ac25a427098d84e1f163f6d49/lib/foreman_maintain/concerns/system_helpers.rb#L161 to generate this part of the command.

kgaikwad: Done with the changes and updated PR.

iNecas:
travis is not happy

kgaikwad:
@iNecas ,
Fixed the test failures and updated PR.

@ntkathole, it would be good to have second round of testing with these changes.

ntkathole:
@kgaikwad Just want to sure..if dhcp config file is not present, the error should be only for 'foreman-proxy-verify-dhcp-config-syntax' check.
Currently foreman-maintain become fully-non functional if config file is missing.Is this correct behaviour?

```
Couldn't find settings file at /etc/foreman-proxy/settings.yml, /usr/local/etc/foreman-proxy/settings.yml
# ./bin/foreman-maintain upgrade check
Couldn't find settings file at /etc/foreman-proxy/settings.yml, /usr/local/etc/foreman-proxy/settings.yml
# ./bin/foreman-maintain health list
Couldn't find settings file at /etc/foreman-proxy/settings.yml, /usr/local/etc/foreman-proxy/settings.yml
```

kgaikwad:
Thank you @ntkathole!
Absolutely it is not correct behavior. 
Removed condition from confine block which checks whether proxy features list contains `dhcp` using API /features. 
Modified code so that
- if config file absent then it will give failure while running that check and 
- if feature is absent then it will just skip that check by saying - `DHCP feature is not enabled`.

ntkathole:
ACK looks good!

iNecas:
One nitpick, but otherwise good to go from my side.

kgaikwad:
@iNecas, 
Updated the PR with suggested changes.

PR #164, opened 2018-03-29
Fixes #22456 - Backup command

mbacovsky:
This is ongoing work on migration of katello-backup to foreman-maintain.
The goal is to provide the same functionality + handling of remote databases during backup.

- [ ] fix all TODO items
- [x] check changes that happened in katello-buckup since I started
- [x] make rubocop happy (refactoring)
- [x] remove WIP
- [x] add some tests

PR #169, opened 2018-04-11
Fixes #22886 - Add katello-restore

johnpmitsch:
Moving from https://github.com/theforeman/foreman-packaging

- [x] Add confirmation step
- [x] add tests
- [x] test manually on standalone foreman, foreman proxy w/ content, and satellite versions
- [x] block any satellite version < 6.3
- [x] add remote db support
- [x] add new checks for remote db actions
- [x] check mongo 3.4 support
- [x] rebase on latest updates from https://github.com/theforeman/foreman_maintain/pull/164


johnpmitsch:
Output currently looks like:
```
Running Restore backup
Validate backup has appropiate files:                                 [OK]
Validate hostname is the same as backup:                              [OK]
Setting file security: 
\ Restoring SELinux context                                           [OK]      
Restore configs from backup and reset with installer: 
| Installer reset                                                     [OK]      
Stop applicable services: 
postgresql, mongod, qdrouterd, qpidd, squid, pulp_resource_manager, pulp_streamer, foreman-proxy, pulp_celerybeat, pulp_workers, tomcat, foreman-tasks, httpd, puppetserver

\ All services stopped                                                [OK]      
Extract any existing tar files in backup: 
| Extracting Files                                                    [OK]      
Restore any existing postgresql dumps from backup: 
\ All services stopped                                                [OK]      
Restore mongo dump if present in backup: 
| All services stopped                                                [OK]      
Migrate pulp db: 
- Migrating pulp database                                             [OK]      
Start applicable services: 
postgresql, mongod, qdrouterd, qpidd, squid, pulp_resource_manager, pulp_streamer, foreman-proxy, pulp_celerybeat, pulp_workers, tomcat, foreman-tasks, httpd, puppetserver

/ All services started                                                [OK]      
Run daemon reload:                                                    [OK]
```

definitions/features/tar.rb
```diff
@@ -0,0 +1,86 @@
...
+    tar_command << '--selinux'
+    tar_command << '--overwrite' if options[:overwrite]
+    tar_command << '--gzip' if options[:gzip]
+    tar_command << "--#{options.fetch(:command, 'create')}"
+    tar_command << "--file=#{options.fetch(:archive)}"
+
+    if volume_size
+      split_tar_script = default_split_tar_script
+      tar_command << "--tape-length=#{volume_size}"
+      tar_command << "--new-volume-script=#{split_tar_script}"
```
johnpmitsch: I'll need this without --tape-length (see [here](https://git.io/vxxh7)), can you make --volume-script a separate option? I can do it in my PR, but will make things easier if you change here.

johnpmitsch:
@mbacovsky and @iNecas: The restore and backup PRs both require some complicated logic, I wonder if we can loosen up some of the restraints on rubocop rules? I can work within the present restraints or disable rules, but I think this is going to cause some unnecessary refactoring leading to unnecessary abstraction and more complexity. I also see some of these rules disabled elsewhere. Here are my proposed new settings:

```
+ Metrics/AbcSize:
+   Max: 20
+ 
+ Metrics/CyclomaticComplexity:
+   Max: 10
  
  Metrics/PerceivedComplexity:
    Max: 10
```

johnpmitsch:
@iNecas thanks, its mostly for the valid_*_backup? methods in `lib/foreman_maintain/utils/backup.rb` where we need the logic, but I was also able to remove disabling the cops in other places.

I also removed the double negation cop as I know @mbacovsky had to disable it in places too, I don't see a reason to keep it. I can disable it as needed if others disagree and would like to keep the cop.

definitions/checks/restore/validate_backup.rb
```diff
@@ -0,0 +1,72 @@
+require 'foreman_maintain/utils/backup'
+
+module Checks::Restore
+  class ValidateBackup < ForemanMaintain::Check
+    metadata do
+      description 'Validate backup has appropiate files'
```
ntkathole: Typo `appropiate`

ntkathole:
I tried to restore on sat 63 instance, faced wrong number of arguments (0 for 1) for extracting pulp data.
```
# ls /tmp/backup/satellite-backup-20180413023034
candlepin.dump  config_files.tar.gz  foreman.dump  metadata.yml  mongo_dump  pg_globals.dump  pulp_data.tar
```
```
# ./bin/foreman-maintain restore /tmp/backup/satellite-backup-20180413023034
Running Restore backup
Validate backup has appropiate files:                                 [OK]
Validate hostname is the same as backup:                              [OK]
Setting file security:
/ Restoring SELinux context                                           [OK]
Restore configs from backup and reset with installer:
/ Installer reset                                                     [OK]
Stop applicable services:

postgresql, mongod, squid, qpidd, qdrouterd, tomcat, pulp_streamer, smart_proxy_dynflow_core, pulp_resource_manager, pulp_celerybeat, pulp_workers, foreman-proxy, httpd, foreman-tasks

- All services stopped                                                [OK]
Extract any existing tar files in backup:
\ Extracting pulp data                                                [FAIL]
wrong number of arguments (0 for 1)
Scenario [Restore backup] failed.

  [restore-extract-files]

use --whitelist="restore-extract-files"

```

.rubocop.yml
```diff
@@ -1,6 +1,18 @@
...
+Metrics/AbcSize:
+  Max: 20
+
+Metrics/CyclomaticComplexity:
+  Max: 10
+
+Metrics/PerceivedComplexity:
+  Max: 10
+
+Style/DoubleNegation:
```
mbacovsky: This was already disabled in hammer feature, could you rebase please?

johnpmitsch: Thanks, rebased

johnpmitsch:
@ntkathole thanks, that should be fixed now

johnpmitsch:
Removing WIP so we can review even though I'm still testing manually.

Remote db support has not been added yet, depending on how long the review of this PR and https://github.com/theforeman/foreman_maintain/pull/164 takes, I will add it here or in a separate PR.

johnpmitsch:
Added https://github.com/theforeman/foreman-packaging/pull/2311

johnpmitsch:
@mbacovsky I implemented the remote mongo support, check my latest commit. I'm open to any suggestions, this was just the first way I thought of :)

johnpmitsch:
Updated for remote DB support

johnpmitsch:
@mbacovsky thanks for the mongo updates, I made a small update to something like this:

```
  # options[:config] (Hash) mongo configuration
  # options[:database] (String) database to connect to
  # options[:args] (String) additional arguments
  def base_command(command, options)
    options[:config] ||= configuration
    options[:args] ||= ""

    config = options[:config]

    command += " #{options[:database]}" if options[:database]
    command += " -u #{config['username']}" if config['username']
    command += " -p #{config['password']}" if config['password']
    command += " --host #{config['host']} --port #{config['port']}"

    if config['ssl']
      command += " --ssl"
      command += " --sslCAFile #{config['ca_path']}" if config['ca_path']
    end

    cmd = "#{command} #{options[:args]}"
  end
```

johnpmitsch:
This is fully functional now, once https://github.com/theforeman/foreman_maintain/pull/164 is wrapped up I'll rebase and fix any remaining rubcop issues

mbacovsky:
@johnpmitsch I've finished the mongo feature with remote db support. I've added some improvements from the installer `--reset` patch.

definitions/features/mongo.rb
```diff
@@ -0,0 +1,133 @@
...
+  def mongo_command(args, database = 'pulp_database')
+    base_command('mongo', :args => args, :database => database)
+  end
+
+  def restore(file)
+    execute!(base_command('mongorestore', :args => " #{file}"))
+  end
+
+  def dropdb
+    execute!(mongo_command(" --eval 'db.dropDatabase()'"))
```
alda519: need to read database name from /etc/pulp/server.conf and not use pulp_database

alda519:
Tried to restore offline backup of external database
```
# foreman-maintain restore /tmp/backup/foreman-backup-2018-04-23-08-25-29/
...
Validate backup has appropriate files:                                [FAIL]

The given directory does not contain the required files or has too many files

All backup directories contain: config_files.tar.gz
An online backup directory contains: mongo_dump, candlepin.dump, foreman.dump, pg_globals.dump
An offline backup directory contains: mongo_data.tar.gz, pgsql_data.tar.gz
A logical backup directory contains: mongo_dump, candlepin.dump, foreman.dump, pg_globals.dump, mongo_data.tar.gz, pgsql_data.tar.gz
Including pulp_data.tar is optional and will restore pulp data to the filesystem if included.

Only the following files were found: pulp_data.tar, foreman.dump, candlepin.dump, mongo_dump, config_files.tar.gz, metadata.yml
```

johnpmitsch:
@alda519 that looks like an online backup, was there a `pg_globals.dump` file present?

johnpmitsch:
@alda519 just remembered we don't pg_globals for remote databases, I'll have to fix that in the check (which I think is your point) :smile:

alda519:
it was offline `I, [2018-04-23 09:41:47-0400 #30068]  INFO -- : Running foreman-maintain command with arguments [["backup", "offline", "/tmp/backup/"]]`
see 09:41:47, and this is content of it
```
ls /tmp/backup/foreman-backup-2018-04-23-09-41-47/
candlepin.dump  config_files.tar.gz  foreman.dump  metadata.yml  mongo_data.tar.gz  pulp_data.tar
```
yes, my point is that the content does not fit any of online, offline backups described there

johnpmitsch:
@alda519 nice catch, will update

mbacovsky:
@johnpmitsch I've fixed the tests, rubocop, some bugs and commit messages.

alda519:
```
foreman-maintain restore /tmp/backup/offline/foreman-backup-2018-04-24-05-00-05
...
Restore mongo dump: 
\ Restoring mongo dump                                                [FAIL]    
Failed executing mongorestore --host  --port  , exit status 255:
 2018-04-24T05:28:04.427-0400 getaddrinfo("--port") failed: Name or service not known
couldn't connect to [--port] couldn't connect to server --port:27017 (0.0.0.0), address resolved to 0.0.0.0

```

johnpmitsch:
I have to rebase on @mbacovsky's latest commits and update a few things so please hold off functional reviews here for now, the priority is on getting in https://github.com/theforeman/foreman_maintain/pull/164 to unblock this PR

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -52,6 +52,24 @@ def dump_db(file, config = configuration)
...
+        if localdb
+          dump_cmd = "runuser - postgres -c 'pg_restore -C -d postgres #{file}'"
+          execute!(dump_cmd)
+        else
+          # TODO: figure out how to completely ignore errors. Currently this
+          # sometimes exits with 1 even though errors are ignored by pg_restore
+          dump_cmd = base_command(config, 'pg_restore') +
+                     ' --no-privileges --clean --disable-triggers -n public ' \
+                     "-d #{config['database']} #{file}"
+          execute!(dump_cmd, :valid_exit_statuses => [0, 1])
```
mbacovsky: add :hidden_patterns => [config['password']] to hide credentials from the logs

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -52,6 +52,24 @@ def dump_db(file, config = configuration)
...
+        end
+      end
+
+      def restore_pg_globals(pg_globals, config = configuration)
+        puts base_command(config, 'psql') + " -f #{pg_globals} postgres 2>/dev/null"
```
mbacovsky: Did you meant to execute this? If so add also `:hidden_patterns => [config['password']]` to hide credentials from the logs

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -184,18 +190,44 @@ def get_lv_path(lv_name)
...
-
       def find_dir_containing_file(directory, target)
         result = nil
         Find.find(directory) do |path|
           result = File.dirname(path) if File.basename(path) == target
         end
         result
       end
+
+      def foreman_proxy_with_content?
```
mbacovsky: I'd suggest move all these to `feature(:instance)` See comment above for reasons

definitions/features/pulp.rb
```diff
@@ -30,6 +30,7 @@ def config_files
       '/etc/crane.conf',
       '/etc/default/pulp_workers',
       '/var/lib/qpidd'
```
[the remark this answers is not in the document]
mbacovsky: it makes foreman-maintain not executable at the moment

alda519:
```
# foreman-maintain health list
/usr/share/gems/gems/foreman_maintain-0.1.5/lib/foreman_maintain/param.rb:36:in `process': Param backup_dir is required but no value given (ArgumentError)
	from /usr/share/gems/gems/foreman_maintain-0.1.5/lib/foreman_maintain/executable.rb:35:in `block in setup_params'
...
```

alda519:
breaks backup
```
# foreman-maintain backup offline --skip-pulp-content -y /tmp/backup/offline
...
Backup mongo offline: 
- Collecting Mongo data                                               [FAIL]    
Failed executing tar --selinux --create --file=/tmp/backup/offline/katello-backup-2018-05-04-10-52-43/mongo_data.tar --create --file=/tmp/backup/offline/katello-backup-2018-05-04-10-52-43/mongo_data.tar --exclude=mongod.lock --listed-incremental=/tmp/backup/offline/katello-backup-2018-05-04-10-52-43/.mongo.snar --transform 's,^,var/lib/mongodb/,S', exit status 2:
 tar: Multiple archive files require `-M' option
Try `tar --help' or `tar --usage' for more information.
```

definitions/procedures/restore/installer_reset.rb
```diff
@@ -0,0 +1,61 @@
...
+        :directory => '/',
+        :archive => backup.file_map[:config_files][:path],
+        :gzip => true
+      }
+
+      feature(:tar).run(tar_options)
+    end
+
+    def installer_cmd
+      installer = "yes | #{installer_name} --scenario #{scenario_name} -v --reset"
```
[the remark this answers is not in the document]
johnpmitsch: I was able to move some of the logic to the installer feature, but we do need the `yes` pipe

johnpmitsch:
Updated for mongo 3.4, this should be working fine, but I'm leaving it WIP as I still have to add the preparation steps checks for the database procedures. This will most likely require me to break up the postgresql dumps procedure as well.

ntkathole:
```
# ./bin/foreman-maintain upgrade check
/root/foreman_maintain/lib/foreman_maintain/param.rb:36:in `process': Param backup_dir is required but no value given (ArgumentError)
        from /root/foreman_maintain/lib/foreman_maintain/executable.rb:35:in `block in setup_params'
        from /root/foreman_maintain/lib/foreman_maintain/executable.rb:34:in `each_value'
```

johnpmitsch:
@ntkathole @mbacovsky @iNecas what  is the best way to handle the required param errors? I'm confused as to what I am doing to create them, is there some sort of best practices I am neglecting?

I wonder if the process for finding checks/procedures should change now that foreman-maintain is growing to have more functionality?

johnpmitsch:
@mbacovsky @ntkathole I broke apart the procedures and added the remote db checks in any relevant procedure. I also move the "does this file exist?" type logic to the scenario, so steps are only added if the file exists. These changes are in a separate commit.

Currently the tests are failing with: `Restore subcommand is supported by Satellite 6.3+. Please use katello-restore instead.`. What kind of box do we use for travis? Are we assuming downstream features are present?

johnpmitsch:
@ntkathole fixed the autodetection issue by using manual detection.

One thought for the future, seems like this is something that shouldn't be so inclusive any more? As in, instead of excluding procedures w/ manual_detection, maybe we should only include them with an auto_detection type param in metadata? Foreman-maintain is growing to cover many use cases besides upgrades.

johnpmitsch:
manual_detection fixed the tests as well, so please ignore my comment about the tests failing :)

definitions/procedures/restore/candlepin_dump.rb
```diff
@@ -0,0 +1,32 @@
...
+      confine do
+        feature(:candlepin_database)
+      end
+    end
+
+    def run
+      backup = ForemanMaintain::Utils::Backup.new(@backup_dir)
+
+      with_spinner('Restoring candlepin postgresql dump') do |spinner|
+        feature(:service).handle_services(spinner, 'start', :only => ['postgresql'])
```
alda519: this does not work on remote database, because postgresql service is not installed at all, getting error
```
E, [2018-05-09 11:00:00+0200 #32464] ERROR -- : No services found matching your parameters (RuntimeError)
/usr/share/gems/gems/foreman_maintain-0.2.1/definitions/features/service.rb:44:in `filtered_services'
/usr/share/gems/gems/foreman_maintain-0.2.1/definitions/features/service.rb:15:in `handle_services'
/usr/share/gems/gems/foreman_maintain-0.2.1/definitions/procedures/restore/candlepin_dump.rb:18:in `block in run'
```

the same thing fails also in `definitions/procedures/restore/foreman_dump.rb`
mongo works `definitions/procedures/restore/mongo_dump.rb` but I would fix it too to prevent failures one mongod was removed

definitions/procedures/restore/mongo_dump.rb
```diff
@@ -0,0 +1,34 @@
...
+        feature(:service).handle_services(spinner, 'stop', :only => feature(:mongo).services.keys)
+      end
+    end
+
+    def drop_and_restore_mongo(backup, spinner)
+      spinner.update('Dropping pulp_database')
+      feature(:mongo).dropdb
+
+      spinner.update('Restoring mongo dump')
+      feature(:mongo).restore(File.join(backup.file_map[:mongo_dump][:path], 'pulp_database'))
```
alda519: restore remote mongo database does not work
```
E, [2018-05-09 12:04:03+0200 #10368] ERROR -- : Failed executing scl enable rh-mongodb34 -- mongorestore -u pulp1 -p [FILTERED] --host mongo.host.redhat.com --port 27017 /root/backup/data/offline/mongo_dump/pulp_database, exit status 1:
 2018-05-09T12:04:03.472+0200   Failed: error connecting to db server: server returned error on SASL authentication step: Authentication failed. (ForemanMaintain::Error::ExecutionError)
```
 * hardcoded `pulp_database` name
 * missing `--db <database>` option

pondrejk:
Checked with some modified robottelo tests at https://github.com/SatelliteQE/robottelo/pull/5973/files on nightly build, results looked ok

johnpmitsch:
@alda519 good catch, I updated not start services when dbs are local and to use the mongo db name from the configs.

ntkathole:
```
# ./bin/foreman-maintain restore /tmp/satellite-backup-2018-05-10-02-34-58
ERROR: Restore subcommand is supported by Satellite 6.3+. Please use katello-restore instead.
# echo $?
1
```
Returns 1...Is this intentional?

mbacovsky:
@johnpmitsch as for the autodetection - IMO the main goal for f-m is still upgrades and I hope the number of checks applicable there will grow. How we use checks is slightly deviating from the original purpose of the checks so my preference to have checks as opt-out. Maybe we can update detector to exclude checks with required params as it is safe indicator that the check is not intended to run automatically.

For scenarios I'm not sure yet.

definitions/procedures/restore/candlepin_dump.rb
mbacovsky: The goal is to fix f-m service to handle remote DBs correctly. It should not restart db services for remote dbs and only run db ping there for `start` and `status`. I'll prepare PR soon.

johnpmitsch:
@ntkathole seems like non-zero exit is appropriate there, the command didn't run properly as it is not supported on that machine.

alda519:
mongorestore still needs `--db` option, it cannot authenticate without it
```
D, [2018-05-11 10:24:43+0200 #8766] DEBUG -- : Running command scl enable rh-mongodb34 -- mongorestore -u pulp1 -p [FILTERED] --host ....redhat.com --port 27017 /tmp/backup/data/offline/mongo_dump/pulp1db with stdin nil
D, [2018-05-11 10:24:43+0200 #8766] DEBUG -- : output of the command:
 2018-05-11T10:24:43.790+0200   Failed: error connecting to db server: server returned error on SASL authentication step: Authentication failed.
```

lib/foreman_maintain/utils/backup.rb
```diff
@@ -206,6 +206,12 @@ def sql_dump_files_exist?
           file_map[:candlepin_dump][:present]
       end
 
+      def handle_postgres_service(spinner, action)
```
mbacovsky: Why class representing Backup data cares about DB?

mbacovsky:
For the last commit handling the pg service I prefer the original version with service starts outside the the restore procs. The only issue I can see there is we should add Service start proc instead of the feature method call as it is evaluated during the compose and not during the scenario run. 

I'm preparing pr fixing the service on remote dbs so the check for local DBs could be removed.

johnpmitsch:
@mbacovsky @alda519 Updated to fix both the mongo db issue and the handling of services

alda519:
well, I retested it and looks pretty good now
I hit only these two issues
http://projects.theforeman.org/issues/23498
http://projects.theforeman.org/issues/23592

definitions/procedures/restore/installer_reset.rb
```diff
@@ -0,0 +1,39 @@
...
+        else
+          spinner.update('Installer reset')
+          execute!(installer_cmd)
+        end
+      end
+    end
+
+    def installer_cmd
+      installer = "yes | #{feature(:installer).installer_command} "
+      installer << "--scenario #{feature(:installer).scenario_name} -v --reset"
```
mbacovsky: @johnpmitsch, why do we compute the scenario name here? At this point the configs are restored so the installer can figure it out on its own if `--scenario` is omitted. Or we can use `feature(:installer).last_scenario` if necessary.

mbacovsky:
@johnpmitsch the last changes looks good. I found one potential issue in the installer_reset, see comment inline. I'm looking into the mongo issue @alda519 found.

mbacovsky:
The SCL only Mongo DB should be fixed in https://github.com/theforeman/foreman_maintain/pull/179

johnpmitsch:
@mbacovsky updated to remove the scenario part of the installer command

johnpmitsch:
@ntkathole can you confirm all your issues were addressed?

ntkathole:
Looks good to me...deferring to @alda519 to ack from remote dbs side.

alda519:
Remote dbs are good as well.

iNecas:
Regarding http://projects.theforeman.org/issues/23592, here is the change on the packaging side https://github.com/theforeman/foreman-packaging/commit/b0bc45ff81, which lead to the directory changes. I think we should backup `/var/lib/foreman-proxy/.ssh` if exists, and backup `/usr/share/foreman-proxy/.ssh` only when it's not a symlink. This way, we should be able to cover both cases, before and after the `.ssh` dir location change.

The installer needs to be changed anyway.

iNecas:
Anyway, I would probably not block this PR on this issue, and I'm ok with handling that in separate one.

mbacovsky:
@iNecas, @johnpmitsch, it seems we do not backup `/var/lib/foreman-proxy/ssh` at all. This should IMO be fixed before merging. I'm not sure the strategy Ivan suggest would prevent the failures. I'd backup both paths and if there is failure during restore `rm -rf /usr/share/foreman-proxy/.ssh` should work as workaround.  

@johnpmitsch , during restoring from regular offline backup I noticed one confusing thing
```
Drop postgresql databases:                                            [SKIPPED]
No postgresql dump files present in backup or postgresql databases are remote.
```
Why is the db drop restricted only for local DBs - we have dumps mainly for remote db?  Could be the step not included if the dump files are not present (by moving the test to the scenario as it is done for other procs)?

johnpmitsch:
Currently updating, please don't review just yet

@mbacovsky I'm moving the logic to the scenario, but how do you suggest I drop the databases remotely? Currently we use the 'dropdb' command

johnpmitsch:
@mbacovsky thanks, updated

PR #171, opened 2018-04-16
Fixes #23289 - Add note about puppetserver

johnpmitsch:
both 'puppetserver' and 'puppet-server' are puppet server
packages depending on puppet 3/4

### DIFF
--- a/definitions/features/puppet_server.rb
+++ b/definitions/features/puppet_server.rb
@@ -2,6 +2,9 @@ class Features::PuppetServer < ForemanMaintain::Feature
   metadata do
     label :puppet_server
 
+    # We only check puppetserver and not puppet-server, as puppet-server
+    # is a part of httpd and relies on httpd service to restart, therefore
+    # not requiring a separate service to restart
     confine do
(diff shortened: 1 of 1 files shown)


ntkathole:
Might need to update service name as well 
```
D, [2018-04-16 10:10:37-0400 #68873] DEBUG -- : Running command systemctl is-enabled ```
puppetserver 2>&1 | tail -1 with stdin nil
D, [2018-04-16 10:10:37-0400 #68873] DEBUG -- : output of the command:
 Failed to get unit file state for puppetserver.service: No such file or directory
```

johnpmitsch:
@ntkathole I marked this WIP as I was still investigating the issue, it turns out we don't have to restart puppet-server, its included in httpd. I added a comment for clarity, let me know if it makes sense.

ntkathole:
@johnpmitsch thanks..makes sense!

johnpmitsch:
@kgaikwad can I get a merge?

kgaikwad:
Thank you @johnpmitsch

PR #174, opened 2018-04-23
Fixes #23367 - add 6.4 target version

iNecas:
For development of next Sat release.

### DIFF
--- a/definitions/scenarios/upgrade_to_satellite_6_4.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_4.rb
@@ -1,0 +1,79 @@
+module Scenarios::Satellite_6_4
+  class Abstract < ForemanMaintain::Scenario
+    def self.upgrade_metadata(&block)
+      metadata do
+        tags :upgrade_to_satellite_6_4
+        confine do
+          feature(:downstream) && feature(:downstream).current_minor_version == '6.3'
+        end
+        instance_eval(&block)
+      end
+    end
+  end
+
+  class PreUpgradeCheck < Abstract
+    upgrade_metadata do
+      description 'Checks before upgrading to Satellite 6.4'
+      tags :pre_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+      add_step(Checks::Repositories::Validate.new(:version => '6.4'))
+    end
+  end
+
+  class PreMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures before migrating to Satellite 6.4'
+      tags :pre_migrations
+    end
+
+    def compose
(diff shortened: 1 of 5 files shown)


definitions/scenarios/upgrade_to_satellite_6_4.rb
```diff
@@ -0,0 +1,80 @@
...
+    upgrade_metadata do
+      description 'Checks before upgrading to Satellite 6.4'
+      tags :pre_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+      add_step(Checks::RemoteExecution::VerifySettingsFileAlreadyExists.new)
```
kgaikwad: For this check, we have added [confine](https://github.com/theforeman/foreman_maintain/blob/master/definitions/checks/remote_execution/verify_settings_file_already_exists.rb#L8) block which checks for `downstream` feature plus condition `feature(:downstream).current_minor_version == '6.2'`.

If this check required here for sat 6.4 then we will need to modify this condition.

iNecas: Good point, removed. Not needed after it was migrated already

ntkathole:
@iNecas if you want to cover this BZ as well https://bugzilla.redhat.com/show_bug.cgi?id=1571368

iNecas:
Good idea, updated

ntkathole:
@iNecas might be we need some more checks and modifications regarding puppet. As satellite 6.3. puppet3 -> sat6.4 update failed due to dependency
```
Error: Package: puppet-server-3.8.6-4.el7sat.noarch (@rhel-7-server-satellite-6.3-rpms)
           Requires: puppet = 3.8.6-4.el7sat
           Removing: puppet-3.8.6-4.el7sat.noarch (@rhel-7-server-satellite-6.3-rpms)
               puppet = 3.8.6-4.el7sat
           Obsoleted By: puppet-agent-5.5.0-1.el7sat.x86_64 (sat6)
               puppet >= 4.0.0-1.el7sat
```
and upgrade path will be sat 63 with puppet4 -> sat64. So possible twicks can be if 6.3 with puppet3, raise warning (do not allow ) upgrade to 6.4 or add support to f-m to update sat 63 to puppet4.

iNecas:
Can we handle that in separate PR?

iNecas:
Please raise a BZ for that

ntkathole:
@iNecas yes sure. Here it is https://bugzilla.redhat.com/show_bug.cgi?id=1571859

iNecas:
I hear ack then? @kgaikwad mind having a final look and merging eventually?

kgaikwad:
@iNecas, 
Ack, good to go :thumbsup:

kgaikwad:
Thank you @iNecas & @ntkathole

PR #175, opened 2018-04-30
Fixes #23411 - use dynflowd in services for foreman-tasks

johnpmitsch:
In foreman 1.17+, foreman-tasks was renamed to dynflowd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ test/definitions/config/foreman-maintain-hammer-default.yml
 data.yml
 .ruby-version
 .ruby-gemset
+*~
+*.swp
+*.swo
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -108,7 +108,11 @@ class Features::ForemanTasks < ForemanMaintain::Feature
   end
 
   def services
-    { 'foreman-tasks' => 30 }
+    if check_min_version('foreman', '1.17')
+      { 'dynflowd' => 30 }
+    else
+      { 'foreman-tasks' => 30 }
+    end
   end
 
   private


jturel:
@johnpmitsch could you add a description of what an acceptance test for this change would be for someone who has never used foreman_maintain before? :)

johnpmitsch:
@jturel yes, there is a reproducer in the issue. The main thing is after you disable and enable services you can restart tasks fine.

From the foreman_maintain directory:
```
./bin/foreman-maintain service disable
./bin/foreman-maintain service enable
./bin foreman-maintain service restart
```

This should work on katello/satellite systems using foreman > 1.17 and < 1.17 (that is where the service name changed) so its worth checking both of these cases (can use older Katello or Satellite to check < 1.17)

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -56,6 +56,11 @@ def execute?(command, input = nil)
         $CHILD_STATUS.success?
       end
 
+      def rpm_greater_than_version?(name, target_version)
```
jturel: nitpick: this method is checking >= not just > so maybe `rpm_gte_version` ?

jturel:
Works nicely. I tested 3.6 and 3.5. Left one comment that you may or may not want to address, but I will ACK it now!

lib/foreman_maintain/concerns/system_helpers.rb
jturel: i'm good with that!

johnpmitsch: updated the method name :+1:

jturel: still works - ship it 👍

johnpmitsch:
@iNecas of course that already exists :smile: updated to use `check_min_version`

iNecas:
Thanks @johnpmitsch

PR #178, opened 2018-05-14
Fixes #23589, #23590 - Handle remote dbs during service actions

mbacovsky:
When service command is used on deployment with remote dbs we
do not have access to the DB services. We try to access the DB
and log the status. On service start we make sure the remote DB
is up and running.

### DIFF
--- a/definitions/features/instance.rb
+++ b/definitions/features/instance.rb
@@ -29,6 +29,10 @@ class Features::Instance < ForemanMaintain::Feature
     end
   end
 
+  def database_remote?(feature)
+    !!feature(feature) && !feature(feature).local?
+  end
+
   def database_local?(feature)
     !!feature(feature) && feature(feature).local?
   end
--- a/definitions/features/service.rb
+++ b/definitions/features/service.rb
@@ -66,6 +66,42 @@ class Features::Service < ForemanMaintain::Feature
   end
 
   def perform_action_on_service(action, service)
+    if service == 'postgresql'
+      if feature(:instance).postgresql_local?
+        perform_action_on_local_service(action, service)
+      end
+      if feature(:instance).database_remote?(:candlepin_database)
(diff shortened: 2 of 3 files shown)


iNecas:
Ack codewise, deferring to @ntkathole with testing.

alda519:
Postgres looks quite good. If only one of [foreman, candlepin] is remote, then
```
# foreman-maintain service status --only postgresql
...
● postgresql.service - PostgreSQL database server
   Loaded: loaded (/etc/systemd/system/postgresql.service; enabled; vendor preset: disabled)
   Active: active (running) since Mon 2018-05-14 09:42:00 CEST; 36min ago
...
For Foreman DB the service is on remote host. The DB is UP.
```
If both are remote, then postgresql service is not installed at all and `foreman-maintain service` skips postgres completely, which seems ok to me.

Remote mongo works only if it is up
```
# foreman-maintain service status --only rh-mongodb34-mongod 
...
For Pulp DB the service is on remote host. The DB is UP.
```
because it cannot check version and use [mongo 2.6 by default](https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/mongo.rb#L40) (quite related  to http://projects.theforeman.org/issues/23498)

And I noticed that
`# foreman-maintain service start --only postgresql,rh-mongodb34-mongod` does not show 
`...the service is on remote host. The DB is UP.`  lines in terminal because they are overwritten by `\r| All services started...`

mbacovsky:
@alda519 oh, good catch with the mongo being down. That is a nasty gap in my plan. I'll try to address this in the other PR (https://github.com/theforeman/foreman_maintain/pull/179).

The overwritten messages are intentional as this is the same behavior for other services. The message should be also logged for eventual troubleshooting. I believe that when the service is down the message is  visible in the output.  When everything is okay users do not care much about the details.

iNecas:
Thanks @mbacovsky and @alda519

PR #179, opened 2018-05-16
Fixes #23498 - run with mongo 3.4 from scl

mbacovsky:
To get version of the db server (may be remote) we assume the
'mongo' command is always available. If there is only mongo from SCL
present the mongo client has to be called from  within a scl.

With current state of things it is safe to assume there is at least
some mongo client installed. For the future We will need to figure out
how to install a correct version of the client.

### DIFF
--- a/definitions/features/mongo.rb
+++ b/definitions/features/mongo.rb
@@ -31,16 +31,31 @@ class Features::Mongo < ForemanMaintain::Feature
     @configuration = load_db_config(config_file)
   end
 
+  # guess mongo version from installed packages
+  # safe method to run mongo commands prior we know
+  # what version uses the target DB
+  def available_core
+    return @core unless @core.nil? # return correct mongo ver if known
+    if @available_core.nil?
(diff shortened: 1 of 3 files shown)


iNecas:
ACK codewise.  alda519 mind testing to check it fixes the issue for you?

alda519:
Sure.
And it does. But there is one more problem - if the mongo service is down, then `server_version` cannot connect to it to get the version.
What about simplifying it, and use @core = available_core directly? But I am not sure if we can rely on that the server is the same version as client and what happens to mongo 2.6 packages after upgrade to 3.4)

alda519:
Yup, now it works when mongo is down. Perfect.

iNecas:
One last thing: rubocop is not happy. Going to merge after that's resolved.

mbacovsky:
@iNecas rubocop seems satisfied now.

iNecas:
Thanks @mbacovsky

PR #180, opened 2018-05-16
Fixes #22198 - maintenance-mode subcommand with start/stop/status

kgaikwad:


definitions/features/service.rb
```diff
@@ -53,6 +53,16 @@ def action_past_tense(action)
     action_word_modified(action) + 'ed'
   end
 
+  def perform_action_on_service(action, service)
+    command = service_command(action, service)
+    if action == 'status'
+      status = execute(command)
+      puts "\n\n#{status}\n\n"
```
iNecas: We should not do printing from features, but rather returning the status as return value and printing from procedure, as there is the reporter that should be used for reporting the data to the user.

iNecas: On the other hand, I see that the change was already in, so leaving this as optional for this PR to change that.

definitions/features/iptables.rb
```diff
@@ -0,0 +1,40 @@
+class Features::Iptables < ForemanMaintain::Feature
```
iNecas: I wonder if we should have separate features for `iptables`, `cron` + concern for the maintenance file, vs. having `maiintenance mode` feature, implementing all of those methods?

kgaikwad:
@iNecas , @ntkathole,
Resolved the conflicts and updated the PR.

johnpmitsch:
@kgaikwad does this work on upstream katello?

kgaikwad:
@johnpmitsch,

I haven't tested this scenario with upstream Katello.
In maintenance-mode command, we are only dealing with `iptables` and `crond` service so it should work in case of upstream Katello as well but still I will verify this on my setup to be make sure.

It would be great if you review [custom-rules](https://github.com/theforeman/foreman_maintain/pull/180/files#diff-92403856c814894e53c5bcb0423a0b85R44) which we are adding for our custom chain in `iptables` and please let me know if I need to handle any specific requests in case of upstream Katello.

ntkathole:
- Remove wrong options in help for start/stop/status
```
# ./bin/foreman-maintain maintenance-mode start -h
Usage:
    foreman-maintain maintenance-mode start [OPTIONS]

Options:
    --exclude EXCLUDE             A comma-separated list of services to skip
    --only ONLY                   A comma-separated list of services to include
    -h, --help                    print help
```

- Performing maintenance-mode start will not stop crond service but while perfoming upgrade crond service stop. @kgaikwad should Stop cron service run also on maintenance-mode start command?

```
./bin/foreman-maintain maintenance-mode start
Running Start maintenance-mode
Add chain to iptables:                                                [OK]
Create maintenance_file:                                              [OK]
```
```
 ./bin/foreman-maintain upgrade run --target-version 6.4.z 
Add chain to iptables:                                                [OK]
Stop cron service:                                                    [OK]
Create maintenance_file:                                              [OK]
```
```

**Comment by ntkathole** [comment]
- Pre_migration/post_migrations only runs sync_plans_disable/sync_plans_enable

```
# ./bin/foreman-maintain advanced procedure by-tag pre-migrations
disable active sync plans: 
/ Total 0 sync plans are now disabled.                                [OK]      
```

earlier it was 
```
# ./bin/foreman-maintain advanced procedure by-tag

post-migrations               Run procedures tagged #post_migrations: maintenance_mode_disable, sync_plans_enable
pre-migrations                Run procedures tagged #pre_migrations: maintenance_mode_enable, sync_plans_disable
```

**Comment by kgaikwad** [comment]
@ntkathole,

Thanks for pointing incorrect options. Removed those and updated PR.
For second scenario, I need to check on reproducer as I am getting the step which stops crond service.

**Comment by kgaikwad** [comment]
@ntkathole,

Added confine block inside procedures related to cron where it checks the value of ` :enable_cron_stop:` config before running procedure. By default it is set to false. 

Previously I have added it in scenario only that's why you were getting it in upgrade path.

```
# ./bin/foreman-maintain advanced procedure by-tag pre-migrations
disable active sync plans: 
/ Total 0 sync plans are now disabled.                                [OK]      
```
Currently, there is no single procedure to run for maintenance-mode enable/disable.
There are three procedures which I kept hidden from `advanced procedure` command because now we have maintenance-mode command.

@iNecas, @mbacovsky,
Please confirm above behavior is appropriate or should I change it?
And about default value for `:enable_cron_stop:`.

**Comment by iNecas** [comment]
I'm ok with hiding the procedures, although it might be useful in some corner cases. I vote for `:enable_cron_stop` to be set to true by default.

**Comment by kgaikwad** [comment]
@iNecas,

Thanks for your inputs. I will make `:enable_cron_stop` to true by default.

```
One thing I'm not sure about is the Create maintenance_file procedure, as it might be confusing
what it actually does. An alternative could be updating info in storage.

Alternatively, the maintenance mode procedures could have a noop mode,
ale the status could be determined by running enable with noop and looking if any procedure
would be applied.
```

Please correct me if I misunderstood anything here.
You would like to have `noop mode` for every procedure from maintenance-mode which actually checks whether that procedure is applied or not. 
Example for iptables, I need to check whether chain really exists or not.
Depending upon this mode value per procedure, I need to display status of maintenance-mode, Right?

**Comment by iNecas** [comment]
yes

**Comment by johnpmitsch** [comment]
@kgaikwad Katello is requiring Foreman-maintain https://github.com/theforeman/foreman-packaging/pull/2496 starting in 3.7. If you don't plan to test and maintain this functionality on Katello, I think we should hide the command on any upstream katello system. Otherwise, I think its a great addition for Katello as well as Satellite.

I'm not sure about the iptable rules, but imagine they would be similar to satellite. I'll ask around to be sure.

**Comment by iNecas** [comment]
@johnpmitsch there is no reason this should not work with katello

**Comment by johnpmitsch** [comment]
The iptables rules should be fine for Katello as well. Does this disable sync plans and cron jobs as well?

**Comment by bbuckingham** [comment]
If a user has custom iptable rules (that aren't part of the normal foreman/katello rules), will they be retained when in/out of maintenance mode?   I am assuming those rules would not be lost, but confirming.

**Comment by iNecas** [comment]
What we do is creating a custom chain to introduce the rules, and then clean up the chain - no changes to other existing rules.

**Comment by kgaikwad** [comment]
@johnpmitsch,

It will not disable `sync plans`. In maintenance-mode we are only adding custom chain and rules in `iptables` and stopping `crond` service.

If user don't want to stop `crond` service then they can configure it using option `:enable_cron_stop:`  in config file i.e foreman_maintain.yml

**Comment by johnpmitsch** [comment]
@kgaikwad We've ran into sync plans causing issues on clones (using sat-clone), an option to disable them would be a great feature to add to foreman-maintain maintenance mode. Was this ever discussed?

**Comment by kgaikwad** [comment]
@johnpmitsch,

I have tested it on my dev setup where command runs perfectly fine. The only thing that I missed is that I did some back-porting to get access my server using `localhost`.
And that's why I failed to test iptables changes.

I will test this on production server and will let you know the test results.

**Comment by kgaikwad** [comment]
@johnpmitsch,

IIRC, we haven't discussed this before to add it under `maintenance-mode` command.
Disabling/enabling sync-plans come under only pre/post-migrations phases while doing upgrade.

But this can be achievable using 'advanced procedure' command. I might be wrong but I think we should not include it under `maintenance-mode` command.

**Comment by johnpmitsch** [comment]
@kgaikwad would there be a way to add this as a flag? i.e. `foreman-maintain maintenance-mode --disable-sync-plans`. I could see this being helpful in some scenarios, and makes it more user-friendly than using advanced procedures.

**Comment by kgaikwad** [comment]
@johnpmitsch,
I will check on this. Adding this flag will not be difficult.
In my opinion adding flag may create confusion at the time of checking status of maintenance-mode. right?

Or then it may be better to include disabling/enabling sync-plans under maintenance-mode.

**Comment by johnpmitsch** [comment]
@kgaikwad Unless I'm missing something here, managing sync plans seems like a pretty important part of keeping a katello/satellite box from performing unwanted communication/actions. I think disabling them should just be part of the maintenance mode as a default, maybe with a flag to keep the sync plans if that is desired?

**Comment by iNecas** [comment]
+1 for including the sync plan disable in maintenance mode.

File: definitions/scenarios/maintenance_mode.rb [hunk 3]
```diff
@@ -0,0 +1,44 @@
...
+    metadata do
+      description 'Start maintenance-mode'
+      tags :maintenance_mode_start
+      label :maintenance_mode_start
+      manual_detection
+    end
+
+    def compose
+      add_step(Procedures::Iptables::AddChain.new)
+      add_step(Procedures::Cron::Stop.new)
```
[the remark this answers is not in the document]
kgaikwad: Yes, doing these changes for `crond` service.

Just to confirm that to achieve this, I have to change logic inside [service](https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/service.rb#L10) feature because using `katello-service`, actions can be performed on fix set of services not on all services. So it would be good to check list from katello-service and use else block otherwise which uses either `service-wait` or `systemctl`.

kgaikwad:
@iNecas, @mbacovsky,

I have added `sync-plans` under maintenance-mode  and added `services` method to `cron` feature to consider `crond` service.

Please share your suggestions on `status` sub-command of maintenance-mode which returns status string and `is-enabled` sub-command  under maintenance-mode which will return integer value i.e 1 for ON and  0 for OFF.

definitions/scenarios/maintenance_mode.rb
```diff
@@ -24,8 +28,11 @@ class MaintenanceModeStop < ForemanMaintain::Scenario
 
     def compose
       add_step(Procedures::Iptables::RemoveChain.new)
-      add_step(Procedures::Cron::Start.new)
-      add_step(Procedures::MaintenanceFile::Remove.new)
+      add_step(Procedures::SyncPlans::Enable.new)
+      if feature(:cron)
```
mbacovsky: While cron_service can only be `crond` it should be enough to
```ruby
add_step(Procedures::Service::Start.new(:only => 'crond')) if feature(:cron)
```
or do I miss something?

definitions/procedures/iptables/add_chain.rb
```diff
@@ -2,15 +2,15 @@ module Procedures::Iptables
...
+      for_feature :iptables
       description 'Add chain to iptables'
       tags :pre_migrations
       advanced_run false
       after :sync_plans_disable
     end
 
     def run
-      feature(:maintenance_mode).perform_action(:iptables, 'add_chain')
+      feature(:iptables).perform_action('add_chain')
```
mbacovsky: Why the action is called indirectly? Are there any benefits over `feature(:iptables).add_chain`?

definitions/features/iptables.rb
```diff
@@ -0,0 +1,58 @@
...
+  metadata do
+    label :iptables
+  end
+
+  def perform_action(action_name)
```
mbacovsky: This feature looks quite "maintenance_mode oriented". May be more flexible if we have more generic feature with e.g. 
`add_chain`, `remove_chain`, `chain_exist?`, `rule_exist?` and some shortcuts like `add_maintenance_mode_chain`, `remove_maintenance_mode_chain`. What do you think?

mbacovsky:
@iNecas, @kgaikwad I'm glad we split the status and is_enabled which is IMO more consistent with the rest of the f-m. 
However the code is getting complex. I wanted to comment in-line but I'd like to make clear what is the use case for this feature and how it should behave. I'm a bit concerned about how we handle the 'partly turned on' maintenance mode. Isn't it an error state that should not happen and be treated as such?

What I'd expect as a user the status is either on or off nothing in between. Therefore we should have single point of truth for telling if the maint-mode is on and that should be the iptables chain present. In the code there are various situations indicated to the user but solution is up to her. I believe we can do more. 
How about adding mode consistency check to the status command? The check would detect and suggest execution of fixes for invalid states. If chain is present and sync plans enabled lets assume it is invalid state and suggest running procedure that disables them. Similarly for other cases.

For `is-enabled` I'd suggest just to check for presence of the chain.

It could be also helpful to list the individual components and their status as part of `maintenance-mode status`, e.g.:
```
# ./bin/foreman-maintain maintenance-mode status
<Running Check of maintennace mode consistency>
Running Status of maintenance-mode
==============================================================================                                     
Status of maintenance-mode: On                                        [OK]
 - Iptables chain: present
 - sync plans: disabled
 - cron jobs: not managed
------------------------------------------------------------------------------
```
Does it make sense?

definitions/features/iptables.rb
kgaikwad: +1  for a suggested change. It becomes more readable than the current change.

definitions/procedures/maintenance_mode/enable.rb
```diff
@@ -1,13 +0,0 @@
-module Procedures::MaintenanceMode
-  class Enable < ForemanMaintain::Procedure
```
mbacovsky: We should keep these procedures as they are already used in backup/restore. And it is indirectly used in upgrades.

iNecas:
@mbacovsky makes sense for me, that might simplify the code a bit.

definitions/procedures/maintenance_mode/enable.rb
kgaikwad: @mbacovsky,

For the upgrades, we are running steps using `:pre_migrations and :post_migrations` tags.
Is that ok if I added additional tags i.e `:maintenance_mode_on, :maintenance_mode_off` and find procedures using those tags in backup like 
[find_procedures(:maintenance_mode_on)](https://github.com/theforeman/foreman_maintain/pull/180/files#diff-a0ee368c0fb1cccf764aeade5e245985R133?).

In backup, it is already doing `Procedures::Service::Stop` and `Procedures::Service::Start` so I have not handled `crond` service part there separately.

Or you want me to move procedures from `MaintenanceModeStart and MaintenanceModeStop` scenarios into Procedures::MaintenanceMode::Enable and Procedures::MaintenanceMode::Disable?

Any suggestions?

definitions/checks/maintenance_mode/check_consistency.rb
```diff
@@ -0,0 +1,65 @@
+module Checks::MaintenanceMode
+  class CheckConsistency < ForemanMaintain::Check
+    metadata do
+      label :check_maintenance_mode_consistency
+      description 'Check of maintenance-mode consistency'
+      confine do
+        feature(:downstream)
```
mbacovsky: Why these checks are downstream only?

kgaikwad: My bad. No need to limit check for :downstream here. I will remove it.

definitions/features/cron.rb
```diff
@@ -0,0 +1,25 @@
...
+      ForemanMaintain.config.manage_crond && !(
+        feature(:downstream) && feature(:downstream).less_than_version?('6.3')
+      )
+    end
+  end
+
+  def services
+    # TODO: For debian, add cron as service
+    [
+      system_service('crond', 5)
```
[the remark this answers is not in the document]
kgaikwad: Hmm.. Thank you @mbacovsky  for pointing this.

I don't think it will bring any harm for other services but the only thing I could see here is that whenever command  `foreman-maintain service restart` will run, it won't run jobs scheduled at that moment.
Again, this needs to be confirmed.

definitions/procedures/maintenance_mode/enable.rb
mbacovsky: I think find_procedures could be acceptable solution.

Pros:
- it works out of the box
- it reads well

Cons:
- it is new concept. When composing new scenario you have to know there is no single procedure for turning on maint mode and you have look around how to do that
- it is hard to tell what is going on during 'maint mode on'. You have to search code for the tags

Other options:
- use Scenarios::MaintenanceMode::Start and let compose expand the steps. I'm not sure if we can nest scenarios. It would be easy to do.
- extract useful sets of steps in some module and re-use it later

definitions/scenarios/maintenance_mode.rb
```diff
@@ -0,0 +1,57 @@
...
+  class MaintenanceModeStatus < ForemanMaintain::Scenario
+    metadata do
+      description 'Status of maintenance-mode'
+      tags :maintenance_mode_status
+      label :maintenance_mode_status
+      manual_detection
+    end
+
+    def compose
+      add_step(Checks::MaintenanceMode::CheckConsistency.new)
```
mbacovsky: `.new` is not necessary here as add_step create the instance on its own

mbacovsky:
@kgaikwad thanks for keeping this in progress despite delays between my reviews. Sorry for that.

It seems we are close to merge now. I commented with my concerns on the missing enable/disable procedures but won't block the PR on that. We can refactor that later if there is better way an it is needed. I appreciate all the effort you've put into this PR :ice_cream:. :+1: for using the system_service.

I did some testing and overall it looks good. I found a case that is not clear to me as an user:

I have basic Sat 6.4 setup. Maintenance-mode is off.
```
# foreman-maintain maintenance-mode is-enabled
Running Show only status code of maintenance-mode
Showing status code for maintenance_mode:                             [OK]

0

# echo $?
0
```
The output is a bit verbose but missing the result. The [OK] means that the scenario didn't failed.
The result is probably the `0` but without the context it is not saying much. Would you mind adding some translation such as `Maintenance mode is off`?
The exit_code works fine (if `0` means off). It would deserve short paragraph in the README perhaps.
 
```
# ./bin/foreman-maintain maintenance-mode status
Running Status of maintenance-mode
Check of maintenance-mode consistency: 
/ Running status of maintenance-mode                                            

Status of maintenance-mode: Off
- Iptables chain: absent
- sync plans: empty data
- cron jobs: running                                                  [FAIL]
You can follow remediation procedure(s) to fix maintenance-mode
1) Add maintenance_mode chain to iptables
2) Stop applicable services
Select step to continue, [n(next), q(quit)] q       
Scenario [Status of maintenance-mode] failed.                                   

  [check-maintenance-mode-consistency]

use --whitelist="check-maintenance-mode-consistency"
```
This was run immediately after the first command so the result should be similar. It detected the main mode as `Off` properly but failed and offered remediation steps. From the output I'm not sure what happened. Is my maint-mode status inconsistent or the check wants me to turn it on (it shouldn't if the status is consistent)? The point of confusion may be I have no sync-plans at all.

I'd would expect it to print:
```
Check of maintenance-mode consistency: 
/ Running status of maintenance-mode                                            

Status of maintenance-mode: Off
- Iptables chain: absent
- sync plans: empty data
- cron jobs: running                                                  [OK]
```

@kgaikwad please let me know if you need more details about my environment.

kgaikwad:
@mbacovsky,

One thing, I forgot to mention.
Previously, I was overridden exit code of foreman-maintain in status command. To avoid that, we have introduces is-enable subcommand which returns integer value i.e 1/0.

# foreman-maintain maintenance-mode is-enabled

will return 0 when maintenance-mode OFF and 1 when it's ON.

I haven't overridden exit codes of foreman-maintain. In both the cases exit code is 0 as there is no failure. Is it correct?
Or do you want me to override exit code of these commands according to return value?

mbacovsky:
Very nice @kgaikwad, I tested the updated version and like the UX! Thank you. 

For the `is-enabled` I'd prefer to reflect the status in the exit-code accordingly. This command should be here mostly for scripts and testing the exit code is common and easy in bash. 

The last piece we should figure out is how to handle the `cron`. From the limited feedback on the list I got the impression that we shouldn't touch cron if possible. I'd propose the following:

I think the current cron handling in `maintenance-mode` is a nice and useful feature so I'd keep it in but had it opt-in - disabled by default.

If we ship something like
```bash
cat > /etc/profile.d/foreman-maintain.sh << EOF
#!/bin/bash 
function quit_if_in_maintenance_mode {
  /usr/bin/foreman-maintain maintenance-mode is-enabled
  if [ $? -ne 0 ]
  then
    echo "Maintenance mode is on can not continue"
    exit 1
  fi
}

function is_maintenance_mode_enabled {
  /usr/bin/foreman-maintain maintenance-mode is-enabled
  echo $?
}
EOF
```
Then we could add the function to relevant cron jobs we ship as well as the user can with his custom ones

```bash
#!/bin/bash
quit_if_in_maintenance_mode
...
if [ $(is_maintenance_mode_enabled) -ne 0 ]
then
  # Do maint-mode work
else
  # The usual work
fi
```
Does it make sense?

kgaikwad:
> Very nice @kgaikwad, I tested the updated version and like the UX! Thank you.
> 
> For the `is-enabled` I'd prefer to reflect the status in the exit-code accordingly. This command should be here mostly for scripts and testing the exit code is common and easy in bash.
> 
> The last piece we should figure out is how to handle the `cron`. From the limited feedback on the list I got the impression that we shouldn't touch cron if possible. I'd propose the following:
> 
> I think the current cron handling in `maintenance-mode` is a nice and useful feature so I'd keep it in but had it opt-in - disabled by default.
> 
> If we ship something like
> 
> ```shell
> cat > /etc/profile.d/foreman-maintain.sh << EOF
> #!/bin/bash 
> function quit_if_in_maintenance_mode {
>   /usr/bin/foreman-maintain maintenance-mode is-enabled
>   if [ $? -ne 0 ]
>   then
>     echo "Maintenance mode is on can not continue"
>     exit 1
>   fi
> }
> 
> function is_maintenance_mode_enabled {
>   /usr/bin/foreman-maintain maintenance-mode is-enabled
>   echo $?
> }
> EOF
> ```
> 
> Then we could add the function to relevant cron jobs we ship as well as the user can with his custom ones
> 
> ```shell
> #!/bin/bash
> quit_if_in_maintenance_mode
> ...
> if [ $(is_maintenance_mode_enabled) -ne 0 ]
> then
>   # Do maint-mode work
> else
>   # The usual work
> fi
> ```
> 
> Does it make sense?

Thank you @mbacovsky! Makes sense :+1:

kgaikwad:
Updated PR with suggested changes.

definitions/features/cron.rb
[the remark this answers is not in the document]
mbacovsky: @kgaikwad I think this still needs some resolution. I believe the general opinion was not to touch the crond. We made it opt-in for maintenance mode but `crond` is also stopped during `formena-maintain service stop`. Do we want to keep this behavior? If so we should update the comments in the configs and warn the users.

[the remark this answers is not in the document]
kgaikwad: @mbacovsky, 
Thank you for review. Yes, right!
With these changes,  `crond` will get stop when `foreman-miantain service stop` when user set `manage_crond:true`. I will update PR which will fix this behaviour.

mbacovsky:
@kgaikwad, thanks again for incorporating my comments! I went through the last update and and I'm concerned about some changes it introduces. The concept of registering 'non-registrable' services doesn't feel correct. (Why should I register service that I don't want to be registered?) I'd much rather see the logic in the  Start/Stop procedure or better the underlying Service feature. E.g.
```ruby
add_step(Procedures::Service::Stop.new(:force => 'crond')
# or
add_step(Procedures::Service::Stop.new(:force => system_service('crond', 5)
```  
which would add the service regardless of the `existing_services` result. The logic then would be cleanly encapsulated in `filter_services`. Also then it is not necessary to add hidden parameters as the new `force` parameter is quite clear in its actions. We don't want to expose it in CLI but it is IMO okay to have it in `advanced procedure run`.

What do you think?

kgaikwad:
> @kgaikwad, thanks again for incorporating my comments! I went through the last update and and I'm concerned about some changes it introduces. The concept of registering 'non-registrable' services doesn't feel correct. (Why should I register service that I don't want to be registered?) I'd much rather see the logic in the Start/Stop procedure or better the underlying Service feature. E.g.
> 
> ```ruby
> add_step(Procedures::Service::Stop.new(:force => 'crond')
> # or
> add_step(Procedures::Service::Stop.new(:force => system_service('crond', 5)
> ```
> 
> which would add the service regardless of the `existing_services` result. The logic then would be cleanly encapsulated in `filter_services`. Also then it is not necessary to add hidden parameters as the new `force` parameter is quite clear in its actions. We don't want to expose it in CLI but it is IMO okay to have it in `advanced procedure run`.
> 
> What do you think?

I would like to summarize further upcoming changes to this PR and one query regarding outside/ unregistered services.

- As Foreman Maintain already have an option --force, instead of force option introducing an `--include` option here. 
- As per @mbacovsky  mentioned, this option won't be available in case of `service command` but it will be visible under `advanced procedure run`
- In case of `cron` feature, I am going to pass cron service in `--only` so that it will going to consider ONLY `crond service` at the time of start/stop
- Currently, we have  defined priority to every service. so if any user pass service names into either in `--only` or `--include` and those service names are not belongs to any feature's service list then I have to set priority for each service as its mandatory attribute which will decide when to run.
Currently, I am adding priority for those services in such a way that those will run prior to existing features services.

@mbacovsky, @iNecas 
Please let me know your thoughts on last point regarding priority.

mbacovsky:
@kgaikwad could you please also rebase with your updates?

kgaikwad:
@mbacovsky,

Done with the rebase.
- Modified service feature to consider unregistered service i.e either by name or system_service objects inside `--include` option.
- Currently, set default value of priority to `50` to method `system_service` inside concern file
- I think `maintenance_file` config no where unused so removed those changes from config file
- Modified few and added new tests to service_test

The only point is currently for satellite 6.2, we are still using `katello-service` command so I haven't modified changes there. Just appended note inside current message that services that are not listed by `katello-service` will get ignored.
If you think we should handle this scenario as well then I would prefer a separate PR for that change.  
Please let me know your thoughts on this.

lib/foreman_maintain/param.rb
```diff
@@ -3,7 +3,8 @@ class Param
     attr_reader :name, :description, :options
 
     def initialize(name, description, options, &block)
-      options.validate_options!(:description, :required, :flag, :array, :allowed_values, :default)
+      options.validate_options!(:description, :required, :flag, :array,
+                                :allowed_values, :default, :hidden)
```
mbacovsky: Shouldn't be here when hidden was removed

lib/foreman_maintain/utils/service.rb
```diff
@@ -11,4 +11,8 @@ def self.system_service(name, priority, options = {})
       Service::Systemd.new(name, priority, options)
     end
   end
+
+  def self.valid_sys_service?(service)
+    service.instance_of?(Service::RemoteDB) || service.instance_of?(Service::Systemd)
```
mbacovsky: Is it safe to list the possibilities? If somebody makes new subclass we would have to include in here.
`service.class < Service::Abstract` should work too.

definitions/features/service.rb
```diff
@@ -96,17 +100,39 @@ def allowed_action?(action)
...
+    return service_list unless services_filter.any?
+
+    services_filter = services_filter.reject do |obj|
+      service_list.any? { |service| service.matches?(obj) }
+    end
+
+    unregistered_service_list = services_filter.map do |obj|
+      if obj.is_a? String
+        system_service(obj)
+      elsif valid_sys_service?(obj)
```
mbacovsky: We don't do type check in other places so it is probably unnecessary. In the reject above we expect Service type or string anyway.

kgaikwad:
@mbacovsky,

> * the change in the --only behavior will remove the priorities and add services at the end of the list with priority 50 (didn't tested yet)

Behavior of `--only` option won't remove priorities of existing services and will add non-feature services with priority `50` at the end of list.

> * --only xyz will attempt to manipulate with the service while previously it would check if the service exists

Yes, previously `--only` would check if the service exists but modified its behavior for `crond` service.
As we are going to provide an `--include` option which will accept `string/system_service` then I feel it would be good to have consistency between these two options.
For `--exclude`, we can't use the same behavior as `exclude` itself indicates that some sort of service list exists already. 

> * the --include does not reflect the priority and can duplicate existing services

I have tested this but I didn't found that --include duplicates existing services. Any specific scenario that I missed out?

> That would mean you need to --include unregistered service first and then use --only to filter it out but that is IMO okay. 

Here, I am assuming that `unregistered service` means service_name/system_service which is not available inside any feature's services from foreman-maintain and registered service means that available under one of feature's services method. Please correct me if I am wrong.  

> I can see the benefits of using just --only as you are suggesting but is it good behavior if you use registered service that is not present on the system and it tries to start that service?

Sorry but I didn't get this point clearly but the only point I thought is to have consistency between these two options. The scenario that you have mentioned can happen with `--include` as well, right?

definitions/features/service.rb
kgaikwad: In the reject above, it matches string/Service type against existing service_list of available features.

For crond service, after maintenance -mode status command, foreman-maintain suggests procedure to fix maintenance-mode for crond i.e  either to stop/start.
I have used argument as `{ :only => system_service('crond') }` in this case that wouldn't be part of existing service_list of available features. That's why added type checking here.

mbacovsky:
@kgaikwad I have to apologize, I did the last review in a hurry and misread the condition and while I didn't properly tested some of my assumptions may be incorrect. I'll do proper testing after I'm back. 
The case I have tested and which needs to be fixed is `Procedures::Service::Start.new(:include => system_service('crond', 1))` which does not execute the action on services in correct order.

kgaikwad:
@mbacovsky,
I have checked on `Procedures::Service::Start.new(:include => system_service('crond', 1))`.

Found that with array value to `:include or :only` options then it works i.e `Procedures::Service::Start.new(:include => [system_service('crond', 1)])` 
I think, its because we have overridden `to_s` method  in `Service::Abstract` and additionally I need to apply `.sort` method on service_list , is it ok?

lib/foreman_maintain/param.rb
mbacovsky: `:hidden` i still present

lib/foreman_maintain/utils/service.rb
mbacovsky: Still seems unresolved

lib/foreman_maintain/param.rb
kgaikwad: Done!

kgaikwad:
@mbacovsky, 
Fixed the code at two places as per suggestions.

definitions/scenarios/backup.rb
```diff
@@ -219,7 +219,7 @@ class BackupRescueCleanup < ForemanMaintain::Scenario
 
     def compose
       add_step_with_context(Procedures::Service::Start) if strategy != :online
-      add_step_with_context(Procedures::MaintenanceMode::Disable) if strategy != :online
+      add_step_with_context(find_procedures(:maintenance_mode_off)) if strategy != :online
```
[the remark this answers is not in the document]
mbacovsky: And also tagged :pre-upgrade and :post_upgrade to be part of upgrade scenario.

mbacovsky:
@kgaikwad Thanks for the update! I almost hit the merge button but unfortunately I found out backup is failing and cron is not managed properly in neither backup nor upgrades. Please see my comments inline.

kgaikwad:
@mbacovsky,
Good Catch! :ok_hand: :+1: 

Updated PR with the suggested changes. 
Extracted out `crond` actions into separate procedures.

kgaikwad:
@jameerpathan111, ping.
Please have test cycle around both scenarios i.e backup & upgrade with latest changes.

mbacovsky:
@kgaikwad thanks for fixing those issues. The feature looks good now. It passed my tests. Well done!
I'll wait for @jameerpathan's green light before I squash and merge.

jameerpathan111:
@kgaikwad @mbacovsky I will start testing this PR.

jameerpathan111:
@mbacovsky I have run Testfm automation with this PR and there were no failures. (Backup and restore run successfully)

@kgaikwad I have tried following scenarios:
1. 
- Create and enable a sync plan 
- Run `./bin/foreman-maintain maintenance-mode status`
- Here status command give `sync plans: empty data`
- `sync plans: empty data` feels little confusing, can we do something about it. Its because of `/var/lib/foreman-maintain/data.yml` contains no data.
2. 
- Run `./bin/foreman-maintain maintenance-mode start`
- Run `./bin/foreman-maintain maintenance-mode stop`
- Run `hammer sync-plan update --id 1 --enabled 0 --organization-id 1`
- Run `./bin/foreman-maintain maintenance-mode start`
- Run `./bin/foreman-maintain maintenance-mode status`
- Here `check-maintenance-mode-consistency` is keep on failing.
- Is this expected behaviour? 
I have attached pastebin link [1], it contains all the steps which I have followed.


[1] https://paste.fedoraproject.org/paste/w~cEf1DHMj4NNsZK49A7yQ
There are still some scenarios remaining which I have to test.

kgaikwad:
@jameerpathan111,

In foreman-maintain, we are keeping sync-plans ids in data.yml in disabling sync_plans procedure and in other sync_plans command(s) it is just fetching status from data.yml.
Modifying state of any sync_plan(s) using hammer outside of the foreman-maintain would any way cause such inconsistent behaviors. Currently, there is no way for user to clear existing data.

While testing scenarios, I found that on sat 6.2, I am now getting below errors on disabling sync_plans which modifies data in DB but failed to update data in foreman-maintain due to error raised.

```
./bin/foreman-maintain maintenance-mode start
Running Start maintenance-mode
===============================================================================
disable active sync plans: 
| disabling sync plans                                                [FAIL]    
Could not update the sync plan:
  Required lock is already taken by other running tasks.
  Please inspect their state, fix their errors and resume them.
  
  Required lock: read
  Conflicts with tasks:
  - https://satellite.example.com/foreman_tasks/tasks/63780182-b933-4395-baa2-639a6fcbd081
```
Even I tried to run using hammer but it is also giving same errors. which I think needs to fix but I haven't observed this behavior before. 

@iNecas & @mbacovsky,
Any suggestions on this?

jameerpathan111:
> Even I tried to run using hammer but it is also giving same errors. which I think needs to fix but I haven't observed this behavior before.

@kgaikwad 
The error you mentioned seems to be occurring because of a same task is executed twice before first task is completed.

jameerpathan111:
@kgaikwad 
I have tested satellite upgrade (satellite 6.3 and 6.4) and it was successful.

- there is small change/issue which I found that `'--only'` option regarding service does not fail/warn when wrong service name is provided.

with current changes:
```

# ./bin/foreman-maintain advanced procedure run service-list --only http
List applicable services: 


All services listed                                                   [OK]
```
Previously:
```

# foreman-maintain advanced procedure run service-list --only http
List applicable services:                                             [FAIL]
```

kgaikwad:
@mbacovsky, 
Updated PR with discussed changes related to `sync_plans` and  valid service checking for service commands.

@jameerpathan111, please have a testing round for updated changes.

jameerpathan111:
> @mbacovsky,
> Updated PR with discussed changes related to `sync_plans` and valid service checking for service commands.
> 
> @jameerpathan111, please have a testing round for updated changes.
> 
> **EDIT** - Modified tests and added one as those are failing before.

@kgaikwad 
- Issue related to service command seems to be resolved.
- I am still able to reproduce issue regarding `sync-plan`

kgaikwad:
@jameerpathan111,
Thank you for quick testing!
Updated PR. Sorry, with previous changes, it was broken.
I have tried to test few sync-plans scenarios. 
Hopefully, updated changes will cover all of its scenarios this time :(

jameerpathan111:
> @jameerpathan111,
> Thank you for quick testing!
> Updated PR. Sorry, with previous changes, it was broken.
> I have tried to test few sync-plans scenarios.
> Hopefully, updated changes will cover all of its scenarios this time :(

Yes, tested new changes, PR works as expected. 
Thanks @kgaikwad :+1:

mbacovsky:
:+1: That was a long ride! Thanks everybody involved! Big thanks to @kgaikwad for the hard work! :fireworks: :cake: :champagne:

iNecas:
@mbacovsky mind doing a new release? I think this deserves 0.4.0

mbacovsky:
@iNecas can do one tomorrow.

PR #181, opened 2018-05-18
Fixes #23637 - backup and restore ssh keys for proxy

mbacovsky:
Added missing directory with ssh keys for proxy and fixing
restore failure during file extractions. On some instances
the symlink can be replaced with the actual content which prevents
tar to extract.


iNecas:
Thanks. I think it makes sense.

PR #182, opened 2018-05-18
Fixes #27790 - Provide interface to foreman-rake foreman_tasks:cleanup

adamruzicka:


lib/foreman_maintain/cli/advanced/task_cleanup_command.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      DEFAULT_RAKE_COMMAND = 'foreman-rake'.freeze
+
+      option %w(-B --batch-size), 'BATCH_SIZE', 'Process tasks in batches of BATCH_SIZE, 1000 by default' do |s|
+        Integer(s)
+      end
+      option %w(-S --states), 'STATES', 'Operate on tasks in STATES, comma separated list of states, set to all to operate on tasks in any state' do |s|
+        s.split(',')
+      end
+      option %w(-a --after), 'AGE', "Operate on tasks older than AGE. Expected format is a number followed by the time unit (s,h,m,y), such as '10d' for 10 days"
```
adamruzicka: Validation would be nice here

lib/foreman_maintain/cli/advanced/task_cleanup_command.rb
```diff
@@ -0,0 +1,65 @@
...
+          format_kv('STATES', states == %w(all) ? [] : states),
+          format_kv('AFTER', after),
+          format_kv('TASK_SEARCH', search),
+          format_kv('TASK_BACKUP', backup?),
+          format_kv('NOOP', noop?),
+          format_kv('VERBOSE', verbose?)
+        ].flatten.compact
+      end
+
+      def format_kv(key, value)
```
iNecas: I would suggest expanding this to `format_key_value` for readability purposes

iNecas:
I'm ok with this approach. @mbacovsky/@kgaikwad thoughts/review comments?

lib/foreman_maintain/cli/advanced/task_cleanup_command.rb
```diff
@@ -0,0 +1,65 @@
+module ForemanMaintain
+  module Cli
+    class TaskCleanupCommand < Base
+      include Concerns::SystemHelpers
```
kgaikwad: `SystemHelpers` concern also have `execute` method that's why creating a conflict between these two methods.

Giving below error on `foreman-maintain advanced task-cleanup`:

```
foreman_maintain/lib/foreman_maintain/cli/advanced/task_cleanup_command.rb:22:in `execute': wrong number of arguments (1 for 0) (ArgumentError)
	from /root/foreman_maintain/lib/foreman_maintain/concerns/system_helpers.rb:89:in `find_package'
	from /root/foreman_maintain/lib/foreman_maintain/concerns/system_helpers.rb:100:in `server?'
	from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/task_cleanup_command.rb:27:in `execute'
```

adamruzicka:
Pending rubocop

ntkathole:
- When katello-services are down, handle error.
```
# ./bin/foreman-maintain advanced task-cleanup -v
Failed executing foreman-rake foreman_tasks:cleanup VERBOSE\=1, exit status 1:
 /usr/share/foreman/lib/foreman.rb:8: warning: already initialized constant Foreman::UUID_REGEXP
/usr/share/foreman/lib/foreman.rb:8: warning: previous definition of UUID_REGEXP was here
/usr/share/foreman/lib/core_extensions.rb:182: warning: already initialized constant ActiveSupport::MessageEncryptor::DEFAULT_CIPHER
/opt/theforeman/tfm-ror51/root/usr/share/gems/gems/activesupport-5.1.4/lib/active_support/message_encryptor.rb:22: warning: previous definition of DEFAULT_CIPHER was here
rake aborted!
PG::ConnectionBad: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```


- Currently running `# ./bin/foreman-maintain advanced task-cleanup` won't give verbose output.
good to see output as follow:
```
Running ForemanMaintain::TaskCleanupCommand 
Performing task cleanup:                                           [OK]
```

kgaikwad:
@adamruzicka,
Sorry fot delay on this.
Created redmine ticket [27790](https://projects.theforeman.org/issues/27790) for this change.
Could you please rebase this branch and add issue ticket number in your commit?

lib/foreman_maintain/cli/advanced_command.rb
```diff
@@ -8,6 +8,12 @@ class AdvancedCommand < Base
       subcommand 'prebuild-bash-completion',
                  'Prepare map of options and subcommands for Bash completion',
                  PrebuildBashCompletionCommand
+
+      if defined?(Procedures::ForemanTasks)
+        procedure = Procedures::ForemanTasks::Cleanup
+        klass = Class.new(Procedure::AbstractProcedureCommand) { params_to_options(procedure.params) }
+        subcommand(dashize(procedure.label), procedure.description, klass)
```
kgaikwad: Any specific reason behind having this as a subcommand  of `advanced`? 

The reason I am asking this because Foreman Maintain already has `foreman-maintain advanced procedure` command which allow users to run any available procedures.
Example - foreman-maintain advanced procedure run foreman-tasks-cleanup --help

definitions/procedures/foreman_tasks/cleanup.rb
```diff
@@ -0,0 +1,44 @@
+puts "LOADING"
+
+module Procedures::ForemanTasks
+  class Cleanup < ForemanMaintain::Procedure
+    PERMITTED_STATES = %w[all pending scheduled planning planned running paused stopped].freeze
+    metadata do
```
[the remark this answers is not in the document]
adamruzicka: Well, that depends. Originally I wanted this to have two modes of operation one which would actually perform the cleanup (and therefore has to be run on foreman with foreman-tasks). The other would print the command the user could then take and run on a foreman machine.

upadhyeammit:
Hello @adamruzicka

We are planning to add this feature upcoming versions, can you rebase this?

Thank You!

adamruzicka:
Rebased

definitions/procedures/foreman_tasks/cleanup.rb
```diff
@@ -0,0 +1,53 @@
...
+      @search = 'label != a_label' if !@states.empty? && @search.nil?
+      args = [@rake_command, @batch_size, @states, @after, @search, @backup, @noop, @verbose]
+
+      cleanup_command = f.generate_task_cleanup_command(*args)
+      if @generate
+        puts cleanup_command
+      else
+        message = 'Performing task cleanup'
+        with_spinner(message) do |spinner|
+          execute!(cleanup_command) { |line| spinner.update line }
```
upadhyeammit, line 47: I can see we are passing the block,
~~~
Perform task cleanup: 
\ Performing task cleanup                                             [OK]       "2021-04-24 11:18:56") AND (state ^ (stopped))
~~~

This is not the problem of your changes but the `@max_length` of CLIReporter is static. I have faced this issue few times. I have below options as proposal to handle this, let me know your thoughts,

* Use Highline.output_cols as `@max_length`. This will be dynamic value and will be set as per the terminal size, however this changes all the output. Which means we stretch the decorators `=`, `-'  and the status output too,

    With `Highline.output_cols`
    ~~~
    =============================================================================================================================================================================
    Make sure Foreman DB is up: 
    / Checking connection to the Foreman DB                                                                                                                            [OK]      
    ~~~

    Existing output,
    ~~~
    Make sure Foreman DB is up: 
    ~~~

* The other way is we flush the line conditionally, I think this is better because first one will require to change the lots of messages which were written to fit in 80 charsize?
    ~~~
    class CLIReporter < Reporter
    ..
    ..
          def clear_line
            flush_size = @last_line.length > @max_length ? @last_line.length : @max_length
            print "\r" + ' ' * flush_size + "\r"
          end
    ~~~

definitions/procedures/foreman_tasks/cleanup.rb
```diff
@@ -0,0 +1,53 @@
...
+             "\nAvailable states: " + PERMITTED_STATES.join(', '),
+            :array => true, :allowed_values => PERMITTED_STATES
+      param :batch_size,
+            'Process tasks in batches of BATCH_SIZE, 1000 by default' do |s|
+        Integer(s) unless s.nil?
+      end
+      param :after,
+            'Operate on tasks older than AFTER. Expected format is a number ' \
+            "followed by the time unit (s,h,d,m,y), such as '10d' for 10 days"
+      param :backup, 'Backup deleted tasks', :flag => true
```
upadhyeammit, line 23: I tried to pass the backup option but it did not work. I also tried to change the value of `TASK_BACKUP` to `true`instead of `1` but no effect. Request to check this one.

upadhyeammit (changes requested):
Hello @adamruzicka I gave this a try. I yet to test few remaining options of the procedure, added few observations. Request to check, thank you!

definitions/procedures/foreman_tasks/cleanup.rb
adamruzicka: Should be fixed now

definitions/procedures/foreman_tasks/cleanup.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def run
+      f = feature(:foreman_tasks)
+
+      # If states are provided, then a search must be provided as well
+      # We assume if the user didn't provide a search query, they want to match
+      # all tasks in given stat so we provide a bogus search value which matches
+      # all tasks
+      @search = 'label != a_label' if !@states.empty? && @search.nil?
```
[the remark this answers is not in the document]
adamruzicka: Should be fixed now

lib/foreman_maintain/utils/command_runner.rb
```diff
@@ -94,6 +94,15 @@ def full_command
         "#{@command} 2>&1"
       end
 
+      def with_line_streaming(io)
+        return io.read unless block_given?
+
+        io.each_line.lazy.map do |line|
```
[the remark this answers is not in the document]
adamruzicka: To be honest I don't remember at all why I did this and I don't really see why it would make sense for it to be lazy here. Will drop

definitions/procedures/foreman_tasks/cleanup.rb
adamruzicka: If I understand the issue correctly:
- When the task export is running, we are displaying its output line by line on the line with the spinner
- The lines are often too long leading to broken display

Is this correct?

upadhyeammit: Yes!

adamruzicka: Sorry for the delay, done

upadhyeammit:
Hello @adamruzicka this looks good to merge, I like to know if we are confirm to get this into fm instead another pr https://github.com/theforeman/foreman-tasks/pull/601? if yes, I ask you to rebase last time! thank you!

adamruzicka:
Closing in favor of https://github.com/theforeman/foreman-tasks/pull/601

PR #184, opened 2018-06-05
Fixes #1582215 - added preparation_steps to install facter package

kgaikwad:


ntkathole:
@kgaikwad @iNecas I think this solution is not going to work here. 

When satellite is 6.3 with puppet 4 fresh installed, it shows
```
Procedures::Packages::Install: 
Package 1:facter-2.4.6-3.el7sat.x86_64 is obsoleted by puppet-agent-1.10.9-1.el7sat.x86_64 which is already installed
                                                                      [OK]
```
still check for validating repository failed showing  
```

| Validating availability of repositories for 6.3                     [FAIL]    
Failed executing facter operatingsystemmajrelease, exit status 127:
 sh: facter: command not found
```

kgaikwad:
@ntkathole,

Thank you. Then, probably I have to remove dependency on facter command.

ntkathole:
@gnurag Thanks... `source /etc/profile.d/puppet-agent.sh `works for me. @kgaikwad  I am not sure Should we handle this or not else PR looks good to me!

iNecas:
What about using the absolute path to the facter instead fo doing `source`?  I'm a bit hesitant to source something we don't have that much control about.

kgaikwad:
@iNecas, @ntkathole,

Created `facter` utils module to fetch details like package name, path.

definitions/checks/repositories/validate.rb
```diff
@@ -2,6 +2,9 @@ module Checks::Repositories
   class Validate < ForemanMaintain::Check
     metadata do
       description 'Validate availability of repositories'
+      preparation_steps do
+        Procedures::Packages::Install.new(:packages => [ForemanMaintain::Utils::Facter.package])
```
iNecas: Do we actually need this now? There is also a logical issue, as `ForemanMaintain::Utils::Facter.package` actually checks on the presence of the `/usr/bin/facter` which, in case the facter is not installed, would never be true.

We either need to decide which package to install by other ways, assume that some puppet will always be there, or implement the methods we rely on Facter right now with the use of other tools: especially if we need only the os_major_release, there might be simpler ways to achieve that.

kgaikwad: Good catch!
Yeah, I realized that I made mistake in finding which package to install. But it was too late, I saw your review comment :-)
I will fix that.

In case of os_major_release, facter command itself handles many conditions like different distros. 
For other way, we will need to handle those conditions in our logic. Facter command makes  that easy.

kgaikwad: Thank you @iNecas.
I have checked puppet version to decide package to install. Updated the PR.

iNecas:
Works for me code-wise. @ntkathole mind testing?

PR #185, opened 2018-06-05
Fixes #23685 - failure when curr_dir_path equals to settingsd_path

kgaikwad:


definitions/procedures/remote_execution/remove_existing_settingsd.rb
```diff
@@ -11,7 +11,14 @@ class RemoveExistingSettingsd < ForemanMaintain::Procedure
 
     def run
       with_spinner("Removing existing #{@dirpath} directory") do |_spinner|
-        execute!("rm -rf #{@dirpath}")
+        curr_dir = execute('pwd').strip
```
iNecas: I guess you can us `Dir.pwd` instead https://ruby-doc.org/core-2.2.0/Dir.html#method-c-pwd

iNecas:

PR #186, opened 2018-06-11
Fixes #23856 - on sat 6.3 & puppet4, enable puppet4 repo

kgaikwad:


definitions/features/downstream.rb
```diff
@@ -55,24 +55,33 @@ def rhsm_refresh
...
   end
 
+  def main_rh_repos(rh_version_major)
+    ["rhel-#{rh_version_major}-server-rpms",
+     "rhel-server-rhscl-#{rh_version_major}-rpms",
+     "rhel-#{rh_version_major}-server-satellite-maintenance-6-rpms"]
+  end
+
+  def puppet_repo?(sat_version)
+    sat_version.to_s >= '6.3' && (execute('rpm -q puppet') =~ /puppet-4/)
```
iNecas: Please use `package_version` helper instead of direct `execute` https://github.com/theforeman/foreman_maintain/blob/1ae71c3ad827503eb370310a6b27232f80289351/lib/foreman_maintain/concerns/system_helpers.rb#L88

definitions/features/downstream.rb
```diff
@@ -55,24 +55,33 @@ def rhsm_refresh
...
-    rh_repos = ["rhel-#{rh_version_major}-server-rpms",
-                "rhel-server-rhscl-#{rh_version_major}-rpms",
-                "rhel-#{rh_version_major}-server-satellite-maintenance-6-rpms",
-                sat_tools_repo_id,
-                sat_repo_id]
-    rh_repos << 'rhel-7-server-ansible-2-rpms' if sat_version.to_s == '6.4'
+      rh_repos << 'rhel-7-server-ansible-2-rpms'
+    end
+    rh_repos.concat([sat_tools_repo_id, sat_repo_id])
+    rh_repos << 'rhel-7-server-satellite-tools-6.3-puppet4-rpms' if puppet_repo?(sat_version)
```
iNecas: Shouldn't here be `sat_version` instead of hardcoded `6.3`?

kgaikwad: Ohh.. my mistake. I will update PR.

ntkathole: @iNecas @kgaikwad i think no problem if it is 6.3 hardcoded anyhow puppet4 repo is with 6.3 version only.

kgaikwad:
@iNecas, @ntkathole,

Sorry for unnecessary confusion. I have updated the PR.
Please have a look.

ntkathole:
@kgaikwad for testing, i will need PR #184 merged. Facing facter command not found error.

definitions/features/downstream.rb
kgaikwad: @iNecas, @ntkathole,

We have considered `puppet4` repository here. 
In similar way, do we need to check about how `puppet5` will be available through main repository or in different repository?

ntkathole: @kgaikwad puppet5 will be available from satellite 6.4 repo itself (no separate enabling of puppet5 repo) so no need to check.

definitions/features/downstream.rb
```diff
@@ -55,24 +55,32 @@ def rhsm_refresh
...
     end
+    rh_repos.concat([sat_tools_repo_id, sat_repo_id])
 
-    rh_repos = ["rhel-#{rh_version_major}-server-rpms",
-                "rhel-server-rhscl-#{rh_version_major}-rpms",
-                "rhel-#{rh_version_major}-server-satellite-maintenance-6-rpms",
-                sat_tools_repo_id,
-                sat_repo_id]
-    rh_repos << 'rhel-7-server-ansible-2-rpms' if sat_version.to_s == '6.4'
+    if current_minor_version == '6.3' && (find_package('puppet') =~ /puppet-4/)
```
mbacovsky: Moving the logic to the puppet feature may make it more readable and easier to change
```ruby
if current_minor_version == '6.3' && feature(:puppet) && feature(:puppet).version.major == 4
```
also on none of my instances (sat 6.4, katello 3.6) the package `puppet` is not installed. 
`puppet --version` works everywhere. Might be more flexible to use something like `version(execute!('puppet --version'))`

kgaikwad:
@mbacovsky,  @ntkathole,

Updated the PR. Please have a look.

iNecas:
https://github.com/theforeman/foreman_maintain/pull/184 got merged: @kgaikwad mind rebasing so that @ntkathole can test again?

kgaikwad:
@ntkathole, 
Done with rebasing 👍

definitions/features/downstream.rb
```diff
@@ -55,22 +55,39 @@ def rhsm_refresh
...
     rh_version_major = ForemanMaintain::Utils::Facter.os_major_release
-    sat_version_full = "#{sat_version.major}.#{sat_version.minor}"
 
+    rh_repos = main_rh_repos(rh_version_major)
+
+    rh_repos.concat(sat_and_tools_repos(rh_version_major, sat_version))
+
+    rh_repos << 'rhel-7-server-ansible-2-rpms' if sat_version.to_s == '6.4'
+
+    if current_minor_version == '6.3' && (
```
ntkathole: @kgaikwad @iNecas The PR looks good to me...except one question while upgrading from sat 6.3 puppet 4 to sat 6.4 we are enabling puppet 4 repo which is not needed also i am not sure will it affect anything or not...shall we add one more condition here that to version != 6.4 ?

kgaikwad: I have updated the PR with this change. Please have a look!

iNecas:
Thanks @kgaikwad

PR #187, opened 2018-06-11
Fixes #23819 - upgrade guide for puppet if sat 6.3 & puppet3

kgaikwad:


definitions/checks/puppet/provide_upgrade_guide.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      manual_detection
+    end
+
+    def run
+      with_spinner('Verifying puppet version before upgrade') do |spinner|
+        puppet_package = find_package('puppet')
+        spinner.update "current puppet version: #{puppet_package}"
+        if puppet_package =~ /puppet-3/
+          fail!(<<-MESSAGE.strip_heredoc)
```
iNecas: Shouldn't we extend https://github.com/theforeman/foreman_maintain/blob/master/definitions/procedures/knowledge_base_article.rb to support docs as well as kcs and use it here?

kgaikwad: @iNecas,
Done with the changes and updated the PR.

definitions/checks/puppet/provide_upgrade_guide.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def run
+      with_spinner('Verifying puppet version before upgrade') do |spinner|
+        puppet_package = find_package('puppet')
+        spinner.update "current puppet version: #{puppet_package}"
+        result = (puppet_package =~ /puppet-3/)
+        curr_sat_version = feature(:downstream).current_minor_version
+        assert(
+          result.nil?,
```
iNecas: It would read a bit better to not use the result variable, but use the condition directly here

```ruby
assert(
  (puppet_package !~ /puppet-3/),
  …
```

iNecas:
Small suggestion in the code: deferring to @ntkathole regarding full testing.

definitions/checks/puppet/provide_upgrade_guide.rb
kgaikwad: Done!

iNecas:

PR #188, opened 2018-06-12
Fixes #23714 - Bash completion for foreman-maintain

mbacovsky:
See README for details on how to install and use

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,34 +387,38 @@ foreman-maintain config
 
 ## Bash completion
 
+The completion offers suggestion of possible command-line subcommands and their options
+as usual. It can also suggest values for options and params where file
+or directory path is expected.
+
 Bash completion is automatically installed by RPM. To use it for development setup 
 `cp ./config/foreman-maintain.completion /etc/bash_completion.d/foreman-maintain` 
 and load it to the current shell `source /etc/bash_completion.d/foreman-maintain`.
+Make sure the `$PWD/bin` is in `PATH` or there is full path to `foreman-maintain-complete`
+executable specified in `/etc/bash_completion.d/foreman-maintain`.
 
 Bash completion for foreman-maintain needs pre-built cache that holds description of 
 all subcommands and its parameters. The cache is located by default in 
 `~/.cache/foreman_maintain_completion.yml`. The location can be changed in foreman-maintain's
-config file. The cache is pre-built automatiacally during RPM install, can be built manually
-with `foreman-maintain advanced prebuild-bash-completion` or is built automatically when
-completion is used and the cache is missing (this may cause slight delay).
-
-The completion offers suggestion of possible command-line subcommands and their options
-as usual. It can also suggest values for options and params where it should be file
-or directory path.
+config file. The cache can be built manually with 
+`foreman-maintain advanced prebuild-bash-completion` or is built automatically when
+completion is used and the cache is missing (this may cause slight delay). The cache expires 
+after installer scenario answer file changed (it indicates that the features on the instance
+may have changed which has impact on foreman-maintain CLI options and subcommands).
 
 ####  Available value types
 
 Completion of values is dependent on CLI option and prameter settings, e.g.:
 
 ```ruby
-  parameter 'BACKUP_DIR', 'Path to backup dir', :completion => { :directory => {} }
+  parameter 'BACKUP_DIR', 'Path to backup dir', :completion => { :type => :directory }
 ```
 
 Possible options for the `:completion` attribute are:
- - `{}` option has no value, default for flags
- - `{ :value => {} }` option has value of unknown type, no suggestions for the value, default
- - `{ :directory => {} }` value is directory, suggestions follow directory structure
- - `{ :file => { :filter => '\.txt$' } }` value is file, suggestions follow directory structure, 
+ - `{ :type => :flag }` option has no value, default for flags
+ - `{ :type => :value }` option has value of unknown type, no suggestions for the value, default
+ - `{ :type => :directory }` value is directory, suggestions follow directory structure
+ - `{ :type => :file, :filter => '\.txt$' }` value is file, suggestions follow directory structure, 
  	optional `:filter` is regexp to filter the results.
  
 
--- a/README.md
+++ b/README.md
@@ -385,6 +385,39 @@ foreman-maintain console
 foreman-maintain config
 ```
 
+## Bash completion
+
+Bash completion is automatically installed by RPM. To use it for development setup 
+`cp ./config/foreman-maintain.completion /etc/bash_completion.d/foreman-maintain` 
+and load it to the current shell `source /etc/bash_completion.d/foreman-maintain`.
+
+Bash completion for foreman-maintain needs pre-built cache that holds description of 
+all subcommands and its parameters. The cache is located by default in 
+`~/.cache/foreman_maintain_completion.yml`. The location can be changed in foreman-maintain's
+config file. The cache is pre-built automatiacally during RPM install, can be built manually
+with `foreman-maintain advanced prebuild-bash-completion` or is built automatically when
+completion is used and the cache is missing (this may cause slight delay).
+
+The completion offers suggestion of possible command-line subcommands and their options
+as usual. It can also suggest values for options and params where it should be file
+or directory path.
+
+####  Available value types
+
+Completion of values is dependent on CLI option and prameter settings, e.g.:
+
+```ruby
+  parameter 'BACKUP_DIR', 'Path to backup dir', :completion => { :directory => {} }
+```
+
+Possible options for the `:completion` attribute are:
+ - `{}` option has no value, default for flags
+ - `{ :value => {} }` option has value of unknown type, no suggestions for the value, default
+ - `{ :directory => {} }` value is directory, suggestions follow directory structure
+ - `{ :file => { :filter => '\.txt$' } }` value is file, suggestions follow directory structure, 
+ 	optional `:filter` is regexp to filter the results.
+ 
+
 ## How to contribute?
 
 Generally, follow the [Foreman guidelines](https://theforeman.org/contribute.html). For code-related contributions, fork this project and send a pull request with all changes. Some things to keep in mind:
--- a/bin/foreman-maintain-complete
+++ b/bin/foreman-maintain-complete
@@ -1,0 +1,39 @@
+#!/usr/bin/env ruby
+require 'English'
+project_root = File.expand_path('../..', __FILE__)
+$LOAD_PATH.unshift(File.join(project_root, 'lib'))
+CONFIG_FILE = '/etc/foreman-maintain/foreman_maintain.yaml'.freeze
+system('which foreman_maintain > /dev/null 2>&1')
+found = $CHILD_STATUS.success?
+FOREMAN_MAINTAIN = found ? 'foreman-maintain' : File.join(project_root, 'bin/foreman-maintain')
+
+require 'yaml'
+require 'foreman_maintain/utils/bash'
+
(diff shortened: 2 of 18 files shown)


README.md
```diff
@@ -385,6 +385,39 @@ foreman-maintain console
...
+
+Completion of values is dependent on CLI option and prameter settings, e.g.:
+
+```ruby
+  parameter 'BACKUP_DIR', 'Path to backup dir', :completion => { :directory => {} }
+```
+
+Possible options for the `:completion` attribute are:
+ - `{}` option has no value, default for flags
+ - `{ :value => {} }` option has value of unknown type, no suggestions for the value, default
```
iNecas: The empty hash in `{ :value => {} }` looks a bit odd. What about somehing like: `{ :type => :value }`, for file it could be `{ :type => :file, :filter => '\.txt$'}`

iNecas: Or, for simpler cases, maybe `{ :completion => :directory }`

lib/foreman_maintain/utils/bash.rb
```diff
@@ -0,0 +1,154 @@
...
+          return [] unless path.empty? # lost during traversing
+
+          if @complete_line
+            next_word(dict)
+          else
+            finish_word(dict, full_path.last)
+          end
+        end
+
+        def self.load_description(path)
```
iNecas: Can be simplified to:

```ruby
def self.load_description(path)
  YAML.load(File.open(path))
rescue Errno::ENOENT
  {}
end
```

lib/foreman_maintain/utils/bash.rb
```diff
@@ -0,0 +1,154 @@
...
+          end
+        end
+
+        def next_word(dict)
+          next_option_value(dict) || (next_option_or_subcommand(dict) + next_param(dict))
+        end
+
+        def next_param(dict)
+          if dict.key?(:params)
+            if dict[:params].first.key?(:directory)
```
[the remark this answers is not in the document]
iNecas: The type would be either `:param` or `:option`

README.md
```diff
@@ -385,6 +385,39 @@ foreman-maintain console
 foreman-maintain config
 ```
 
+## Bash completion
+
+Bash completion is automatically installed by RPM. To use it for development setup 
+`cp ./config/foreman-maintain.completion /etc/bash_completion.d/foreman-maintain` 
+and load it to the current shell `source /etc/bash_completion.d/foreman-maintain`.
```
iNecas: When following this instructions in development, I was getting:

```
foreman-maintain-complete: command not found
```
as it assumes it's on the path. We should probably mention that the the PATH needs to be set accordingly

README.md
iNecas: One suggestion - doesn't have to be solved in this PR, but would be good to think about this now: It would be great if `foreman-maintain upgrade check --target-version ` could produce a list of versions that the system can be upgraded to: since this list could be dynamic, we would need to generate it on actual setup (rather than in packaging time). I can imagine the setting by something like `{ :type => :value, :generator => lambda { Upgrades.available_versions } }` - warping to lambda so that it would not be evaluated every time the file is sourced.

README.md
iNecas: We could also assume that in development, the executable will bi in `$PWD/bin`

iNecas:
Tests are currently filing.

iNecas:
Tested and works flawlessly. Nice work @mbacovsky. Some suggestions inline. I'm wondering if we should even bother with building the cache in build phase of packaging… the first regeneration was not that terrible. And with things like dynamically generated completion lists, we could not even do it beforehand. What we could/should do is running that in post-inst/post-upgrade phase. Maybe that was what you already meant, but just in case…

mbacovsky:
@sean797 that's exactly my plan. I will move it to separate gem if this concept proves itself. It could be re-used for other Clamp based apps we have in the stack i.e. hammer and installer as you suggest.

README.md
mbacovsky: @inecas I'll re-check on notation, what you suggest  feels more natural.
Regarding the target version value, I have some plans for this already. The cache is never build during packaging but rather during rpm %post. Then we can have 
```ruby
{ :type => :enum, :values => Upgrades.available_versions } 
```
We should probably call the cache rebuild as part of upgrade scenario.

The lambda probably won't work as we don't load whole ForemanMaintain (for performance reasons) just the completion parser, so no FM objects are available during the parsing.

Part of more ambitious plans is to add `:function` type. In the project you would inherit from the `Completion` class and implement there methods returning the more dynamic data on the fly e.g. `organizations', `products`, `capsules`, etc.. In the option you could do
```ruby
{ :type => :function, :name => :capsules }
```

mbacovsky:
@iNecas for the cache build I meant to do that on RPM @post. As you said it makes no sense to do it during package build. What didn't occurred to me is the case when f-m is installed sooner then installer finishes. It may be better to omit the pre-build completely. 

We should trigger the cache build after the installer is run to update after possible changes. I'll add something to expire the cache on installer answers change.

README.md
iNecas: > The lambda probably won't work as we don't load whole ForemanMaintain

The idea is that the lambda would be called only on cache-creation phase, to store the values in the cache, not in the completion parser.

mbacovsky:
@iNecas, updated.
- notation updated as suggested
- refactoring of the code
- fixed completion of files and directories to feel more bash-like
- cache expiration, cache expires when scenario answer file changed
- updated docs

I see now why the tests are failing in Jenkins, will update shortly.

lib/foreman_maintain/utils/bash.rb
```diff
@@ -0,0 +1,138 @@
...
+
+        def next_word(dict)
+          next_option_value(dict) || (next_option_or_subcommand(dict) + next_param(dict))
+        end
+
+        def next_option_or_subcommand(dict)
+          dict.keys.select { |k| k.is_a?(String) }.map { |k| k + ' ' }
+        end
+
+        def complete_value(value, partial, is_param)
```
iNecas: the `value` name is misleading: what about something like `completion_options` instead?

mbacovsky: how about `value_description`?

lib/foreman_maintain/cli/backup_command.rb
```diff
@@ -44,7 +45,8 @@ def common_backup_options
             self.class.valid_tape_size(size)
           end
           option ['-i', '--incremental'], 'PREVIOUS_BACKUP_DIR',
-                 'Backup changes since previous backup' do |dir|
+                 'Backup changes since previous backup',
+                 :completion => { :type => :file } do |dir|
```
alda519: directory type?

bin/foreman-maintain-complete
```diff
@@ -0,0 +1,46 @@
...
+rescue
+end
+# rubocop:enable Lint/RescueWithoutErrorClass, Lint/HandleExceptions
+
+config_file = if File.exist?(CONFIG_FILE)
+                CONFIG_FILE
+              else
+                File.join(project_root, 'config/foreman_maintain.yml')
+              end
+completion_cache_file = cache_file(config_file) || '~/.cache/foreman_maintain_completion.yaml'
```
alda519: there are places with `yaml` and other places with `yml` file extenstion, I would make it consistent

iNecas:
I've hit one issue during testing:

```
foreman-maintain backup online  /var/lib/ <TAB>/root/foreman_maintain/lib/foreman_maintain/utils/bash.rb:52:in `complete_value': undefined method `[]' for nil:NilClass (NoMethodError)
	from /root/foreman_maintain/lib/foreman_maintain/utils/bash.rb:77:in `next_param'
	from /root/foreman_maintain/lib/foreman_maintain/utils/bash.rb:44:in `next_word'
	from /root/foreman_maintain/lib/foreman_maintain/utils/bash.rb:20:in `complete'
	from /root/foreman_maintain/bin/foreman-maintain-complete:44:in `<main>'

```

bin/foreman-maintain-complete
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env ruby
+require 'English'
+project_root = File.expand_path('../..', __FILE__)
+$LOAD_PATH.unshift(File.join(project_root, 'lib'))
+CONFIG_FILE = '/etc/foreman-maintain/foreman_maintain.yaml'.freeze
```
alda519: should be `/etc/foreman-maintain/foreman_maintain.yml`?

mbacovsky:
@iNecas, @alda519 updated:
- error on completion fixed
- using 'yml' instead of 'yaml'
- fixed type in --incremental option
- some more refactoring and code reduction

mbacovsky:
@iNecas it will need changes in packaging, probably just installing the completion file. Should I prepare the PR?

iNecas:
Ideally yes, if you can't make it before the time off, it will probably wait for you later, or we will try to find someone else for the packaging part.

mbacovsky:
@iNecas, I've updated gemspec. Incomplete attempt to update packaging can be found here [1] if needed. [1] https://github.com/mbacovsky/foreman-packaging/commit/3822d09ccc7be6dc8efb39d006bdd55918bfbc42

mbacovsky:
@iNecas, I've added fixes for completion issues found during testing packaged version on a fresh box.

I've also created a test build and it seems it installs fine, bash-completion works out of the box (almost, you have to load the completion defs by e.g. starting new shell). 

Not sure how to proceed with the packaging PR. We can either copy my changes when releasing 0.2.7-1 or add them in 0.2.7-2 as a separate PR. What do you think? My packaging changes are in https://github.com/mbacovsky/foreman-packaging/commit/3173c096ee202bc73f9eea26205c23e3745069d1. I was not able to do scratch build in Koji so did local mockbuild, you can check it here https://drive.google.com/open?id=1w51YNsZIdvYCy9jYPdHfYhof5KxJhOci.

iNecas:
Left some nitpicks. Overall reads better now. I've also tested and worked well for me: the only obstacle was the fact that you've built it for tfm software collection so that there were some packaging conflicts and I needed to `rpm -e --nodeps rubygem-foreman_maintain` first before I could install it. After that, everything worked as expected. @alda519 / @sean797 want to give it a try before merging?

Regarding packaging, I'm ok with doing the changes as part of 0.2.7-1 release.

bin/foreman-maintain-complete
```diff
@@ -0,0 +1,46 @@
...
+found = $CHILD_STATUS.success?
+FOREMAN_MAINTAIN = found ? 'foreman-maintain' : File.join(project_root, 'bin/foreman-maintain')
+
+require 'yaml'
+require 'foreman_maintain/utils/bash'
+
+# rubocop:disable Lint/RescueWithoutErrorClass, Lint/HandleExceptions
+def cache_file(config_file)
+  config = YAML.load(File.open(config_file))
+  config[:completion_cache]
```
iNecas: Missed rename?

mbacovsky:
@sean797 we decided not to build the completions during a build. The cache depends on configuration of the system it is used with so the build would produce outdated cache. The cache is currently built during a first attempt to complete something and is expired by installer run. The build on my testing system takes less then a second. This can fixed in the future if too obtrusive. Sounds reasonable?

iNecas:
@mbacovsky I would like to merge it as soon as you have time to finish the packaging part: when is the right time?

mbacovsky:
@iNecas I consider the packaging part finished already. It is in my branch mentioned in the previous comment. I'm not sure what I'm supposed to provide now. My expectation was there will be new f-m release after my PR is merged and during the packaging  we will add the changes from my branch. Or is the plan different?

iNecas:
Ok, I read it as you'll open a PR to packaging right after I merge this. To here you go :)

iNecas:
Thanks @mbacovsky and all involved in this beautiful change.

PR #189, opened 2018-06-13
Fixes #22221 - RAM check before upgrade

upadhyeammit:
This adds feature to check RAM for system, it checks,
1. If its upstream and if katello is enabled then 8 GB of RAM required.
2. If its upstream and if katello not enabled then 4GB ram.
3. For downstream its always 8 gb.

### DIFF
--- a/definitions/checks/check_available_memory.rb
+++ b/definitions/checks/check_available_memory.rb
@@ -1,0 +1,24 @@
+class Checks::CheckAvailableMemory < ForemanMaintain::Check
+  metadata do
+    label :check_available_memory
+    description 'Check if system has sufficient RAM available'
+    tags :pre_upgrade
+  end
+
+  def run
+    with_spinner('Checking if system has sufficient RAM') do
+      total_ram = execute("grep MemTotal /proc/meminfo | awk '{print $2}'").to_i
+      curr_feature = available_feature
+      disp_mem = curr_feature.display_mem
+      mem_in_gb = (curr_feature.min_mem / 1024) / 1024
+      msg = "System has total #{mem_in_gb} GB RAM, at least #{disp_mem} GB RAM required."
+      assert(curr_feature.min_mem < total_ram, msg)
+    end
+  end
+
+  private
+
+  def available_feature
+    feature(:downstream) || feature(:katello) || feature(:upstream)
+  end
+end
--- a/definitions/features/downstream.rb
+++ b/definitions/features/downstream.rb
@@ -1,10 +1,21 @@
 class Features::Downstream < ForemanMaintain::Feature
+  MIN_MEM = 7_900_000
+  DISPLAY_MEM = 8
+
   metadata do
     label :downstream
 
     confine do
       downstream_installation?
     end
+  end
+
+  def min_mem
+    MIN_MEM
+  end
+
+  def display_mem
+    DISPLAY_MEM
   end
 
   def less_than_version?(version)
--- a/definitions/features/katello.rb
+++ b/definitions/features/katello.rb
@@ -1,10 +1,21 @@
 class Features::Katello < ForemanMaintain::Feature
+  MIN_MEM = 7_900_000
+  DISPLAY_MEM = 8
+
   metadata do
     label :katello
 
     confine do
       find_package('katello')
     end
+  end
+
+  def min_mem
+    MIN_MEM
+  end
+
+  def display_mem
+    DISPLAY_MEM
   end
 
   def data_dirs
--- a/definitions/features/upstream.rb
+++ b/definitions/features/upstream.rb
@@ -1,10 +1,21 @@
 class Features::Upstream < ForemanMaintain::Feature
+  MIN_MEM = 4_100_000
+  DISPLAY_MEM = 4
+
   metadata do
     label :upstream
 
     confine do
       !downstream_installation?
     end
+  end
+
+  def min_mem
+    MIN_MEM
+  end
+
+  def display_mem
+    DISPLAY_MEM
   end
 
   def setup_repositories(_version)


definitions/checks/check_available_memory.rb
```diff
@@ -0,0 +1,29 @@
...
+      feature(:downstream) || feature(:upstream)
+    end
+  end
+
+  def run
+    system_memory_check
+  end
+
+  def system_memory_check
+    katello_min_ram = 7_900_000
```
kgaikwad: I would suggest to move these values to respective feature i.e `upstream/downstream`. Based on conditions, just return min. value from feature and then have comparison here.

definitions/checks/check_available_memory.rb
```diff
@@ -0,0 +1,29 @@
...
+    total_ram = `grep MemTotal /proc/meminfo | awk '{print $2}'`.to_i
+    if (feature(:upstream) && katello_plugin_enabled?) || feature(:downstream)
+      assert(katello_min_ram < total_ram, 'This system has less than 8 GB of RAM.')
+    elsif feature(:upstream) && !katello_plugin_enabled?
+      assert(foreman_min_ram < total_ram, 'This system has less than 4 GB of RAM.')
+    end
+  end
+
+  def katello_plugin_enabled?
+    Features::Installer.new.answers.key? 'katello'
```
kgaikwad: Instead of `Features::Installer.new`, you can directly use `feature(:installer)`

sean797:
Should we also remove the check from katello-installer? It seems unnecessary to have this check in both tools, especially considering foreman-maintain is required in a katello install now.

https://github.com/Katello/katello-installer/blob/master/hooks/pre/17-memory_check.rb

iNecas:
The trick is we're currently not running the f-m checks from the installer: IMO we can remove the check from the installer once we run the f-m checks from the installer for non-upgrade case.

definitions/features/downstream.rb
```diff
@@ -1,4 +1,7 @@
 class Features::Downstream < ForemanMaintain::Feature
+  MIN_MEM = 7_900_000
+  DISPLAY_MEM = 8
```
iNecas: It would be good if we had just one value for specifying the minum memory, and the display one would be just calculated from there. If the concern is that we want to be a bit tolerant and not fail, if the real memory is 7.9GB when we require 8GB sharp, it could be handled with incorporating some tolerance (ether in absolute value of let's say 0.5GB, or 10% or something like that).

Also, please avoid using abbreviations for readability purposes, especially for variables with larger scope, so `MIN_MEMORY` would work better here.

iNecas:
@mbacovsky mind having look at this PR as well?

definitions/features/downstream.rb
upadhyeammit: Thank you! I am working on changes and will send those again.

mbacovsky:
I think this deserves some design discussion. To start here is my thinking aloud: The original issue is that the upgrade fails on installer if the installer conditions are not met. Failing on installer run is too late when maintenance mode is on etc. From this perspective the only issue I can see with the current PR (1) is we are duplicating the check implementation and have to keep it in sync with installer. And still we can fail on any other condition.

If for next Katello version we will require 16GB we will have to update f-m and also behave depending on the versions. 

(2) It may be better to extend installer to be able to just run the checks and execute it form a foreman-maintain check. This solution has a caveat though - we run the old installer not the one that will be run after yum update.

Unless I missed something with any of the solutions above we are not able to upgrade system that was installed with `--disable-system-checks`.

(3) From the perspective of the original bug the best solution would be to add extra checks just to the upgrade scenarios where we know the conditions changed with the new version. In other cases we could assume the system was installed and run thus meets the conditions. Then we could run the installer upgrade with `--disable-system-checks` to prevent any failures. It could tricky to not miss any of the condition changes though.

(4) to avoid duplication by running f-m from an installer seems to be the most tricky way. We would have to be able to perform the checks not only for upgrade but also for install. f-m is not able to run on a clean system yet. We would not be able to use features to distinguish the different installer scenarios. Probably  checks with special tags would need to be used and installer scenarios would have to know the tags. This sounds pretty awkward.

Any thoughts anyone?

mbacovsky:
@iNecas, @kgaikwad  any thoughts on my previous comment?

kgaikwad:
@mbacovsky, 

By seeing all aspects that you mentioned and especially point(4), I will prefer to have changes in installer and not having in foreman-maintain.

>  It may be better to extend installer to be able to just run the checks and execute it form a foreman-maintain check.

+1 for this suggestion. 
In this way, installer will not have dependency on foreman-maintain and it will work during both upgrade and install scenario. 
With above change, foreman-maintain will run this check using installer command. Once done, we need to allow users to pass extra option i.e `--disable-system-checks` to installer command during upgrade.
I think, this way we will able to avoid other challenges that you mentioned and it will work with minimal dependencies.

I would like to hear suggestions from @iNecas.

mbacovsky:
@kgaikwad thanks for your feedback!

I investigated this a bit more and have some more input. To sum up the original BZ 
- user wants to install capsule and it runs ok on 4GB
- we require 8GB per both (server and proxy) and stated in the BZ we are not going to loosen the restriction for proxy
- using `--disable-system-checks` seems acceptable solution
- BZ is closed as not a bug
- NACK for duplicating the checks in foreman-maintain

When we add support for capsule upgrades and there were suggestions to increase the limit to 16GB I think the scenario become more prominent. I see the following usability problems:
1. we do not support passing `--disable-system-checks` to installer so there is no way to upgrade system installed with this option via foreman-maintain
2. upgrade fails too late (after `yum update`) 

The reference use case could be: 
- as an admin I want to upgrade sever installed with --disable-system-checks`
- as an admin I want to upgrade server installed without the option but my HW does not meet increased reqs of the new version

If we take the route of calling `foreman-installer --checks-only` from foreman-maintain then there is a problem when the checks are different in the updated version. We have no updated installer at the pre-upgrade phase.

To address `2.` we can either duplicate at least some checks or run the outdated checks as a best effort.

To address 1. we could squeeze between `yum update` and `installer --upgrade` one more step that would run `foreman-installer --checks-only` with updated installer and offer to continue with the upgrade with `--disable-system-checks` if the checks failed. The bad thing is you find out you need HW upgrade too late (same as before) The win is you have a way to perform the upgrade if you wish. I guess the docs are descriptive about the requirements and we can assume a responsible admin checked that before upgrading production  machine. If this is the way to go we should consider if this check should response to `--assumeyes` or it deserves own option `--disable-system-checks`. Both options have its pros and cons.

I'd be glad for any input...

upadhyeammit:
closing this pr as former issue has been closed and this change is stuck between two different components, maybe later on something like this can be implemented when we have common code in installer and maintain.

PR #190, opened 2018-06-14
Fixes #23940 - Change backup metadata keys to have colon

johnpmitsch:
Katello-backup used ':key' in the metadata.yaml file. We
should continue to use this to not break any existing functionality
around the backup metadata

### DIFF
--- a/definitions/procedures/backup/metadata.rb
+++ b/definitions/procedures/backup/metadata.rb
@@ -11,11 +11,11 @@ module Procedures::Backup
     def run
       with_spinner('Collecting metadata') do |spinner|
         metadata = {}
-        metadata['os_version'] = release_info(spinner)
-        metadata['plugin_list'] = plugin_list(spinner) || []
-        metadata['proxy_features'] = proxy_feature_list(spinner) || []
-        metadata['rpms'] = rpms(spinner)
-        metadata['incremental'] = @incremental_dir || false
+        metadata[':os_version'] = release_info(spinner)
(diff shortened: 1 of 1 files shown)


jturel:
Is this for consistency's sake or do we run the risk of problems by not having this?

johnpmitsch:
@jturel for consistency, I ran into this with satellite-clone, when I was using `:rpms`. I updated it to handle both.

I don't know how much is set up in the Satellite/Katello ecosystem that works off of the backup metadata, and I don't know how many users have their own scripts that do. Ideally we can do without the colons since its pretty ugly IMO. To be honest I'm torn between switching back to the colons or just leaving as is

jturel:
I don't think this is the output you were expecting in terms of the format of the yaml keys. However I am using a katello 3.5 box so maybe the metadata changed in later versions of katello-backup?

your PR
```
--
':os_version': CentOS Linux release 7.4.1708 (Core)
':plugin_list': []
':proxy_features': ''
':rpms':
```

katello-backup
```
---
:os_version: 'CentOS Linux release 7.3.1611 (Core) '
:plugin_list:
- bastion-6.1.5
- foreman-tasks-0.10.9
- foreman_docker-3.2.1
- katello-3.5.1.1
:rpms:
```

johnpmitsch:
After some off-thread discussion, I think its better to keep the keys as they are in the yaml format as its cleaner syntax. As far as I can tell, there isn't much tooling that parses the metadata file, and its trivial to update any that does. If this causes any issues for anyone, we can revisit

PR #192, opened 2018-06-18
Fixes #22578 - fail! if server is self-registered

kgaikwad:
I've taken the @swapab's original PR #142.
Added knowledge base article procedure and resolved few review comments.

### DIFF
--- a/definitions/checks/system_registration.rb
+++ b/definitions/checks/system_registration.rb
@@ -11,20 +11,29 @@ class Checks::SystemRegistration < ForemanMaintain::Check
 
   def run
     assert(
-      !rhsm_hostname_eql_hostname?,
+      rhsm_hostname.downcase != hostname.downcase,
       notification_message,
-      :next_steps => [
-        Procedures::KnowledgeBaseArticle.new(:doc => 'migrate_self_registered_satellite_63')
-      ]
+      :next_steps => steps_to_follow
     )
   end
 
+  def steps_to_follow
+    if feature(:instance).downstream.current_minor_version == '6.3'
+      [Procedures::KnowledgeBaseArticle.new(:doc => 'migrate_self_registered_satellite_63')]
+    else
+      []
+    end
+  end
+
   def notification_message
-    <<-MESSAGE.strip_heredoc
-      Satellite is self registered.
-      Self registered satellite are not supported as of Satellite version 6.3.
-      Please follow steps to migrate.
-    MESSAGE
+    product_name = feature(:instance).product_name
+
+    msg = "\n#{product_name} is self registered."
+    msg += "\nSelf registered #{product_name} is not supported from version 6.3 onwards."
+    if feature(:instance).downstream.current_minor_version == '6.3'
+      msg += "\nPlease follow steps to migrate."
+    end
+    msg
   end
 
   def rhsm_hostname
@@ -33,9 +42,5 @@ class Checks::SystemRegistration < ForemanMaintain::Check
 
   def rhsm_conf_file
     '/etc/rhsm/rhsm.conf'
-  end
-
-  def rhsm_hostname_eql_hostname?
-    rhsm_hostname.casecmp(hostname).zero?
   end
 end
--- a/definitions/checks/system_registration.rb
+++ b/definitions/checks/system_registration.rb
@@ -10,6 +10,7 @@ class Checks::SystemRegistration < ForemanMaintain::Check
   end
 
   def run
+    puts "TEST:: #{rhsm_hostname} -> #{hostname}"
     assert(
       rhsm_hostname.downcase != hostname.downcase,
       notification_message,
--- a/definitions/checks/system_registration.rb
+++ b/definitions/checks/system_registration.rb
@@ -10,22 +10,21 @@ class Checks::SystemRegistration < ForemanMaintain::Check
   end
 
   def run
-    if rhsm_hostname_eql_hostname?
-      if feature(:downstream)
-        notification_message = <<-MESSAGE
-  Satellite is self registered
+    assert(
+      !rhsm_hostname_eql_hostname?,
+      notification_message,
+      :next_steps => [
+        Procedures::KnowledgeBaseArticle.new(:doc => 'migrate_self_registered_satellite_63')
+      ]
+    )
+  end
 
-  Self registered satellite won't be supported as of Satellite version 6.3. Please follow following documentation to resolve.
-
-  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/installation_guide/preparing_your_environment_for_installation#system_requirements
-
-  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite#migrating_a_self_registered_satellite
+  def notification_message
+    <<-MESSAGE.strip_heredoc
+      Satellite is self registered.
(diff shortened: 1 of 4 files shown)


upadhyeammit:
Request to use correct message style for second commit of this PR - https://github.com/theforeman/foreman_maintain/pull/192/commits/66698ea0c6560728b102f70e2cd701642ad6a2db

definitions/checks/system_registration.rb
```diff
@@ -5,15 +5,26 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
+      if feature(:downstream)
+        notification_message = <<-MESSAGE
+  Satellite is self registered
+
+  Self registered satellite won't be supported as of Satellite version 6.3. Please follow following documentation to resolve.
+
+  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/installation_guide/preparing_your_environment_for_installation#system_requirements
+
+  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/upgrading_and_updating_red_hat_satellite/upgrading_red_hat_satellite#migrating_a_self_registered_satellite
+    MESSAGE
```
upadhyeammit: As per old PR it has been suggested to move fail! method outside of if else block so we always fail if Satellite is self registered.

definitions/checks/system_registration.rb
```diff
@@ -5,15 +5,26 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
   def run
     if rhsm_hostname_eql_hostname?
-      warn! 'System is self registered'
+      if feature(:downstream)
+        notification_message = <<-MESSAGE
+  Satellite is self registered
+
+  Self registered satellite won't be supported as of Satellite version 6.3. Please follow following documentation to resolve.
+
+  https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/installation_guide/preparing_your_environment_for_installation#system_requirements
```
upadhyeammit: It was suggested to remove this link and just give user guide on how to migrate from self registered system. Request to change this too.

definitions/checks/system_registration.rb
```diff
@@ -5,16 +5,26 @@ class Checks::SystemRegistration < ForemanMaintain::Check
     tags :default
 
     confine do
-      file_exists?('/etc/rhsm/rhsm.conf') &&
-        !feature(:foreman_server) &&
-        feature(:foreman_proxy)
+      feature(:downstream) && file_exists?('/etc/rhsm/rhsm.conf')
```
mbacovsky: This check was for proxy only before and now is satellite server only. Is it okay?

mbacovsky: IMO feature(:downstream) has confusing behavior and should be fixed (probably not as part of this PR)

definitions/checks/system_registration.rb
```diff
@@ -5,16 +5,26 @@ class Checks::SystemRegistration < ForemanMaintain::Check
...
     end
   end
 
   def run
-    if rhsm_hostname_eql_hostname?
-      warn! 'System is self registered'
-    end
+    assert(
+      !rhsm_hostname_eql_hostname?,
```
mbacovsky: wouldn't be `rhsm_hostname.downcase != hostname.downcase` more readable?

definitions/checks/system_registration.rb
kgaikwad: Yes, previously this check was for proxy. According to BZ details, it should restrict server as well as proxy upgrading in that case and that's why previous condition was removed in initial commit.

My bad, due to my changes, it is now restricting only on server.  My intention behind adding `feature(:downstream)` in config was because there is nothing like self registered term in upstream.

definitions/checks/system_registration.rb
kgaikwad: Yes but `rubocop` wouldn't be happy.

definitions/checks/system_registration.rb
mbacovsky: And it made perfect sense to use the `fetaure(:downstream)`. And that's why I suggested there is a bug in the feature and it should behave as you expected. `:downstream` shouldn't mean `:downstream_server` and I think that should be fixed. What do you think @iNecas?

@kgaikwad technically I think you can have self-registered Katello, but it does not make sense to talk about support there. Shouldn't the downstream detection stay?

kgaikwad: @mbacovsky, yes it is needed. Reverting commit back to previous :-)

iNecas: I think `feature(:downsteam)` should be fixed to handle both the server and the capsule. Let's fix that behavior in separete PR, but doing this PR in a way that handles the proxy properly

definitions/checks/system_registration.rb
mbacovsky: Thanks for making it a bit better, still I think we should disable this performance cop in the project :) but surely won't block the PR on this. I was checking if it is just me who has problem with the ugly notation and I found in ruby >= 2.4 there is  `str.casecmp?(other_str)` which is better but that we can not use due to the ruby version.

definitions/checks/system_registration.rb
kgaikwad: For now, I am removing `feature(:downstream)` from confine as it won't allow to run this check on capsule.

definitions/checks/system_registration.rb
kgaikwad: +1,  Disabled casecmp performance cop in rubocop file.

definitions/checks/system_registration.rb
mbacovsky: I'm tracking the :downstream feature problem in https://projects.theforeman.org/issues/25549.

@kgaikwad, @iNecas  is this check acceptable in upstream? I think we can have Katello registered to itself with subscription-manager. In that case we shouldn't refer to the KBA. If it makes no sense in upstream, we should fix the confine condition or at least add #FIXME and reference the issue.

upadhyeammit: I think in long term we should also have capsule as a package in capsule repositories as we have satellite as package. This will make it very easy to detect that its downstream capsule.

kgaikwad: @mbacovsky, @upadhyeammit,
Updated condition in confine block. Please verify and let me know if it's serve purpose for now. or any better way to handle it.

definitions/checks/system_registration.rb
```diff
@@ -5,16 +5,29 @@ class Checks::SystemRegistration < ForemanMaintain::Check
     tags :default
 
     confine do
-      file_exists?('/etc/rhsm/rhsm.conf') &&
-        !feature(:foreman_server) &&
-        feature(:foreman_proxy)
+      # FIXME: Once this https://projects.theforeman.org/issues/25549 issue
+      # gets fix then please fix code here. Only feature(:downstream) expected.
+      (feature(:downstream) || (!feature(:katello) && feature(:foreman_proxy))) &&
```
mbacovsky: @kgaikwad it seems it does not exclude upstream proxy. Wouldn't something like `feature(:downstream) || (execute('rpm -q foreman-proxy') =~ /sat.noarch/)` describe better the options?
Thanks for the #FIXME comment!

bkearney:
@kgaikwad this is an older bug, can you please reviive this and push it through?

upadhyeammit:
Closing this pr as check is not required and former bugzilla has been closed!

PR #193, opened 2018-06-18
Fixes #23984 - Update PATH in initialization

johnpmitsch:
We use commands like 'runuser' that live in /sbin, so we need
to ensure the proper PATH is loaded when running commands. This
is important when using foreman-maintain in a cron job.

### DIFF
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -53,6 +53,18 @@ module ForemanMaintain
       self.config = Config.new(options)
       load_definitions
       init_logger
+      update_path
+    end
+
+    # Appending PATH with expected paths needed for commands we run
+    def update_path
+      paths = ['/sbin']
+      existing_paths = ENV['PATH'].split(':')
+      paths = paths - existing_paths
+      if paths.any?
+        paths = paths.join(':').chomp(':')
+        ENV['PATH'] = "#{ENV['PATH']}:#{paths}"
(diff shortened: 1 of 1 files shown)


lib/foreman_maintain.rb
```diff
@@ -53,6 +53,12 @@ def setup(options = {})
       self.config = Config.new(options)
       load_definitions
       init_logger
+      update_path(['/sbin'])
+    end
+
+    def update_path(paths)
```
iNecas: I don't think this needs actually be parametrized. It should be a responsibility of this method itself to set the path correctly - otherwise we have the logic at two places, even though close to each other but still.

Also please add some explanatory comment on why we are doing this.

lib/foreman_maintain.rb
```diff
@@ -53,6 +53,12 @@ def setup(options = {})
...
+      paths = paths.join(':')
```
iNecas: I guess we should check and add each path only if it's not already present.

johnpmitsch:
@iNecas updated

iNecas:
Works for me, @alda519 mind testing?

alda519:
sure, tested
```
export PATH="/usr/bin:/bin"
foreman-maintain ...
```
and everything works fine

iNecas:
Thanks @johnpmitsch and @alda519

PR #195, opened 2018-06-21
Refs #23984 - resolved rubocop warning

kgaikwad:


iNecas:
Please add `Refs #23984` to the commit message, as that's were the change was introduced, probably not that important here, but in general, it helps in case we would need to do some cherry-picking later.

kgaikwad:
@iNecas,
done!!

iNecas:
Thanks @kgaikwad

PR #197, opened 2018-06-29
Fixes #24101 - execute confine for puppet3 check for sat 6.4

kgaikwad:


definitions/scenarios/upgrade_to_satellite_6_4.rb
```diff
@@ -21,8 +21,8 @@ class PreUpgradeCheck < Abstract
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:pre_upgrade))
+      add_steps(find_checks(:puppet_upgrade_guide))
```
kgaikwad: I think, this check should be before validating repositories so moved line above repo. validate check.

kgaikwad: If upgrade path goes from sat 6.4 to sat 6.4.z not from sat 6.3 to sat 6.4.z then we don't need it in sat 6.4.z scenario. But here, it is needed for sat 6.4.

kgaikwad: @iNecas, 
Still I think,  it may happen that upgrade path from sat 6.3.z to sat 6.4.z so would like to know your suggestion on this?

iNecas: I think it should probably not be in 6.4.z, but only in 6.4 (upgrading from 6.3)

kgaikwad: Thank you @iNecas.
Updated PR with this modification.

iNecas:

PR #198, opened 2018-07-05
Fixes #23969 - clean contents of directory

upadhyeammit:
Instead of deleting directory this will delete contents of yum and dnf cache.

### DIFF
--- a/lib/foreman_maintain/concerns/system_helpers.rb
+++ b/lib/foreman_maintain/concerns/system_helpers.rb
@@ -116,13 +116,9 @@ module ForemanMaintain
                  :interactive => true)
       end
 
-      def clean_all_packages(options = {})
-        options.validate_options!(:assumeyes)
-        yum_options = []
-        yum_options << '-y' if options[:assumeyes]
-        execute!("yum #{yum_options.join(' ')} clean all", :interactive => true)
-        execute!('rm -rf /var/cache/yum')
-        execute!('rm -rf /var/cache/dnf')
+      def clean_all_packages
+        execute!('dnf clean all') if find_package('dnf')
+        execute!('yum clean all') if find_package('yum')
       end
 
       def package_version(name)
--- a/definitions/procedures/packages/update.rb
+++ b/definitions/procedures/packages/update.rb
@@ -7,7 +7,7 @@ module Procedures::Packages
 
     def run
       assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
-      clean_all_packages(:assumeyes => assumeyes_val)
+      clean_all_packages
       packages_action(:update, @packages, :assumeyes => assumeyes_val)
     end
 


gnurag:
lgtm

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -121,8 +121,8 @@ def clean_all_packages(options = {})
         yum_options = []
         yum_options << '-y' if options[:assumeyes]
         execute!("yum #{yum_options.join(' ')} clean all", :interactive => true)
-        execute!('rm -rf /var/cache/yum')
-        execute!('rm -rf /var/cache/dnf')
+        execute!('rm -rf /var/cache/yum/*')
+        execute!('rm -rf /var/cache/dnf/*')
```
sean797: Why are we even deleting a directory directly? Is `yum/dnf clean all` not enough for our purposes?

kgaikwad: @sean797, 

Yes, it is not enough to run `yum clean all` as it doesn't delete `/var/cache/yum` when you are using `dnf` package.
For more details, please check [PR-119](https://github.com/theforeman/foreman_maintain/pull/119#issue-149447067)

kgaikwad: @sean797,

I agree with what you mentioned. Yes, we can use `yum-deprecated clean all`.
IIRC, above commands only clear the cache for the current version of OS and there might be a chance where some files may present under those directories after clean all.
This is also one of the reason that I have manually deleted content of those two directories.

If there is any other way so that we can handle this, please let us know.

sean797: @kgaikwad if there are any files that are present in the cache directory after running `clean all` that is fine. Any files left over are done so on purpose by yum or dnf.

Why would you need to clean any files present under those directories? We just need to clear the cache and install the latest pkg - which is what `clean all` is designed for.

If you have some edge case where that isn't working please raise a bug with yum/dnf and add a comment explaining why we are temporarily adding a workaround.

gnurag: Hello @sean797, we were initially inclined to manually delete the directories since we had metadata from disabled repos, and yum nudged to use `rm`.

```
[root@host cache]# yum clean all
Loaded plugins: fastestmirror
Cleaning repos: base centos-sclo-rh centos-sclo-sclo epel extras foreman foreman-plugins puppet5 tfm-ror51 updates
Cleaning up everything
Maybe you want: rm -rf /var/cache/yum, to also free up space taken by orphaned data from disabled or removed repos
Cleaning up list of fastest mirrors
[root@host cache]#
```
However, upon close look it looks like manually running `rm` would also fail in case the user sets `cachedir=/some/other/dir` in `/etc/yum.conf`. We should therefore revert to running `yum/dnf clean all` only.

sean797: @gnurag if you are worried about users who have changed their `cachedir` I'm sure you could get the configured value if you wanted to.

My point is actually clearly explained in the output you have posted above:

```
Maybe you want: rm -rf /var/cache/yum, to also free up space taken by orphaned data from disabled or removed repos
```
Your output tells you the reason why you might want to `rm -rf $cachedir`, foreman-maintain's goal is not to `free up space taken by orphaned data from disabled or removed repos` is it to upgrade Foreman/Katello/Satellite, `freeing up space taken by orphaned data from disabled or removed repos` is not required to upgrade Foreman/Katello/Satellite.

gnurag: Taking the checks present in `definitions/checks` into consideration, I would hazard a guess that foreman-maintain's goal is to *successfully* drive a foreman/satellite upgrade. That said, I agree with your observation that freeing up space taken by yum metadata may be an overreach of its purpose. We should revert to running `yum/dnf clean all` only.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -116,13 +116,12 @@ def packages_action(action, packages, options = {})
...
-        options.validate_options!(:assumeyes)
-        yum_options = []
-        yum_options << '-y' if options[:assumeyes]
-        execute!("yum #{yum_options.join(' ')} clean all", :interactive => true)
-        execute!('rm -rf /var/cache/yum')
-        execute!('rm -rf /var/cache/dnf')
+      def clean_all_packages
+        if find_package('dnf')
+          execute!('dnf clean all')
+        elsif find_package('yum')
```
mbacovsky: Would it make sense to clean both if both are installed?
```ruby
execute!('dnf clean all') if find_package('dnf')
execute!('yum clean all') if find_package('yum')
```

mbacovsky:
Thanks @upadhyeammit!

PR #199, opened 2018-07-06
Fixes #24184 - fix typo for clear_register method

kgaikwad:
@iNecas,

While going through the code found this so thought to fix it.
I didn't get any place where `clear_register` method is being called. So I am not sure whether we really need this method. If not, may I remove this method?


iNecas:
Thanks. It might be still useful to have it, especially for tests. Thanks for fixing.

PR #200, opened 2018-07-12
Fixes #23673 - handle services that are remote only

mbacovsky:
This patch fixes the situation when the service being worked with
is configured as remote (mongo, postgresql) and is not
installed on the local system.


iNecas:
Travis seems to be stuck, but the tests are passing locally: I'm pulling this in. Thanks @mbacovsky

PR #204, opened 2018-07-30
Fixes #23500 - upstream repositories check

upadhyeammit:
This checks if katello, pulp, candlepin, foreman, foreman-plugin, puppet repositories enabled on system and provides a procedure to disable those.

### DIFF
--- a/definitions/checks/repositories/check_upstream_repository.rb
+++ b/definitions/checks/repositories/check_upstream_repository.rb
@@ -13,9 +13,9 @@ class Checks::CheckUpstreamRepository < ForemanMaintain::Check
 
   def run
     with_spinner('Checking for presence of upstream repositories') do
-      enabled_upstream_repos = feature(:system_repos).upstream_repos_id
+      enabled_upstream_repos = feature(:system_repos).upstream_repos_ids
       assert(enabled_upstream_repos.empty?,
-             "System has upstream #{enabled_upstream_repos} repositories enabled",
+             "System has upstream #{enabled_upstream_repos.join(',')} repositories enabled",
              :next_steps => Procedures::Repositories::Disable.new(:repos => enabled_upstream_repos))
     end
   end
--- a/definitions/checks/repositories/check_upstream_repository.rb
+++ b/definitions/checks/repositories/check_upstream_repository.rb
@@ -13,48 +13,10 @@ class Checks::CheckUpstreamRepository < ForemanMaintain::Check
 
   def run
     with_spinner('Checking for presence of upstream repositories') do
-      enabled_upstream_repos = run_compare
-      repos_to_disable = comma_separated_repositories(enabled_upstream_repos)
+      enabled_upstream_repos = feature(:system_repos).upstream_repos_id
       assert(enabled_upstream_repos.empty?,
-             "System has upstream #{repos_to_disable} repositories enabled",
-             :next_steps => Procedures::Repositories::Disable.new(:repos => repos_to_disable))
+             "System has upstream #{enabled_upstream_repos} repositories enabled",
+             :next_steps => Procedures::Repositories::Disable.new(:repos => enabled_upstream_repos))
     end
-  end
-
-  def comma_separated_repositories(repos)
-    repos.map! { |r| r.gsub(%r{Repo-id:|\/+\w*}, '') }.join(',')
-  end
-
-  def run_compare
-    enabled_upstream_repos = []
-    repo_id = ''
-    system_repos.each do |repo|
-      repo_id = repo if repo =~ /Repo-id/
-      next unless repo =~ /Repo-baseurl/
-
-      upstream_repo.each do |regex|
-        enabled_upstream_repos.push(repo_id) if repo =~ regex
-      end
-    end
-    return enabled_upstream_repos
-  end
-
-  def system_repos
-    repos = execute("yum repolist enabled -d 6 -e 0 2> /dev/null | grep -E 'Repo-id|Repo-baseurl'")
-    return [] if repos.empty?
-
-    repos.delete!(' ').split("\n")
-  end
-
-  def upstream_repo
-    repo_urls = { :Foreman => %r{yum.theforeman.org\/},
-                  :Katello => %r{fedorapeople.org\/groups\/katello\/releases\/yum\/[\/|\w|.]*} }
-    [/#{repo_urls[:Foreman]}+releases/,
-     /#{repo_urls[:Foreman]}+plugins/,
-     /#{repo_urls[:Katello]}+katello/,
-     /#{repo_urls[:Katello]}+client/,
-     /#{repo_urls[:Katello]}+candlepin/,
-     /#{repo_urls[:Katello]}+pulp/,
-     %r{yum.puppetlabs.com\/el\/[\/|\w|.]*}]
   end
 end
--- a/definitions/checks/repositories/check_upstream_repository.rb
+++ b/definitions/checks/repositories/check_upstream_repository.rb
@@ -1,0 +1,60 @@
+class Checks::CheckUpstreamRepository < ForemanMaintain::Check
+  metadata do
+    label :check_upstream_repository
+    description 'Check if any upstream repositories are enabled on system'
+    tags :pre_upgrade
+    preparation_steps do
+      Procedures::Packages::Install.new(:packages => %w[yum-utils])
+    end
+    confine do
+      feature(:downstream)
+    end
+  end
+
+  def run
+    with_spinner('Checking for presence of upstream repositories') do
+      enabled_upstream_repos = run_compare
+      repos_to_disable = comma_separated_repositories(enabled_upstream_repos)
+      assert(enabled_upstream_repos.empty?,
+             "System has upstream #{repos_to_disable} repositories enabled",
+             :next_steps => Procedures::Repositories::Disable.new(:repos => repos_to_disable))
+    end
+  end
+
+  def comma_separated_repositories(repos)
+    repos.map! { |r| r.gsub(%r{Repo-id:|\/+\w*}, '') }.join(',')
+  end
+
+  def run_compare
+    enabled_upstream_repos = []
+    repo_id = ''
+    system_repos.each do |repo|
+      repo_id = repo if repo =~ /Repo-id/
+      next unless repo =~ /Repo-baseurl/
+
+      upstream_repo.each do |regex|
+        enabled_upstream_repos.push(repo_id) if repo =~ regex
+      end
+    end
+    return enabled_upstream_repos
+  end
+
+  def system_repos
+    repos = execute("yum repolist enabled -d 6 -e 0 2> /dev/null | grep -E 'Repo-id|Repo-baseurl'")
+    return [] if repos.empty?
+
+    repos.delete!(' ').split("\n")
+  end
+
+  def upstream_repo
+    repo_urls = { :Foreman => %r{yum.theforeman.org\/},
+                  :Katello => %r{fedorapeople.org\/groups\/katello\/releases\/yum\/[\/|\w|.]*} }
+    [/#{repo_urls[:Foreman]}+releases/,
+     /#{repo_urls[:Foreman]}+plugins/,
+     /#{repo_urls[:Katello]}+katello/,
+     /#{repo_urls[:Katello]}+client/,
+     /#{repo_urls[:Katello]}+candlepin/,
+     /#{repo_urls[:Katello]}+pulp/,
+     %r{yum.puppetlabs.com\/el\/[\/|\w|.]*}]
+  end
+end
--- a/definitions/features/system_repos.rb
+++ b/definitions/features/system_repos.rb
@@ -11,7 +11,7 @@ class Features::SystemRepos < ForemanMaintain::Feature
         repositories[repo] = url if url =~ regex
       end
     end
-    return repositories
+    repositories
   end
 
   def enabled_repos_hash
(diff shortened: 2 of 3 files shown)


definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -0,0 +1,65 @@
...
+      next unless repo =~ /Repo-baseurl/
+      upstream_repo.each do |regex|
+        enabled_upstream_repos.push(repo_id) if repo =~ regex
+      end
+    end
+  end
+
+  def system_repos
+    repos = execute("yum repolist enabled -d 6 -e 0 2> /dev/null | grep -E 'Repo-id|Repo-baseurl'")
+    if !repos.empty?
```
kgaikwad: Instead of `if not.. else`, you can directly use - 

```
return [ ] if repos.empty?
repos.delete!(' ').split("\n")
```

definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -0,0 +1,65 @@
...
+           :next_steps => next_steps)
+  end
+
+  def repo_label(repos)
+    repos.map! { |r| r.gsub(%r{Repo-id:|\/+\w*}, '') }
+    repos.join(',')
+  end
+
+  def run_compare(enabled_upstream_repos)
+    repo_id = ''
```
kgaikwad: Instead of passing `enabled_upstream_repos`, could you please modify method and return list of enabled upstream repositories from it.

definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -0,0 +1,65 @@
...
+      run_compare(enabled_upstream_repos)
+    end
+    repos_to_disable = repo_label(enabled_upstream_repos)
+    next_steps = []
+    next_steps.concat(
+      [Procedures::Repositories::Disable.new(:repos => repos_to_disable)]
+    )
+    assert(enabled_upstream_repos.empty?,
+           "System has upstream #{repos_to_disable} repositories enabled",
+           :next_steps => next_steps)
```
kgaikwad: No need to declare extra variable `next_steps` here. 
You can directly assign array `[Procedures::Repositories::Disable.new(:repos => repos_to_disable)]` to `:next_steps` key.

definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  def repo_label(repos)
```
kgaikwad: As this method returns comma separated list of repositories labels, please modify name of it.
e.g. `comma_separated_repositories` or `comma_separated_repositories_labels`


Also, you can modify code to `repos.map { |r| r.gsub(%r{Repo-id:|\/+\w*}, '') }.join(',')`

upadhyeammit:
@iNecas @mbacovsky request you to review changes once.

mbacovsky:
@upadhyeammit nice, LGTM! Just a few comments...

Why not to do this in upstream too? 

It may be worth considering to move `system_repos` and `disable` to a `feature(:system_repos)` with `list` and `disable` methods. List could return hash of {<repo_id> => {name =>, baseurl => ,...}} and other info from ` yum repolist enabled -d 6`. IT would make it more re-usable and easier to find later. Also when the repolist data are parsed at the beginning, the code might be cleaner down the road. But just nice to have and I don't think wee need to have it to merge this.

mbacovsky:
@upadhyeammit forget about the question about doing this in upstream ;), I realized it does not make sense ...

definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -0,0 +1,58 @@
...
+      enabled_upstream_repos = run_compare
+      repos_to_disable = comma_separated_repositories(enabled_upstream_repos)
+      assert(enabled_upstream_repos.empty?,
+             "System has upstream #{repos_to_disable} repositories enabled",
+             :next_steps => Procedures::Repositories::Disable.new(:repos => repos_to_disable))
+    end
+  end
+
+  def comma_separated_repositories(repos)
+    repos.map! { |r| r.gsub(%r{Repo-id:|\/+\w*}, '') }.join(',')
```
adamruzicka: Since you're immediately using the result of the `map!` I'd recommend using `map` which does not "mutate" the original array

definitions/procedures/repositories/disable.rb
```diff
@@ -0,0 +1,11 @@
+module Procedures::Repositories
+  class Disable < ForemanMaintain::Procedure
+    metadata do
+      param :repos, 'List of repositories to disable'
```
mbacovsky: This would be better to pass as a list instead of csv

upadhyeammit:
@mbacovsky create new feature as requested and other required changes, request to review. Thank You!

jameerpathan111:
Tested for bellow given scenario.

scenario:
-  run "./bin/foreman-maintain health check --label check-upstream-repository"
```
# ./bin/foreman-maintain health check --label check-upstream-repository
Check if any upstream repositories are enabled on system: 
- Checking for presence of upstream repositories                      [OK]      

```

- create upstream repositories
-  run "./bin/foreman-maintain health check --label check-upstream-repository"
```

# ./bin/foreman-maintain health check --label check-upstream-repository
Check if any upstream repositories are enabled on system: 
/ Checking for presence of upstream repositories                      [FAIL]    
System has upstream candlepin_repo,client_repo,foreman_repo,katello_repo,plugins_repo,pulp_repo
repositories enabled
Continue with step [Disable repositories]?, [y(yes), n(no), q(quit)] y
Disable repositories:                                                 [OK]      
Check if any upstream repositories are enabled on system: 
| Checking for presence of upstream repositories                      [OK]      

```

@upadhyeammit pr works for katello, pulp, candlepin, foreman, foreman-plugin repositories 
and disables them as well but it does not work for puppet repository.
puppet repository remains enabled.(http://yum.puppetlabs.com/el/6.4/x86_64/)

definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -0,0 +1,58 @@
...
+  def upstream_repo
+    repo_urls = { :Foreman => %r{yum.theforeman.org\/},
+                  :Katello => %r{fedorapeople.org\/groups\/katello\/releases\/yum\/[0-9].[0-9]\/} }
+    [/#{repo_urls[:Foreman]}+releases/,
+     /#{repo_urls[:Foreman]}+plugins/,
+     /#{repo_urls[:Katello]}+katello/,
+     /#{repo_urls[:Katello]}+client/,
+     /#{repo_urls[:Katello]}+candlepin/,
+     /#{repo_urls[:Katello]}+pulp/,
+     %r{yum.puppetlabs.com\/el\/\d+[\w|\.]*+[\/\w]*/}]
```
kgaikwad: There is error `nested *?+ in regexp: /yum.puppetlabs.com\/el\/\d+[\w|\.]*+[\/\w]*\// (SyntaxError)` with ruby 1.8.7 environment.
Regex `%r{yum.puppetlabs.com\/el\/\d+[\w|\.]*[\/\w]*/}` would work as expected.
It matches patterns like 
`http://yum.puppetlabs.com/el/6.4/x86_64/`, 
`http://yum.puppetlabs.com/el/6/x86_64/`.

definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -0,0 +1,58 @@
...
+    enabled_upstream_repos = []
+    repo_id = ''
+    system_repos.each do |repo|
+      repo_id = repo if repo =~ /Repo-id/
+      next unless repo =~ /Repo-baseurl/
+      upstream_repo.each do |regex|
+        enabled_upstream_repos.push(repo_id) if repo =~ regex
+      end
+    end
+    return enabled_upstream_repos
```
kgaikwad: Return value of a ruby method is the return value of the last statement in the method body
so no need of `return` keyword here. Please remove it.

upadhyeammit:
> @upadhyeammit @kgaikwad
> Tested and pr works as expected.
> But only thing i am concerned about is that, if we choose any non version specific
> repository then pr won't be able to handle this kind of repositories.
> 
> [foreman_maintain/definitions/checks/repositories/check_upstream_repository.rb](https://github.com/theforeman/foreman_maintain/blob/f925c532a47c6909fca9ce7b5043e0cb99531650/definitions/checks/repositories/check_upstream_repository.rb#L49)
> 
> Line 49 in [f925c53](/theforeman/foreman_maintain/commit/f925c532a47c6909fca9ce7b5043e0cb99531650)
>  :Katello => %r{fedorapeople.org\/groups\/katello\/releases\/yum\/[0-9].[0-9]\/} } 
> 
> for ex. https://fedorapeople.org/groups/katello/releases/yum/latest/katello
> is it ok to not consider those kind of repositories?

Nice suggestion, I changed the regex for katello repo and now it should consider non version specific repo too. Thank You. Pushing change for same.

jameerpathan111:
@upadhyeammit 
Tested
Looks good!:+1:

mbacovsky:
@upadhyeammit sorry for leaving this intact for so long.

Above you wrote that you've added the feature I requested but it is missing in the code. Any idea why it was lost? There is also bunch of unaddressed comments from other reviewers, could you take a look and fix/comment?

I'm missing some of my comments that were probably lost due to force-push. For this reason and also to better see the changes between the commits I prefer to push all the changes in separate commits that are squashed in the end. With this approach you can still see the whole contribution and don't have to revisit the whole code for eventual changes. Wouldn't you mind sending the updates in separate commits instead of force-pushing?

upadhyeammit:
> @upadhyeammit sorry for leaving this intact for so long.
> 
> Above you wrote that you've added the feature I requested but it is missing in the code. Any idea why it was lost? There is also bunch of unaddressed comments from other reviewers, could you take a look and fix/comment?
> 
> I'm missing some of my comments that were probably lost due to force-push. For this reason and also to better see the changes between the commits I prefer to push all the changes in separate commits that are squashed in the end. With this approach you can still see the whole contribution and don't have to revisit the whole code for eventual changes. Wouldn't you mind sending the updates in separate commits instead of force-pushing?

Hello,

Sorry for mess, I changed laptop and as its old work had bad time to carry it over, anyways I pushed the correct changes in new commit request you to have a look.

Thank You!

ntkathole:
@mbacovsky @iNecas Changes looks good from testing side. ACK.

definitions/features/system_repos.rb
```diff
@@ -0,0 +1,48 @@
...
+    return {} if repos.empty?
+
+    Hash[*repos.delete!(' ').split("\n")]
+  end
+
+  def upstream_repos_ids
+    trim_repoids(upstream_repos.keys)
+  end
+
+  def trim_repoids(repos)
```
mbacovsky: @upadhyeammit Could you make this `private`?

definitions/features/system_repos.rb
```diff
@@ -0,0 +1,48 @@
...
+    repos.map { |r| r.gsub(%r{Repo-id:|\/+\w*}, '') }
+  end
+
+  def disable_repos(repo_ids)
+    execute!("yum-config-manager --disable #{repo_ids.join(',')}")
+  end
+
+  def upstream_repo_urls
```
mbacovsky: Could you make this private?

mbacovsky:
@upadhyeammit two single nits inline. Otherwise it LGTM.
@kgaikwad could please check all your comments were addressed and update your review?

definitions/features/system_repos.rb
upadhyeammit: done!

definitions/features/system_repos.rb
upadhyeammit: done!

mbacovsky:
We can merge it once @kgaikwad is okay with the changes.

definitions/features/system_repos.rb
```diff
@@ -0,0 +1,50 @@
...
+    repositories = {}
+    list.each do |repo, url|
+      upstream_repo_urls.each do |regex|
+        repositories[repo] = url if url =~ regex
+      end
+    end
+    return repositories
+  end
+
+  def list
```
kgaikwad: This method is returning enabled repos.
It would be good to have more specific name than just mentioning it as a `list`.
example - `enabled_repos_hash or repo_urls_by_repo`.

What do you think?

definitions/features/system_repos.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def upstream_repos
+    repositories = {}
+    list.each do |repo, url|
+      upstream_repo_urls.each do |regex|
+        repositories[repo] = url if url =~ regex
+      end
+    end
+    return repositories
```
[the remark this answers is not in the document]
upadhyeammit: @kgaikwad if I dont explicitly return repositories then its returning list/enabled_repos_hash so I kept the return.

definitions/features/system_repos.rb
upadhyeammit: I changed the method name as suggested. Thank You !

definitions/features/system_repos.rb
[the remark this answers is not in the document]
kgaikwad: Keep repositories as last statement but without 'return' keyword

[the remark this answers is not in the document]
upadhyeammit: Changed it, thank you :)

mbacovsky:
Okay lets merge it. Thanks everybody involved!

PR #209, opened 2018-09-04
Fixes #24797 - Use dynflowd on Foreman >= 1.17

mbacovsky:
In foreman 1.17 the Dynflow became part of Foreman and the dynflowd
service is used instead of foreman-tasks. We keep mapping of resources
from 'hammer ping' to related services and it was missing the change.

IMO this should be just a temporary fix and we need to change the HammerPing check to use the services directly as tracked in https://projects.theforeman.org/issues/24793

### DIFF
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -108,11 +108,11 @@ class Features::ForemanTasks < ForemanMaintain::Feature
   end
 
   def services
-    if check_min_version('foreman', '1.17')
-      { 'dynflowd' => 30 }
-    else
-      { 'foreman-tasks' => 30 }
-    end
+    { service_name => 30 }
+  end
+
+  def service_name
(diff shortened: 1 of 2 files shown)


ntkathole:
ACK! looks good :+1: 

For 6.3,

```
Check whether all services are running using hammer ping:             [FAIL]
foreman_tasks resource(s) are failing.
Continue with step [Restart applicable services]?, [y(yes), n(no), q(quit)] y
Restart applicable services:                                                    

foreman-tasks

| All services stopped                                                          

foreman-tasks

\ All services started
```

For 6.4,
```

Check whether all services are running using hammer ping:             [FAIL]
foreman_tasks resource(s) are failing.
Continue with step [Restart applicable services]?, [y(yes), n(no), q(quit)] y 
Restart applicable services:                                                    

dynflowd

\ All services stopped                                                          

dynflowd

- All services started 
```

iNecas:
Thanks @mbacovsky and @ntkathole

PR #210, opened 2018-09-11
Fixes #24793 - Add check for running services that does not rely on hammer

mbacovsky:
To run check on all the services got complicated with remote databases.
Technically Candlepin and Foreman DBs can be two different services.
To address this I added new object representing the service called
SystemService. It has name of the service and priority and also component
Foreman or Candlepin in case of the postgresql. It alows us to keep
the service specific code isolated at one place. It is easy to extend
and we can tune the output of the individual actions separately.
Furthermore actions can return the exit codes so we can propagate
them correctly in the service commands.

The changes affected the service commands output but I tried to keep
it as close to the original state as possible. Currently the service
commands handle the service for each of the component separately.
E.g. `foreman-maintain service status --only postgresql` does check on
both the Candlepin and the Foreman DBs. Internally the Service::Status
procedure accept also SystemService object so can do status check only
for one component (this is not possible from CLI).

New check ServicesUp was added and it has no deps. It is always run
as part of upgrade and health check. If Katello is installed the restart
during the remediation step waits for hammer ping to finish.
Also if Katello is installed the original HammerPing check is perfomed too
as it has some additional value.


lib/foreman_maintain/utils/system_service.rb
```diff
@@ -0,0 +1,156 @@
...
+    end
+
+    private
+
+    def execute(action, options = {})
+      @sys.execute_with_status(command(action, options))
+    end
+  end
+
+  class RemoteDBService < SystemService
```
iNecas: The `RemoteDBService` name is misleading: when I was reading the code, my first impression was this represents only remote database (and was expecting `LocalDBService` counterpart).

What about `RemotableDBService`, or just `DBService`, and add some additoinal doc comment describing briefly the purpose.

definitions/features/service.rb
```diff
@@ -10,37 +10,16 @@ def handle_services(spinner, action, options = {})
...
-    service_list.select { |service| service_remote?(service) || service_exists?(service) }
-  end
-
-  def service_remote?(service)
-    (service == 'postgresql' && !feature(:instance).postgresql_local?) ||
-      (service == 'mongod' && feature(:instance).database_remote?(:mongo))
+    ForemanMaintain.available_features.map(&:services).
+      flatten(1).
+      sort.
+      uniq(&:to_s).
```
iNecas: This leads to foreman's posgres and canclepin's service being treated as separate, even though especially with local, they are the same service:

```
- starting postgresql (foreman)Redirecting to /bin/systemctl start postgresql.service

| starting postgresql (candlepin)Redirecting to /bin/systemctl start postgresql.service
```

Also, there is missing space before `Redirecting…` string

iNecas:
I like the direction. Some comments inline. Also inviting @ntkathole and @alda519 to the testing party :)

lib/foreman_maintain/utils/system_service.rb
mbacovsky: I was not sure about this concept either. I'm thinking of making it remote DB only and move the remote/local condition to the system_service factory.

definitions/features/service.rb
mbacovsky: This was intentional. I see in this case it could be confusing. I'll try to change the logic to use the component only for non-local DBs which should fix the situation while not breaking the partly-remote scenario.

mbacovsky:
@iNecas however the detector refresh may imply that evaluation of the confine blocks at the scenario compose time may not be a safe thing, right? :-/  Probably not worth 6.4 but should be addressed sooner then later.

iNecas:
Regarding the detector refresh: let's put it outside of this PR

definitions/features/service.rb
iNecas: yes, that would be good enough solution for me

mbacovsky:
@iNecas updated. RemoteDBService is remote DB only. SystemServices do not have component and duplicit components of the same name are not allowed. Detector untouched.
I couldn't reproduce the missing space, looking into it. Also there was some failing test on 1.8.7 which needs to be fixed.

ntkathole:
@mbacovsky 

- The output of service command doesn't look good. \n before Redirecting statement.
```

mongod, postgresql, qdrouterd, qpidd, squid, tomcat, pulp_streamer, pulp_resource_manager, pulp_celerybeat, smart_proxy_dynflow_core, pulp_workers, foreman-tasks, httpd, foreman-proxy
/ starting mongodRedirecting to /bin/systemctl start mongod.service             

- starting postgresqlRedirecting to /bin/systemctl start postgresql.service     

\ starting qdrouterdRedirecting to /bin/systemctl start qdrouterd.service       

| starting qpiddRedirecting to /bin/systemctl start qpidd.service               

| starting squidRedirecting to /bin/systemctl start squid.service               

| starting tomcatRedirecting to /bin/systemctl start tomcat.service             

/ starting pulp_streamerRedirecting to /bin/systemctl start pulp_streamer.service

```
Second issue is when foreman-tasks service is stopped and online backup started:
```
# ./bin/foreman-maintain service stop --only foreman-tasks
# ./bin/foreman-maintain backup online /tmp
Starting backup: 2018-09-14 05:48:24 -0400
Make sure Foreman DB is up: 
Make sure Mongo DB is up: 
| Checking connection to the Mongo DB                                 [FAIL]    
undefined method `keys' for [SystemService(mongod [5])]:Array
undefined method `keys' for [SystemService(mongod [5])]:Array
```

ntkathole:
Also performing service restart when services are stopped, now gives 
```
# ./bin/foreman-maintain service stop
# ./bin/foreman-maintain service restart
Setup hammer: 
Hammer admin password: 
                                                                      [FAIL]
Hammer configuration failed: Is the admin password correct? (it was stored in /root/foreman_maintain/config/foreman-maintain-hammer.yml)Is the server down?
Scenario [preparation steps required to run the next scenarios] failed.
```

mbacovsky:
@ntkathole are you sure this is a new thing introduced by my patch? I can reproduce it on master.

ntkathole:
> @ntkathole are you sure this is a new thing introduced by my patch? I can reproduce it on master.

I thought as it is handled on capsule with this PR...I expect service restart wont look for hammer on server as well.

mbacovsky:
@ntkathole I fixed the hammer issue with check if katello is not installed. In that case it makes no sense to query hammer ping and that is why it works on capsule. On server this is another bug when we try to setup and check hammer connection when the server is down (https://projects.theforeman.org/issues/24784). It is expected to fail but should be communicated better. I'll try to postpone the setup when the httpd service is down till when it is really needed and services can be up. But I willl do that in separate PR. I can do it in separate commit in this PR though if you prefer to test it once.

@iNecas would it be a reasonable way to speed up the testing to add commits for the other related issues in this PR? (hammer setup on server down, status exit code)

mbacovsky:
@iNecas, @ntkathole : I'm curious where these messages come from while I don't see it on my system:

> | starting squidRedirecting to /bin/systemctl start squid.service               
> | starting tomcatRedirecting to /bin/systemctl start tomcat.service  

It suggests the SysV `service <name> start` is used but I'm not sure how. Do you have `service-wait` on your system? Could that be the cause? How do I install it? It is not no my clean snap install. I checked service-wait sources and it should call `systemctl`anyway. 

I've fixed the output. Is it okay to display the messages or should it be hidden?

mbacovsky:
@ntkathole the mongo failures were fixed, good catch.

mbacovsky:
Sorry for last minute reorganization of the code but I hope this way it is more maintainable and extensible in the furture

lib/foreman_maintain/utils/service.rb
```diff
@@ -0,0 +1,14 @@
+require 'foreman_maintain/utils/service/abstract'
+require 'foreman_maintain/utils/service/systemd'
+require 'foreman_maintain/utils/service/remote_db'
+
+module ForemanMaintain::Utils
+  def self.system_service(name, priority, options = {})
+    db_feature = options.fetch(:db_feature, nil)
+    if name =~ /^(postgresql|.*mongod)$/ && db_feature && db_feature.local? == false
```
iNecas: Nitpick: `!db_feature.local?`. Checking on `== false` is not covering the potential case, where `local?` would return nil, which is also considered false. But the main reason to fix are the conventions.

mbacovsky:
@iNecas updated.
@johnpmitsch could you please take a look? I guess you are most familiar with the feature and related caveats.

iNecas:
@mbacovsky the messages about the redirects are coming from `service-wait` (I was testing on sat 6.3).

Regarding the other changes, I would probably prefer separate PRs for them: if you already have something and you don't want to wait, you can open a separate PR and include the unmerged commits (we will get rid of them once rebased to master after this is merged).

johnpmitsch:
@mbacovsky yes, sorry for the delay, I will look today

lib/foreman_maintain/utils/service.rb
```diff
@@ -0,0 +1,14 @@
+require 'foreman_maintain/utils/service/abstract'
+require 'foreman_maintain/utils/service/systemd'
+require 'foreman_maintain/utils/service/remote_db'
+
+module ForemanMaintain::Utils
+  def self.system_service(name, priority, options = {})
+    db_feature = options.fetch(:db_feature, nil)
+    if name =~ /^(postgresql|.*mongod)$/ && db_feature && !db_feature.local?
```
johnpmitsch: Does it make sense to use regex here or could this be an option that is passed in?

mbacovsky: @johnpmitsch could you please explain how it should look like and where the option should go from? I'm not sure I understand.

johnpmitsch: i.e. (untested)
```ruby
      system_service('postgresql', 10, :component => 'foreman',
                                       :db_feature => feature(:foreman_database),
                                       :db_service => true,
```

```ruby
module ForemanMaintain::Utils
  def self.system_service(name, priority, options = {})
    db_feature = options.fetch(:db_feature, nil)
    db_service = options.fetch(:db_service, false)
    if db_service && db_feature && !db_feature.local?
      Service::RemoteDB.new(name, priority, options)
    else
      Service::Systemd.new(name, priority, options)
    end
  end
end
```
I'm not sure if I'm missing some other reason for the pattern matching, but it seems like if there is an option hash we should use that to set if its a postgres or mongo db related service. The pattern matching will probably be fine, but we don't know what kind of services will be introduced in the future and what they will be called.

mbacovsky: Thanks a lot for the explanation, now I see what you mean. This make sense. I wanted the factory to be black box where all the logic around choosing the right Service class is concentrated. I'm still not sure if telling the factory what the service should be is a good thing. I'll have to think about it.
Any opinion @iNecas?

iNecas: What about something like:

```ruby
if db_feature && db_feature.respond_to?(:local?) && !db_feature.local?
```

mbacovsky: @iNecas okay, lets give it a try. I'll update.

mbacovsky: @iNecas, @johnpmitsch updated, please check.

iNecas:
Thanks @mbacovsky @ntkathole and @johnpmitsch

mbacovsky:
Thanks for reviews to everybody involved! :rainbow:

PR #212, opened 2018-09-13
Fixes #24923 - Add check for duplicate puppet classes

xprazak2:


xprazak2:
Base on the discussion in https://github.com/theforeman/foreman/pull/6019, we decided to ask users to remove the duplicates manually if there are any before we add uniqueness to the db column.

iNecas:
I would ask for separate issue for foreman-maintain, to make it easier to track the releases.

definitions/checks/foreman/puppet_class_duplicates.rb
```diff
@@ -0,0 +1,45 @@
+module Checks
+  module Foreman
+    class PuppetClassDuplicates < ForemanMaintain::Check
+      metadata do
+        label :puppet_class_duplicates
+        for_feature :foreman_database
+        description 'Check for duplicate Puppet class records'
+        tags :pre_upgrade
```
iNecas: It would be good to wait for the core PR to be merged, and then add a confine block here to enable this only on versions < than the version that has the fix: once the unique constraint is there, ther is no point on running this check anymore

iNecas:
Nice work in general. Would it be possible to provide some production reproducible steps that @ntkathole could use for verifying in production?

definitions/checks/foreman/puppet_class_duplicates.rb
```diff
@@ -0,0 +1,48 @@
...
+        confine do
+          feature(:foreman_database) && feature(:foreman_database).check_max_version('foreman', '1.20')
+        end
```
xprazak2: I added a confine to restrict the checks only for version 1.20 and lower.

xprazak2:
https://projects.theforeman.org/issues/24633 contains the full reproducing steps on how to add duplicate puppet classes through foreman/katello. But it would be much quicker just to add records with duplicate names directly into the `puppetclasses` table in db for the purposes of testing this.

definitions/checks/foreman/puppet_class_duplicates.rb
iNecas: The `check_max_version` should be available directly in current context., no need to call it on `feautre(:foreman_database)`. Also, I guess the check should be only on `1.19` and below (provided the fix will be in `1.20`?

iNecas: On the other hand, it probably doesn't hurt, if we had is on 1.20 as well

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -188,6 +193,19 @@ def find_dir_containing_file(directory, target)
...
+      private
+
+      def check_version(name)
+        current_version = package_version(name)
+        if current_version
+          yield current_version if block_given?
+        end
+      end
+
+      def version_cmp(version, other_version)
```
iNecas: For readability purposes, actually I think this method is redundant, and just keeping `>=` in the `check_{min,max}_version` methods would be easier to understand.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -188,6 +193,19 @@ def find_dir_containing_file(directory, target)
         end
         result
       end
+
+      private
+
+      def check_version(name)
+        current_version = package_version(name)
+        if current_version
+          yield current_version if block_given?
```
iNecas: Since there is no reason to call `check_version` without a block, I would be ok not checking on presence of the block here.

definitions/checks/foreman/puppet_class_duplicates.rb
ares: it's already merged and present will be present in 1.20, does the confine block below address it?

ares:
I opened a PR which addresses comments at https://github.com/theforeman/foreman_maintain/pull/215, please continue the review there, so we don't have to wait until @xprazak2 is back. I'm fine with squashing my changes if needed.

definitions/checks/foreman/puppet_class_duplicates.rb
iNecas: I think it actually could be `check_max_version('foreman', '1.19')`, as in `1.20` there would already be the index present

xprazak2:
Closing in favour of #215

PR #213, opened 2018-09-14
Fixes #22229 - Check Available Disk Space

upadhyeammit:
This adds the root partition disk space check, I have taken 4GB as threshold here.

### DIFF
--- a/definitions/checks/disk/available_space.rb
+++ b/definitions/checks/disk/available_space.rb
@@ -1,0 +1,23 @@
+module Checks
+  module Disk
+    class AvailableSpace < ForemanMaintain::Check
+      metadata do
+        label :available_space
+        description 'Check to make sure root(/) partition has enough space'
+        tags :pre_upgrade
+      end
+
+      MIN_SPACE_IN_MB = 4096
+
+      def run
+        assert(enough_space?, "System has less than #{MIN_SPACE_IN_MB / 1024}GB space available"\
+              ' on root(/) partition')
+      end
+
+      def enough_space?
+        io_obj = ForemanMaintain::Utils::Disk::IODevice.new('/')
+        io_obj.available_space > MIN_SPACE_IN_MB
+      end
+    end
+  end
+end
--- a/lib/foreman_maintain/utils/disk/io_device.rb
+++ b/lib/foreman_maintain/utils/disk/io_device.rb
@@ -18,6 +18,10 @@ module ForemanMaintain
           @unit ||= 'MB/sec'
         end
 
+        def available_space
+          convert_kb_to_mb(execute!("df #{dir}|awk {'print $4'}|tail -1").to_i)
+        end
+
         private
 
         # In fio command, --direct option bypass the cache page


definitions/checks/disk/available_space.rb
```diff
@@ -0,0 +1,22 @@
...
+      metadata do
+        label :available_space
+        description 'Check if root(/) partition has enough space to proceed with upgrade'
+        tags :pre_upgrade
+      end
+
+      MIN_SPACE_IN_MB = 4096
+
+      def run
+        assert(!enough_space?, 'System has less than 4GB space available on root(/) partition')
```
mbacovsky: The `4GB` should be computed from `MIN_SPACE_IN_MB`

lib/foreman_maintain/utils/disk/io_device.rb
```diff
@@ -32,6 +32,12 @@ def fio
         def convert_kb_to_mb(val)
           val / 1024
         end
+
+        public
```
mbacovsky: Could the method be moved up closer to other public methods?

definitions/checks/disk/available_space.rb
mbacovsky: Or easier from `MIN_SPACE_IN_GB`

definitions/checks/disk/available_space.rb
```diff
@@ -0,0 +1,22 @@
+module Checks
+  module Disk
+    class AvailableSpace < ForemanMaintain::Check
+      metadata do
+        label :available_space
+        description 'Check if root(/) partition has enough space to proceed with upgrade'
+        tags :pre_upgrade
```
mbacovsky: The original Issue requests this check also in health check which runs chcks tagged :default

definitions/checks/disk/available_space.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+
+      MIN_SPACE_IN_MB = 4096
+
+      def run
+        assert(!enough_space?, "System has less than #{MIN_SPACE_IN_MB / 1024}GB space available"\
+              ' on root(/) partition')
+      end
+
+      def enough_space?
```
kgaikwad: By name of the method 'enough_space?', anyone expects that this method should return true when there is enough same.
Currently, it is returning nil for positive scenario. 
Please make a correction to method and call `enough_space?` in assert.

definitions/checks/disk/available_space.rb
```diff
@@ -0,0 +1,23 @@
...
+        io_obj = ForemanMaintain::Utils::Disk::IODevice.new('/')
+        return true if io_obj.available_space < MIN_SPACE_IN_MB
```
kgaikwad: `?` mark at the end of method name indicates that a method returns a `boolean` value.
Please modify method so that it returns only boolean value.

upadhyeammit:
@kgaikwad changes done as requested. @mbacovsky I kept this for pre_upgrade only as 4GB space assumption has been calculated for upgrade scenario only https://bugzilla.redhat.com/show_bug.cgi?id=1527171#c3 . Thank You!

kgaikwad:
Codewise looks good to me.

mbacovsky:
@upadhyeammit thanks for the updates!

PR #214, opened 2018-09-14
Fixes #24942 - set UTF-8 explicitly

iNecas:
To make sure everything works as usual even with LC_ALL=C

### DIFF
--- a/bin/foreman-maintain
+++ b/bin/foreman-maintain
@@ -1,5 +1,8 @@
 #!/usr/bin/env ruby
 
+# Ensure we use UTF-8 by defualt, even when run with LC_ALL=C
+Encoding.default_internal = Encoding.default_external = Encoding::UTF_8
+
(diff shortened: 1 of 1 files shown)


mbacovsky:
:+1: Thanks, @iNecas! Codewise it looks reasonable and it fixed the issue for me. Should I merge it or it needs @ntkathole's ack?

ntkathole:
@iNecas I see respective bugzilla is closed just...do we still need this change?

iNecas:
Let's see how that get resolved. On the other hand, I wonder if there might be some users that would actually hae this locale as default… I think it's still good to be on the safe side here.

mbacovsky:
Regardless on the BZ I think it is reasonable to have this in upstream. Is there any scenario where this could be harmful?

iNecas:
I can't think of any: it's just reinforcing the defaults we rely on anyway.

mbacovsky:
Thanks @iNecas, @ntkathole!

PR #215, opened 2018-09-18
Fixes #24923 - Add check for duplicate puppet classes

ares:


iNecas:
Rubocop is not happy.

ares:
Rubocop now jumps up and down overdosed with happiness

definitions/checks/foreman/puppet_class_duplicates.rb
```diff
@@ -0,0 +1,49 @@
...
+  module Foreman
+    class PuppetClassDuplicates < ForemanMaintain::Check
+      metadata do
+        label :puppet_class_duplicates
+        for_feature :foreman_database
+        description 'Check for duplicate Puppet class records'
+        tags :pre_upgrade
+        confine do
+          feature(:foreman_database) &&
+            feature(:foreman_database).check_max_version('foreman', '1.20')
```
iNecas: actually, this confine block could be eliminated to:

```
confine do
  check_max_version('foreman', '1.20')
end
```

iNecas: The `for_feature` already implies that the feature is there, and the `check_max_version` should be available in this context already

ares:
updated

iNecas:
Tested with:

```
$ for x in {1,2,3}; do sudo su postgres - -c 'psql -d foreman -c "INSERT INTO puppetclasses (name) VALUES ('\''myclass'\'');"'; done

$ foreman-maintain health check --label puppet_class_duplicates
Check for duplicate Puppet class records:                             [FAIL]
There are 1 Puppet classes with duplicities:
myclass - 6
Please head over to Configure -> Classes and make sure there is only 1 Puppet class record for each name.

$ sudo su postgres - -c 'psql -d foreman -c "DELETE FROM puppetclasses WHERE name = ('\''myclass'\'');"';
# sudo su postgres - -c 'psql -d foreman -c "INSERT INTO puppetclasses (name) VALUES ('\''myclass'\'');"'

$ foreman-maintain health check --label puppet_class_duplicates
Check for duplicate Puppet class records:                             [OK]

```

PR #216, opened 2018-09-18
Fixes #24891 - return non-zero exit code when services are down

mbacovsky:


mbacovsky:
@kgaikwad thanks for review, updated

iNecas:
Thanks @mbacovsky

PR #217, opened 2018-09-19
Fixes #24784 - Do not pre-setup Hammer when httpd is down

mbacovsky:
Usualy when Hammer is needed we try to pre-setup it
at the beginning of the scenario so that the user is not asked for
credentials in the middle of execution. If the httpd is down the setup
fail and the whole scenario stops. This patch allow skipping of
the pre-setup and do it when Hammer is really needed.

### DIFF
--- a/definitions/procedures/hammer_setup.rb
+++ b/definitions/procedures/hammer_setup.rb
@@ -5,8 +5,13 @@ class Procedures::HammerSetup < ForemanMaintain::Procedure
   end
 
   def run
-    result = feature(:hammer).setup_admin_access
-    logger.info 'Hammer was configured successfully.' if result
+    if feature(:foreman_server) && ForemanMaintain::Utils.system_service('httpd', 30).running?
+      puts 'Configuring Hammer CLI...'
+      result = feature(:hammer).setup_admin_access
+      logger.info 'Hammer was configured successfully.' if result
+    else
+      skip("#{feature(:instance).product_name} server is not running. Hammer can't be setup now.")
+    end
   end
 
   def necessary?
--- a/definitions/checks/foreman_tasks/not_paused.rb
+++ b/definitions/checks/foreman_tasks/not_paused.rb
@@ -5,6 +5,7 @@ module Checks::ForemanTasks
       for_feature :foreman_tasks
       description 'Check for paused tasks'
       tags :default
+      after :services_up, :hammer_ping
     end
 
     def run
--- a/definitions/checks/hammer_ping.rb
+++ b/definitions/checks/hammer_ping.rb
@@ -5,6 +5,7 @@ class Checks::HammerPing < ForemanMaintain::Check
     for_feature :hammer
     description 'Check whether all services are running using hammer ping'
     tags :default
+    after :services_up
 
     confine do
       feature(:katello)
--- a/test/definitions/procedures/hammer_setup_test.rb
+++ b/test/definitions/procedures/hammer_setup_test.rb
@@ -34,6 +34,8 @@ describe Procedures::HammerSetup do
 
   context 'there is a default configuration with valid credentials' do
     it 'uses the default config and sets the feature' do
+      assume_feature_present(:foreman_server)
+      assume_service_running('httpd')
       hammer_ins.stubs(:_check_connection).returns(true)
       result = run_procedure(subject)
       assert result.success?, 'the procedure was expected to succeed'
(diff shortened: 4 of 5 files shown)


mbacovsky:
I bundled the fix with another trivial commit that makes the testing easier and health check usable.
I'd like to know your opinion on how we should behave with the Hammer setup. options I see are:

1. just skip the procedure if we are on the server and httpd is down. Hammer will be setup when used first with a chance the httpd is up at that time
2. make a check from the setup procedure and let user confirm she wants to continue and setup hammer later or restart httpd and retry or abort the scenario and solve the problem manually. 

I implemented the first option as it was much easier but if we agree the second is worth it I can update the PR. Is there any other worthy option I missed?

mbacovsky:
I've fixed the failing test and added some mocking capabilities fro services that are in line with how we mock features.

@kgaikwad thanks for review. Any opinion on how it should behave and if approach #1 is sufficient? (please see my comment above)

kgaikwad:
@mbacovsky,

I like the approach #1.
In addition to approach #1, it would be good if we provide option for end user to start/restart httpd service. I could not think of how retrying of a failed scenario/procedure will happen after running a check. Does it complicate the existing logic of retrying a failed scenario/procedure?

mbacovsky:
@kgaikwad, IMO we would have to turn the procedure into check and add the remediation step which will result in approach #2. I have no problem going this route. 

I didn't checked but I don't think it would play well if we prepend extra preparation step (check) to the existing procedure and call it both in preparation phase.

iNecas:
Have not tested, but the changes look reasonable

mbacovsky:
I'll give approach no.2 a try to see what is involved. If it looks reasonable, I'll update in separate commit.

mbacovsky:
@iNecas, @kgaikwad I've added a commit with Hammer setup detecting the server is down and offering remediation. It works quite nice if you choose to restart httpd, With other flows the UX is not great.

Happy path:
```
# systemctl stop httpd
Setup hammer:                                                         [WARNING]
Satellite server is not runnig. Hammer can't be setup.
You can either start httpd service and retry
or skip and postpone hammer setup for later when services may be up
or quit and resolve the issue manualy.
Continue with step [Start applicable services]?, [y(yes), n(no), q(quit)] y
Start applicable services:                                                      

httpd
- starting httpd                                                                

\ All services started                                                [OK]      
Setup hammer: 
\ Configuring Hammer CLI                                              [OK]      


Check for verifying syntax for ISP DHCP configurations:               [SKIPPED]
DHCP feature is not enabled
```

I'm not sure what to start to do the hammer setup check when all the services are down. Should we start everything and not only httpd? Another problem is the spinner I've added to make clear that the script does not wait for a user input. It clashes with interactive password input. When the setup is waiting for rails to restart without the spinner `Setup hammer:` shines on the screen as it was waiting for user input.

kgaikwad:
Thank you @mbacovsky for considering scenario #2 and adding this check.

Currently, we are using hammer commands inside sync_plans, foreman_task and restart services.
If other services like foreman-tasks down then it may affect the execution, right?
+1 for restarting all services than restarting httpd only.

But I don't want to block this PR as original issue has already been addressed in your previous changes.
Above change i.e restarting all services is a quite big step in current flow of hammer setup and I am not sure whether its worth doing this at this point, so I would suggest to have this change in separate PR. 

@iNecas, I would like to hear your thoughts on this.

mbacovsky:
I updated the PR to solve the issue with spinner and restart. The first approach gives cleaner UX but won't let you restart the services. The second approach lets you restart but if you want NOT to do that the instructions are not very clear. We need to improve our failure/remediation steps handling UI.
I'll wait for @iNecas opinion but in current situation I slightly prefer the approach one.

kgaikwad:
@mbacovsky,
Could you please updated PR as per first approach.
After @ntkathole's ack, we can merge this PR :+1:

mbacovsky:
@iNecas, @kgaikwad, @ntkathole updated to approach 1. 
Please do not squash during the merge as we address two different issues.

mbacovsky:
I updated the commit message after I found out wrong issue was referenced. Updated the links in BZ and upstream tracker.

definitions/procedures/hammer_setup.rb
```diff
@@ -5,8 +5,13 @@ class Procedures::HammerSetup < ForemanMaintain::Procedure
...
 
   def run
-    result = feature(:hammer).setup_admin_access
-    logger.info 'Hammer was configured successfully.' if result
+    if feature(:foreman_server) && ForemanMaintain::Utils.system_service('httpd', 30).running?
+      puts 'Configuring Hammer CLI...'
+      result = feature(:hammer).setup_admin_access
+      logger.info 'Hammer was configured successfully.' if result
+    else
+      skip("#{feature(:instance).product_name} server is not runnig. Hammer can't be setup now.")
```
ntkathole: typo s/runnig/running

mbacovsky:
@thanks a lot for your testing @ntkathole, typo fixed.

iNecas:
Thanks @mbacovsky, @kgaikwad and @ntkathole, I'm going to release a new version with all the recent changes.

iNecas:
0.2.10 pushed to rubygems, packaging PR is here https://github.com/theforeman/foreman-packaging/pull/3033

PR #219, opened 2018-09-19
Fixes #24983 - set MAX_TERMINATION to 1

lzap:
I was thinking if it's safer to kill just one by one by default. There is no guarantee that Katello is able to restart after the GRACEFUL_TIMEOUT period.


kgaikwad:
@lzap, 
Please change issue id in commit message to [24983](https://projects.theforeman.org/issues/24983).

iNecas:
Thanks @lzap

PR #223, opened 2018-09-24
Fixes #25022 - skip repos validation if subscribed using act_key

kgaikwad:


definitions/checks/repositories/validate.rb
```diff
@@ -18,19 +18,29 @@ class Validate < ForemanMaintain::Check
...
-        unless absent_repos.empty?
-          spinner.update('Some repositories missing, calling `subscription-manager refresh`')
-          feature(:downstream).rhsm_refresh
-          absent_repos = feature(:downstream).absent_repos(@version)
-        end
-        unless absent_repos.empty?
-          fail!(
-            "Following repositories are not available on your system: #{absent_repos.join(', ')}"
-          )
+      if feature(:instance).subscribed_using_activationkey?
```
mbacovsky: Why is this check in `:instance` and not in `:downstream` feature?

iNecas: Yes, this could be downstream only. I'm ok with moving that

kgaikwad: @mbacovsky, updated PR with a suggested change.

mbacovsky:
Thanks @kgaikwad! The failing tests were caused by some Travis hiccup. I restarted the test and everything is green now.

PR #224, opened 2018-09-24
Fixes #25019 - GA repositories enabled for 6.4

kgaikwad:


mccun934:
ACK - Looks good, same as we did for 6.3

yes we need to refactor this, but we are short on time to tie this in now.

iNecas:
Thanks @kgaikwad

PR #225, opened 2018-09-26
Fixes #24858 - Fixed formatting of katello-service params

mbacovsky:
This fixes the formatting of --only and --exclude parameters
when passed to katello-service.

Before: katello-service start --only ["postgresql", "squid"]
After:  katello-service start --only postgresql,squid


iNecas:
Thanks @mbacovsky

PR #226, opened 2018-10-18
Fixes #25032 - Setup hammer on system with defaults set

mbacovsky:
When organization is set in hammer defaults foreman-maintain
is not able to setup hammer properly. The original cause is
API controller fails for architecture list with org argument set.
As a workaround in foreman-maintain we use different API endpoint
to confirm hammer auth is properly set.

### DIFF
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -212,6 +212,6 @@ class Features::Hammer < ForemanMaintain::Feature
   end
 
   def _check_connection
-    `#{command_base} architecture list 2>&1` && $CHILD_STATUS.exitstatus == 0
+    `#{command_base} user list --per-page 1 2>&1` && $CHILD_STATUS.exitstatus == 0
   end
 end


iNecas:
Tested and works well with org set up in defaults. It would be slightly better, if hammer commands used from f-m would be not influenced by the user defaults, as there might be more potential issues with other hammer commands alter. But as an immediate fix, I'm ok with this one.

mbacovsky:
@iNecas, I agree. I'll create new BZ for the API, one for Hammer to get `--ignore-defaults` or something similar and one to prevent f-m hammer commands from using the defaults. For a start we can use `--organization-id NIL` to prevent defaults from being used.

Thanks for the review!

iNecas:
So I think we might split this issue into two: a failure for a valid org_id (high serverity, that this PR is fixing) and failure for non-valid org_id, lower severity, out of scope of this PR

mbacovsky:
If we want to be sure we can add --organization NIL --location NIL to the user list. That will prevent the defaults from applying.Should I prepeare another PR?

iNecas:
@mbacovsky would that work for older satellite releases? I don't think so

mbacovsky:
@iNecas it is hammer 0.12 feature so maybe Satellite 6.3 but surely not 6.2. The NIL params should be conditioned by hammer version. Hammer feature could add that automatically.

iNecas:
I would rather like to see a long-term solution, that would avoid default completely, when used from foreman-maintain

PR #228, opened 2018-11-01
Fixes #23563 - check and correct hammer cli_config.yml

upadhyeammit:
This adds the check for hammer configuration, it checks for :host parameter and if its set as localhost then check fails. I have used hammer feature itself to get :host parameter value and compared that with localhost

### DIFF
--- a/definitions/checks/check_hammer_config.rb
+++ b/definitions/checks/check_hammer_config.rb
@@ -1,0 +1,30 @@
+require 'uri'
+class Checks::CheckHammerConfig < ForemanMaintain::Check
+  metadata do
+    label :check_hammer_config
+    description 'Check if hammer configuration file is using FQDN of system'
+    tags :pre_upgrade
+    confine do
+      feature(:downstream)
+    end
+  end
+
+  def run
+    with_spinner('Checking hostname of hammer configuration') do
+      assert(compare_hostname?, 'The :host parameter configured for hammer should not be'\
+        " 'localhost'")
+    end
+  end
+
+  def hammer_host_url
+    if feature(:hammer).configuration[:foreman][:host]
+      URI.parse(feature(:hammer).configuration[:foreman][:host]).host
+    end
+  end
+
+  def compare_hostname?
+    return hammer_host_url != 'localhost' unless hammer_host_url.nil?
+
+    true
+  end
+end
--- a/definitions/checks/check_hammer_config.rb
+++ b/definitions/checks/check_hammer_config.rb
@@ -3,16 +3,19 @@ class Checks::CheckHammerConfig < ForemanMaintain::Check
   metadata do
     label :check_hammer_config
     description 'Check if hammer configuration file is using FQDN of system'
-    tags :pre_upgrade
+    tags :post_upgrade
     confine do
       feature(:downstream)
     end
   end
 
   def run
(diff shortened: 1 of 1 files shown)


mbacovsky:
@upadhyeammit, I'm not sure how to proceed with this PR. I tried to reproduce the issue and was not able to do so as it was already fixed by  https://github.com/theforeman/foreman_maintain/pull/165. The original issue should be closed as a duplicate of https://bugzilla.redhat.com/show_bug.cgi?id=1547912.

I know we discussed the added value - the check for localhost in hammer setup, but as I'm testing it I'm concerned about the false positive failures. 

If the user doesn't use hammer as a root user but with his ordinary account his setup in `/root` having `localhost` is perfectly okay because nobody is using it. If he start using hammer as a root hammer will complain and guide to configure properly. The problem is we prevent upgrade until the unused config is fixed. Not sure how common this scenario is but should be recommended practice. Another minor flase positive case is when the user has custom config and run hammer with `-c custom_config.yaml`.

A few thoughts on how to improve this feature if we decide to keep it:
- make it a post upgrade check
- make it a warning only
- make better instructions on how to fix it - make sure :host: <server_url> is included in your ~/.hammer/cli.modules.d/foreman.yml or /etc/hammer/cli.modules.d/foreman.yml or something similar.

@upadhyeammit what do you think?

upadhyeammit:
> @upadhyeammit, I'm not sure how to proceed with this PR. I tried to reproduce the issue and was not able to do so as it was already fixed by #165. The original issue should be closed as a duplicate of https://bugzilla.redhat.com/show_bug.cgi?id=1547912.
> 
> I know we discussed the added value - the check for localhost in hammer setup, but as I'm testing it I'm concerned about the false positive failures.
> 
> If the user doesn't use hammer as a root user but with his ordinary account his setup in `/root` having `localhost` is perfectly okay because nobody is using it. If he start using hammer as a root hammer will complain and guide to configure properly. The problem is we prevent upgrade until the unused config is fixed. Not sure how common this scenario is but should be recommended practice. Another minor flase positive case is when the user has custom config and run hammer with `-c custom_config.yaml`.
> 
> A few thoughts on how to improve this feature if we decide to keep it:
> 
>     * make it a post upgrade check
> 
>     * make it a warning only
> 
>     * make better instructions on how to fix it - make sure :host: <server_url> is included in your ~/.hammer/cli.modules.d/foreman.yml or /etc/hammer/cli.modules.d/foreman.yml or something similar.
> 
> 
> @upadhyeammit what do you think?

Hello, if I see https://github.com/theforeman/foreman_maintain/pull/165 then yes we do handle all of the false config scenarios and then if we find issue with config we correct it and we use it as custom config for that specific hammer run. This is correct we should not touch the user config. Here https://projects.theforeman.org/issues/23563 is for running hammer standalone, even with root user if I run hammer with below config it fails with SSL errors,

:foreman:
  :username: 'admin'
  :password: 'admin'
  :host: 'https://localhost/'

As per this we either need to fix the host in hammer feature itself, which we already discussed and we are not going to do. Next is host attribute could be at many locations, if we specifically need to print that abc file has incorrect host attribute then it will be like adding complexity which wont benefit much. Accordingly I think I agree with below things,

We put this in post_run
We put it as warning message
We embed default file locations in warning message so user can fix it

PR #229, opened 2018-11-07
Fixes #25329 - add 6.5 target version

iNecas:
For development of next Sat release.


mbacovsky:
I did some testing and it is good to go.

PR #231, opened 2018-11-13
Fixes #25433 - Mark data dir for later detection

mbacovsky:
Original detection based on existing files was error prone due to the nature of the files.
The procedure was changed to create distinct mark/file before taking the snapshot and removing it after to not pollute the fs with our helper files.

### DIFF
--- a/definitions/procedures/backup/snapshot/mount_pulp.rb
+++ b/definitions/procedures/backup/snapshot/mount_pulp.rb
@@ -13,10 +13,12 @@ module Procedures::Backup
       def run
         skip if @skip
         with_spinner('Creating snapshot of Pulp') do |spinner|
-          lv_info = get_lv_info(feature(:pulp).data_dir)
-          create_lv_snapshot('pulp-snap', @block_size, lv_info[0])
-          spinner.update("Mounting snapshot of Pulp on #{mount_location('pulp')}")
-          mount_snapshot('pulp', lv_info[1])
+          feature(:pulp).with_marked_directory(feature(:pulp).data_dir) do
+            lv_info = get_lv_info(feature(:pulp).data_dir)
+            create_lv_snapshot('pulp-snap', @block_size, lv_info[0])
+            spinner.update("Mounting snapshot of Pulp on #{mount_location('pulp')}")
+            mount_snapshot('pulp', lv_info[1])
+          end
         end
       end
     end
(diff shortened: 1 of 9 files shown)


definitions/procedures/backup/snapshot/mount_pulp.rb
```diff
@@ -13,10 +13,12 @@ class MountPulp < MountBase
       def run
         skip if @skip
         with_spinner('Creating snapshot of Pulp') do |spinner|
+          feature(:pulp).mark_base_directory
```
iNecas: Having a block-versoin of this method would proabely feel a bit better, so that we don't forget to unmark the file later (and we should probably do that in `ensure` block). This way, we could actually check when calling `find_base_directory(mount_point)`, that we're actually inside the block (by using some instance variable inside the feature). I'm not sure this is the best explanation of my suggestion - we can discuss it offline in more details if needed.

mbacovsky: I was thinking of a block-version too so I'll update. About the check I'm not sure because the find is happening on the snapshot much later and the original is no longer neither marked nor inside the block.

iNecas: Can't we leverage the fact that both methods are defined on the same object?

mbacovsky:
@iNecas updated, how about this version?

lib/foreman_maintain/concerns/directory_marker.rb
```diff
@@ -0,0 +1,35 @@
...
+      def unmark_directory(directory)
+        filename = mark_file(directory)
+        File.delete(filename) if File.exist?(filename)
+      end
+
+      def mark_directory(directory)
+        File.open(mark_file(directory), 'a') {}
+      end
+
+      def mark_file(directory)
```
iNecas: Nitpick: it was a bit confusing for me to see `mark_file` and `mark_directory`, as from the name, they would look like they do the same, but one with dir, the other with file. `mark_file_path` would probably address this problem.

iNecas:
I like it. One nitpick left. I would like to invite @alda519 for a testing party, if possible, as he has been doing some testing in the past for this feature.

mbacovsky:
@iNecas, thanks for the commants, updated.

alda519:
ACK, tested and did not hit any issues.

iNecas:
Thanks @mbacovsky and @alda519

PR #233, opened 2018-11-16
Fixes #23430 - checks if TMOUT is set or not

upadhyeammit:
The TMOUT variable can cause session to timeout and to avoid happening that at time of upgrade this checks if TMOUT environment variable is set or not ? if set check fails.


definitions/checks/check_tmout.rb
```diff
@@ -0,0 +1,16 @@
...
+    tags :pre_upgrade
+  end
+
+  def run
+    assert(tmout_set?, 'The TMOUT environment variable is set on system.'\
+          " Run 'unset TMOUT' command to unset this variable.")
+  end
+
+  def tmout_set?
+    execute('printenv TMOUT').empty?
```
mbacovsky: Is there any difference from using ENV['TMOUT']?
Is the condition correct? empty? returns true when the var is not set, right?
Also consider TMOUT=0 as valid way to turn off auto-logout.

mbacovsky:
So I searched a bit and root can have read-only TMOUT too. 
Would it make sense to call installer like
`exec env TMOUT=0 bash -c <installer>` or it would logout regardless on the inner bash TMOUT?

mbacovsky:
Any insight on this @sean797?

upadhyeammit:
> So I searched a bit and root can have read-only TMOUT too.
> Would it make sense to call installer like
> `exec env TMOUT=0 bash -c <installer>` or it would logout regardless on the inner bash TMOUT?

I feel like then we should append TMOUT=0 for foreman-maintain itself ? I am not sure about if we execute foreman-maintain and then installer from that will both of the processes be in same bash environment ? If they are not then foreman-maintain will timeout.

mbacovsky:
@upadhyeammit yes, I was afraid of that. So just the message and exit. May be useful to include the actual timeout to the message.

upadhyeammit:
> @upadhyeammit yes, I was afraid of that. So just the message and exit. May be useful to include the actual timeout to the message.

Added ENV and added actual timeout to the message. Request to check.

Thank You!

PR #234, opened 2018-11-20
Fixes #25504 - Backup databases without password

mbacovsky:
Backup now should proceed without hard failures
om special setups such as server development
and databases with no passwords.


ntkathole:
@mbacovsky I tested this on same instance, but facing error

```
# ./bin/foreman-maintain backup online /tmp/
Starting backup: 2018-11-22 03:33:41 -0500
Make sure Foreman DB is up: 
/ Checking connection to the Foreman DB                               [FAIL]    
Foreman DB is not responding. It needs to be up and running to perform the following steps
Continue with step [Start applicable services]?, [y(yes), n(no), q(quit)] y
Start applicable services:                                                      
                                   [FAIL]
```

mbacovsky:
@ntkathole, I'm not sure how to behave here. We found you have foreman package installed, there is foreman DB configuration found but the DB is not present. So the message is kind of correct and gives a clue where to check for the cause. This is invalid setup and I guess what you see is best we can do about it.

iNecas:
@ntkathole anything to add regarding the reproducer setup? I agree with @mbacovsky that this seems like issue unrelated to this BZ

iNecas:
@ntkathole ping

ntkathole:
Tested on latest satellite setup and it works fine. @iNecas @mbacovsky yea showing message explains the cause. ACK.

iNecas:
Thanks @mbacovsky

iNecas:
and @ntkathole :)

PR #235, opened 2018-11-22
Fixes #25535 - changed the config_files definition

abradshaw:
Before this patch, the entire /var/www/html/pub directory was backed up 
In somecases leading to a config_files.tar.gz of 10's og Gigs

This change ensures that only files that begin with katello- are added during
 this phase

### DIFF
--- a/definitions/features/foreman_server.rb
+++ b/definitions/features/foreman_server.rb
@@ -26,7 +26,7 @@ module ForemanMaintain
       def config_files
         [
           '/etc/httpd',
-          '/var/www/html/pub',
+          '/var/www/html/pub/katello-*',
           '/etc/squid',
           '/etc/foreman',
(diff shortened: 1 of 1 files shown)


sean797:
Why did you close this @abradshaw ?

abradshaw:
> Why did you close this @abradshaw ?

Because I didnt have time/motivation to wrestle with 

- length of the first commit message line for 8987296 exceeds 65 characters
- commit message for 8987296 is not wrapped at 72nd column
- commit message for e25338b is not wrapped at 72nd column
- e1c0f2b must be in the format fixes #redmine_number - brief description
- length of the first commit message line for e1c0f2b exceeds 65 characters

for just a 10 char merge. 

If commiting were easier I suspect you'd get more commits

PR #236, opened 2018-11-29
Fixes #25186 - hotfix check

kgaikwad:


definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,92 @@
...
+    # )
+
+    repoquery_cmd = "repoquery -a --installed --qf '%{ui_from_repo} %{name}'"
+    repoquery_cmd += " | awk '/#{repolist_regexstr}/ {print $2}'"
+    execute!(repoquery_cmd)
+    repoquery_cmd
+  end
+
+  def regex_for_files_check
+    '^(..5....T.){1}(.{2}[^c])+.+\.(rb|py|erb|js)$'
```
iNecas: This thing is hard to understand: would it be possible to put some explanatory comments on how that works?

definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+      message += files_modifications[0..9].join("\n")
+    end
+    message += "\nBefore continuing upgrade, please verify above hotfix(es) details\n"
+    message
+  end
+
+  # only for downstream
+  def find_hotfix_rpms_installed
+    cmd = "rpm -qa 'HOTFIXRHBZ*'"
```
iNecas: this doesn't seem to match packages, that have `HOTFIX` state in release version (see https://bugzilla.redhat.com/show_bug.cgi?id=1642369 as an example for testing with some hotfix installed)

iNecas: It doesn't seem lik `rpm -qa` provides the capability. We could perhaps use the `find_installed_packages` and then search for presence of `HOTFIX` there.

definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,92 @@
...
+    return [] unless execute?(cmd) # handle echo $? = 1
+
+    output = execute!(cmd).strip
+    return [] if output.empty?
+
+    output.strip.split("\n")
+  end
+
+  def find_installed_packages(version)
+    repolist_regexstr = feature(:downstream).repolist_for_hotfix_verify(version).join('|')
```
iNecas: Could we make it in a way, that the `version ` is not needed: for this case, we could rather look at the enabled repos and filter those that include `satellite` or `rhscl`

kgaikwad:
Thank you @iNecas
I will address above comments. Addition to this, @mbacovsky suggested to execute linux commands using ruby library like `Open3.popen3/exec` which provides kind of streams.

definitions/checks/check_hotfix_installed.rb
mbacovsky: @kgaikwad here is my take on the repoqueries https://gist.github.com/mbacovsky/889459a71244d04135bcd71cd62fd692. I didn't tested how the performance differ and it should be subject for further refactoring. Memory-wise it should be okay.

kgaikwad: Now it will work without `version`. I have considered all repos that includes `satellite/rhscl` instead of enabled one. It might be chance that user disabled repos.

definitions/checks/check_hotfix_installed.rb
kgaikwad: Modified search using `HOTFIX`. 
I found few issues where hot fixes delivered using RPM which not included `HOTFIX` string in release. I am not sure what is the reason behind that but in such cases, this check will fail to find out hot fixes applied using RPM.
Any suggestions on this?

definitions/checks/check_hotfix_installed.rb
kgaikwad: Updated PR with the suggested changes. Please have look.

While testing, with other scenarios it would be good to test this check multiple times with some time interval for performance check.

definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,84 @@
...
+        arr_output = line.chomp.split
+        flags = arr_output.first
+        filename = arr_output.last
+        changed_files << filename if flags.include?('5') && filename =~ /\.(rb|py|erb|js)$/
+      end
+    end
+    changed_files
+  end
+
+  def installed_packages
```
upadhyeammit: I got another way to get these details,

1. To get packages from satellite and rhscl repos you can avoid if statement for each package name,

> yumdb search from_repo *satellite* *rhscl*
rh-ruby23-runtime-2.2-7.el7.x86_64
     from_repo = rhel-server-rhscl-7-rpms
rubygem-ansi-1.4.3-3.el7sat.noarch
     from_repo = rhel-7-server-satellite-6.3-rpms

The **yumdb** comes with **yum-utils** only, and this is very quick :)

> time yumdb search from_repo *satellite* *rhscl*
real	0m0.563s
user	0m0.356s
sys	0m0.197s

We just need to disable yum plugins before query and enable it after getting the packages to make it faster else its unnecessary waits for plugins which we dont use while querying yum database,

to disable,

> sed -i -e 's/plugins=1/plugins=0/g' /etc/yum.conf

to enable,

> sed -i -e 's/plugins=0/plugins=1/g' /etc/yum.conf

2. To get the hotfix packages installed you can use,

> rpm -qa release="*HOTFIX*"

kgaikwad: I have updated command to get hotfix rpms using `rpm -qa release="*HOTFIX*"` command in pull-request.
Kept `repoquery` command as `yumdb` is not giving packages names but having extra input lines.

definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,90 @@
...
+      Procedures::Packages::Install.new(:packages => %w[yum-utils])
+    end
+
+    confine do
+      feature(:downstream)
+    end
+  end
+
+  def run
+    if feature(:downstream) && feature(:downstream).subscribed_using_activationkey?
```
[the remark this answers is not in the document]
kgaikwad: Good catch :+1: 
Forgot to remove first condition.

definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,90 @@
...
+  def msg_for_hotfix_rpms(rpms_list)
+    message = "Found below HOTFIX rpm(s) applied on this system.\n"
+    message += rpms_list.join(',')
+    message
+  end
+
+  def msg_for_modified_files(files_modified)
+    message = "\n\nFound #{files_modified.length} file(s) modified on this system.\n"
+    if files_modified.length > 10
+      message += 'Here, it shows only 10 records. For complete result, please check a log file.'
```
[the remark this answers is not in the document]
kgaikwad: +1 
Yes, it would be ok I think if we display it in output only.

kgaikwad:
@mbacovsky, Thank you for review! 
Updated PR with suggested changes and removed unused code.

definitions/checks/check_hotfix_installed.rb
```diff
@@ -12,8 +12,8 @@ class Checks::CheckHotfixInstalled < ForemanMaintain::Check
   end
 
   def run
-    if feature(:downstream) && feature(:downstream).subscribed_using_activationkey?
-      skip 'Your system is subscribed using custom activationkey'
+    if feature(:downstream).subscribed_using_activationkey?
+      skip "Hotfixes can't be detected"
```
mbacovsky: @kgaikwad, sorry for the confusion, I think the original part is important too:
`Your system is subscribed using custom activationkey. Hotfixes can't be detected.`.
This way the user knows the cause and can understand the result.

mbacovsky:
@kgaikwad thanks for the update. I'll do some more testing and this can get merged.

definitions/checks/check_hotfix_installed.rb
kgaikwad: No problem! my bad I was thinking the same :-) but then I thought you expected this as main message.I will do that

definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    confine do
+      feature(:downstream)
+    end
+  end
+
+  def run
+    if feature(:downstream).subscribed_using_activationkey?
+      skip "Your system is subscribed using custom activationkey. Hotfixes can't be detected."
```
mbacovsky: space is missing in "activation key"

definitions/checks/check_hotfix_installed.rb
```diff
@@ -0,0 +1,87 @@
+class Checks::CheckHotfixInstalled < ForemanMaintain::Check
+  metadata do
+    label :check_hotfix_installed
+    description 'Check to verify if any hotfix installed on system'
+    preparation_steps do
```
mbacovsky: Shouldn't this be run during pre-upgrade checks? `tags :pre_upgrade`

mbacovsky:
@kgaikwad I have a few more findings from my testing.
- the check is not tagged to run during pre-upgrade checks
- "Before update make sure the updated packages contain the listed modifications otherwise these fixes will be lost. It is also recommended to backup the modified files prior update." is too long and breaks the text flow. We need to add newlines:
Before update make sure the updated packages contain the listed modifications\notherwise these fixes will be lost.\nIt is also recommended to backup the modified files prior update.
- the result should be rather non-blocking warning as we don't expect the condition will be resolved (hotfixes removed) before upgrade
- I'm not sure if we can move the message before the `[FAIL]` but it would IMO look better. After some investigation it is not this PR related and should be fixed separately. See, this way the message is sort of disconnected from the check:
```
Check to verify if any hotfix installed on system: 
\ Checking for presence of hotfix(es). It may take some time to verify.         
                                                                      [FAIL]

HOTFIX rpm(s) applied on this system:
pulp-selinux-2.13.4.9-2.HOTFIXRHBZ1590906.el7sat.noarch

Before update make sure the updated packages contain the listed modifications otherwise these fixes will be lost. It is also recommended to backup the modified files prior update.
```

jameerpathan111:
@kgaikwad 
Tested PR for following scenarios : 
1) with hotfix installed
2) with modified_files
3) with custom activation key
apart from issues mentioned by @mbacovsky in above comment, PR looks good to me.
Still testing it for other scenarios as well.

definitions/checks/check_hotfix_installed.rb
kgaikwad: Yes, it is needed. updated!

kgaikwad:
Thank you @mbacovsky and @jameerpathan111
Updated PR with suggested changes by @mbacovsky.
In new commit, I modified failure type to "WARNING" instead of "FAIL" and also few changes to warning message. 
@mbacovsky, please let me know if you have any comments/suggestions on new changes.

jameerpathan111:
> Thank you @mbacovsky and @jameerpathan111
> Updated PR with suggested changes by @mbacovsky.
> In new commit, I modified failure type to "WARNING" instead of "FAIL" and also few changes to warning message.
> @mbacovsky, please let me know if you have any comments/suggestions on new changes.

@kgaikwad I have retested PR 
-the check is now tagged to run during pre-upgrade check.
-failure type is "WARNING" instead of "FAIL".
-changes in warning message .
new changes looks good to me. :+1:

mbacovsky:
This looks and works perfectly now. Well done @kgaikwad and thanks for your patience :)!
Thanks @jameerpathan111 for testing!

PR #240, opened 2019-01-10
Fixes #25808 - Split filters for corrupted roles

xprazak2:
When we change resource type for permission which is
already assigned to role, it causes filter to have
permissions with multiple resource types associated,
which does not pass validations. It is impossible to
recover from UI, because the changed permission does
not show when editing filter even though it is still
assigned to that filter.

Added check detects such roles and related procedure
splits the filters to remove validation errors.

### DIFF
--- a/definitions/checks/foreman/check_corrupted_roles.rb
+++ b/definitions/checks/foreman/check_corrupted_roles.rb
@@ -1,0 +1,56 @@
+module Checks
+  module Foreman
+    class CheckCorruptedRoles < ForemanMaintain::Check
+      metadata do
+        label :corrupted_roles
+        for_feature :foreman_database
+        description 'Check for roles that have filters with multiple resources attached'
+        tags :pre_upgrade
+      end
+
+      def run
+        items = find_filter_permissions
+        assert(items.empty?,
+               'There are user roles with inconsistent filters',
+               :next_steps => Procedures::Foreman::FixCorruptedRoles.new)
+      end
+
+      def find_filter_permissions
+        feature(:foreman_database).query(self.class.inconsistent_filter_perms)
+      end
+
+      def self.inconsistent_filter_perms
+        subquery = <<-SQL
+          SELECT filters.id AS filter_id,
+                 filters.role_id,
+                 filters.search,
+                 filters.taxonomy_search,
+                 filters.override,
+                 filterings.id AS filtering_id,
+                 permissions.id AS permission_id,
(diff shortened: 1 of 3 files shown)


definitions/procedures/foreman/corrupted_roles.rb
```diff
@@ -0,0 +1,102 @@
+module Procedures::Foreman
+  class CorruptedRoles < ForemanMaintain::Procedure
+    metadata do
+      param :items, 'Attributes of the roles with inconsistent filters', :required => true
```
mbacovsky: For this procedure to not depend that much on result from the check could it get the records using the same call and not by passing through the param?

definitions/procedures/foreman/corrupted_roles.rb
```diff
@@ -0,0 +1,102 @@
+module Procedures::Foreman
+  class CorruptedRoles < ForemanMaintain::Procedure
```
mbacovsky: Could this be called FixCorruptedRoles or SplitCorruptedRoles or to see what it does?

definitions/procedures/foreman/corrupted_roles.rb
```diff
@@ -0,0 +1,102 @@
...
+    metadata do
+      param :items, 'Attributes of the roles with inconsistent filters', :required => true
+      for_feature :foreman_database
+      tags :pre_migration
+      advanced_run false
```
mbacovsky: When there is no complex param this could be advanced_run can be true (default)

xprazak2:
I renamed the classes to better represent what they do and procedure no longer depends on the results from check but it fetches its own data.

mbacovsky:
@xprazak2 thanks for addressing my comments. If rubocop is happy we can merge this.

mbacovsky:
@xprazak2 thanks! Well done. :beer:

jameerpathan111:
@xprazak2 @mbacovsky @kgaikwad 

corrupted-roles/fix-corrupted-roles check does not work on satellite 6.2
here is what I found during upgrade testing of satellite 6.2

[https://paste.fedoraproject.org/paste/PZDNpaelaqjFZwHjybNPLA](https://paste.fedoraproject.org/paste/PZDNpaelaqjFZwHjybNPLA)

The error message I see in log is `"ERROR:  column filters.override does not exist"`

xprazak2:
Maybe https://github.com/theforeman/foreman_maintain/pull/243 could help?

PR #241, opened 2019-01-17
Refs #25186 - downcase repo_name before matching regex

kgaikwad:
@mbacovsky,
Added small change in hotfix_check.

@jameerpathan111 noticed that in latest snap sat 6.5, there are repositories contain `Satellite` word in names.  Applied `downcase` on repo_name before matching with `/satellite|rhscl/` for safer side.


mbacovsky:
@kgaikwad thanks. ACK from me. @jameerpathan111 will you have time for testing this?

jameerpathan111:
> @kgaikwad thanks. ACK from me. @jameerpathan111 will you have time for testing this?

tested and works as expected
ACK from me as well

PR #242, opened 2019-01-30
Fixes #25939 - Fix mongo engines during restore

mbacovsky:
When we restore backup of MongoDB using MMapv1 storage engine
on a machine where MongoDB with WiredTiger exists at some point
we end up with old config and new data which breaks the service.
This patch makes sure this situation is detected and remediated.

### DIFF
--- a/definitions/procedures/restore/ensure_mongo_engine_matches.rb
+++ b/definitions/procedures/restore/ensure_mongo_engine_matches.rb
@@ -1,0 +1,28 @@
+module Procedures::Restore
+  class EnsureMongoEngineMatches < ForemanMaintain::Procedure
+    metadata do
+      description 'Ensure restored MongoDB storage engine matches the current DB'
+      for_feature :mongo
+    end
+
+    def run
+      if feature(:mongo).local? && engine_mismatch?
+        with_spinner('') do |spinner|
+          feature(:service).handle_services(spinner, 'stop', :only => feature(:mongo).services)
+          spinner.update('Clean MongoDB data')
+          data_path = Dir[feature(:mongo).data_dir + '/*']
+          FileUtils.rm_rf(data_path)
+          feature(:service).handle_services(spinner, 'start', :only => feature(:mongo).services)
+        end
+      end
+    end
+
+    private
+
+    def engine_mismatch?
+      tiger_file = File.join(feature(:mongo).data_dir, 'WiredTiger.wt')
+      config_file = feature(:mongo).core.server_config_files.first
+      File.exist?(tiger_file) && File.open(config_file).grep(/^storage.engine:\s*mmapv1/).any?
+    end
+  end
+end
--- a/definitions/procedures/restore/ensure_mongo_engine_matches.rb
+++ b/definitions/procedures/restore/ensure_mongo_engine_matches.rb
@@ -5,6 +5,7 @@ module Procedures::Restore
       for_feature :mongo
     end
 
+    # rubocop:disable Metrics/AbcSize
     def run
       if feature(:mongo).local? && mongo_data_dir_exists? && engine_mismatch?
         with_spinner('Clean MongoDB data') do |spinner|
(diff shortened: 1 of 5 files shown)


mbacovsky:
@alda519 could you please take a look?

definitions/procedures/restore/ensure_mongo_engine_matches.rb
```diff
@@ -0,0 +1,28 @@
...
+      for_feature :mongo
+    end
+
+    def run
+      if feature(:mongo).local? && engine_mismatch?
+        with_spinner('') do |spinner|
+          feature(:service).handle_services(spinner, 'stop', :only => feature(:mongo).services)
+          spinner.update('Clean MongoDB data')
+          data_path = Dir[feature(:mongo).data_dir + '/*']
+          FileUtils.rm_rf(data_path)
```
iNecas: I'm always a bit cautions when I see `rm_rf` somewhere. I would rather double-check, that the `feature(:mongo).data_dir` doesn't return `nil?` or `empty?` so that we don't end up with `rm_rf("/*")`, as for example https://github.com/theforeman/foreman_maintain/blob/3da185523aba488b4118a991ca783e609429f583/definitions/procedures/backup/offline/mongo.rb#L27 has some corner cases.

iNecas:
Two small comments code-wise, relying on QEs with the functional testing.

mbacovsky:
@iNecas thanks for review; updated in separate commit

definitions/procedures/restore/ensure_mongo_engine_matches.rb
```diff
@@ -0,0 +1,32 @@
...
+    private
+
+    def engine_mismatch?
+      tiger_file = File.join(feature(:mongo).data_dir, 'WiredTiger.wt')
+      config_file = feature(:mongo).core.server_config_files.first
+      File.exist?(tiger_file) && File.open(config_file).grep(/^storage.engine:\s*mmapv1/).any?
+    end
+
+    def mongo_data_dir_exists?
+      File.directory?(feature(:mongp).data_dir)
```
iNecas: typo

alda519:
restore of mmapv1 backup works (tested both online and offline)
but after that upgrade to wiredtiger fails

```
# satellite-installer --upgrade-mongo-storage-engine
Starting disk space check for upgrade
Resetting puppet server version param...
Running Stop Services
Stop applicable services: Stopping the following service(s):

qdrouterd, qpidd, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, smart_proxy_dynflow_core, goferd, httpd
- All services stopped                                                [OK]      

foreman-maintain service stop --exclude "rh-mongodb34-mongod","postgresql","tomcat","dynflowd","foreman-proxy","puppetserver" finished successfully!
/usr/share/ruby/fileutils.rb:245:in `mkdir': File exists - /var/tmp/mongodb_engine_upgrade (Errno::EEXIST)
	from /usr/share/ruby/fileutils.rb:245:in `fu_mkdir'
	from /usr/share/ruby/fileutils.rb:174:in `block in mkdir'
	from /usr/share/ruby/fileutils.rb:173:in `each'
	from /usr/share/ruby/fileutils.rb:173:in `mkdir'
	from /usr/share/katello-installer-base/hooks/pre/31-mongo_storage_engine.rb:19:in `migration'
	from /usr/share/katello-installer-base/hooks/pre/31-mongo_storage_engine.rb:65:in `block (4 levels) in load'
	from /usr/share/gems/gems/kafo-2.1.0/lib/kafo/hooking.rb:34:in `instance_eval'
	from /usr/share/gems/gems/kafo-2.1.0/lib/kafo/hooking.rb:34:in `block (4 levels) in load'
```

mbacovsky:
@alda519 Good catch, updated!

mbacovsky:
Thanks @alda519, @iNecas!

PR #245, opened 2019-02-20
Fixes #26043 - Remote Mongo SSL connection improvements

mbacovsky:
This patch fixes improper hamdling of empty entries in Pulp's
Mongo ssl configuration. Empty ssl_certfile and ca_path are ignored.
verify_ssl = false is now supported.

'service status' is printing the remote service is DOWN when it
can not connect to the remote DB. This is not helpful when the state
is not expected so some hints to look for the cause in logs were added.

### DIFF
--- a/definitions/features/mongo.rb
+++ b/definitions/features/mongo.rb
@@ -72,19 +72,24 @@ class Features::Mongo < ForemanMaintain::Feature
     ['localhost', '127.0.0.1', hostname].include?(configuration['host'])
   end
 
+  # rubocop:disable Metrics/AbcSize
   def base_command(command, config = configuration, args = '')
     if config['ssl']
       ssl = ' --ssl'
-      if config['ca_path']
+      if config['ca_path'] && !config['ca_path'].empty?
         ca_cert = " --sslCAFile #{config['ca_path']}"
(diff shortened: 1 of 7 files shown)


mbacovsky:
@ntkathole or @jameerpathan111 could you take a look please?

kgaikwad:
Code wise looks good to me! :+1:

mbacovsky:
Rebased

mbacovsky:
@jameerpathan111, I fixed the failing test. Also we have confirmed by customer that this patch unblocked their upgrade which was finally finished. So maybe it is just okay to run the automation to check we didn't break anything else.

jameerpathan111:
> @jameerpathan111, I fixed the failing test. Also we have confirmed by customer that this patch unblocked their upgrade which was finally finished. So maybe it is just okay to run the automation to check we didn't break anything else.

@mbacovsky That's great, it works for customer
I will run automation with this PR.

jameerpathan111:
@mbacovsky I have run automation around this PR and everything looks ok.

mbacovsky:
@jameerpathan111 Thanks for the testing!
@kgaikwad could you please merge the PR if there are no other objections?

kgaikwad:
Thank you, @mbacovsky!

PR #248, opened 2019-03-06
Fixes #26610 - offline backup config file fix

pjgunst:
…d tar backups due to Candlepin activemq files being modified

### DIFF
--- a/definitions/scenarios/backup.rb
+++ b/definitions/scenarios/backup.rb
@@ -27,7 +27,6 @@ module ForemanMaintain::Scenarios
       prepare_directory
       logical_volume_confirmation
       add_step_with_context(Procedures::Backup::Metadata)
-      add_step_with_context(Procedures::Backup::ConfigFiles)
 
       case strategy
       when :online
@@ -132,6 +131,7 @@ module ForemanMaintain::Scenarios
       add_steps_with_context(
         find_procedures(:maintenance_mode_on),
         Procedures::Service::Stop,
+      	Procedures::Backup::ConfigFiles,
         Procedures::Backup::Pulp,
         Procedures::Backup::Offline::Mongo,
         Procedures::Backup::Offline::CandlepinDB,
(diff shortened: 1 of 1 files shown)


mbacovsky:
@pjgunst thanks for the patch! The code looks good and works well. The only drawback of this solution is the the downtime required for backup is expanded with the time of backup of configs. It is about 30s on my system and I'd assume it is acceptable with regard to the time of whole backup. 

This is good to merge but we have to satisfy some formal conditions. 
- Is the original problem documented  under some issue? If not could you please create one in https://projects.theforeman.org/projects/foreman-maintain/issues/new? 
- Also the commit message should be in format `Fixes #<issue_no> - backup configs when services are down` or similar.

Let me know if you need any assistance.

pjgunst:
Hi,
Redmine issue created: https://projects.theforeman.org/issues/26610

definitions/scenarios/backup.rb
```diff
@@ -132,6 +131,7 @@ def add_offline_backup_steps
       add_steps_with_context(
         find_procedures(:maintenance_mode_on),
         Procedures::Service::Stop,
+      	Procedures::Backup::ConfigFiles,
```
mbacovsky: Tab instead of spaces

definitions/scenarios/backup.rb
```diff
@@ -162,6 +162,7 @@ def add_snapshot_backup_steps
         Procedures::Backup::Snapshot::PrepareMount,
         find_procedures(:maintenance_mode_on),
         Procedures::Service::Stop,
+      	Procedures::Backup::ConfigFiles,
```
mbacovsky: Tab instead of spaces

mbacovsky:
@pjgunst thanks for the update! 
Rubocop is not yet happy, could you please check the log in https://travis-ci.org/theforeman/foreman_maintain/jobs/520140519? You can also test locally by running `rake rubocop`. Removing the line from compose made the method fit the rubocop limits and we don't need the rule to be turned off. Please remove the disable comments on L22 and L41. Also you've used tab instead of spaces for indentation and our rubocop is sensitive about that (marked inline).

mbacovsky:
@pjgunst, I've amended the commit and merged manually. Thanks for your contribution!

PR #249, opened 2019-03-08
Fixes #26216 - Lock package versions to protect against unwanted upgrades

mbacovsky:
Adds ability to lock versions of packages of Foreman/Katello/Satellite
and its dependences to prevent it from partial upgrade during
common system update.

The locking consists of three parts:
 - make sure the tooling is installed and configured properly
 - collect list of packages to lock
 - do the actual locking

New commands were added to foreman_maintain:

 $ foreman-maintian installation lock
 $ foreman-maintian installation unlock

New feature :package_manager is introduced. It serves as a proxy
for package manipulation commands and dispatches the calls to
the PackageManager implementation that match what is installed.
In the initial version only yum is supported. Apt is in progress though
and will be added soon. The system helpers were redirected to
the package manager which should allow seamless support for
debian once the apt manager is added.

The version locking itself is implemented in the package manager
and should match what system supports The same pays for checks
and instalation of the related tooling.

### DIFF
--- a/lib/foreman_maintain/cli.rb
+++ b/lib/foreman_maintain/cli.rb
@@ -9,6 +9,7 @@ require 'foreman_maintain/cli/advanced_command'
 require 'foreman_maintain/cli/service_command'
 require 'foreman_maintain/cli/restore_command'
 require 'foreman_maintain/cli/maintenance_mode_command'
+require 'foreman_maintain/cli/packages_command'
 
 module ForemanMaintain
   module Cli
@@ -20,6 +21,7 @@ module ForemanMaintain
       subcommand 'service', 'Control applicable services', ServiceCommand
       subcommand 'backup', 'Backup server', BackupCommand
       subcommand 'restore', 'Restore a backup', RestoreCommand
+      subcommand 'packages', 'Lock/Unlock installed packages', PackagesCommand
       subcommand 'advanced', 'Advanced tools for server maintenance', AdvancedCommand
       subcommand 'maintenance-mode', 'Control maintenance-mode for application',
                  MaintenanceModeCommand
--- a/lib/foreman_maintain/package_manager/yum.rb
+++ b/lib/foreman_maintain/package_manager/yum.rb
@@ -1,0 +1,126 @@
+module ForemanMaintain::PackageManager
+  class Yum < Base
+    VERSIONLOCK_START_CLAUSE = '## foreman-maintain - start'.freeze
+    VERSIONLOCK_END_CLAUSE = '## foreman-maintain - end'.freeze
+    VERSIONLOCK_CONFIG_FILE = '/etc/yum/pluginconf.d/versionlock.conf'.freeze
+    VERSIONLOCK_DEFAULT_LIST_FILE = '/etc/yum/pluginconf.d/versionlock.list'.freeze
+
+    def self.parse_envra(envra)
+      # envra format: 0:foreman-1.20.1.10-1.el7sat.noarch
+      parsed = envra.match(/\d*:?(?<name>.*)-[^-]+-[^-]+\.[^.]+/)
+      parsed ? Hash[parsed.names.zip(parsed.captures)].merge(:envra => envra) : nil
+    end
+
+    def foreman_related_packages
+      query = "repoquery -a --qf='%{envra} %{repo.id}' --search foreman-installer |head -n1"
+      foreman_repo = sys.execute(query).split[1]
+      query_installed = "repoquery -a --qf='%{envra}' --repoid='#{foreman_repo}'"
+      sys.execute(query_installed). split("\n").map do |pkg|
+        self.class.parse_envra(pkg)
+      end
+    end
+
+    def version_locking_packages
+      %w[yum-utils yum-plugin-versionlock]
+    end
+
+    def lock_versions(package_list)
+      unlock_versions
+      File.open(versionlock_file, 'a') do |f|
+        f.puts VERSIONLOCK_START_CLAUSE
+        f.puts '# The following packages are locked by foreman-maintain. Do not modify!'
+        package_list.each { |package| f.puts "#{package[:envra]}.*" }
+        f.puts '# End of list of packages locked by foreman-maintain'
+        f.puts VERSIONLOCK_END_CLAUSE
+      end
+    end
+
+    def unlock_versions
+      lock_file = versionlock_file
+      content = File.read(lock_file)
+      content = content.gsub(/#{VERSIONLOCK_START_CLAUSE}.*#{VERSIONLOCK_END_CLAUSE}\n/m, '')
+      File.open(lock_file, 'w') { |f| f.write content }
+    end
+
+    def versions_locked?
+      lock_file = versionlock_file
+      return false if lock_file.nil?
+      content = File.read(lock_file)
+      !!content.match(/#{VERSIONLOCK_START_CLAUSE}.*#{VERSIONLOCK_END_CLAUSE}\n/m)
+    end
+
+    def version_locking_enabled?
+      installed?(version_locking_packages) && versionlock_config =~ /^\s*enabled\s+=\s+1/ \
+        && File.exist?(versionlock_file)
+    end
+
+    # make sure the version locking tools are configured
+    #  enabled = 1
+    #  locklist = <list file>
+    # we can assume it is already installed
+    def configure_version_locking
+      config = versionlock_config
+      config += "\n" unless config[-1] == "\n"
+      enabled_re = /^\s*enabled\s*=.*$/
+      if enabled_re.match(config)
+        config = config.gsub(enabled_re, 'enabled = 1')
+      else
+        config += "enabled = 1\n"
+      end
+      unless config =~ /^\s*locklist\s*=.*$/
+        config += "locklist = #{VERSIONLOCK_DEFAULT_LIST_FILE}\n"
+      end
+      File.open(versionlock_config_file, 'w') { |file| file.puts config }
+      FileUtils.touch(versionlock_file)
+    end
+
+    def installed?(packages)
+      packages_list = [packages].flatten(1).map { |pkg| "'#{pkg}'" }.join(' ')
+      sys.execute?(%(rpm -q #{packages_list}))
+    end
+
+    def find_installed_package(name)
+      status, result = sys.execute_with_status(%(rpm -q '#{name}'))
+      if status == 0
+        result
+      end
+    end
+
+    def install(packages, assumeyes: false)
+      yum_action('install', packages, :assumeyes => assumeyes)
+    end
+
+    def update(packages = [], assumeyes: false)
+      yum_action('update', packages, :assumeyes => assumeyes)
+    end
+
+    def clean_cache
+      yum_action('clean', 'all')
+    end
+
+    private
+
+    def versionlock_config
+      File.exist?(versionlock_config_file) ? File.read(versionlock_config_file) : ''
+    end
+
+    def versionlock_config_file
+      VERSIONLOCK_CONFIG_FILE
+    end
+
+    def versionlock_file
+      result = versionlock_config.match(/^\s*locklist\s*=\s*(\S+)/)
+      result.nil? ? nil : File.expand_path(result.captures[0])
+    end
+
+    def yum_action(action, packages, assumeyes: false)
+      yum_options = []
+      packages = [packages].flatten(1)
+      yum_options << '-y' if assumeyes
+      yum_options_s = yum_options.empty? ? '' : ' ' + yum_options.join(' ')
+      packages_s = packages.empty? ? '' : ' ' + packages.join(' ')
+      sys.execute!("yum#{yum_options_s} #{action}#{packages_s}",
+                   :interactive => true)
+    end
+  end
+end
--- a/definitions/checks/version_locking_enabled.rb
+++ b/definitions/checks/version_locking_enabled.rb
@@ -1,0 +1,14 @@
+module Checks
+  class VersionLockingEnabled < ForemanMaintain::Check
+    metadata do
+      description 'Check if tooling for package version locking is installed'
+    end
+
+    def run
+      enabled = feature(:package_manager).version_locking_enabled?
+      enable_locking = Procedures::Packages::EnableVersionLocking.new(:assumeyes => assumeyes?)
+      assert(enabled, 'Tools for package version locking are not available on this system',
+             :next_steps => enable_locking)
+    end
+  end
+end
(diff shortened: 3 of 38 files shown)


mbacovsky:
This is WIP for initial review of the concept. Apt support will be added in separate PR but the ground is prepared.

TODO:
[ ] tests/rubocop
[ ] migration of remainig pacakage related system helpers to pacakage manager
[ ] tuning of the way how we collect packages to lock

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -0,0 +1,108 @@
...
+      %w[yum-utils yum-plugin-versionlock]
+    end
+
+    def lock_versions(package_list)
+      unlock_versions
+      open(versionlock_file, 'a') do |f|
+        f.puts VERSIONLOCK_START_CLAUSE
+        f.puts '# The following packages are locked by foreman-maintain. Do not modify!'
+        package_list.each { |package| f.puts "#{package[:envra]}.*" }
+        f. puts '# End of list of packages locked by foreman-maintain'
```
kgaikwad: Minor typo correction - remove whitespace before `puts`

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -0,0 +1,108 @@
...
+    end
+
+    private
+
+    def versionlock_config
+      File.read(VERSIONLOCK_CONFIG_FILE) if File.exist?(VERSIONLOCK_CONFIG_FILE)
+    end
+
+    def versionlock_file
+      /^\s*locklist\s*=\s*(\S+)/.match(versionlock_config).captures[0]
```
kgaikwad: In case of `versionlock_config_file` not present then Nil case handling required.

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    def sys
+      ForemanMaintain::Utils::SystemHelpers
+    end
+
+    def yum_action(action, packages, assumeyes: false)
+      yum_options = []
+      yum_options << '-y' if assumeyes
+      sys.execute!("yum #{yum_options.join(' ')} #{action} #{packages.join(' ')}",
```
upadhyeammit: If by chance yum-plugin-versionlock is not there or repository not enabled then we need to catch exception of execute! and I feel terminate the execution with error message. I know its corner case when base repository is not enabled but it fails and goes into loop to install package unless terminated with user. This also then fails in method configure_version_locking and line no 60 with "undefined method `[]' for nil:NilClass" .

mbacovsky:
Fixed the tests that were failing on debian travis workers. It was caused by missing package manager for deb packages. I added mocked package manager to mimic existing manager in tests. It allows us easily mock what packages are installed on the system and separate our tests from dependency on the real  packages on the system.

lib/foreman_maintain/package_manager/yum.rb
mbacovsky: @kgaikwad good point. I made it a bit more robust.

lib/foreman_maintain/package_manager/yum.rb
mbacovsky: @kgaikwad yay!, fixed

mbacovsky:
After @ekohl's  valid comments in https://github.com/theforeman/foreman-installer/pull/336 I changed 
exit codes for is-locked command to '0' for positive result an '1' for negative one. 

We should reconsider this also in `maintenance-mode is-enabled` and I filed an issue for that https://projects.theforeman.org/issues/26456. @kgaikwad what do you think?

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -0,0 +1,129 @@
...
+    VERSIONLOCK_START_CLAUSE = '## foreman-maintain - start'.freeze
+    VERSIONLOCK_END_CLAUSE = '## foreman-maintain - end'.freeze
+    VERSIONLOCK_CONFIG_FILE = '/etc/yum/pluginconf.d/versionlock.conf'.freeze
+    VERSIONLOCK_DEFAULT_LIST_FILE = '/etc/yum/pluginconf.d/versionlock.list'.freeze
+
+    def foreman_related_packages
+      installed = sys.execute("repoquery -a --qf='%{envra} %{repo.baseurl}'").
+                  split("\n"). map(&:split)
+      repos_to_lock = []
+      foreman_repo = installed.find { |package| package[0] =~ /foreman-installer-\d+/ }[1]
```
upadhyeammit: I feel this find iteration can be avoided by simple grep command ? 
# repoquery -a --qf='%{envra} %{repo.baseurl}'|grep foreman-installer

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -0,0 +1,129 @@
...
+      repos_to_lock << foreman_repo
+      installed.inject([]) do |related, package|
```
upadhyeammit: I think you can also save this iteration, something like below,

Query repoid,
~~~
repoquery -a --qf='%{envra} %{repo.id}' |grep foreman-installer
1:foreman-installer-1.18.0.3-1.el7sat.noarch rhel-7-server-satellite-6.4-rpms
0:foreman-installer-katello-3.7.0.15-1.el7sat.noarch rhel-7-server-satellite-6.4-rpms
~~~

and then directly query all packages which are there from that repoid,

~~~
repoquery -a --repoid=rhel-7-server-satellite-6.4-rpms
~~~

upadhyeammit: I also feel that we should also lock rhscl, ansible and puppet repositories because some of Satellite bits should be depending on these repositories, but I am not very sure how frequently we update these three repositories, if those are not having frequent upgrade and just locking packages from Satellite repository does not break things then I think there is no need.

lib/foreman_maintain/package_manager/yum.rb
mbacovsky: This method is trying to find the repo foreman-installer is comming from. Then it selects for locking all installed packages coming from that repo.
This part could be simplified as you described but then I'd need to run another repoquery to get the list of packages from the detected repo. The repoquery is pretty expensive call so it is more effective to run it once and process the result twice in ruby.

lib/foreman_maintain/cli.rb
```diff
@@ -20,6 +21,7 @@ class MainCommand < Base
       subcommand 'service', 'Control applicable services', ServiceCommand
       subcommand 'backup', 'Backup server', BackupCommand
       subcommand 'restore', 'Restore a backup', RestoreCommand
+      subcommand 'installation', 'Lock/Unlock installed packages', InstallationCommand
```
iNecas: Per offline discussion, suggesting renaming the command to `foreman-maintain packages lock/unlock`

test/definitions/scenarios/upgrade_satellite_6_2_test.rb
```diff
@@ -10,6 +10,7 @@ module Scenarios
       end
       assume_feature_present(:foreman_tasks)
       stub_foreman_proxy_config
+      stub_package_manager
```
iNecas: As per offline discussion: probably makes sense to move this stubbing to test helper

iNecas:
:+1: conceptual-wise, no other comments from me.

lib/foreman_maintain/package_manager/yum.rb
mbacovsky: I did some more experiments and optimized repoquery so that we can save 10-15% of time and also a few megs of memory with the suggested approach. But that is only in case we are locking packages from single repo. For more then one repo it is 20+% slower. So I'd wait till we figure out if we need to lock also scl/tools/other. If we won't I'll rewrite using two repoqueries. Is it acceptable @upadhyeammit.

mbacovsky:
Updated. I renamed the `installation` command to `packages` and updated tests mocking. I removed the WIP flag. We need to get this merged to get installer bits in on time.

Things that needs to be done here:
- tuning of what packages  we lock (scl/tools/other) (can be in separate commit)
- more tests (can be in separate commit)
- fixing the case @upadhyeammit found when version lock plugin or  repo are not present

mbacovsky:
[test]

mbacovsky:
[test foreman_maintain]

jameerpathan111:
@mbacovsky 
found a issue while performing satellite 6.4 upgrade using this PR.
here is pastebin link for error:
[https://paste.fedoraproject.org/paste/iGVA9TS1qSNEkR5YS5TIIQ](https://paste.fedoraproject.org/paste/iGVA9TS1qSNEkR5YS5TIIQ)

jameerpathan111:
@mbacovsky I have mentioned few things which we need to do. If you can confirm it and add anything which we are missing or is wrong.

- [ ]  We have to file documentation bugzilla regarding how this feature works, what are supported and unsupported scenarios.
- [ ]  Currently we are missing automatic unlocking of locked packages, which I guess will affect upgrade through FM. I think we should add this part into feature.
- [ ]  We also will have this feature present on capsule's FM(with foreman-installer part as well), we need to take care of this part as well. Like how it will be used/not used on capsule.
- [ ]  It will be good if we have output of 'foreman-maintain packages is-locked' command similar to 'foreman-maintain maintenance-mode status' command.

lib/foreman_maintain/package_manager/yum.rb
upadhyeammit: @mbacovsky this approach is acceptable.

jameerpathan111:
@mbacovsky 
- the new change introduced through https://github.com/theforeman/foreman_maintain/pull/249/commits/86b87fc0d504743fbfa6103e64eff6b20765ef34 solves the upgrade issue.
- we have added unlocking versions before yum upgrade https://github.com/theforeman/foreman_maintain/pull/249/commits/269076d072895c008eeb60dac7e7a3fe16177855, 
so my question is, should we also lock package versions after upgrade?
currently package versions gets unlocked while upgrade but does not get locked again after upgrade.

jameerpathan111:
@mbacovsky @upadhyeammit @kgaikwad 
For testing version locking PR we copied 3 files of foreman-installer PR onto satellite, can we say that doing this is same as applying a hotfix?
In this case should hotfix check should pass or fail?
currently check gets passed even when I have copied those files onto satellite.

mbacovsky:
@jameerpathan111, thanks for testing. 
What you are experiencing with the unlock can happen only when you lock manually and automatic locking is not turned on. Locking after upgrade is a job of installer and I'm not sure if that should happen when auto-locking is turned off. Correct me if I'm wrong.

As for the hot fixes - this patch is just addition of the files so I guess the change won't be removed with package update and therefore does not need to be detected.

jameerpathan111:
> What you are experiencing with the unlock can happen only when you lock manually and automatic locking is not turned on. Locking after upgrade is a job of installer and I'm not sure if that should happen when auto-locking is turned off. Correct me if I'm wrong.

@mbacovsky you are right, but it's just that we won't have foreman-installer part available for previous satellite versions. In this case if user just manually locks packages on satellite (using FM having version-locking feature) and run upgrade through FM, then after upgrade user will expect that packages will get automatically locked. @mbacovsky what do you think about this scenario?
> As for the hot fixes - this patch is just addition of the files so I guess the change won't be removed with package update and therefore does not need to be detected. What do you think about this scenario?

:+1: Thank you for the clarification.

iNecas:
What is the status here?

mbacovsky:
@jameerpathan111 there are more weird scenarios when you are using locking without the installer part. E.g. if you lock packages manually and then run satellite-installer to update your configuration the packages stay locked as it is the installer job to unlock. The more I think about it the more I tend to think we should at least disable the locking commands on Sat < 6.6. Any preference @iNecas?

@iNecas the TODO here is:
- [ ] optimization of package detection as @upadhyeammit suggested
- [ ] add status command similar to what maint mode has
- [ ] resolve the behavior on Sat < 6.6, probably one of
  - hide commands
  - keep commands but print 6.6 is required and exit
  - backport the installer hooks and support the feature fully
  - keep it as is - unsupported but possible with manual lock/unlock and the logical hiccups mentioned above 
- [ ] add some tests

jameerpathan111:
> @jameerpathan111 there are more weird scenarios when you are using locking without the installer part. E.g. if you lock packages manually and then run satellite-installer to update your configuration the packages stay locked as it is the installer job to unlock. The more I think about it the more I tend to think we should at least disable the locking commands on Sat < 6.6. Any preference @iNecas?

@mbacovsky  you are right, missing installer part for satellite < 6.6 makes feature implementation much more complicated. We should discuss about this with others and figure out what we should do.

lib/foreman_maintain/package_manager/yum.rb
mbacovsky: For a record: after some discussion we decided not to lock collections and other repos for now.

mbacovsky:
Packages lookup was optimized, status command added. As for the support for earlier Sat versions there are preferences to keep it as is - no change for the upgrade/install on Sat < 6.6 but manual locking would be possible with extra effort.

From my perspective all intended development is done so please give it a final look. In the meantime I'm working on tests for package manger layer and documentation.

lib/foreman_maintain/package_manager/yum.rb
mbacovsky: @upadhyeammit I fixed the attempt to configure when packages were not installed properly.
For the install loop I see that after the remediation procedure the check is re-run and fail. The output explains what happened and it is up to the user if he wants to retry or quit. That seems to be correct. If run with `--assumeyes` it ends up after one round which prevents endless loop.

Is the behavior acceptable now when configure error is handled?

jameerpathan111:
@mbacovsky 
- I have tried Satellite 6.3 to Satellite 6.4 and Satellite 6.4 to Satellite 6.5( on different satellite) upgrade using this PR. I did not face any major issue while upgrade. Scenarios related to version locking worked as expected. 
- while testing I found one minor issue so filled [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1710305) for it.

mbacovsky:
Squashed and rebased, everything is green!
@jameerpathan111 thanks for all the effort you've put into testing this repeatedly :beer:.
@iNecas, @kgaikwad, @upadhyeammit thanks for all the reviews!

lib/foreman_maintain/package_manager/yum.rb
upadhyeammit: This is good to go !

upadhyeammit:
LGTM!

kgaikwad:
Merged!
Thank you @mbacovsky for adding a big feature :tada:

@jameerpathan111, thank you for testing end to end scenarios.

PR #252, opened 2019-03-29
Fixes #26456 - modified exit codes of maintenance-mode is-enabled

kgaikwad:


kgaikwad:
> Thanks @kgaikwad! Looks good so far. We still need to update the docs in
> 
> ```
> maintenance-mode is-enabled returns 0 or 1 output depending upon the maintenance-mode status. Here, 1=ON & 0=OFF.
> ```
> Other then that it is good to go.

Good Catch :+1:

mbacovsky:
:+1: thanks @kgaikwad!!  :cake:

PR #253, opened 2019-04-01
Fixes #26017 - treat tar packing response code 1 as success

pmoravec:
During online backup, tar packages files changing on the fly.
It returns 1 if detecting so, while this is not a fatal error
for the backup. Therefore we should treat response code 1 as
success.

### DIFF
--- a/definitions/procedures/backup/config_files.rb
+++ b/definitions/procedures/backup/config_files.rb
@@ -11,6 +11,8 @@ module Procedures::Backup
       param :backup_dir, 'Directory where to backup to', :required => true
       param :proxy_features, 'List of proxy features to backup (default: all)',
             :array => true, :default => ['all']
+      param :ignore_changed_files, 'Should packing tar ignore changed files',
+            :flag => true, :default => false
     end
 
     def run
@@ -18,9 +20,10 @@ module Procedures::Backup
       increments = File.join(@backup_dir, '.config.snar')
       with_spinner('Collecting config files to backup') do
         configs = config_files.join(' ')
+        statuses = @ignore_changed_files ? [0, 1] : [0]
         execute!("tar --selinux --create --gzip --file=#{tarball} " \
           "--listed-incremental=#{increments} --ignore-failed-read " \
-          "#{configs}")
+          "#{configs}", :valid_exit_statuses => statuses)
       end
     end
 
--- a/definitions/scenarios/backup.rb
+++ b/definitions/scenarios/backup.rb
@@ -183,7 +183,7 @@ module ForemanMaintain::Scenarios
     # rubocop:enable  Metrics/MethodLength
 
     def add_online_backup_steps
-      add_step_with_context(Procedures::Backup::ConfigFiles)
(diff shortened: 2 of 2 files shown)


pmoravec:
Thanks @mbacovsky for the original idea.

definitions/procedures/backup/config_files.rb
```diff
@@ -20,7 +20,7 @@ def run
         configs = config_files.join(' ')
         execute!("tar --selinux --create --gzip --file=#{tarball} " \
           "--listed-incremental=#{increments} --ignore-failed-read " \
-          "#{configs}")
+          "#{configs}", :valid_exit_statuses => [0, 1])
```
mbacovsky: @pmoravec Is it safe? I don't care much about online backup. But it could let some inconsistency sneak in offline backup. I guess that doing config backup after we stop services (https://github.com/theforeman/foreman_maintain/pull/248) might help but still there is a chance of something unexpected. 
Could we make this change optional by a new parameter that is set for online backups only?
Does it make sense or I'm I too paranoid?

pmoravec: It makes sense to be tolerant to changed configs _only_ on online backups - will update the PR.

(that's why I am not a great developer, I fix just my (broken) call flow)

pmoravec:
PR updated.

pmoravec:
Anything is pending here, @mbacovsky ?

I see similar problem (online backup fails the same way) in #248 is fixed, so let merge this one as well to have this kind of issues (and rhbz 1673908) resolved :) .

mbacovsky:
@thanks for the update. As we discussed outside of this thread, this could be pushed a bit and have the procedure independent on the strategy parameter. With #248 merged it should be easy to do, I'll put some hints inline.

definitions/procedures/backup/config_files.rb
```diff
@@ -11,16 +11,18 @@ class ConfigFiles < ForemanMaintain::Procedure
       param :backup_dir, 'Directory where to backup to', :required => true
       param :proxy_features, 'List of proxy features to backup (default: all)',
             :array => true, :default => ['all']
+      param :strategy, 'Backup strategy'
```
mbacovsky: This could be renamed to :ignore_changed_files or similar with  `:flag => true, :default => false` options

definitions/scenarios/backup.rb
```diff
@@ -69,6 +69,8 @@ def set_context_mapping
                   Procedures::Backup::Metadata => :incremental_dir)
       context.map(:proxy_features,
                   Procedures::Backup::ConfigFiles => :proxy_features)
+      context.map(:strategy,
```
mbacovsky: The context map should be replaced by passing the :ignore_changed_files param in compose e.g.
```ruby
add_step_with_context(Procedures::Backup::ConfigFiles, :ignore_changed_files => true)
```
in `add_online_backup_steps`.

pmoravec:
Thanks for feedback, PR updated (and rebased to upstream to resolve conflicts with #248).

PR #254, opened 2019-04-03
Fixes #26522 - sosreport procedure

lzap:
Do not merge until we settle up on Debian packaging.

https://community.theforeman.org/t/rfc-deprecate-foreman-debug-in-favor-of-sosreport/12465

### DIFF
--- a/definitions/procedures/sos/sosreport.rb
+++ b/definitions/procedures/sos/sosreport.rb
@@ -1,0 +1,33 @@
+module Procedures::Sos
+  class Sosreport < ForemanMaintain::Procedure
+    metadata do
+      label :sos
+      description 'Perform extra log collection for sosreport'
+      param :exclude, 'List of actions to exclude (comma separated)'
+    end
+
+    def run
+      # Put new actions here and create method below. Make sure to create
+      # patch into sosreport with "--exclude new_method" as well.
+      @all_actions = %w(create_directory excluded_actions)
+      @excluded_actions = @exclude.split(',').map(&:strip)
+      @actions = @all_actions - @excluded_actions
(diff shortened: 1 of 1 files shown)


mbacovsky:
@lzap Thanks for the PR, I like the idea overall. The Debian packaging is a matter of time. Most of the necessary preparation is in https://github.com/theforeman/foreman_maintain/pull/249 implementation of apt manager will follow.

I'd suggest some design changes to better leverage features we have in the framework. What the procedure currently does is it wraps the individual actions and makes it possible to exclude some. We have this functionality already in FM and it is called Scenario. 

If the individual actions are turned into Procedures (tagged with e.g. :sosreport) Scenario can detect and run all available procedures with the tag. Advantage of this approach is we can let FM automatically discover if the action is applicable on the instance i.e. it won't run Pulp related actions on pure foreman, it can call different actions on downstream etc. With scenario we can exclude some steps  with `--whitelist <steps>` out of the box. 

It would also require to create new CLI command to wrap it such as `foreman-maintain sos report|upload|extras` or similar. The `report` could call sosreport, we could add other related helper commands like `upload` to have everything at one place.  `foreman-maintain sos extras --whitelist a,b` would be the command called by sosreport to get the extra bits. If we don't want let users call sosreport via foreman-maintain we can just add `foreman-maintain advanced sos-extras` to generate the extras. However the original idea was to make the different Foreman/Satellite related maintenance tasks reachable from foreman-maintain.
 
Same approach to collect and run the tagged procedures you can find in the `foreman-maintain health` command (https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/cli/health_command.rb)
Makes sense?

ehelms:
Given the age of the PR, I am opting to close it for now.

PR #255, opened 2019-04-12
Fixes #26598 - add support for ansible 2.8 repo

upadhyeammit:
This adds support of ansible 2.8 repository for Satellite 6.6 . With this enables ansible 2.6 for Satellite 6.4 and 6.5 versions only.

### DIFF
--- a/definitions/features/downstream.rb
+++ b/definitions/features/downstream.rb
@@ -64,7 +64,9 @@ class Features::Downstream < ForemanMaintain::Feature
 
     rh_repos.concat(sat_and_tools_repos(rh_version_major, sat_version))
 
-    enable_ansible_repo(sat_version, rh_repos, rh_version_major)
+    if sat_version > version('6.3')
+      rh_repos << enable_ansible_repo(sat_version, rh_version_major)
+    end
 
     if current_minor_version == '6.3' && sat_version.to_s != '6.4' && (
       feature(:puppet_server) && feature(:puppet_server).puppet_version.major == 4)
@@ -74,11 +76,11 @@ class Features::Downstream < ForemanMaintain::Feature
     rh_repos
   end
 
-  def enable_ansible_repo(sat_version, rh_repos, rh_version_major)
+  def enable_ansible_repo(sat_version, rh_version_major)
     if sat_version >= version('6.6')
-      rh_repos << "rhel-#{rh_version_major}-server-ansible-2.8-rpms"
+      "rhel-#{rh_version_major}-server-ansible-2.8-rpms"
     elsif sat_version >= version('6.4')
-      rh_repos << "rhel-#{rh_version_major}-server-ansible-2.6-rpms"
+      "rhel-#{rh_version_major}-server-ansible-2.6-rpms"
     end
   end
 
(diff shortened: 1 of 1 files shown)


definitions/features/downstream.rb
```diff
@@ -74,6 +74,14 @@ def rh_repos(sat_version)
     rh_repos
   end
 
+  def enable_ansible_repo(sat_version, rh_repos, rh_version_major)
+    if [version('6.4'), version('6.5')].include?(sat_version)
+      rh_repos << "rhel-#{rh_version_major}-server-ansible-2.6-rpms"
+    elsif sat_version == version('6.6')
```
ntkathole: @upadhyeammit Will be good if we keep it  >= 6.6. This way we can avoid further changes for 6.7 as well if we keep using same ansible version.

definitions/features/downstream.rb
```diff
@@ -74,6 +74,14 @@ def rh_repos(sat_version)
     rh_repos
   end
 
+  def enable_ansible_repo(sat_version, rh_repos, rh_version_major)
+    if [version('6.4'), version('6.5')].include?(sat_version)
```
kgaikwad: This condition is going to fail for z-stream versions i.e `6.4.z & 6.5.z`

Could you please modify this condition to `sat_version >= version('6.4') && sat_version < version('6.6')` or I would suggest to modify if..else block similar to 

```

if sat_version >= version('6.6')
   rh_repos << "rhel-#{rh_version_major}-server-ansible-2.8-rpms"
elsif sat_version >= version('6.4')
   rh_repos << "rhel-#{rh_version_major}-server-ansible-2.6-rpms"
end
```

definitions/features/downstream.rb
upadhyeammit: @ntkathole handled this condition. Marking it as fixed. Thank You !

definitions/features/downstream.rb
upadhyeammit: @kgaikwad Thank you for help, changed this and its working perfectly!

kgaikwad: @upadhyeammit, 
I would except that new method should return appropriate ansible_repo name instead of modifying `rh_repos` inside new method. 
For example -
`rh_repos <<  {your_new_method_call(sat_version)} if sat_version > version('6.3')`

ntkathole:
Looks good to me :+1:

definitions/features/downstream.rb
```diff
@@ -74,6 +76,14 @@ def rh_repos(sat_version)
     rh_repos
   end
 
+  def enable_ansible_repo(sat_version, rh_version_major)
```
mbacovsky: @upadhyeammit just a nitpick: this method has misleading name as it does not enable anything but rather returns the repo so IMO it should be `ansible_repo`.

mbacovsky:
Besides the small comment it looks good to go.

definitions/features/downstream.rb
upadhyeammit: changed method name, @mbacovsky thank you!

PR #256, opened 2019-04-23
Fixes #26009 - clear login sessions

patilsuraj767:
This will check if the last_value in sessions_id_seq table is less than 2151782769. If it goes above it the there is procedures to clear the sessions.

### DIFF
--- a/definitions/checks/foreman/verify_login_session_count.rb
+++ b/definitions/checks/foreman/verify_login_session_count.rb
@@ -8,7 +8,8 @@ module Checks
         for_feature :foreman_database
       end
 
-      # As session ID is of type integer, server can't add a new session record beyond 2151782969. Therefore assigning MAX_SESSION_ID to 2151782969/2.
+      # As session ID is of type integer, server can't add a new session record beyond 2151782969.
+      # Therefore assigning MAX_SESSION_ID to 2151782969/2.
       MAX_SESSION_ID = 1_075_891_484
 
       def run
--- a/definitions/checks/foreman/verify_login_session_count.rb
+++ b/definitions/checks/foreman/verify_login_session_count.rb
@@ -8,8 +8,7 @@ module Checks
         for_feature :foreman_database
       end
 
-      # The max id value that we can't go beyond is 2151782969.
-      # Therefore setting MAX_SESSION_ID = 2151782969/2
+      # As session ID is of type integer, server can't add a new session record beyond 2151782969. Therefore assigning MAX_SESSION_ID to 2151782969/2.
       MAX_SESSION_ID = 1_075_891_484
 
       def run
--- a/definitions/checks/foreman/verify_login_session_count.rb
+++ b/definitions/checks/foreman/verify_login_session_count.rb
@@ -1,28 +1,32 @@
-class Checks::VerifyLoginSessionCount < ForemanMaintain::Check
-  metadata do
-    label :foreman_login_session
-    tags :default
-    description 'Check the login session count'
-    for_feature :foreman_database
-  end
+module Checks
+  module Foreman
+    class Checks::VerifyLoginSessionCount < ForemanMaintain::Check
+      metadata do
+        label :foreman_login_session
+        tags :default
+        description 'Check the login session count'
+        for_feature :foreman_database
+      end
 
-  # The max id value that we can't go beyond is 2151782969.
-  # Therefore setting MAX_SESSION_ID = 2151782969/2
-  MAX_SESSION_ID = 1_075_891_484
+      # The max id value that we can't go beyond is 2151782969.
+      # Therefore setting MAX_SESSION_ID = 2151782969/2
+      MAX_SESSION_ID = 1_075_891_484
 
-  def run
-    login_session_count = count
-    assert(login_session_count <= MAX_SESSION_ID,
-           "Login session count reached to #{login_session_count} i.e greater than"\
-           " #{MAX_SESSION_ID} which will cause a problem while creating a new session."\
-           ' It needs to reset login session sequence',
-           :next_steps => Procedures::ResetLoginSessionIds.new)
-  end
+      def run
+        login_session_count = count
+        assert(login_session_count <= MAX_SESSION_ID,
+               "Login session count reached to #{login_session_count} i.e greater than"\
+               " #{MAX_SESSION_ID} which will cause a problem while creating a new session."\
+               ' It needs to reset login session sequence',
+               :next_steps => Procedures::ResetLoginSessionIds.new)
+      end
 
-  def count
-    sql = <<-SQL
-    SELECT last_value FROM sessions_id_seq
-    SQL
-    feature(:foreman_database).query(sql).first['last_value'].to_i
+      def count
+        sql = <<-SQL
+        SELECT last_value FROM sessions_id_seq
+        SQL
+        feature(:foreman_database).query(sql).first['last_value'].to_i
+      end
+    end
   end
 end
--- a/definitions/checks/foreman/verify_login_session_count.rb
+++ b/definitions/checks/foreman/verify_login_session_count.rb
@@ -1,6 +1,6 @@
 module Checks
   module Foreman
-    class Checks::VerifyLoginSessionCount < ForemanMaintain::Check
+    class VerifyLoginSessionCount < ForemanMaintain::Check
       metadata do
         label :foreman_login_session
         tags :default
--- a/definitions/checks/foreman/verify_login_session_count.rb
+++ b/definitions/checks/foreman/verify_login_session_count.rb
@@ -13,7 +13,9 @@ class Checks::VerifyLoginSessionCount < ForemanMaintain::Check
   def run
     login_session_count = count
     assert(login_session_count <= MAX_SESSION_ID,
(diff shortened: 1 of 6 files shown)


definitions/checks/foreman_login_session.rb
```diff
@@ -0,0 +1,21 @@
+class Checks::ForemanLoginSession < ForemanMaintain::Check
+  metadata do
+    label :foreman_login_session
+    tags :foreman_login_session
```
mbacovsky: the `tags` define when the check should be run. I guess we want to run this as part of the `health check` command and also with the `pre_upgrade` checks. For that  the tags should be set to `:default`.

definitions/checks/foreman_login_session.rb
```diff
@@ -0,0 +1,21 @@
...
+    description 'Check the login session count'
+  end
+
+  def run
+    login_session_count = count
+    assert(login_session_count <= 2_151_782_769,
```
mbacovsky: @patilsuraj767 Is the value the max id that we can't pass? Would it make sense to set the value to something smaller to increase the chance we catch the situation before it is too late e.g. during the upgrade or regular health checks? Can we count how quickly it grows and estimate when it is going to fail?

Also could we make the value named constant (e.g. MAX_SESSION_ID) to see what we actually match against?

definitions/checks/foreman_login_session.rb
patilsuraj767: Thanks, I will change the tag to `:default`

definitions/checks/foreman_login_session.rb
patilsuraj767: Yes, I will make it as a constant (MAX_SESSION_ID).
Max value we can't pass is 2151782969. Should we set it to 1075891484 (2151782969/2)?

upadhyeammit:
Hello,

I am having few thoughts on this,

1. If I consider 10000 sessions per day and calculate still it takes theoretically many years to reach this max count 2151782969,

Per month - 10000 x 30 = 300000
Per year - 300000 x 12 = 3600000
Years to reach max - 2151782969 / 3600000 = 597.717491389

So is it still good to add this check and procedure here ? Because I feel it is not required.

2. If we are going to add it then I feel we should not delete existing sessions, user could be doing something which is important and deleting session means halting that in between ? Instead I feel we can just restart the sequence ?

3. Just restarting the sequence should not harm because 2151782969 or even half of max 1075891484 is very large number and new session numbers wont reach there before existing sessions are deleted automatically.

patilsuraj767:
@upadhyeammit Thanks for your inputs. 
I agree with you and your calculations. User will surely not encounter this issue/condition if OAuth authentication is used(as it is used in foreman webUI) but what if we make an API call with basic authentication. 
I have tried to fire a GET request with basic auth and observe that sessions_id_seq was increased by 2 and I think count will be more in case of POST request. If any user is using heavy automation with basic auth may reach the limit. 

Do you think that we should only do this check during the upgrade? we can change the tag to  `pre_upgrade` 
Changing the tag to `pre_upgrade` will also take care of the 2nd and 3rd point which you mention. During the upgrade, the user will not perform anything important so we can clean the session at that time. 

Also, I am not sure but if during the upgrade foreman-maintain or foreman-installer is performing `foreman-rake db:sessions:clear` then we can remove the `delete_all_sessions` from the PR and only do `alter_sessions_id_seq`

upadhyeammit:
@patilsuraj767 even if I consider 100 times more queries per day 10000×100 = 1000000(one million) still it takes almost 6 years to reach this max value. I am not against adding check here but what I feel is if its really worst of the worst case then we should not add it. @mbacovsky @kgaikwad what you think ?

mbacovsky:
The customer hit the issue so it may be real case. If we consider large number of hosts sending reports frequently the sessions sequence can grow really quickly so basically I'm not against having the check in general. It does not seem to be expensive code.

The question is what should the remediation step do. I'd like to know opinion of field-eng team as they know best what the customer expects and what are the pitfails of various solution. Could you please check with them @patilsuraj767? I agree with @upadhyeammit that deleting sessions may be on the risky side. It may not hurt to just reset the sequence when it reach MAX/2. That would be my preference.

patilsuraj767:
Hello @mbacovsky and @upadhyeammit ,
Sorry for the delay. I had a word with Alexey Masolov who has filed the bug, and he suggested that to only alter the seq, and no need to delete the existing sessions.
I have made the requisite and commit, Please let me know if there are any more suggestions.

definitions/procedures/foreman_login_session.rb
```diff
@@ -0,0 +1,19 @@
+class Procedures::ForemanLoginSession < ForemanMaintain::Procedure
```
mbacovsky: @patilsuraj767 I'm sorry for missing that earlier but the name and description of the remediation procedure should provide more precise information about what we are going to do. As a customer I may think that 'Clean login sessions' may do more than I want. Would 'Reset login session id sequence' sound appropriate? Also from class name the label of the procedure is deduced so It may be handy to have a verb in it for better orientation (ResetLoginSessionIds perhaps?). Also the file name would need some update.

mbacovsky:
@patilsuraj767 thanks for the research it made things much clearer! The code looks good to go now. I just noticed the procedure name I'm concerned about, please see the details inline. Do you think it is something worth fixing?

definitions/procedures/foreman_login_session.rb
patilsuraj767: @mbacovsky Thanks for suggesting that. Should I give the file name as reset_login_session_id_sequence.rb and the class name as ResetLoginSessionIds? Also, Should I change the description with the same?

ares:
what is the status on this one?

kgaikwad:
@patilsuraj767, any updates on requested changes?

patilsuraj767:
@ares @kgaikwad I have submitted the required changes. Sorry for the delay.

definitions/checks/foreman/verify_login_session_count.rb
```diff
@@ -1,8 +1,9 @@
-class Checks::ResetLoginSessionIds < ForemanMaintain::Check
+class Checks::VerifyLoginSessionCount < ForemanMaintain::Check
```
kgaikwad: In addition to this, you will need to move your check class under `foreman` module.

```
module Checks
  module Foreman
      class VerifyLoginSessionCount < ForemanMaintain::Check
          ....
      end
  end
end
```

definitions/procedures/foreman/reset_login_session_id_sequence.rb
```diff
@@ -1,11 +1,12 @@
 class Procedures::ResetLoginSessionIds < ForemanMaintain::Procedure
```
kgaikwad: Similarly,

```
module Procedures::Foreman
  class ResetLoginSessionIds < ForemanMaintain::Procedure
      ...
  end
end
```

definitions/checks/foreman/verify_login_session_count.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Foreman
+    class VerifyLoginSessionCount < ForemanMaintain::Check
+      metadata do
+        label :foreman_login_session
+        tags :default
+        description 'Check the login session count'
+        for_feature :foreman_database
+      end
+
+      # The max id value that we can't go beyond is 2151782969.
```
kgaikwad:
```suggestion
      # As session ID is of type integer, server can't add a new session record beyond 2151782969. Therefore assigning MAX_SESSION_ID to 2151782969/2.
```

definitions/procedures/foreman/reset_login_session_id_sequence.rb
```diff
@@ -0,0 +1,23 @@
+module Procedures::Foreman
+  class ResetLoginSessionIds < ForemanMaintain::Procedure
+    metadata do
+      description 'Reset Login Session Ids'
+      for_feature :foreman_database
+    end
+
+    def run
+      with_spinner('Reset the login session sequence:') do |spinner|
+        spinner.update 'Reseting sessions id sequence'
```
kgaikwad:
```suggestion
        spinner.update 'Resetting sessions id sequence'
```

juliantaylor:
Wouldn't it be simpler to set the sequence to the same width as the session id (4 bytes currently) and have it cycle?
https://www.postgresql.org/docs/9.5/sql-createsequence.html

mbacovsky:
@kgaikwad, cycle on 4 bytes sounds like a reasonable compromise. It should fix the issue and 4 bytes should give us enough room not to have conflicting session ids.

ares:
What is the status here? I see 2 requesting changes reviews, but they seem rather small. It would be awesome to get this in soon.

ezr-ondrej:
So now with the MySQL dropped could we just do: theforeman/foreman#7441?

ezr-ondrej:
> So now with the MySQL dropped could we just do: [theforeman/foreman#7441](https://github.com/theforeman/foreman/pull/7441)?

This is in, can we close this then? Or do we want to have both? The sequence should automatically cycle now.

kgaikwad:
Thank you @patilsuraj767 for your contribution and thanks to all who has been involved in reviewing.

Having cycle for this column at DB level sounds legitimate and will solve the issue so closing this PR on behalf of [theforeman/foreman#7441](https://github.com/theforeman/foreman/pull/7441). For older versions, there is a knowledge based [document](https://access.redhat.com/solutions/3660801) for reference. Thank you @ezr-ondrej

PR #259, opened 2019-04-29
Fixes #26698 - Add steps to drop foreman_docker

xprazak2:
This will need to be a part of 6.6 upgrade scenario once we have it.


ares:
Is there something @san7ket could test for us here? When do we create such scenario? Who usually adds it?

mbacovsky:
@ares without the scenario the procedures can be tested manually with `foreman-maintain advanced procedure run remove-foreman-docker` (not tested).

The scenario is usually added manually by f-m devs during the dev cycle.

san7ket:
@jameerpathan111 could you be able to add the coverage for this in FM automation https://github.com/SatelliteQE/testfm

jameerpathan111:
> @jameerpathan111 could you be able to add the coverage for this in FM automation https://github.com/SatelliteQE/testfm

@san7ket yes, I can add coverage for this.

xprazak2:
I added a separate commit that aims to add the docker removal scenario to the upgrade, but I am not sure what the best approach is. @mbacovsky, will the tags be enough or should I rather add the steps directly into `compose` method in `PostMigrations`? Or is there a way to "compose" whole scenarios?

jameerpathan111:
@xprazak2 can you fix conflicts?

xprazak2:
I rebased and added `remove` for the package manager.

mbacovsky:
@xprazak2 Thanks for the update! I quite like the idea of creating some sub-scenarios as you suggest. Unfortunately this is not possible at the moment and will need some more investigation what caveats nesting of the scenarios brings in. 

If the docker removal is just a upgrade-to-6.6 thing I'd suggest to add the steps directly to compose of the upgrade scenario. At current state of things I believe the new `post_migration` scenario will break the upgrade as exactly one post_migration scenario is allowed per upgrade.

xprazak2:
Thank you for the hints @mbacovsky! I added the steps directly to the upgrade scenario.

jameerpathan111:
@xprazak2 @mbacovsky 

I have tried Docker remove PR in Satellite 6.5 to Satellite 6.6 upgrade.
Satellite was upgraded successfully, I didn't face any issues while upgrade process.
After satellite upgrade, Docker removal part was executed, and was completed successfully.
The Satellite on which I did upgrade was not populated though.
Output of docker removal part and it's log: [docker_remove_FM_pr.txt](https://github.com/theforeman/foreman_maintain/files/3236599/docker_remove_FM_pr.txt)

ares:
@mbacovsky anything else to get this in?

ares:
awesome news, could you please also let us know, when the new release with this will be made available?

mbacovsky:
@ares tomorrow at latest so it should make it to the next snap.

PR #261, opened 2019-05-16
Fixes #26779 - Check server response via non-auth call

mbacovsky:
This patch reimplements hammer ping to use simple net::http
call instead of hammer. The main advantage is we can easily
skip the hammer auth setup which is not required for the call.
Another advantage is we can extend ping to pure foreman
and proxy.

Internaly the ping implementation moved to the instance feature,
related CLI option was kept for backward compatibility. The ping?
now returns true/false and the additional details such as error
message and failing services can be checked using additional
methods. Restrictions to :katello feature were removed as
the ping can query all types of instances now.


definitions/features/instance.rb
```diff
@@ -44,4 +49,101 @@ def postgresql_local?
...
+
+  def ping?
+    if feature(:katello)
+      katello_ping
+    elsif external_proxy?
+      proxy_ping
+    else
+      foreman_ping
+    end
+    last_ping_result
```
iNecas: Why not passing the result as a return of `{katello,proxy,foreman}_ping`? Reads a bit weird this way

definitions/features/instance.rb
```diff
@@ -44,4 +49,101 @@ def postgresql_local?
...
+      'candlepin_auth' =>  %w[candlepin candlepin_database],
+      'candlepin' => %w[candlepin candlepin_database],
+      'pulp_auth' => %w[pulp mongo],
+      'pulp' => %w[pulp mongo],
+      'foreman_tasks' => %w[foreman_tasks]
+    }
+  end
+
+  def component_services(components)
+    components = [components].flatten(1)
```
iNecas: If the intend is to make sure the `components` is always array, using `Array(components)` would feel a bit better to me

definitions/features/instance.rb
```diff
@@ -44,4 +49,101 @@ def postgresql_local?
...
+
+  def proxy_ping
+    feature(:foreman_proxy).features
+    set_ping_result(true, 'Success', nil)
+  rescue StandardError => e # server error, proxy down
+    set_ping_result(false, "Couldn't connect to the proxy: #{e.message}", nil)
+  end
+
+  def pick_failing_components(components)
+    components.inject([]) do |failing, (name, data)|
```
iNecas: What about using `each_with_object` instead?:

```ruby
compoents.each_with_object([]) do |(name, data), failing|
  failling << name unless data['status'] == 'ok'
end
```

definitions/features/instance.rb
```diff
@@ -44,4 +49,101 @@ def postgresql_local?
...
+    net
+  end
+
+  private
+
+  # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
+  def katello_ping
+    res = server_connection.get('/katello/api/ping')
+    logger.debug('Called /katello/api/ping')
+    logger.debug("Response: #{res.code}, #{res.body}")
```
iNecas: I love the debugging messages :+1:

definitions/features/instance.rb
```diff
@@ -1,8 +1,13 @@
+require 'net/http'
+require 'json'
+
 class Features::Instance < ForemanMaintain::Feature
   metadata do
     label :instance
   end
 
+  attr_reader :last_ping_failing_services, :last_ping_status, :last_ping_result
```
iNecas: Perhaps we could encapsulate this three into one class and make the `ping` method return it, if that's what we need to use it outside of the `ping?` method

mbacovsky: I decided not to go this way originally but I agree it does not look that well in the end. Updated.

definitions/features/instance.rb
mbacovsky: Updated

definitions/features/instance.rb
mbacovsky: Updated

definitions/features/instance.rb
mbacovsky: Updated

mbacovsky:
@iNecas thanks for your valuable input. I've updated the PR. Could you please re-review?

mbacovsky:
Rebased and fixed conflicts

jameerpathan111:
> Ack code-wise. @jameerpathan111 mind verifying from QE perspective: the focus should be on `foreman-maintain service` commands and upgrade workflows working without regression + the foreman-maintain service not requiring the hammer credentials to be stored.

yeah sure.

jameerpathan111:
@iNecas I have tested this PR on various Satellite versions and didn't faced any issues so far.
Now the `foreman-maintain service` command  doesn't require hammer credentials to be stored.
I have also tried upgrade workflow, I did Satellite 6.3 to Satellite 6.4 to Satellite 6.5 to Satellite 6.6
upgrade and they all completed successfully.

To test that service command doesn't require hammer credentials to be stored any more, one of the scenario which I tried is as follow: 
- Have Invalid password in `.hammer/cli.modules.d/foreman.yml `,  `/etc/foreman-maintain/foreman-maintain-hammer.yml` and `/etc/foreman-installer/scenarios.d/satellite-answers.yaml
`
- Run `./bin/foreman-maintain service restart -p`

jameerpathan111:
Capsule related testing is still in progress, will add comment once finished testing this PR on capsule.

jameerpathan111:
> Capsule related testing is still in progress, will add comment once finished testing this PR on capsule.

Tested PR on satellite 6.6 capsule as well and it LGTM.

iNecas:
Everything good from my side. @kgaikwad anything to add from you? Feel free to merge when no additional changes would be requested

PR #262, opened 2019-05-22
fix erros in README

sean797:
Couple of small error I noticed.


README.md
```diff
@@ -1,6 +1,6 @@
 # Foreman Maintenance [![Build Status](https://travis-ci.org/theforeman/foreman_maintain.svg?branch=master)](https://travis-ci.org/theforeman/foreman_maintain) <a href="https://codeclimate.com/github/theforeman/foreman_maintain"><img src="https://codeclimate.com/github/theforeman/foreman_maintain/badges/gpa.svg" /></a>
 
-`foreman_maintain` aims to provide various features that helps keeping the
+`foreman_maintain` aims to provide various features that helps keeps the
```
iNecas: What is wrong with the original phrasing, and is the new one actually correct?

sean797: > What is wrong with the original phrasing

It doesn't make grammatical sense

>  and is the new one actually correct?

No `keeps` should be `keep`.Updated..

iNecas:
Thanks @sean797

PR #263, opened 2019-05-23
Fixes #25549 - [WIP] differentiate between satellite and capsule

kgaikwad:
Changes are still in progress but I would like to share code changes.
In case of queries, I can add those here.


definitions/features/downstream.rb
```diff
@@ -16,13 +20,14 @@ def at_least_version?(version)
...
+    @current_version ||= current_feature.send(:current_version)
   end
 
   def current_minor_version
     current_version.to_s[/^\d+\.\d+/]
   end
 
+  # TODO: Modify activation_key changes as per server
   def setup_repositories(version)
     activation_key = ENV['EXTERNAL_SAT_ACTIVATION_KEY']
```
kgaikwad: Previous [conversation](https://github.com/theforeman/foreman_maintain/commit/4af44714aae88cc5c291a5fd2d0c5d1c5d6527ae#commitcomment-33718141) erased:

> @ntkathole, @jameerpathan111,
Could confirm that does this step work same for capsule as well in case of dogfod servers?Also confirm if ENV variables are same/different for Capsule?

>@kgaikwad It will work if capsule is registered with satellite using activation key and organization.We need to consider following scenarios:
>1. Capsule registered with satellite using environment and organization
subscription-manager register --org="Default_Organization" --environment="Library"
>2. Capsule registered with satellite using activation-key and organization
subscription-manager register --org="Default_Organization" --activationkey="<Activation Key Name>"
>3. Capsule can also be registered with directly CDN.
In this case we also need to take care right red hat repos should be enabled on capsule.

kgaikwad:
@mbacovsky,
Restore [scenario](https://github.com/kgaikwad/foreman_maintain/blob/25549_diff_between_sat_and_capsule/definitions/scenarios/restore.rb#L70) - Does this applicable to capsule as well or should I replace :downstream to :satellite?

definitions/features/capsule.rb
```diff
@@ -0,0 +1,15 @@
+require 'features/foreman_proxy'
+
+class Features::Capsule < Features::ForemanProxy
+  metadata do
+    label :capsule
+
+    confine do
+      smart_proxy? && feature(:installer) && feature(:installer).last_scenario.eql?('capsule')
```
mbacovsky: This confine won't work for installer without scenarios (<=6.2). We should confirm if we still want to support those.

definitions/features/smart_proxy.rb
```diff
@@ -0,0 +1,11 @@
+require 'features/foreman_proxy'
+
+class Features::SmartProxy < Features::ForemanProxy
```
mbacovsky: What is the difference between SmartProxy and ForemanProxy?

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -44,7 +44,7 @@ def check_max_version(name, maximal_version)
         end
       end
 
-      def downstream_installation?
+      def satellite_installation?
```
mbacovsky: This should be moved to the satellite feature and should be using feature(:package_manager) instead of direct usage of rpm

definitions/procedures/candlepin/delete_orphaned_records_from_env_content.rb
```diff
@@ -6,7 +6,7 @@ class DeleteOrphanedRecordsFromEnvContent < ForemanMaintain::Procedure
 
       confine do
         feature(:candlepin_database) &&
-          feature(:downstream) &&
+          feature(:downstream) && feature(:downstream).satellite? &&
```
mbacovsky: Why not `feature(:satellite) && feature(:downstream).current_minor_version == '6.2'`

definitions/features/capsule.rb
mbacovsky: The `smart_proxy?` limits this to external capsules only. Is this intentional? IMO we should get rid of `smart_proxy?` helper completely once this is covered by the features. We don't want to have duplicity in this logic. When we inherit the from the feature can we call the 'super' confine?

definitions/checks/candlepin/validate_db.rb
```diff
@@ -12,7 +12,8 @@ class ValidateDb < ForemanMaintain::Check
     def run
       result, result_msg = feature(:candlepin_database).execute_cpdb_validate_cmd
       next_steps = []
-      if feature(:downstream) && feature(:downstream).current_minor_version == '6.2'
+      if feature(:downstream) && feature(:downstream).satellite? &&
```
mbacovsky: Why not `feature(:satellite) && feature(:satellite).current_minor_version == '6.2'`?

definitions/procedures/candlepin/delete_orphaned_records_from_env_content.rb
mbacovsky: Or better `feature(:satellite) && feature(:satellite).current_minor_version == '6.2'` if we remove the :downstream.

definitions/features/capsule.rb
kgaikwad: @mbacovsky
Sure, thank you for pointing it.
I will confirm if we still want to support capsule scenarios for (<= 6.2).

Verified that `smart_proxy?` method not used anywhere so we can remove this completely. 
One query - do you want me to modify behavior of `capsule` in such way which will handle both external and in-built proxy in case of downstream?
I thought for external capsule we have different upgrade as well as installation setup than in-built proxy which is why I kept foreman-proxy feature as it is and inherited it.

definitions/features/smart_proxy.rb
kgaikwad: Please correct me if my understanding is wrong.
I thought we have different flows to have external and in-built proxy so it thought to keep foreman_proxy feature. We can keep most of common behavior in foreman-proxy and after inheriting we can override and extend other things in smart_proxy feature (nothing but external proxy). 
I would like to know if there is scope of improvement or better way to handle possible scenarios. I have considered four:
- in-built and external smart-proxy in case of upstream
- same in case of downstream

kgaikwad:
Thank you @mbacovsky for review. 

+1,  for a separate module to avoid code duplication between `satellite` & `capsule`. I will modify the current code modifications. 
I didn't implement it before because I still feel that we might need some sort of single entry point to refer `downstream` implementation as there are many places (For example - checks) which applies to both satellite and capsule. What do you think?

mbacovsky:
@kgaikwad, I understand the point with single entry point but am not sure about it. It sounds reasonable but I'm short of use cases supporting need for existence of such thing. Do you have any? For the checks I can see a check running only on downstream - that could be easily solvable by `feature(:instance).downstream?` which would check if it is Satellite or Capsule. Or we can test `feature(:satellite) || feature (:capsule)` directly. Can you see any drawbacks with this approach?

kgaikwad:
> For the checks I can see a check running only on downstream - that could be easily solvable by `feature(:instance).downstream?` which would check if it is Satellite or Capsule. Or we can test `feature(:satellite) || feature (:capsule)` directly. Can you see any drawbacks with this approach?

+1 sounds good to me. I can use `feature(:instance).downstream?` as I don't want to write condition `feature(:satellite) || feature (:capsule)` at everyplace which is currently using feature(:downstream) and applicable for both features.
May I create separate module to keep common code/methods between these two (example - repo setup)?

definitions/features/smart_proxy.rb
mbacovsky: Ah now I see, thanks for explanation. I believe we decided to stick with Foreman Proxy everywhere to not confuse users with the two names and external proxy is called external proxy.  Do you have any specific flows on your mind? Maybe for the sake of this PR I'd keep just :foreman_proxy and :capsule and add `feature(:foreman_proxy).external?`. If we have any use case where it makes sense to split the proxy to more features we can proceed then. Does it make sense?

kgaikwad:
@mbacovsky, 
I have pushed suggested changes.

Pending work - Working on how to execute confine of parent class with current confine block.

kgaikwad:
@mbacovsky,
I was able to execute parent confine block inside `capsule` by passing parent's block as argument.
```
confine do |super_confine = superclass.metadata[:confine_blocks][0]|      
  instance_exec(&super_confine) && feature(:instance).downstream?
end
 
```
Block is executing properly but while testing this I realized that `capsule` is not considered as a feature. `Capsule` feature class is listed under `foreman-proxy` feature as it is inheriting `foreman-proxy` feature. 
Any suggestions? Should I remove inheritance and create concern for common things?

definitions/features/package_manager.rb
```diff
@@ -28,6 +28,22 @@ def manager
                  end
   end
 
+  def satellite_installed?
+    installed?([satellite_package]) || (find_package('foreman') =~ /sat.noarch/)
```
upadhyeammit: The satellite_package method returns 'satellite' statically, so above line will never end up using find_package method ?

kgaikwad: I am not sure from what version of satellite, it contains rpm package called `satellite`. If it is from satellite 6.2 then probably then we don't need a second condition.

mbacovsky: @kgaikwad I don't think PackageManager is proper place for these methods. The class should be generic tool for package manipulation and we should avoid putting inside methods with specific packages.

Would it be possible to make feature(:satellite) the single point of truth and doing the check in the confine?
Then in the feature(:instance) we could have `confine do feature(:satellite) || feature(:capsule) end`
Wouldn't it make better sense? Similar for capsule.

definitions/features/satellite.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def current_version
+    @current_version ||= rpm_version(package_name) || version_from_source
+  end
+
+  private
+
+  def package_name
+    feature(:package_manager).satellite_package
```
mbacovsky: The satellite feature should be primary source for the satellite package name. Not the package manager.

definitions/features/capsule.rb
```diff
@@ -0,0 +1,31 @@
...
+    confine do
+      find_package('foreman-proxy') && feature(:instance).downstream?
+    end
+  end
+
+  def internal?
+    !!feature(:foreman_server)
+  end
+
+  def external?
```
mbacovsky: Is there anything else besides internal and external? Why external? is not !internal? Do we need it at all?

definitions/features/capsule.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def current_version
+    @current_version ||= rpm_version(package_name)
+  end
+
+  private
+
+  def package_name
+    feature(:package_manager).capsule_package
```
mbacovsky: package name shouldn't be part of package_manager

definitions/features/instance.rb
```diff
@@ -45,7 +45,19 @@ def postgresql_local?
...
+
+  def downstream?
+    feature(:package_manager).satellite_installed? || feature(:package_manager).capsule_installed?
+  end
+
+  def downstream
+    @downstream ||= (feature(:satellite) || feature(:capsule))
+  end
+
+  def proxy_feature
```
mbacovsky: I'm not very convinced about the concept of the proxy_feature. Wouldn't it be clearer if we have capsule feature as more specific foreman_proxy feature with most of the functionality shared. We would have foreman_proxy feature on both upstream and downstream proxy and capsule feature only on the downstream one. 
This way we could use foreman_proxy if we don't care about up or downstream and test for the capsule feature if we do. We wouldn't need proxy_feature then.

@kgaikwad What do you think?

definitions/features/package_manager.rb
kgaikwad: Yes, makes sense! :+1:

kgaikwad: >Then in the feature(:instance) we could have confine do feature(:satellite) || feature(:capsule) end

Should I add feature(:katello) as well in confine condition? I think few users still use foreman_maintain for `backup`?

kgaikwad: Sorry, now I remember why I have moved these methods i.e `satellite_installed? as well as capsule_installed?` to feature package_manager. The reason behind this is that I need to check `feature(:instance).downstream?` in the confine of feature capsule.

definitions/features/capsule.rb
kgaikwad: Inside `internal?` method, I wanted to avoid checking installer's last_scenario. As we have most of the checks common to internal as well as external capsule where having a feature capsule or downstream? will be enough.
But upgrade scenarios will be valid in case of `feature(:capsule).external?` where it's worth checking installer's last_scenario. 
Please correct me if I am missing anything here.

definitions/features/package_manager.rb
mbacovsky: IMO `feature(:capsule)` should be lower level then the `:instance.downstream?` which is to me rather helper. So `:capsule` should do the work and `downstream?` should just check for existence of `:satellite` or `:capsule`. Is there any use-case where this won't work?

mbacovsky: @kgaikwad I don't get the question about `:katello` and backup. That may be due to my wrong comment - by `confine do feature(:satellite) || feature(:capsule) end` I meant `def downstream?  feature(:satellite) || feature(:capsule) end` - was that it? Sorry for the confusion. `feature(:instance)` is rather helper feature and should be present always with no confine block.

definitions/features/instance.rb
kgaikwad: Understood the concern but I thought, having both features foreman_proxy and capsule on downstream will create lot of confusion for others. Every time, we will need to think of when to use what.

definitions/features/package_manager.rb
kgaikwad: Yes, got it.

definitions/features/capsule.rb
mbacovsky: I'm not that familiar with what we need to distinguish so let me sum up my expectation and then work together on clarifying it:

| Case             | feature                   | detection                                        |                                  |
|------------------|---------------------------|--------------------------------------------------|----------------------------------|
| Proxy            | :foreman_proxy            | package 'foreman-proxy'                          | Includes also downstream proxies |
| Internal Proxy   | :foreman_proxy.internal?  | package 'foreman-proxy' && package 'foreman'     |                                  |
| External Proxy   | !:foreman_proxy.internal? |                                                  |                                  |
| Capsule          | :capsule                  | package 'satellite-capsule'                      | Downstream only                  |
| Internal Capsule | :capsule.internal?        | package 'satellite-capsule' && package 'foreman' |                                  |
| External Capsule | !:capsule.internal?       |                                                  |                                  |

Please help me fill the gaps. 
- in what version was 'satellite-capsule' introduced?
- does satellite-capsule imply foreman-proxy and installer scenario 'Capsule'

upadhyeammit: in what version was 'satellite-capsule' introduced? 
-> The package is available since Capsule 6.2 version

does satellite-capsule imply foreman-proxy and installer scenario 'Capsule' 
-> yes its foreman-proxy but only external one. In downstream we only have Capsule distinguishable when its external. This is because on Satellite we by default have foreman-proxy. Yes and the installer scenario Capsule will be for external system with 'satellite-capsule' installed.

definitions/features/instance.rb
upadhyeammit: @mbacovsky @kgaikwad it seems both of you are having valid statements. If we consider @mbacovsky way then in future if we want to share anything between upstream(foreman-proxy) and downstream(Capsule) then this will make it easier. 

As we are still going to have Capsule exclusively for downstream I think we can go with @mbacovsky approach. I am open for more suggestion if any.

definitions/features/capsule.rb
mbacovsky: @upadhyeammit thanks, that means that :capsule.internal? makes no sense. I've updated the table. 

With current deployment model we probably have to care about Sat <= 6.1, correct? Prior 6.2 we had 'capsule-installer' package so I'll update the table again.

definitions/features/instance.rb
mbacovsky: @kgaikwad my hope was the opposite ;).
While you extracted the common stuff and it is shared among both you can use whatever is present. The rule of thumb should be:
- if you do not care about capsule specific features use `:foreman_proxy` if you are interrested in capsule specifically use `:capsule`.

Does it make things any clearer?

definitions/features/capsule.rb
upadhyeammit: When we consider having support of Capsule upgrade for 6.7 I think we can skip considering the less than 6.2 versions ? But anyway having condition like below should not be problematic and costly,

package 'satellite-capsule' || package 'capsule-installer'

So I agree to have above condition for Capsule.

Yes internal Capsule does not make sense because we treat that as Satellite.

definitions/features/instance.rb
kgaikwad: After rule of thumb, all my concerns are resolved :-) 

Does that mean now there is no need to create a separate concern file for common things from foreman_proxy & capsule?

kgaikwad: Thank you @mbacovsky , @upadhyeammit for review.
Done with the code modifications as per suggestions.

kgaikwad:
@mbacovsky, I have done with the modifications as per suggestions.
may I merge @upadhyeammit's [PR](https://github.com/kgaikwad/foreman_maintain/pull/2) into my branch so that we can proceed with the further review?
Or should I close this PR and @upadhyeammit will create new PR?

definitions/features/instance.rb
mbacovsky: @kgaikwad, thanks for the update. The code looks way cleaner now. It seems ready to get merged except for one thing I don't understand. The thing is purpose of this `proxy_feature` method. If I'm not mistaken capsule depends on foreman-proxy package therefore `feature(:foreman_proxy)` is present always on upstream and downstream proxy. Actually it is present always as it represents also the internal proxy. So it seems the `proxy_feature` is always true and  equals `feature(:foreman_proxy)`, correct? Or do I miss something?

Also the only usage of the method is probably incorrect. Can we remove this method completely?

definitions/procedures/backup/config_files.rb
```diff
@@ -35,7 +35,8 @@ def config_files
       exclude_configs = []
       ForemanMaintain.available_features.each do |feature|
         # exclude proxy as it has special handling later
-        next if feature == feature(:foreman_proxy)
+        next if feature == feature(:instance).proxy_feature
```
mbacovsky: My guess is `feature(:instance).proxy_feature` will always be equal to `feature(:foreman_proxy)` but we need to exclude also `feature(:capsule)` as we want to avoid collecting its configs at this moment. So the code should be:
```ruby
next if feature == feature(:foreman_proxy)
next if feature(:capsule) && feature == feature(:capsule)
```

mbacovsky:
@kgaikwad the updated PR looks pretty well. I found one little inconvenience and commented in the code. When this is corrected/explained we can merge this.

definitions/procedures/backup/config_files.rb
kgaikwad: Sorry, my mistake! I will correct that.

definitions/features/instance.rb
kgaikwad: Decided to refactor this code at the end but I forgot to update :-(
Removed this method completely.

kgaikwad:
Thank you so much @mbacovsky for review 👌 :+1:

PR #270, opened 2019-06-25
Fixes #27122 - Sat 6.6.z scenario

upadhyeammit:
This adds Satellite 6.6.z upgrade scenario.

### DIFF
--- a/definitions/scenarios/upgrade_to_satellite_6_6_z.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_6_z.rb
@@ -1,0 +1,80 @@
+module Scenarios::Satellite_6_6_z
+  class Abstract < ForemanMaintain::Scenario
+    def self.upgrade_metadata(&block)
+      metadata do
+        tags :upgrade_to_satellite_6_6_z
+        confine do
+          feature(:downstream) && feature(:downstream).current_minor_version == '6.6'
+        end
+        instance_eval(&block)
+      end
+    end
+  end
+
+  class PreUpgradeCheck < Abstract
+    upgrade_metadata do
+      description 'Checks before upgrading to Satellite 6.6.z'
+      tags :pre_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+      add_step(Checks::Repositories::Validate.new(:version => '6.6'))
+    end
+  end
+
+  class PreMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures before migrating to Satellite 6.6.z'
+      tags :pre_migrations
+    end
+
+    def compose
+      add_steps(find_procedures(:pre_migrations))
+      add_step(Procedures::Service::Stop.new)
+    end
+  end
+
+  class Migrations < Abstract
+    upgrade_metadata do
+      description 'Migration scripts to Satellite 6.6.z'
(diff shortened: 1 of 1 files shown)


definitions/scenarios/upgrade_to_satellite_6_6_z.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  class PostMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures after migrating to Satellite 6.6.z'
+      tags :post_migrations
+    end
+
+    def compose
+      add_step(Procedures::Service::Start.new)
```
kgaikwad: There is difference in steps of post-migrations phase from scenario 6.6 to 6.6.z.
Below steps are missing which might be needed for upgrading 6.5 to 6.6.z.

```
add_step(Procedures::ForemanDocker::RemoveForemanDocker.new)
add_step(Procedures::Foreman::ApipieCache.new)
```
@xprazak2, please confirm.

xprazak2: Correct, if someone goes from 6.5 directly to 6.6.z, docker should be dropped as well.

upadhyeammit: @kgaikwad @xprazak2 As maintain wont have upgrade path available from 6.5 to 6.6.z I think we can skip these two steps ?

kgaikwad: +1, in that we can skip those steps.

kgaikwad:
@iNecas, @jameerpathan111,
do you have any comments? or may I merge this?

jameerpathan111:
> @iNecas, @jameerpathan111,
> do you have any comments? or may I merge this?

@kgaikwad It is fine with me.

iNecas:
@iNecas no more comments from me

kgaikwad:
Thank you @upadhyeammit

definitions/scenarios/upgrade_to_satellite_6_6_z.rb
ntkathole: @upadhyeammit @kgaikwad Sorry for replying late here. But just a scenario came in mind. Let's consider what if user is performing upgrade from 6.5 to 6.6 and after yum update, somehow installer failed in first try and user wants to retry upgrade. At this point we will have 6.6.z target version and considering installer completed successfully in second run, but, this will not remove docker as we have not defined this steps for 6.6.z scenario. What do you think?

iNecas: The current behavior should remember that the upgrade from 6.5 to 6.6 is in progress, until it finishes successfully, so in this case, it would not have 6.6.z even though there are already packages from 6.6 installed.

ntkathole: yes, true. It may affect only when somehow data.yml is removed/changed.

kgaikwad: Thanks @ntkathole. To avoid this case, merging [PR-272](https://github.com/theforeman/foreman_maintain/pull/272) which will add docker removal steps to 6.6.z scenario.

If user tries to do upgrade for 6.6.z after successful 6.6 upgrade, nothing to worry about it as Docker removal procedure has condition which verifies presence of docker_package before proceeding further.

PR #271, opened 2019-06-27
Fixes #27173 - fixes procedure hammer-setup 

jameerpathan111:
This patch fixes hammer-setup failing because of unable to find password from satellite-answers.yaml file. In previous satellite versions, we were using variable admin_password. Now it has is been changed to initial_admin_password.

### DIFF
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -85,7 +85,7 @@ class Features::Hammer < ForemanMaintain::Feature
     if !ready? && custom_config[:foreman][:password].nil?
       msg = 'Invalid admin password was found in hammer configs. Looking into installer answers'
       logger.info(msg)
-      custom_config[:foreman][:password] = password_from_answers
+      custom_config[:foreman][:password] = feature(:installer).password_from_answers
       save_config_and_check(custom_config)
     end
     custom_config
@@ -107,7 +107,7 @@ class Features::Hammer < ForemanMaintain::Feature
     if admin_password_missing?
       msg = 'Admin password was not found in hammer configs. Looking into installer answers'
       logger.info(msg)
-      custom_config[:foreman][:password] = password_from_answers
+      custom_config[:foreman][:password] = feature(:installer).password_from_answers
     end
     save_config_and_check(custom_config)
     custom_config
@@ -145,15 +145,6 @@ class Features::Hammer < ForemanMaintain::Feature
         ForemanMaintain::Utils::HashTools.deep_merge!(@configuration, config)
         @config_files << file_path
       end
-    end
-  end
-
-  def password_from_answers
-    return nil unless feature(:installer)
-    if feature(:downstream).current_minor_version >= '6.6'
-      feature(:installer).answers['foreman']['initial_admin_password']
-    else
-      feature(:installer).answers['foreman']['admin_password']
     end
   end
 
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -71,7 +71,7 @@ class Features::Hammer < ForemanMaintain::Feature
   end
 
   def admin_username
-    return nil unless feature(:installer)
+    return 'admin' unless feature(:installer)
     if check_min_version('foreman', '1.22')
       feature(:installer).answers['foreman']['initial_admin_username']
     else
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -28,7 +28,7 @@ class Features::Hammer < ForemanMaintain::Feature
   def setup_admin_access
     return true if check_connection
     logger.info('Hammer setup is not valid. Fixing configuration.')
-    custom_config = { :foreman => { :username => admin_username } }
+    custom_config = { :foreman => { :username => 'admin' } }
     custom_config = on_invalid_host(custom_config)
     custom_config = on_missing_password(custom_config) # get password from answers
     custom_config = on_invalid_password(custom_config) # get password from answers
@@ -70,15 +70,6 @@ class Features::Hammer < ForemanMaintain::Feature
     execute("#{command_base} #{args}")
   end
 
-  def admin_username
-    return 'admin' unless feature(:installer)
-    if check_min_version('foreman', '1.22')
-      feature(:installer).answers['foreman']['initial_admin_username']
-    else
-      feature(:installer).answers['foreman']['admin_username']
-    end
-  end
-
   private
 
   def on_invalid_host(custom_config)
@@ -91,7 +82,8 @@ class Features::Hammer < ForemanMaintain::Feature
   end
 
   def on_invalid_password(custom_config)
-    if !ready? && custom_config[:foreman][:password].nil?
+    username = custom_config[:foreman][:username]
+    if !ready? && custom_config[:foreman][:password].nil? && (username == username_from_answers)
       msg = 'Invalid admin password was found in hammer configs. Looking into installer answers'
       logger.info(msg)
       custom_config[:foreman][:password] = password_from_answers
@@ -114,7 +106,8 @@ class Features::Hammer < ForemanMaintain::Feature
   end
 
   def on_missing_password(custom_config)
-    if admin_password_missing?
+    username = custom_config[:foreman][:username]
+    if admin_password_missing? && (username == username_from_answers)
       msg = 'Admin password was not found in hammer configs. Looking into installer answers'
       logger.info(msg)
       custom_config[:foreman][:password] = password_from_answers
@@ -155,6 +148,15 @@ class Features::Hammer < ForemanMaintain::Feature
         ForemanMaintain::Utils::HashTools.deep_merge!(@configuration, config)
         @config_files << file_path
       end
+    end
+  end
+
+  def username_from_answers
+    return nil unless feature(:installer)
+    if check_min_version('foreman', '1.22')
+      feature(:installer).answers['foreman']['initial_admin_username']
+    else
+      feature(:installer).answers['foreman']['admin_username']
     end
   end
 
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -28,7 +28,7 @@ class Features::Hammer < ForemanMaintain::Feature
   def setup_admin_access
     return true if check_connection
     logger.info('Hammer setup is not valid. Fixing configuration.')
-    custom_config = { :foreman => { :username => 'admin' } }
+    custom_config = { :foreman => { :username => username } }
     custom_config = on_invalid_host(custom_config)
     custom_config = on_missing_password(custom_config) # get password from answers
     custom_config = on_invalid_password(custom_config) # get password from answers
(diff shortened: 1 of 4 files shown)


definitions/features/hammer.rb
```diff
@@ -197,11 +197,6 @@ def load_from_file(file_path)
     end
   end
 
-  def password_from_answers
-    return nil unless feature(:installer)
```
kgaikwad: That's good you encapsulated logic inside :installer feature.
In your changes, nil handling is missing.

jameerpathan111: @kgaikwad add nil handling back, and reverted changes of e1369ac

jameerpathan111:
This issue seems to occur because of following[1] change in foreman-installer
[1] [https://github.com/theforeman/foreman-installer/commit/575a642380833b4a4b1ca9d186e8ded26b4fd831](https://github.com/theforeman/foreman-installer/commit/575a642380833b4a4b1ca9d186e8ded26b4fd831)

definitions/features/hammer.rb
```diff
@@ -199,7 +199,11 @@ def load_from_file(file_path)
 
   def password_from_answers
     return nil unless feature(:installer)
-    feature(:installer).answers['foreman']['admin_password']
+    if check_min_version('foreman', '1.22')
+      feature(:installer).answers['foreman']['initial_admin_password']
```
kgaikwad: In latest version, installer uses `initial_admin_username and initial_admin_password`.
Foreman Maintain loads hammer configurations first. If those are invalid then we have [hardcoded](https://github.com/theforeman/foreman_maintain/blob/6b342a5d10186a4e820f95f31899d3a789ba2a57/definitions/features/hammer.rb#L31) username to `admin`. 
I think it would be good if it considers `initial_admin_password` only when initial_admin_username matches with username from foreman_maintain's config.

@iNecas, @mbacovsky, any thoughts?

iNecas: Good call

kgaikwad:
@jameerpathan111, any updates?

definitions/features/hammer.rb
```diff
@@ -28,11 +28,11 @@ def config_directories
   def setup_admin_access
     return true if check_connection
     logger.info('Hammer setup is not valid. Fixing configuration.')
-    custom_config = { :foreman => { :username => 'admin' } }
+    custom_config = { :foreman => { :username => admin_username } }
```
kgaikwad: +1 for this.

definitions/features/hammer.rb
```diff
@@ -28,11 +28,11 @@ def config_directories
...
     custom_config = on_invalid_host(custom_config)
     custom_config = on_missing_password(custom_config) # get password from answers
     custom_config = on_invalid_password(custom_config) # get password from answers
-    ask_password_and_check(custom_config) unless ready?
+    ask_username_password_and_check(custom_config) unless ready?
```
kgaikwad: Do we really need to ask for username?
As foreman-maintain first checks connection using hammer credentials, if that fails to connect it tries to connect using values from installer answers file.

@mbacovsky, any thoughts about this change?

jameerpathan111: @kgaikwad we can remove it and just keep the message saying to check whether the username is correct in `#{custom_config_file})`.

definitions/features/hammer.rb
```diff
@@ -80,6 +80,15 @@ def run(args)
     execute("#{command_base} #{args}")
   end
 
+  def admin_username
+    return nil unless feature(:installer)
```
mbacovsky: shouldn't we return 'admin' instead of nil?

jameerpathan111: makes sense, thanks @mbacovsky 
will update pr.

jameerpathan111: updated pr.

definitions/features/hammer.rb
jameerpathan111: @kgaikwad I have removed this change.

definitions/features/hammer.rb
jameerpathan111: @kgaikwad updated PR according to your suggestions.

definitions/features/hammer.rb
```diff
@@ -92,7 +92,8 @@ def on_invalid_host(custom_config)
   end
 
   def on_invalid_password(custom_config)
-    if !ready? && custom_config[:foreman][:password].nil?
+    username = custom_config[:foreman][:username]
+    if !ready? && custom_config[:foreman][:password].nil? && (username == username_from_answers)
```
mbacovsky: It seems there is error in the original condition. Here if `ready?` is false two things could have happened 1/ there was invalid password in hammer config 2/ There was NO admin password in hammer configs and invalid password in installer answers. This condition should execute the block only in case 1/. (For the 2/ we have already tried the password from answers). So my take on the condition would be: 
`if !ready? && custom_config[:foreman][:password] != password_from_answers`

For username we should test later to not lose the logging.

definitions/features/hammer.rb
```diff
@@ -109,12 +110,14 @@ def ask_password_and_check(custom_config)
...
                   'Is the server down?'
   end
 
   def on_missing_password(custom_config)
-    if admin_password_missing?
+    username = custom_config[:foreman][:username]
+    if admin_password_missing? && (username == username_from_answers)
       msg = 'Admin password was not found in hammer configs. Looking into installer answers'
       logger.info(msg)
       custom_config[:foreman][:password] = password_from_answers
```
mbacovsky: We should test the username in here instead but Id suggest to test the initial_admin_user is the admin we account we need to use (explained above):
`if username_from_answers == feature(:foreman_server).admin_username`

definitions/features/hammer.rb
```diff
@@ -92,7 +92,8 @@ def on_invalid_host(custom_config)
   end
 
   def on_invalid_password(custom_config)
-    if !ready? && custom_config[:foreman][:password].nil?
+    username = custom_config[:foreman][:username]
+    if !ready? && custom_config[:foreman][:password].nil? && (username == username_from_answers)
       msg = 'Invalid admin password was found in hammer configs. Looking into installer answers'
       logger.info(msg)
       custom_config[:foreman][:password] = password_from_answers
```
mbacovsky: We should test the username in here instead but Id suggest to test the initial_admin_user is the admin we account we need to use:
`if username_from_answers == feature(:foreman_server).admin_username`

definitions/features/hammer.rb
```diff
@@ -197,9 +200,22 @@ def load_from_file(file_path)
     end
   end
 
+  def username_from_answers
+    return nil unless feature(:installer)
```
mbacovsky: We may want to move the logic to the installer feature and have here just
```ruby
return nil unless feature(:installer)
feature(:installer).initial_admin_username
```

Also, would it work if we skip the expensive check_min_version and use just
`feature(:installer).answers['foreman']['initial_admin_username'] || feature(:installer).answers['foreman']['admin_username']`?

definitions/features/hammer.rb
```diff
@@ -197,9 +200,22 @@ def load_from_file(file_path)
...
+    if check_min_version('foreman', '1.22')
+      feature(:installer).answers['foreman']['initial_admin_username']
+    else
+      feature(:installer).answers['foreman']['admin_username']
+    end
+  end
+
   def password_from_answers
     return nil unless feature(:installer)
```
mbacovsky: Same suggestions as in username_from_answers.

definitions/features/hammer.rb
mbacovsky: @kgaikwad had good point. I'd just suggest not to use username form the config for comparison (currently it will be always 'admin') but instead use new `feature(:foreman_server).admin_username`.

kgaikwad:
> ```ruby
> ForemanMaintain.config.admin_username || (feature(:installer) && feature(:installer).initial_admin_username) || 'admin'
> ```
@mbacovsky, +1 for the suggested changes. 

Few things that I would like to add here.
I believe that  `ForemanMaintain.config.admin_username` that you mentioned is to refer username from file `foreman-maintain-hammer.yml`. If this the case then @jameerpathan111 needs to load content of this file in `hammer` feature .  Note that  `ForemanMaintain.config` refers to content of file `foreman_maintain.yml`.

In my opinion, `admin_username` method should be present under hammer feature only. Because, when it comes to foreman-server feature then first preference should go to installer username not to what is present under hammer. Here, it needs admin username for hammer connection.

mbacovsky:
@kgaikwad, my thinking was that `foreman-maintain-hammer.yml` can't be the primary source for admin_username because when hammer setup fail we reset the configuration and try to override the file. So it rather serves as a temporary storage for the actual config. From that perspective it feels better to take initial_admin_username from answers as the primary source with possibility to override in `foreman_maintain.yaml` (just for cases when different admin account is used instead of the initial one). And while I suggested to keep the information in the global config I also suggested to extend ForemanMaintain.config with the new attribute. 

In other words, I suggested to add config option to specify what account is the preferred admin account if there is more then one. And while it is not hammer only thing but also usable for any FM feature where we need to refer to the admin account I suggested to add it to the global config. The default still is initial_admin_username. Does it make better sense now?

definitions/features/hammer.rb
```diff
@@ -82,10 +82,12 @@ def on_invalid_host(custom_config)
   end
 
   def on_invalid_password(custom_config)
-    if !ready? && custom_config[:foreman][:password].nil?
+    if !ready? && custom_config[:foreman][:password] != password_from_answers(
```
kgaikwad: Instead of calling `password_from_answers` method twice, it would be good to catch value in a variable and use it.

definitions/features/hammer.rb
```diff
@@ -99,24 +101,24 @@ def ask_password_and_check(custom_config)
 
   def config_error
     raise ForemanMaintain::HammerConfigurationError, 'Hammer configuration failed: '\
-                  "Is the admin password correct? (it was stored in #{custom_config_file})" \
+                  'Is the admin username and password correct? ' \
```
kgaikwad: Can we reword this message something like 
` "Is the admin credential from the file #{custom_config_file} correct?\n" \
   'Is the server down?'`

kgaikwad:
@jameerpathan111, 
Thank you for making it simple. Added two small inline comments

mbacovsky:
@jameerpathan111 the code looks good. I'll need to do some testing. If you need some help with fixing the tests let us know.

jameerpathan111:
> @jameerpathan111 the code looks good. I'll need to do some testing. If you need some help with fixing the tests let us know.

Thanks @mbacovsky :) 
And yes, I do need some help in fixing tests.

definitions/features/hammer.rb
```diff
@@ -99,24 +100,24 @@ def ask_password_and_check(custom_config)
...
     end
     save_config_and_check(custom_config)
     custom_config
   end
 
   def admin_password_missing?
     configuration[:foreman][:password].nil? ||
-      configuration[:foreman][:password].empty? ||
-      configuration[:foreman][:username] != 'admin'
+      configuration[:foreman][:password].empty?
```
kgaikwad: I think, here condition for username comparison is also required. 
Instead of comparing `configuration[:foreman][:username] != admin`, could you please simply modify the condition as `configuration[:foreman][:username] != username` ?

jameerpathan111: Thanks @kgaikwad for helping me in fixing tests.
I have updated the PR and all the tests are fixed now.

jameerpathan111:
I tried to push new commit to this PR but it didn't reflected on github even after waiting for few minutes so I pushed those changes again after waiting for few minutes. Now I see those commits after some time, I think there was some delay/problem. I hope it won't affect PR.

kgaikwad:
@jameerpathan111, @mbacovsky,
Is that okay to squash few commits into single commit?

jameerpathan111:
> @jameerpathan111, @mbacovsky,
> Is that okay to squash few commits into single commit?

yes, I guess.

mbacovsky:
Sure, all are related to one issue. "Squash and merge" should do that.

PR #273, opened 2019-07-09
Fixes #27253 - Check duplicate entries for role(s)

kgaikwad:


ares:
I haven't tested but by quick reading, it feels it finds all duplicated roles, e.g. `Discovery Viewer`, and deletes them if they are not needed. Does it leave at least one instance of this role? If not, it will be recreated on next Foreman app start. It would be nicer if we could keep one instance.

definitions/procedures/foreman/remove_obsolete_roles.rb
```diff
@@ -0,0 +1,29 @@
...
+      for_feature :foreman_database
+      description 'Remove obsolete roles that are not being used by any user or usergroup'
+      confine do
+        check_min_version('foreman', '1.11')
+      end
+    end
+
+    def run
+      duplicate_role_ids = feature(:foreman_database).query(
+        Checks::Foreman::CheckDuplicateRoles.query_to_get_duplicate_roles
```
upadhyeammit: Is it possible to avoid calling SELECT query twice ? maybe we can store the results in check and pass it to procedure if there are any duplicate roles ?

upadhyeammit:
@kgaikwad I have added one comment, apart from that it looks good to me.

kgaikwad:
@ares, @upadhyeammit,

Modified the code as per suggestion. Now it will keep single instances for roles and remove only duplicate obsolete entries.

Here, the output of commands for reference - 

```
[foreman_maintain]# ./bin/foreman-maintain  health check --label duplicate-roles
Check for duplicate roles from DB:                                    [FAIL]
Duplicate entries found for role(s) - Discovery Reader, Invalid Reader in your DB
1) Remove duplicate obsolete roles from DB
2) Show knowledge base article for troubleshooting
Select step to continue, [n(next), q(quit)] 1
Remove duplicate obsolete roles from DB:                              [OK]      
Check for duplicate roles from DB:                                    [OK]

foreman_maintain]# ./bin/foreman-maintain  health check --label duplicate-roles
Check for duplicate roles from DB:                                    [FAIL]
Duplicate entries found for role(s) - Invalid Reader in your DB
1) Remove duplicate obsolete roles from DB
2) Show knowledge base article for troubleshooting
Select step to continue, [n(next), q(quit)] 2
Show knowledge base article for troubleshooting:                                
Go to https://access.redhat.com/solutions/3998941
please follow steps from above article to resolve this issue
press ENTER once done.

[foreman_maintain]# ./bin/foreman-maintain  advanced procedure run  foreman-remove-duplicate-obsolete-roles
Remove duplicate obsolete roles from DB:                              [OK]
```

definitions/procedures/foreman/remove_obsolete_roles.rb
kgaikwad: As using foreman-maintain, it is possible to run procedure directly that's why I have not added parameterized procedure.

kgaikwad:
@jameerpathan111, 
Adding you as it would be good to have feedback from QA :-)

jameerpathan111:
@kgaikwad on which satellite version do we need to run this check?

kgaikwad:
> @kgaikwad on which satellite version do we need to run this check?

Good Catch! 
I have missed confine block to restrict this check only for Satellite versions < 6.5.
In Sat 6.5, indexing is being added on role name column. For Satellite versions >= 6.5, this check is not required.

mbacovsky:
This PR should fix the failing tests: https://github.com/theforeman/foreman_maintain/pull/275

mbacovsky:
[test]

kgaikwad:
> [test]
Done with rebase!

jameerpathan111:
@kgaikwad 
I have tried this PR on Satellite 6.3 and Satellite 6.4, it worked as expected.
Test steps:
- Created duplicate role using `foreman-rake console`
- Run ` ./bin/foreman-maintain  health check --label duplicate-roles`

mbacovsky:
@kgaikwad well done! I have one question. Can there be duplicate roles that are used? If so and they are not removed by the procedure, can they cause the upgrade failure?

ares:
I was thinking about that, but the case would be extremely rare. If that happens, only the user can fix that manually, before the upgrade starts. Given this is aiming for fixing a known issue, I wouldn'y try to solve it here.

mbacovsky:
@kgaikwad, @ares in such case I expect that after the procedure fixed the duplicate roles the check is re-run and fail again. It may be useful to at least print information that it can't be fixed and needs to be sorted out manually.

definitions/procedures/foreman/remove_duplicate_obsolete_roles.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+      duplicate_role_ids = filter_consumed_roles(roles_hash)
+      remove_obsolete_role_records(duplicate_role_ids)
+    end
+
+    private
+
+    def filter_consumed_roles(roles_hash)
+      consumed_role_ids = find_consumed_role_ids
+      roles_hash.values.map do |ids|
```
mbacovsky: how about something like this:
```ruby
roles_hash.map do |role_name, ids|
  consumed_ids = ids & consumed_role_ids
  if consumed_ids.count == 1
    ids = consumed_ids
  elsif consumed_ids > 1
    # here we could print message
    # ids -= consumed_ids
    # puts "Assigned duplicates of role #{role_name} can not be automatically fixed."
    # or (preferred)
    ids.delete(consumed_ids.min)
    "update user_roles set role_id = #{consumed_ids.min} where role_id in (#{consumed_ids})"
    # or
    # ids -= consumed_ids
    # update roles set name = concat(name, "-", id) where id in (#{consumed_ids})
  elsif ids.length > 1
    ids.delete(ids.min)
  end
  ids
end.flatten
```
It doesn't add extra complexity and should fix also the rare case.

kgaikwad:
> @kgaikwad well done! I have one question. Can there be duplicate roles that are used? If so and they are not removed by the procedure, can they cause the upgrade failure?

Thank you @mbacovsky  for raising this.
As @ares mentioned, chances would be extremely rare where consumed duplicate roles are present. That is why, I have added knowledge document in next_steps which explains how to update role records.
 
In my opinion, it would be good if user modifies such DB records not through the procedure. At a same time I agree with you that we need to give proper information so that user will get proper reason behind next time failure i.e due to consumed duplicate roles.

definitions/procedures/foreman/remove_duplicate_obsolete_roles.rb
kgaikwad: When there is no consumed record for any role present, it's easy to keep a role record with `ids.min`. But when there is more than one consumed records for same role then  user have to decide which one he wants to keep or modify.

mbacovsky: @kgaikwad I see now, that is probably because of the associated entities, correct? Then we may want to give unique names to the duplicate roles. Something like #name-#id could work to unblock the upgrade and be non-obtrusive. Or print a message such as in the first commented out option in the snippet. What do you think?

kgaikwad: @mbacovsky, 
+1 
Updated PR with changes using which it will update name of duplicate consumed role records with `name - #id`.

mbacovsky:
[test]

mbacovsky:
https://github.com/theforeman/foreman_maintain/pull/276 should fix the failing tests

PR #274, opened 2019-07-10
Fixes #27072 - remove extra confirm call

upadhyeammit:
This removes the extra confirm_scenario() call in execute_scenario_steps() method . I feel its extra because,

1. The execute_scenario_steps is getting called by two methods, run and run_scenario.
2. The run_scenario is calling confirm_scenario() method.
3. If we see run method, then it calls run_scenario itself, this means we are running confirm_scenario there on condition.

** Only one thing I feel here is when @rescue_scenario is set we are not having confirm_scenario(), so maybe we can call confirm_scenario() when @rescue_scenario is set in execute_scenario_steps() .
** I can see  confirm_scenario() in execute_scenario_steps() added in pr https://github.com/theforeman/foreman_maintain/pull/164
~~~
    def run
      @scenarios.each do |scenario|
        run_scenario(scenario)
        next unless @quit
        if @rescue_scenario
          logger.debug('=== Rescue scenario found. Executing ===')
          execute_scenario_steps(@rescue_scenario, true)
        end
        break
      end
    end

    def run_scenario(scenario, confirm = true)
      return if scenario.steps.empty?
      raise 'The runner is already in quit state' if quit?

      if confirm
        confirm_scenario(scenario)
        return if quit?
      end

      execute_scenario_steps(scenario)
    ensure
      @last_scenario = scenario unless scenario.steps.empty?
      @exit_code = 1 if scenario.failed?
    end
~~~

### DIFF
--- a/lib/foreman_maintain/runner.rb
+++ b/lib/foreman_maintain/runner.rb
@@ -92,7 +92,6 @@ module ForemanMaintain
 
     def execute_scenario_steps(scenario, force = false)
       scenario.before_scenarios.flatten.each { |before_scenario| run_scenario(before_scenario) }
-      confirm_scenario(scenario)
       return if !force && quit? # the before scenarios caused the stop of the execution
       @reporter.before_scenario_starts(scenario)
       run_steps(scenario, scenario.steps)
--- a/lib/foreman_maintain/runner.rb
+++ b/lib/foreman_maintain/runner.rb
@@ -31,6 +31,7 @@ module ForemanMaintain
       @scenarios.each do |scenario|
         run_scenario(scenario)
         next unless @quit
+
         if @rescue_scenario
           logger.debug('=== Rescue scenario found. Executing ===')
           execute_scenario_steps(@rescue_scenario, true)
@@ -60,6 +61,7 @@ module ForemanMaintain
 
     def confirm_scenario(scenario)
       return unless @last_scenario
+
       decision = if @last_scenario.steps_with_error(:whitelisted => false).any? ||
                     @last_scenario.steps_with_abort(:whitelisted => false).any?
                    :quit
@@ -92,7 +94,9 @@ module ForemanMaintain
 
     def execute_scenario_steps(scenario, force = false)
       scenario.before_scenarios.flatten.each { |before_scenario| run_scenario(before_scenario) }
+      confirm_scenario(scenario) if @rescue_scenario
       return if !force && quit? # the before scenarios caused the stop of the execution
+
       @reporter.before_scenario_starts(scenario)
       run_steps(scenario, scenario.steps)
       @reporter.after_scenario_finishes(scenario)


upadhyeammit:
@kgaikwad @mbacovsky request to check this, I hooked up pry to understand how code flows to remove that extra call. If I am missing anything here to understand then request to add details.

lib/foreman_maintain/runner.rb
```diff
@@ -92,7 +92,6 @@ def storage
 
     def execute_scenario_steps(scenario, force = false)
       scenario.before_scenarios.flatten.each { |before_scenario| run_scenario(before_scenario) }
-      confirm_scenario(scenario)
```
kgaikwad: As per you mentioned, it is added for @rescue_scenario. Instead of removing, would it be possible to add condition here which checks if scenario is equal to @rescue_scenario?

upadhyeammit: Added the change.
Thank you !

mbacovsky:
@upadhyeammit thanks for the patch. I appreciated the detailed explanation!. I have no other concern then what @kgaikwad found (good catch!). 

There are failing tests due to updated Clamp gem that introduces new help formating (https://github.com/mdub/clamp/commit/d18c22d07e71d025e190531fc7c21a43701929c2). We need to update the tests or make them less sensitive for the whitespace, preferably in separate commit or PR.

lib/foreman_maintain/runner.rb
kgaikwad: @upadhyeammit, thanks for adding condition which checks for presence of @rescue_scenario.

Currently, `rescue_scenario` is used by only backup_command.
Not sure but I think, the condition that you added is not enough as `run_scenario` method gets invoked by [backup_command](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/cli/backup_command.rb#L25) where it passes backup_scenario & rescue_scenario as arguments.

There is no any before_scenarios i.e preparation steps for `backup_scenario` argument.
While running this backup_scenario, it will invoke [execute_scenario_steps](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/runner.rb#L51) from `run_scenario`. In this case, the condition that you have added will pass successfully.

So I just want to make sure for `backup_scenario` execution, it should not ask confirmation twice and this confirmation should invoke only when `scenario` inside `execute_scenario_steps` matches with the @rescue_scenario.

mbacovsky:
This PR should fix the failing tests: #275

lib/foreman_maintain/runner.rb
upadhyeammit: I tested this and I dont see confirm call twice for backup and rescue scenarios.

upadhyeammit:
@jameerpathan111 request to have a look. The test failures are of Bash Completion so should you can try out this. Thank You!

jameerpathan111:
> @jameerpathan111 request to have a look. The test failures are of Bash Completion so should you can try out this. Thank You!

sure

upadhyeammit:
Rebased to fix bash test failures.

jameerpathan111:
@upadhyeammit I have tried this PR on Satellite 6.5 and Satellite 6.6 and issue seems to be fixed.
Now pre-migration step in foreman maintain doesn't ask for permission twice.

kgaikwad:
Thanks @upadhyeammit!

PR #275, opened 2019-07-12
Fixes #27291 - Make help tests whitespace insensitive

mbacovsky:
New version of clamp (1.3.1) changed help formating which makes our
output tests fail. assert_cmd was changed to optionaly ignore whitespace
to make the tests pass regardless of what clamp version is used.


mbacovsky:
[test]

PR #280, opened 2019-09-12
Fixes #27842 - excludes foreman_auth when < katello-3.2.0

kgaikwad:
Getting an error `Couldn't connect to the server: undefined method 'to_sym' for nil:NilClass` on katello < 3.2.0.  Returning empty result for `foreman-auth` using katello ping.


mbacovsky:
:+1:  @kgaikwad, thank you!

PR #281, opened 2019-09-16
Fixes #27859 - corrected method_name in certs_tar_exist check

kgaikwad:


kgaikwad:
@mbacovsky,
If users are using backup on an external foreman-proxy frequently then we will need this fix.
While working with satellite & capsule differentiation found this issue but no bug is reported by any of the user yet.

mbacovsky:
@kgaikwad, thank you, let's merge it then.

PR #282, opened 2019-09-17
Fixes #27871 - Don't restore /var/lib/foreman/public

jturel:
I (actually @johnpmitsch!) found this while testing satellite-clone for 6.6. Restoring the webpack assets breaks the symlinks and so some pages just don't work. I can't think of a scenario where we'd need to restore the webpack assets - but please let me know if I've overlooked something.

I think this would affect not only satellite-clone but restoring a backup in general if any of the plugins providing webpack assets have been upgraded since the backup was taken as the symlinks would get broken.

### DIFF
--- a/definitions/features/foreman_server.rb
+++ b/definitions/features/foreman_server.rb
@@ -38,7 +38,7 @@ module ForemanMaintain
 
       def config_files_to_exclude
         [
-          '/var/lib/foreman/public/assets'
+          '/var/lib/foreman/public'
         ]
       end
     end


jturel:
@mbacovsky you might be familiar in general with this kind of problem since you introduced 7859a62 - can you take a look here?

johnpmitsch:
I wonder if this should be `/var/lib/foreman/public` to exclude all assets? I'm trying to think of why they would need to be backed up, they should be provided by foreman/katello packaged gems iiuc

jturel:
> I wonder if this should be `/var/lib/foreman/public` to exclude all assets? I'm trying to think of why they would need to be backed up, they should be provided by foreman/katello packaged gems iiuc

I tested this and it seems to work fine. I'll make the change unless there are objections!

mbacovsky:
@jturel I'm not familiar with flows around webpack but if we want to skip it, the implementation is correct. I like @johnpmitsch 's idea of leaving out the public/ completely. Who can confirm it is safe?

johnpmitsch:
I think @tbrisker could: Would a user have any reason to back up the contents of `/var/lib/foreman/public`?

tbrisker:
all of the assets don't need backing up, but iirc katello drops some rpms there for clients to consume that might need backing up?

jturel:
> all of the assets (webpack included) don't need backing up, but iirc katello drops some rpms there for clients to consume that might need backing up?

True, but that's in /var/www/html/pub

johnpmitsch:
@tbrisker thanks for confirming! looks like the client rpms are in `/var/www/html/pub/`, for example `/var/www/html/pub/katello-ca-consumer-latest.noarch.rpm`. They also aren't modified by the user to my knowledge. 

Sounds like it will be safe to skip `/var/lib/foreman/public`

PR #284, opened 2019-10-01
Fixes #27969 - Use GNU tar to list files

mbacovsky:
Rubygems TarReader has some problems with huge archives.
Switching to GNU tar.


johnpmitsch:
It looks fine to me! What do you think about adding a test to validate the regex returns what we expect?

I likely won't have time to test functionality until Friday because of the meetup in Raleigh, so feel free to have someone else test and approve :)

mbacovsky:
@johnpmitsch updated. I finally realized we don't need any regexp as the non-verbose list (tar -tf <arch>) prints just the filenames. Test added anyway. Also removed dependency on a Feature, we probably don't want this in lib where we do not load definitions while testing.

@upadhyeammit, @kgaikwad, could you please take a look?

lib/foreman_maintain/utils/backup.rb
```diff
@@ -213,16 +213,7 @@ def incremental?
...
-        file_list = []
-        File.open(tarball, 'rb') do |file|
-          ::Zlib::GzipReader.wrap(file) do |gz|
-            ::Gem::Package::TarReader.new(gz) do |tar|
-              tar.each { |entry| file_list << entry.full_name }
-            end
-          end
-        end
-        file_list
+        execute("tar -tf #{tarball}").split("\n")
```
kgaikwad: I think this method can be used when there are multiple config files to verify.
But this method is used only by `validate_hostname?` method to verify config file `etc/httpd/conf/httpd.conf` file exists or not. 

So why not to use `tar` command directly?
For example -  `execute?("tar -tf #{tarball}  #{file_name}")`

What do you think?

mbacovsky: I like the idea

mbacovsky:
@kgaikwad I reduced the the tar calls and removed the list method. Is that acceptable?

jameerpathan111:
> @jameerpathan111, could you please test it once to confirm?

Sure, will test it soon.

jameerpathan111:
@kgaikwad 

The PR has been successfully verified on actual environment(Customer DB upgrade setup).
Test steps:
1- Reproduced the issue mentioned in [BZ1754881](https://bugzilla.redhat.com/show_bug.cgi?id=1754881)
2- Used this PR to see if issue is fixed or not.

Observation:
- foreman-maintain restore completed successfully.

mbacovsky:
Thanks @kgaikwad and @jameerpathan111!

PR #285, opened 2019-10-07
Fixes #28012 - Spelling mistake in backup procedure

xBytez:
I noticed when running backups that Postgres was incorrectly referred to as Postgress.

Small contribution but hopefully a useful one :)


mbacovsky:
@xBytez thanks for your contribution it is much appreciated! :+1:

PR #286, opened 2019-10-21
Fixes #28097 - Allow packages update with no packages

mbacovsky:
This patch lets packages update command to be invoked
without any args to update all packages.

### DIFF
--- a/definitions/procedures/packages/update_all_confirmation.rb
+++ b/definitions/procedures/packages/update_all_confirmation.rb
@@ -8,14 +8,15 @@ module Procedures::Packages
 
     def run
       if @packages.nil? || @packages.empty?
-        question = "\nWARNING: No speciffic packages to update were provided\n" \
+        question = "\nWARNING: No specific packages to update were provided\n" \
           "so we are going to update all available packages.\n" \
           "It is recommended to update everything only as part of upgrade\n" \
           "of the #{feature(:instance).product_name} to the next version. \n" \
           "To Upgrade to next version use 'foreman-maintain upgrade'.\n\n" \
+          "NOTE: --assumeyes is not applicable for this check\n\n" \
           "Do you want to proceed with update of everything regardless\n" \
           'of the recommendations?'
-        answer = ask_decision(question, 'y(yes), q(quit)')
+        answer = ask_decision(question, 'y(yes), q(quit)', ignore_assumeyes: true)
         abort! unless answer == :yes
       end
     end
--- a/definitions/procedures/packages/update_all_confirmation.rb
+++ b/definitions/procedures/packages/update_all_confirmation.rb
@@ -8,12 +8,13 @@ module Procedures::Packages
 
     def run
       if @packages.nil? || @packages.empty?
-        question = "\nWARNING: No speciffic packages to update were provided " \
+        question = "\nWARNING: No speciffic packages to update were provided\n" \
           "so we are going to update all available packages.\n" \
-          'It is recommended to update everything only as part of ' \
-          "upgrade of the #{feature(:instance).product_name} to the next version. \n" \
+          "It is recommended to update everything only as part of upgrade\n" \
+          "of the #{feature(:instance).product_name} to the next version. \n" \
           "To Upgrade to next version use 'foreman-maintain upgrade'.\n\n" \
-          'Do you want to proceed with update of everything regardless of the recommendations?'
+          "Do you want to proceed with update of everything regardless\n" \
+          'of the recommendations?'
         answer = ask_decision(question, 'y(yes), q(quit)')
         abort! unless answer == :yes
       end
--- a/definitions/procedures/packages/update_all_confirmation.rb
+++ b/definitions/procedures/packages/update_all_confirmation.rb
@@ -1,0 +1,22 @@
+module Procedures::Packages
+  class UpdateAllConfirmation < ForemanMaintain::Procedure
+    metadata do
+      param :packages, 'List of packages to update', :array => true
+
+      description 'Confirm update all is intentionall'
+    end
+
+    def run
+      if @packages.nil? or @packages.empty?
+        question = "\nWARNING: No speciffic packages to update were provided " \
+          "so we are going to update all available packages.\n" \
+          "It is recommended to update everything only as part of " \
+          "upgrade of the #{feature(:instance).product_name} to the next version. \n" \
+          "To Upgrade to next version use 'foreman-maintain upgrade'.\n\n" \
+          'Do you want to proceed with update of everything regardless of the recommendations?'
+        answer = ask_decision(question, 'y(yes), q(quit)')
+        abort! unless answer == :yes
+      end
+    end
+  end
+end
--- a/definitions/procedures/packages/update_all_confirmation.rb
+++ b/definitions/procedures/packages/update_all_confirmation.rb
@@ -7,10 +7,10 @@ module Procedures::Packages
     end
 
     def run
-      if @packages.nil? or @packages.empty?
+      if @packages.nil? || @packages.empty?
         question = "\nWARNING: No speciffic packages to update were provided " \
           "so we are going to update all available packages.\n" \
-          "It is recommended to update everything only as part of " \
+          'It is recommended to update everything only as part of ' \
           "upgrade of the #{feature(:instance).product_name} to the next version. \n" \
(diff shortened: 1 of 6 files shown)


ntkathole:
Aren't we allowing user here to perform upgrade with `foreman-maintain packages update` command without actually running `foreman-maintain upgrade run` command ?  Currently I am not sure how will this have side effects, but, can we consider showing warning to use `foreman-maintain upgrade run` command rather than allowing it without args?

mbacovsky:
@ntkathole it is already possible we are making it just a bit more easy to use. I see your concern though, updating everything is probably too risky.

What others do think? Should we try to detect attempt to update everything and prevent it with deferring to f-m upgrade? @kgaikwad, @upadhyeammit, @gnurag

upadhyeammit:
> @ntkathole it is already possible we are making it just a bit more easy to use. I see your concern though, updating everything is probably too risky.
> 
> What others do think? Should we try to detect attempt to update everything and prevent it with deferring to f-m upgrade? @kgaikwad, @upadhyeammit, @gnurag

I think we should put warning but should not prevent it,  because we are doing this change to make it easier to update all packages[if required]. I can think of below use cases, if I missed anything then request to add,

1. Update to packages which are of beta, or next major release.
If user is doing this then he knows what are the implications. Because nowhere in document we have mentioned to manually change repositories to either beta or major release and do upgrade on GA.

2. Update to EPEL packages, or upstream repository packages.
In this use case I tried the after and before in confine block and it did not prioritize run. So here also user should be knowing what actions he is doing.

3. Update to next z stream.
Same as 1. but with less impact, we may or may not need to run the installer again.

Considering all these I feel warning should be sufficient.

kgaikwad:
@mbacovsky, 
As it is a risky operation and considering users don't always read the docs before running any command, I agree with @ntkathole & @upadhyeammit on showing a warning message than updating everything. 
Though `foreman-maintain packages update` command is running a installer command after packages update, there might be a chance that manual step(s) will be involved during upgrade from one version to other. That's why it is better not to leave system in broken state.

mbacovsky:
Thanks everybody for your input. I'm going to print a warning that packages update should not be used instead of `upgrade run`. 

@kgaikwad does your suggestion mean we shouldn't allow `packages update` with no  arg?

kgaikwad:
> Thanks everybody for your input. I'm going to print a warning that packages update should not be used instead of `upgrade run`.
> 
> @kgaikwad does your suggestion mean we shouldn't allow `packages update` with no arg?

Yes, by considering the points(1st & 3rd) mentioned by @upadhyeammit and IIRC, in the past few issues were reported by users around `yum update` command. 
If we would like to implement it in this way then it would be good to have more suggestions on this change.

mbacovsky:
I've added the confirmation.

jameerpathan111:
@mbacovsky 
The warning message is not shown properly when I ran `foreman-maintain packages update --assumeyes` command.

```
Confirm update all is intentionall: 

WARNING: No speciffic packages to update were provided so we are going to update all available packages.
It is recommended to update everything only as part of upgrade of the Satellite to the next version. 
To Upgrade to next version use 'foreman-maintain upgrade'.

                                                                      [OK]      ns? (assuming yes)
Confirm installer run is allowed: 
```
Line `Do you want to proceed with update of everything regardless of the recommendations?, [y(yes), q(quit)] y` is missing.

mbacovsky:
@jameerpathan111 the output should be fixed now. I've seen the same behavior elsewhere too so it should be fixed globally.

@kgaikwad, I agree that allowing users to update everything is not the ideal but it still seems better then forcing users to come up with workaround by providing wildcards or turning the locking off completely. So I used the wording indicating that updating everything is risky and what we recommend but allowed to proceed. Is that acceptable for you? Would it make sense to disable --assumeyes for this situation?

jameerpathan111:
> @jameerpathan111 the output should be fixed now. I've seen the same behavior elsewhere too so it should be fixed globally.

Yes, it seems to be fixed now.

kgaikwad:
> Is that acceptable for you? Would it make sense to disable --assumeyes for this situation?

+1 for the warning message and asking user before proceeding further.
As it is the risky and probably won't be possible to bring back to original state once ran so I think we should avoid having --assumeyes option for this at least in the beginning.

jameerpathan111:
@mbacovsky @kgaikwad if we remove --assumeyes option it'll cause problem in customer automation.

mbacovsky:
@jameerpathan111 I guess that is the point. This shouldn't be automated at all. Currently we do not support update without parameters. The --assumeyes should stay with the command we will just ignore --assumeyes on parameterless update. We can inform that the option was ignored on purpose. Would that be acceptable?

jameerpathan111:
> @jameerpathan111 I guess that is the point. This shouldn't be automated at all. Currently we do not support update without parameters. The --assumeyes should stay with the command we will just ignore --assumeyes on parameterless update. We can inform that the option was ignored on purpose. Would that be acceptable?

yes

mbacovsky:
Updated according to our previous discussion.

Also please note I loosend the HashSyntax cop a bit from Hash_rocket to no_mixed_keys. The reason was it fails on keyword argument and it seems there is no remedy for that. I checked the setting on Foreman core and this cop is turned off there. If anybody has concerns with this change please speak up.

definitions/procedures/packages/update_all_confirmation.rb
```diff
@@ -0,0 +1,24 @@
+module Procedures::Packages
+  class UpdateAllConfirmation < ForemanMaintain::Procedure
+    metadata do
+      param :packages, 'List of packages to update', :array => true
+
+      description 'Confirm update all is intentionall'
```
gnurag: `s/intentionall/intentional/`

mbacovsky: @gnurag thanks, updated.

kgaikwad:
@upadhyeammit, @gnurag, any additional comments?

@jameerpathan111, would you like to do final round of testing with updated changes?

upadhyeammit:
> @upadhyeammit, @gnurag, any additional comments?
> 
> @jameerpathan111, would you like to do final round of testing with updated changes?

Nothing else from my side.

jameerpathan111:
> @jameerpathan111, would you like to do final round of testing with updated changes?

@kgaikwad tested this PR and it works as expected.
Observation:
- able to update all packages using `foreman-maintain packages update`
- `--assumeyes` for this specific scenario is disabled, user will still have to manually confirm if he wants to update all packages.( reason [https://github.com/theforeman/foreman_maintain/pull/286#issuecomment-552456307](https://github.com/theforeman/foreman_maintain/pull/286#issuecomment-552456307) )
    - Except when it asks for permission to run satellite-installer.

This is how output of command now looks like:
```
# foreman-maintain packages update --assumeyes 
Check if tooling for package locking is installed:                    [OK]


Running update packages in unlocked session
Confirm update all is intentional: 

WARNING: No specific packages to update were provided
so we are going to update all available packages.
It is recommended to update everything only as part of upgrade
of the Satellite to the next version. 
To Upgrade to next version use 'foreman-maintain upgrade'.

NOTE: --assumeyes is not applicable for this check

Do you want to proceed with update of everything regardless
of the recommendations?, [y(yes), q(quit)] y
                                                                      [OK]      
Confirm installer run is allowed: 

WARNING: This script runs satellite-installer after the yum execution 
to ensure the Satellite is in a consistent state.
As a result some of your services may be restarted. 

Do you want to proceed? (assuming yes)   ---> Here assumeyes is working.
                                                                      [OK]      
Unlock packages:                                                      [OK]
Update package(s) :                                                   [OK]
==============================================================================================================================================================
 Package                                           Arch   Version                          Repository                                                    Size
==============================================================================================================================================================
Updating:
*****************************
****  Output omitted   ****
*****************************
foreman-rake upgrade:run finished successfully!
Upgrade completed!
Package versions are being locked.
       [OK]
Check status of version locking of packages: 
  Automatic locking of package versions is enabled in installer.
  Packages are locked.                                                [OK]

```

kgaikwad:
Thank you @jameerpathan111  for testing this PR.
Merging this  pull-request..

kgaikwad:
Thank you @mbacovsky :tada: :ok_hand:  

Thanks @upadhyeammit, @gnurag  for reviewing this PR

PR #287, opened 2019-10-25
Fixes #28089 - capsule upgrade support

upadhyeammit:
This adds the scenario for Capsule upgrade and below are changes,

1. The performance check is already designed to run on drives for which corresponding directories are available, i.e if pgsql is not available on Capsule skip check for that. The description of check was generic. Its dynamic now to show spinner message for Capsule and Satellite.
2. The original asset check was not guarded to run on system which has foreman package installed, it has changed to avoid running it on Capsule.
3. The downstream concern now handles satellite-tools repo conditionally for Satellite and Capsule.
4. This PR changes the way to register scenarios for upgrade. New changes only allow either Satellite or Capsule scenarios to be registered for upgrade. The scenario registration logic is in one file instead of having it in every scenario file.

### DIFF
--- a/definitions/checks/disk/performance.rb
+++ b/definitions/checks/disk/performance.rb
@@ -46,15 +46,15 @@ module Checks
       end
 
       def check_only_single_device?
-        @default_dirs.values do |dir|
+        default_dirs.values do |dir|
           ForemanMaintain::Utils::Disk::Device.new(dir).name
         end.uniq.length <= 1
       end
 
       def dirs_to_check
-        return @default_dirs.values.first(1) if check_only_single_device?
+        return default_dirs.values.first(1) if check_only_single_device?
 
-        @default_dirs.values
+        default_dirs.values
       end
 
       private
--- a/definitions/checks/disk/performance.rb
+++ b/definitions/checks/disk/performance.rb
@@ -1,16 +1,10 @@
 module Checks
   module Disk
     class Performance < ForemanMaintain::Check
-      DEFAULT_DIRS = [
-        '/var/lib/pulp', '/var/lib/mongodb', '/var/lib/pgsql'
-      ].select { |file_path| File.directory?(file_path) }.freeze
-
       metadata do
         label :disk_performance
-        description "Check recommended disk speed for #{DEFAULT_DIRS.join(',')} directories."
         tags :pre_upgrade
         preparation_steps { Procedures::Packages::Install.new(:packages => %w[fio]) }
-
         confine do
           feature(:instance).pulp
         end
@@ -38,16 +32,29 @@ module Checks
         end
       end
 
+      def default_dirs
+        @default_dirs ||= %i[pulp mongo foreman_database].inject({}) do |dirs, f|
+          if feature(f) && File.directory?(feature(f).data_dir)
+            dirs[feature(f).label_dashed] = feature(f).data_dir
+          end
+          dirs
+        end
+      end
+
+      def description
+        "Check recommended disk speed for #{default_dirs.keys.join(', ')} directories."
+      end
+
       def check_only_single_device?
-        DEFAULT_DIRS.map do |dir|
+        @default_dirs.values do |dir|
           ForemanMaintain::Utils::Disk::Device.new(dir).name
         end.uniq.length <= 1
       end
 
       def dirs_to_check
-        return DEFAULT_DIRS.first(1) if check_only_single_device?
+        return @default_dirs.values.first(1) if check_only_single_device?
 
-        DEFAULT_DIRS
+        @default_dirs.values
       end
 
       private
--- a/definitions/checks/disk/performance.rb
+++ b/definitions/checks/disk/performance.rb
@@ -1,9 +1,13 @@
 module Checks
   module Disk
     class Performance < ForemanMaintain::Check
+      DEFAULT_DIRS = [
+        '/var/lib/pulp', '/var/lib/mongodb', '/var/lib/pgsql'
+      ].select { |file_path| File.directory?(file_path) }.freeze
+
       metadata do
         label :disk_performance
-        description 'Check for recommended disk speed of pulp, mongodb, pgsql dir.'
+        description "Check recommended disk speed for #{DEFAULT_DIRS.join(',')} directories."
         tags :pre_upgrade
         preparation_steps { Procedures::Packages::Install.new(:packages => %w[fio]) }
 
@@ -14,9 +18,6 @@ module Checks
 
       EXPECTED_IO = 60
       DEFAULT_UNIT = 'MB/sec'.freeze
-      DEFAULT_DIRS = [
-        '/var/lib/pulp', '/var/lib/mongodb', '/var/lib/pgsql'
-      ].select { |file_path| File.directory?(file_path) }.freeze
 
       attr_reader :stats
 
--- a/definitions/checks/disk/performance.rb
+++ b/definitions/checks/disk/performance.rb
@@ -33,7 +33,7 @@ module Checks
       end
 
       def default_dirs
-        @default_dirs ||= %i[pulp mongo foreman_database].inject({}) do |dirs, f|
+        @default_dirs ||= %i[pulp2 pulpcore_database mongo foreman_database].inject({}) do |dirs, f|
           if feature(f) && File.directory?(feature(f).data_dir)
             dirs[feature(f).label_dashed] = feature(f).data_dir
           end
--- a/definitions/scenarios/upgrade_to_capsule_6_7.rb
+++ b/definitions/scenarios/upgrade_to_capsule_6_7.rb
@@ -1,23 +1,23 @@
-module Scenarios::Satellite_6_3
+module Scenarios::Capsule_6_7
   class Abstract < ForemanMaintain::Scenario
     def self.upgrade_metadata(&block)
       metadata do
         tags :upgrade_scenario
         confine do
-          feature(:satellite) && feature(:satellite).current_minor_version == '6.2'
+          feature(:capsule) && feature(:capsule).current_minor_version == '6.6'
         end
         instance_eval(&block)
       end
     end
 
     def target_version
-      '6.3'
+      '6.7'
     end
   end
 
   class PreUpgradeCheck < Abstract
     upgrade_metadata do
-      description 'Checks before upgrading to Satellite 6.3'
+      description 'Checks before upgrading to Capsule 6.7'
       tags :pre_upgrade_checks
       run_strategy :fail_slow
     end
@@ -25,14 +25,13 @@ module Scenarios::Satellite_6_3
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:pre_upgrade))
-      add_step(Checks::RemoteExecution::VerifySettingsFileAlreadyExists.new)
-      add_step(Checks::Repositories::Validate.new(:version => '6.3'))
+      add_step(Checks::Repositories::Validate.new(:version => '6.7'))
     end
   end
 
   class PreMigrations < Abstract
     upgrade_metadata do
-      description 'Procedures before migrating to Satellite 6.3'
+      description 'Procedures before migrating to Capsule 6.7'
       tags :pre_migrations
     end
 
@@ -44,12 +43,12 @@ module Scenarios::Satellite_6_3
 
   class Migrations < Abstract
     upgrade_metadata do
-      description 'Migration scripts to Satellite 6.3'
+      description 'Migration scripts to Capsule 6.7'
       tags :migrations
     end
 
     def compose
-      add_step(Procedures::Repositories::Setup.new(:version => '6.3'))
+      add_step(Procedures::Repositories::Setup.new(:version => '6.7'))
       add_step(Procedures::Packages::UnlockVersions.new)
       add_step(Procedures::Packages::Update.new(:assumeyes => true))
       add_step(Procedures::Installer::Upgrade.new)
@@ -58,7 +57,7 @@ module Scenarios::Satellite_6_3
 
   class PostMigrations < Abstract
     upgrade_metadata do
-      description 'Procedures after migrating to Satellite 6.3'
+      description 'Procedures after migrating to Capsule 6.7'
       tags :post_migrations
     end
 
@@ -70,7 +69,7 @@ module Scenarios::Satellite_6_3
 
   class PostUpgradeChecks < Abstract
     upgrade_metadata do
-      description 'Checks after upgrading to Satellite 6.3'
+      description 'Checks after upgrading to Capsule 6.7'
       tags :post_upgrade_checks
       run_strategy :fail_slow
     end
--- a/definitions/scenarios/upgrade_to_capsule_6_7.rb
+++ b/definitions/scenarios/upgrade_to_capsule_6_7.rb
@@ -4,7 +4,9 @@ module Scenarios::Capsule_6_7
       metadata do
         tags :upgrade_scenario
         confine do
-          feature(:capsule) && feature(:capsule).current_minor_version == '6.6'
+          feature(:capsule) &&
+            (feature(:capsule).current_minor_version == '6.6' || \
+            ForemanMaintain.upgrade_in_progress == '6.7')
         end
         instance_eval(&block)
       end
--- a/lib/foreman_maintain/concerns/downstream.rb
+++ b/lib/foreman_maintain/concerns/downstream.rb
@@ -98,16 +98,21 @@ module ForemanMaintain
       end
 
       def common_repos(rh_version_major, full_version)
-        tools_repo_id = "rhel-#{rh_version_major}-server-satellite-tools-#{full_version}-rpms"
-        maintenance_repo_id = "rhel-#{rh_version_major}-server-satellite-maintenance-6-rpms"
+        repos_arrary = common_repos_array(rh_version_major, full_version)
+        return repos_arrary.first(1) if feature(:capsule)
 
-        # Override to use Beta repositories for sat version until GA
-        if ENV['FOREMAN_MAINTAIN_USE_BETA'] == '1'
-          tools_repo_id = "rhel-#{rh_version_major}-server-satellite-tools-6-beta-rpms"
-          maintenance_repo_id = "rhel-#{rh_version_major}-server-satellite-maintenance-6-beta-rpms"
-        end
+        repos_arrary
+      end
 
-        [tools_repo_id, maintenance_repo_id]
+      def common_repos_array(rh_version_major, full_version)
+        ["rhel-#{rh_version_major}-server-satellite-maintenance-6#{use_beta}-rpms",
+         "rhel-#{rh_version_major}-server-satellite-tools-#{full_version}#{use_beta}-rpms"]
+      end
+
+      def use_beta
+        return '-beta' if ENV['FOREMAN_MAINTAIN_USE_BETA'] == '1'
+
+        nil
       end
 
       def main_rh_repos(rh_version_major)
--- a/lib/foreman_maintain/upgrade_runner.rb
+++ b/lib/foreman_maintain/upgrade_runner.rb
@@ -15,29 +15,8 @@ module ForemanMaintain
       def available_targets
         # when some upgrade is in progress, we don't allow upgrade to different version
         return [current_target_version] if current_target_version
-        versions_to_tags.inject([]) do |available_targets, (version, tag)|
-          if !find_scenarios(:tags => [tag]).empty?
-            available_targets << version
-          else
-            available_targets
-          end
-        end.sort
-      end
 
-      def versions_to_tags
-        @versions_to_tags ||= {}
-      end
-
-      # registers target version to specific tag
-      def register_version(version, tag)
-        if versions_to_tags.key?(version) && versions_to_tags[version] != tag
-          raise "Version #{version} already registered to tag #{versions_to_tags[version]}"
-        end
-        @versions_to_tags[version] = tag
-      end
-
-      def clear_register
-        versions_to_tags.clear
+        find_scenarios(:tags => :upgrade_scenario).map(&:target_version).uniq.sort
       end
 
       def current_target_version
@@ -57,25 +36,35 @@ module ForemanMaintain
 
     def initialize(version, reporter, options = {})
       super(reporter, [], options)
-      @tag = self.class.versions_to_tags[version]
-      raise "Unknown version #{version}" unless tag
       @version = version
+      scenarios_present = find_scenarios(:tags => :upgrade_scenario).any?(&matching_version_test)
+      raise "Unknown version #{version}" unless scenarios_present
+
       @scenario_cache = {}
       self.phase = :pre_upgrade_checks
     end
 
     def scenario(phase)
       return @scenario_cache[phase] if @scenario_cache.key?(phase)
-      condition = { :tags => [tag, phase] }
-      matching_scenarios = find_all_scenarios(condition)
+
+      condition = { :tags => [:upgrade_scenario, phase] }
+      matching_scenarios = find_scenarios(condition).select(&matching_version_test)
       raise "Too many scenarios match #{condition.inspect}" if matching_scenarios.size > 1
+
       @scenario_cache[phase] = matching_scenarios.first
+    end
+
+    def matching_version_test
+      proc do |scenario|
+        scenario.respond_to?(:target_version) && scenario.target_version == @version
+      end
     end
 
     def run
       self.class.current_target_version = @version
       PHASES.each do |phase|
         return run_rollback if quit?
+
         if skip?(phase)
           skip_phase(phase)
         else
@@ -127,6 +116,7 @@ module ForemanMaintain
     # deserializes the state of the run from the storage
     def load
       return unless storage[:serialized]
+
       load_from_hash(storage[:serialized])
     end
 
@@ -134,6 +124,7 @@ module ForemanMaintain
       with_non_empty_scenario(phase) do |scenario|
         confirm_scenario(scenario)
         return if quit?
+
         self.phase = phase
         run_scenario(scenario)
         # if we started from the :pre_upgrade_checks, ensure to ask before
@@ -158,6 +149,7 @@ module ForemanMaintain
     # rubocop:disable Metrics/MethodLength
     def rollback_pre_migrations
       raise "Unexpected phase #{phase}, expecting pre_migrations" unless phase == :pre_migrations
+
       rollback_needed = scenario(:pre_migrations).steps.any? { |s| s.executed? && s.success? }
       if rollback_needed
         @quit = false
@@ -198,6 +190,7 @@ module ForemanMaintain
       unless @scenario_cache.empty?
         raise "Some scenarios are already initialized: #{@scenario_cache.keys}"
       end
+
       self.phase = hash[:phase]
       hash[:scenarios].each do |key, scenario_hash|
         @scenario_cache[key] = Scenario.new_from_hash(scenario_hash)
@@ -229,6 +222,7 @@ module ForemanMaintain
 
     def phase=(phase)
       raise "Unknown phase #{phase}" unless PHASES.include?(phase)
+
       @phase = phase
     end
   end
--- a/test/definitions/checks/disk_performance_test.rb
+++ b/test/definitions/checks/disk_performance_test.rb
@@ -7,8 +7,13 @@ describe Checks::Disk::Performance do
 
   before do
     assume_feature_absent(:mongo)
-    assume_feature_present(:pulp2) || assume_feature_present(:pulp3)
-    check_disk_performance.stubs(:default_dirs).returns(:pulp => '/var/lib/pulp')
+    current_feature = if file_nonzero?('/var/lib/pulp')
+                        :pulp2
+                      else
+                        :pulpcore_database
+                      end
+    assume_feature_present(current_feature)
+    check_disk_performance.stubs(:default_dirs).returns(:pulp => feature(current_feature).data_dir)
   end
 
   it 'should confine existence of fio' do
--- a/test/definitions/checks/disk_performance_test.rb
+++ b/test/definitions/checks/disk_performance_test.rb
@@ -7,13 +7,8 @@ describe Checks::Disk::Performance do
 
   before do
     assume_feature_absent(:mongo)
-    current_feature = if file_nonzero?('/var/lib/pulp')
-                        :pulp2
-                      else
-                        :pulpcore_database
-                      end
-    assume_feature_present(current_feature)
-    check_disk_performance.stubs(:default_dirs).returns(:pulp => feature(current_feature).data_dir)
+    assume_feature_present(:pulp2)
+    check_disk_performance.stubs(:default_dirs).returns(:pulp2 => '/var/lib/pulp')
   end
 
   it 'should confine existence of fio' do
--- a/test/definitions/checks/disk_performance_test.rb
+++ b/test/definitions/checks/disk_performance_test.rb
@@ -1,10 +1,15 @@
 require 'test_helper'
-
 describe Checks::Disk::Performance do
   include DefinitionsTestHelper
   include UnitTestHelper
 
   let(:check_disk_performance) { described_class.new }
+
+  before do
+    assume_feature_absent(:mongo)
+    assume_feature_present(:pulp2) || assume_feature_present(:pulp3)
+    check_disk_performance.stubs(:default_dirs).returns(:pulp => '/var/lib/pulp')
+  end
 
   it 'should confine existence of fio' do
     described_class.stubs(:execute?).with('which fio').returns(true)
--- a/test/lib/support/definitions/scenarios/missing_upgrade.rb
+++ b/test/lib/support/definitions/scenarios/missing_upgrade.rb
@@ -17,5 +17,3 @@ module Scenarios::MissingUpgrade
     end
   end
 end
-
-# ForemanMaintain::UpgradeRunner.register_version('1.14', :missing_upgrade)
--- a/test/lib/support/definitions/scenarios/missing_upgrade.rb
+++ b/test/lib/support/definitions/scenarios/missing_upgrade.rb
@@ -1,10 +1,14 @@
 module Scenarios::MissingUpgrade
   class PreUpgradeChecks < ForemanMaintain::Scenario
     metadata do
-      tags :upgrade, :missing_upgrade, :pre_upgrade_checks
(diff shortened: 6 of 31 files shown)


definitions/scenarios/versions_tags_mappings.rb
```diff
@@ -0,0 +1,41 @@
...
+def capsule?
+  pkg_manager.installed?('satellite-capsule')
+end
+
+def call_register_version(version, tag)
+  ForemanMaintain::UpgradeRunner.register_version(version, tag)
+end
+
+if satellite?
+  SATELLITE_MAPPINGS.each do |version, tag|
```
kgaikwad: As constants `SATELLITE_MAPPINGS & CAPSULE_MAPPINGS` used only once and in the same file then instead of declaring these constants at the top, would it be possible to use their values directly or with the help of normal variables inside `if/else` loop?

definitions/scenarios/versions_tags_mappings.rb
```diff
@@ -0,0 +1,41 @@
+SATELLITE_MAPPINGS = { '6.2' => :upgrade_to_satellite_6_2,
```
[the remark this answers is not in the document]
kgaikwad: Thank you @mbacovsky!  :tada:
Overall I like all your ideas and suggestions on the existing implementation. :ok_hand:


> * it duplicates some of the detection logic in `satellite?` and `capsule?` methods

Agreed! By seeing the future scope, these methods will be required outside the feature scope.
> * the pkg_manager is hardcoded to Yum

Yes, noticed this but I thought it is as per offline discussion.

> Now some ideas how to address the issues:
> 
> * during the feature loading we need some logic that is currently embodied in features. We need to move the logic out to some library the features and loading of the features depends on. The rule of thumb would be - if you need to use `feature` in a confine block or outside of the definition then the logic needs to be in the library.
> * it seems package manager as a feature was not a good idea. It is too low level feature and all the logic needs to be in lib. I'll send a PR with refactoring
> * as of the previous rule of thumb I'd suggest to move `satellite?` and `capsule? to a new class e.g `ForemanMaintain::Utils::Instance`. We could make it a concern similar to system_helpers and extend the definitions with it if needed e.g.
> 
> ```ruby
> class SomeFeature < Feature
>   include Concerns::Instance
>   confine do
>     instance.satellite?
>   end
> end
> ```
> 
> * in the scenarios I'd prefer to keep the original registration or even better to avoid the registration at all. If the scenarios inherit from `UpgradeScenario` and metadata have `upgrade_to_version` field, we could easily modify the upgrade runner to select the UpgradeScenarios and distill the upgrade to versions. What others think? @kgaikwad ?

Changes related to `package manager` as well as moving `satellite?` and `capsule?` methods to a new class will help here. 
Using `UpgradeScenario` and with `upgrade_to_version` metadata field, upgrade logic will become much more simpler and clean.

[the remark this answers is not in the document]
mbacovsky: Package Manager refactoring PR was sent, please review https://github.com/theforeman/foreman_maintain/pull/288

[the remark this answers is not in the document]
upadhyeammit: > I can see some conceptual problems with this module:
> 
>     * it abuses autoloading of scenarios to register upgrades and does not define any scenario
> 
>     * if I want to define new upgrade scenario I need to modify two files
> 
>     * it duplicates some of the detection logic in `satellite?` and `capsule?` methods
> 
>     * the pkg_manager is hardcoded to Yum
> 
> 
> Now some ideas how to address the issues:
> 
>     * during the feature loading we need some logic that is currently embodied in features. We need to move the logic out to some library the features and loading of the features depends on. The rule of thumb would be - if you need to use `feature` in a confine block or outside of the definition then the logic needs to be in the library.
> 
>     * it seems package manager as a feature was not a good idea. It is too low level feature and all the logic needs to be in lib. I'll send a PR with refactoring
> 
>     * as of the previous rule of thumb I'd suggest to move `satellite?` and `capsule? to a new class e.g `ForemanMaintain::Utils::Instance`. We could make it a concern similar to system_helpers and extend the definitions with it if needed e.g.
> 
> 
> ```ruby
> class SomeFeature < Feature
>   include Concerns::Instance
>   confine do
>     instance.satellite?
>   end
> end
> ```
> 
>     * in the scenarios I'd prefer to keep the original registration or even better to avoid the registration at all. If the scenarios inherit from `UpgradeScenario` and metadata have `upgrade_to_version` field, we could easily modify the upgrade runner to select the UpgradeScenarios and distill the upgrade to versions. What others think? @kgaikwad ?

Hello Martin,

I liked these suggestions. I tried to have UpgradeScenario separate class in scenarios as below,
~~~
module ForemanMaintain::Scenarios
  class UpgradeScenario < ForemanMaintain::Scenario
    def self.upgrade_metadata(&block)
      metadata do
        tags :upgrade_scenario
        instance_eval(&block)
      end
    end
  end
~~~

I am thinking to keep all the common code in UpgradeScenario class and call the super from actual upgrade scenarios with added steps. But while trying this I can see that UpgradeScenario is not being detected automatically, so I like to know is it always required to manually detect scenario ? because every other scenario is manually detected,

~~~
module Scenarios::Satellite_6_5
  class Abstract < ForemanMaintain::Scenarios::UpgradeScenario
    def self.upgrade_metadata(&block)
~~~

So is this correct approach or I should be having empty definitions like Abstract class in UpgradeScenario and let actual scenarios to implement it ?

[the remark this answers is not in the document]
mbacovsky: @upadhyeammit, I was able to make the autoregistration working. Here is my attempt:

First of all we need to update the upgrade scenarios to be able to identify them and guess what target they aim to:
```diff
diff --git a/definitions/scenarios/upgrade_to_satellite_6_6.rb b/definitions/scenarios/upgrade_to_satellite_6_6.rb
index 990b424..76544e6 100644
--- a/definitions/scenarios/upgrade_to_satellite_6_6.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_6.rb
@@ -2,13 +2,17 @@ module Scenarios::Satellite_6_6
   class Abstract < ForemanMaintain::Scenario
     def self.upgrade_metadata(&block)
       metadata do
-        tags :upgrade_to_satellite_6_6
+        tags :upgrade_scenario
         confine do
           feature(:satellite) && feature(:satellite).current_minor_version == '6.5'
         end
         instance_eval(&block)
       end
     end
+
+    def target_version
+      '6.6'
+    end
   end
```
This was most easy way for my POC it may be cleaner to introduce new class UpgradeScenario that has the target_version method and inherit from it. 

In upgrade runner we can remove all of the code related to `version_to_tags` and need to update how we search for the scenarios. Instead of searching the lookup table we select available scenarios tagged :upgrade_scenario. To find available versions we list distinct target_versions to find our scenarios for the actual upgrade we filter by target_version. It is less efficient then filtering by the original tags but the number of applicable scenarios will be always low. The working POC looks like :
```diff
diff --git a/lib/foreman_maintain/upgrade_runner.rb b/lib/foreman_maintain/upgrade_runner.rb
index ad7bbb5..c129273 100644
--- a/lib/foreman_maintain/upgrade_runner.rb
+++ b/lib/foreman_maintain/upgrade_runner.rb
@@ -15,13 +15,7 @@ module ForemanMaintain
       def available_targets
         # when some upgrade is in progress, we don't allow upgrade to different version
         return [current_target_version] if current_target_version
-        versions_to_tags.inject([]) do |available_targets, (version, tag)|
-          if !find_scenarios(:tags => [tag]).empty?
-            available_targets << version
-          else
-            available_targets
-          end
-        end.sort
+        find_scenarios(:tags => :upgrade_scenario).map(&:target_version).uniq.sort
       end
 
       def versions_to_tags
@@ -30,14 +24,6 @@ module ForemanMaintain
 
       # registers target version to specific tag
       def register_version(version, tag)
-        if versions_to_tags.key?(version) && versions_to_tags[version] != tag
-          raise "Version #{version} already registered to tag #{versions_to_tags[version]}"
-        end
-        @versions_to_tags[version] = tag
-      end
-
-      def clear_register
-        versions_to_tags.clear
       end
 
       def current_target_version
@@ -53,21 +39,21 @@ module ForemanMaintain
       end
     end
 
-    attr_reader :version, :tag, :phase
+    attr_reader :version, :phase
 
     def initialize(version, reporter, options = {})
       super(reporter, [], options)
-      @tag = self.class.versions_to_tags[version]
-      raise "Unknown version #{version}" unless tag
       @version = version
+      scenarios_present = find_scenarios(:tags => :upgrade_scenario).any?(&matching_version_test)
+      raise "Unknown version #{version}" unless scenarios_present
       @scenario_cache = {}
       self.phase = :pre_upgrade_checks
     end
 
     def scenario(phase)
       return @scenario_cache[phase] if @scenario_cache.key?(phase)
-      condition = { :tags => [tag, phase] }
-      matching_scenarios = find_all_scenarios(condition)
+      condition = { :tags => [:upgrade_scenario, phase] }
+      matching_scenarios = find_all_scenarios(condition).select(&matching_version_test)
       raise "Too many scenarios match #{condition.inspect}" if matching_scenarios.size > 1
       @scenario_cache[phase] = matching_scenarios.first
     end
@@ -155,6 +141,12 @@ module ForemanMaintain
 
     private
 
+    def matching_version_test
+      Proc.new do |scenario|
+        scenario.respond_to?(:target_version) && scenario.target_version == @version
+      end
+    end
+
     # rubocop:disable Metrics/MethodLength
     def rollback_pre_migrations
       raise "Unexpected phase #{phase}, expecting pre_migrations" unless phase == :pre_migrations

```

It is necessary to finish cleanup of the methods, perhaps do some refactoring and fix the tests.  I tested available-versions and upgrade run and it seems it works with no issue.

What do you think?

upadhyeammit:
@mbacovsky I have added new commit based on your POC. Tested it with Capsule and Satellite upgrade. Can see tests are failing now which is expected, I am not good at fixing those but will give it a try.

Thank you for POC ! request to review @mbacovsky @kgaikwad

lib/foreman_maintain/upgrade_runner.rb
```diff
@@ -57,25 +36,35 @@ def clear_current_target_version
 
     def initialize(version, reporter, options = {})
       super(reporter, [], options)
-      @tag = self.class.versions_to_tags[version]
-      raise "Unknown version #{version}" unless tag
       @version = version
+      scenarios_present = find_scenarios(:tags => :upgrade_scenario).any?(&matching_version_test)
```
kgaikwad: As for all upgrade scenarios now we have a separate tag i.e `:upgrade_scenario` so can't we load all scenarios with tag ':upgrade_scenario' in the beginning only?
i.e `find_scenarios(:tags => :upgrade_scenario)`

If we will need to apply any additional conditions like at [line 50](https://github.com/theforeman/foreman_maintain/pull/287/commits/d227bc995cbda0630a6dcf9894412d8e5e668b23?file-filters%5B%5D=.rb&hide-deleted-files=true#diff-3a0557c256d67261f49578f70854cb9fR50) then we already have `filter` method which takes scenarios & conditions as arguments. 
Example - detector.filter(upgrade_scenarios, { :tags => [phase] })

jameerpathan111:
> @jameerpathan111, could you please test the new changes.

@upadhyeammit, sure.

lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -98,16 +98,21 @@ def product_specific_repos(rh_version_major, full_version)
...
 
-        # Override to use Beta repositories for sat version until GA
-        if ENV['FOREMAN_MAINTAIN_USE_BETA'] == '1'
-          tools_repo_id = "rhel-#{rh_version_major}-server-satellite-tools-6-beta-rpms"
-          maintenance_repo_id = "rhel-#{rh_version_major}-server-satellite-maintenance-6-beta-rpms"
-        end
+        repos_arrary
+      end
+
+      def common_repos_array(rh_version_major, full_version)
```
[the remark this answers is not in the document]
mbacovsky: What @kgaikwad suggests looks more readable. I'd maybe go a bit further and get rid of the extra return in the middle. I'd keep use_beta as it may get handy for easier mocking.

```ruby
def common_repos(rh_version_major, full_version)
  beta_substitute = use_beta_repos? ? '-beta' : ''
  repos_list = ["rhel-#{rh_version_major}-server-satellite-maintenance-6#{beta_substitute}-rpms"]
  repos_list << "rhel-#{rh_version_major}-server-satellite-tools-#{full_version}#{beta_substitute}-rpms" unless feature(:capsule)
  repos_list
end

def use_beta_repos?
  ENV['FOREMAN_MAINTAIN_USE_BETA'] == '1'
end
```

lib/foreman_maintain/upgrade_runner.rb
mbacovsky: @kgaikwad I'm not sure but it seems we are doing that already. Correct  me if I'm wrong:
`find_scenarios` calls `detector.available scenarios` where detector is singleton and the method is defined as singleton too  https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/detector.rb#L58-L66. In the end it seems it only calls `detector.filter` as you suggested.

Despite all that I noticed we call the expensive confine block in each upgrade scenario multiple times and that is because we are creating new scenario for each upgrade phase and those share the same confine block.
Here we can spare some cycles by either caching the rpm query calls or making the shared confine a singleton.
I can see much more potential in caching of the rpm queries but it is question if it can't kick us back. Maybe cleaning the cache on calling yum update/install/erase would make it safe. But I'd suggest to address this in a separate PR.

definitions/checks/disk/performance.rb
```diff
@@ -1,9 +1,13 @@
 module Checks
   module Disk
     class Performance < ForemanMaintain::Check
+      DEFAULT_DIRS = [
```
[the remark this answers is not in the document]
mbacovsky: @upadhyeammit, any plans to address this? e.g.
```ruby
DEFAULT_DIRS = %i[pulp mongo foreman_database].inject([]) do |dirs, f| 
  dirs << feature(f).data_dir if feature(f) && File.directory?(feature(f).data_dir)
  dirs
end
```

definitions/checks/disk/performance.rb
```diff
@@ -1,9 +1,13 @@
...
+        '/var/lib/pulp', '/var/lib/mongodb', '/var/lib/pgsql'
+      ].select { |file_path| File.directory?(file_path) }.freeze
+
       metadata do
         label :disk_performance
-        description 'Check for recommended disk speed of pulp, mongodb, pgsql dir.'
+        description "Check recommended disk speed for #{DEFAULT_DIRS.join(',')} directories."
```
mbacovsky: This line will be too long and break the output. Maybe if we modify the def above to 
```ruby
DEFAULT_DIRS = %i[pulp mongo foreman_database].inject({}) do |dirs, f| 
  dirs[feature(f).label_dashed] = feature(f).data_dir if feature(f) && File.directory?(feature(f).data_dir)
  dirs
end
```
we could use something like `DEFAULT_DIRS.keys.join(', ')` to print the services we are going to test.
What do you think?

mbacovsky:
@upadhyeammit Thanks for the update! It looks good in general. I based my branch for upstream upgrades (https://github.com/mbacovsky/foreman_maintain/tree/19942_upstream_upgrades) on your branch and everything looks to work fine. It may be handy to rebase to get the package_manager changes in.  Besides the comments above and tests it seems close to get accepted.

lib/foreman_maintain/upgrade_runner.rb
upadhyeammit: @kgaikwad I have checked the code flow and can see available_scenarios is singleton. @mbacovsky I agree to confine block suggestion !

kgaikwad: > @kgaikwad I'm not sure but it seems we are doing that already. Correct me if I'm wrong:
> `find_scenarios` calls `detector.available scenarios` where detector is singleton and the method is defined as singleton too https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/detector.rb#L58-L66. In the end it seems it only calls `detector.filter` as you suggested.


@mbacovsky, you are right about singleton. My point was that we have upgrade scenarios with tag ':upgrade_scenario' and at three places given call to the method `find_scenarios(:tags => :upgrade_scenario)` so would it make sense to extract out this method calling in `upgrade_runner`?


> I can see much more potential in caching of the rpm queries but it is question if it can't kick us back. Maybe cleaning the cache on calling yum update/install/erase would make it safe. But I'd suggest to address this in a separate PR.

+1 for this suggestion.

definitions/scenarios/upgrade_to_capsule_6_7.rb
```diff
@@ -0,0 +1,83 @@
...
+        tags :upgrade_scenario
+        confine do
+          feature(:capsule) && feature(:capsule).current_minor_version == '6.6'
+        end
+        instance_eval(&block)
+      end
+    end
+
+    def target_version
+      '6.6'
```
jameerpathan111: @upadhyeammit I think this should be '6.7' right?

jameerpathan111:
@mbacovsky what should we do about dependency issue while installing fio on capsule?
when packages-locking is enabled.

```
--> Processing Dependency: pciutils for package: rdma-core-22.1-3.el7.x86_64
Error: Package: rdma-core-22.1-3.el7.x86_64 (rhel-7-server-rpms)
           Requires: pciutils
Error: Package: 1:librados2-10.2.5-4.el7.x86_64 (rhel-7-server-rpms)
           Requires: libboost_thread-mt.so.1.53.0()(64bit)
Error: Package: 1:librbd1-10.2.5-4.el7.x86_64 (rhel-7-server-rpms)
           Requires: libboost_thread-mt.so.1.53.0()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
                                                    [FAIL]
Failed executing yum install fio, exit status 1
```
Refs:  https://github.com/theforeman/foreman_maintain/pull/278#issuecomment-527831202

definitions/scenarios/upgrade_to_capsule_6_7.rb
```diff
@@ -0,0 +1,83 @@
...
+  class PostUpgradeChecks < Abstract
+    upgrade_metadata do
+      description 'Checks after upgrading to Capsule 6.7'
+      tags :post_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:post_upgrade))
```
jameerpathan111: @upadhyeammit we don't have any `post_upgrade` check capsule.

jameerpathan111:
Found  issue during `./bin/foreman-maintain upgrade check --target-version 6.7` run, related to check `remote-execution-verify-settings-file-already-exists` which should not have run on capsule, right @kgaikwad ?
```
Check to verify remote_execution_ssh settings already exist:          [FAIL]
Settings related to remote_execution_ssh are already present under /etc/smart_proxy_dynflow_core/settings.d and 
it would conflict with the installer from the next version.
symlinks available - /etc/smart_proxy_dynflow_core/settings.d
Continue with step [Remove existing settings.d directory before installer run.
 The next run of the installer will re-create the directory properly.]?, [y(yes), n(no), q(quit)] n
```

kgaikwad:
> check `remote-execution-verify-settings-file-already-exists` which should not have run on capsule, right @kgaikwad ?

IIRC, we have added this check for Satellite 6.2. 
@upadhyeammit, could you confirm that this check is required for capsule upgrade or not?

definitions/scenarios/upgrade_to_capsule_6_7.rb
```diff
@@ -0,0 +1,83 @@
...
+    upgrade_metadata do
+      description 'Checks before upgrading to Capsule 6.7'
+      tags :pre_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+      add_step(Checks::RemoteExecution::VerifySettingsFileAlreadyExists.new)
```
jameerpathan111: So it was because of this.  
@upadhyeammit so this check is required for capsule upgrade?

jameerpathan111:
@upadhyeammit 
getting following error:
/root/foreman_maintain/definitions/scenarios/upgrade_to_satellite_6_7_z.rb:80:in `<top (required)>': undefined method `register_version' for ForemanMaintain::UpgradeRunner:Class (NoMethodError)

definitions/scenarios/upgrade_to_capsule_6_7.rb
mbacovsky: @jameerpathan111 does it cause any issues? If not I'd be for keeping to make sure the checks added in feature run properly.

jameerpathan111: It doesn't.

mbacovsky:
@upadhyeammit - did you figured out the missing "command" error? Not sure what is the cause but a few thoughts:
- the tests are run on Ubuntu so you may check for differences there
- command should be part of bash and is in `/bin` on rhel based distros. Is the path to the command in the $PATH?

upadhyeammit:
> @upadhyeammit - did you figured out the missing "command" error? Not sure what is the cause but a few thoughts:
> 
>     * the tests are run on Ubuntu so you may check for differences there
> 
>     * command should be part of bash and is in `/bin` on rhel based distros. Is the path to the command in the $PATH?

Its surprising that it did not fail for package_manager pr changes which you have pushed. The command's path is in /usr/bin,
~~~
# which command 
/usr/bin/command
~~~
Is there any way to install or include this command in testing environment before running tests ?

mbacovsky:
@upadhyeammit it should be shipped with bash. The machine where the tests fail has bash installed. It may be worth it to add some line with `puts \`env\``, `puts \`which command`\` and similar and do some debugging on the test machines. It may not failed for me because I was lucky and got machine with different setup.

BTW in 6.7 scenarios the upgrade registration line slipped in during rebase

definitions/checks/disk/performance.rb
```diff
@@ -1,9 +1,16 @@
 module Checks
   module Disk
     class Performance < ForemanMaintain::Check
+      DEFAULT_DIRS = %i[pulp mongo foreman_database].inject({}) do |dirs, f|
+        if feature(f) && File.directory?(feature(f).data_dir)
```
mbacovsky: I found out this actually breaks feature loading. While loading definitions we require this check which tries to evaluate the features. There are none loaded yet at the moment and it sets the feature singleton to empty hash. There is no way to update it later. Therefore we need to evaluate this later.

This worked for me:
```ruby
      def self.default_dirs
        @default_dirs ||= %i[pulp mongo foreman_database].inject({}) do |dirs, f|
          if feature(f) && File.directory?(feature(f).data_dir)
            dirs[feature(f).label_dashed] = feature(f).data_dir
          end
          dirs
        end
      end

      def description
        "Check recommended disk speed for #{self.default_dirs.keys.join(', ')} services."
      end
```
and I had to remove the description from metadata and fix the other references to the DEFAULT_DIRS

upadhyeammit: @mbacovsky thank you for suggestions. The self.default_dirs call does not work, did you miss 'class' keyword inbetween ? I have pushed the changes with 'self.class.default_dirs' which is working for me. Unfortunately there are few tests errors, I should look into that asap.

mbacovsky: @upadhyeammit the typo is possible I was moving the code around. I guess it could work well enough if it is even instance a method (no self.) and not a class method. So feel free to remove it if it feels cleaner.

upadhyeammit:
> Found issue during `./bin/foreman-maintain upgrade check --target-version 6.7` run, related to check `remote-execution-verify-settings-file-already-exists` which should not have run on capsule, right @kgaikwad ?
> 
> ```
> Check to verify remote_execution_ssh settings already exist:          [FAIL]
> Settings related to remote_execution_ssh are already present under /etc/smart_proxy_dynflow_core/settings.d and 
> it would conflict with the installer from the next version.
> symlinks available - /etc/smart_proxy_dynflow_core/settings.d
> Continue with step [Remove existing settings.d directory before installer run.
>  The next run of the installer will re-create the directory properly.]?, [y(yes), n(no), q(quit)] n
> ```

I have checked the issue and bz[1] and rex check is not required in capsule scenario. Removed this in latest commit.
[1] https://projects.theforeman.org/issues/22738

upadhyeammit:
Hello @mbacovsky @kgaikwad,

There are two different test failures, not sure why we did not get this before, as the test code is same? If I set rvm to 2.3.3 it takes 2.6.0 binaries and tests are running fine locally.

A] It says unable to find method for NilDevice, but I do see we have mentioned all correct parameters when mocking it, and then passing that to stub, so it should not be NilDevice at all ?
~~~
On ruby 2.3.3

1) Failure:

Checks::Disk::Performance#test_0003_raise error if disk speed does not meet minimal requirement [/home/travis/build/theforeman/foreman_maintain/test/definitions/checks/disk_performance_test.rb:57]:

--- expected

+++ actual

@@ -1 +1 @@

-"Slow disk"

+"undefined method `slow_disk_error_msg' for #<ForemanMaintain::Utils::Disk::NilDevice:0xXXXXXX>"
~~~

B] the assert_equals use match? method and its not there on 2.0.0, so do we conditionally need to handle this ?
~~~
On ruby 2.0.0

 1) Error:

Checks::Disk::Performance#test_0004_print warning if disk speed does not meet minimal requirement on Sat >= 6.3:

NoMethodError: undefined method `match?' for ":warning":String

    /home/travis/build/theforeman/foreman_maintain/test/definitions/checks/disk_performance_test.rb:83:in `block (2 levels) in <top (required)>'
~~~

kgaikwad:
> There are two different test failures, not sure why we did not get this before, as the test code is same? If I set rvm to 2.3.3 it takes 2.6.0 binaries and tests are running fine locally.
> 
> A] It says unable to find method for NilDevice, but I do see we have mentioned all correct parameters when mocking it, and then passing that to stub, so it should not be NilDevice at all ?
> 
> ```
> On ruby 2.3.3
> 
> 1) Failure:
> 
> Checks::Disk::Performance#test_0003_raise error if disk speed does not meet minimal requirement [/home/travis/build/theforeman/foreman_maintain/test/definitions/checks/disk_performance_test.rb:57]:
> 
> --- expected
> 
> +++ actual
> 
> @@ -1 +1 @@
> 9
> -"Slow disk"
> 
> +"undefined method `slow_disk_error_msg' for #<ForemanMaintain::Utils::Disk::NilDevice:0xXXXXXX>"
> ```
@upadhyeammit,
With exact gems version list like a job of 2.3.3, I am able to reproduce the same errors.
Two things to consider here are 
- For this `performance` check,  you have changed the structure of default directories.
Now, those paths are mapped with respective features. 

- While running those associated tests, it assumes `satellite` feature to be present not any other feature and that's why getting a `NilDevice` object as `io_obj`.

To fix these tests, you can assume any of the feature (let's say `pulp`) then those tests will pass.
I have tested this, those are passed after assuming `pulp` feature. 
Note that - with `pulp` feature present, again you will need to either assume 'mongo` feature to be absent or present. It's up to you. 

With this failed job status, got to know that we should have to handle assertion in case of NilDevice otherwise in real scenario also we will face this kind of exception i.e `undefined method 'slow_disk_error_msg' for #<ForemanMaintain::Utils::Disk::NilDevice:0xXXXXXX>`

kgaikwad:
@upadhyeammit, 
Not sure why but job for 2.3.3 failed with error `The job exceeded the maximum time limit for jobs, and has been terminated.` and didn't run any test.
Could you try once by removing debug mode.

upadhyeammit:
@jameerpathan111 can you test this on priority ?  tests are green now.
Thank You !

upadhyeammit:
I have also added the change to skip system checks from installer as per Martin's changes in upstream upgrade,

https://github.com/mbacovsky/foreman_maintain/commit/2f0578b2257e4ca13150eedbc51c535943695af2

With this we have all the required bits for Capsule upgrade.

kgaikwad:
@upadhyeammit, could you do squashing few commits, removal of minitest commit as #297 is merged  then rebase.

upadhyeammit:
@kgaikwad  The tests were failing because of pulp feature changes, i have fixed that. Squashed and rebased. Request to have a look, thank you !

mbacovsky:
@upadhyeammit sorry for the delay with re-review I hope to get to it soon. 

I'm working on upstream upgrades and am basing the work on changes in this branch. I have working end-to-end scenarios for Katello and Foreman. Everything seems to be fine but I found out a problem introduced with the new detection of upgrade scenarios: If the upgrade fails somewhere after yum update we are not able to re-run the installer because the scenario confine block is no longer valid due to the versions no longer match. In my branch I was able to fix that by adding another [condition for pending upgrade]( https://github.com/mbacovsky/foreman_maintain/blob/19942_upstream_upgrades/definitions/scenarios/upgrade_to_foreman_1_24.rb#L9) but probably won't have time in a near future to fix all the downstream scenarios. @upadhyeammit is there chance you would find time for adding and testing the same for downstream as part of this PR? Regardless could you please check if that is sane solution? @kgaikwad, @johnpmitsch any better ideas?

upadhyeammit:
> @upadhyeammit sorry for the delay with re-review I hope to get to it soon.
> 
> I'm working on upstream upgrades and am basing the work on changes in this branch. I have working end-to-end scenarios for Katello and Foreman. Everything seems to be fine but I found out a problem introduced with the new detection of upgrade scenarios: If the upgrade fails somewhere after yum update we are not able to re-run the installer because the scenario confine block is no longer valid due to the versions no longer match. In my branch I was able to fix that by adding another [condition for pending upgrade](https://github.com/mbacovsky/foreman_maintain/blob/19942_upstream_upgrades/definitions/scenarios/upgrade_to_foreman_1_24.rb#L9) but probably won't have time in a near future to fix all the downstream scenarios. @upadhyeammit is there chance you would find time for adding and testing the same for downstream as part of this PR? Regardless could you please check if that is sane solution? @kgaikwad, @johnpmitsch any better ideas?

Thank you for insights ! I tested this situation on downstream and its valid issue. This should be correct way to handle it, I have updated the scenarios and added the new method. @jameerpathan111 request you to test it.

upadhyeammit:
@kgaikwad rebased done

test/lib/support/definitions/scenarios/missing_upgrade.rb
```diff
@@ -14,4 +18,4 @@ def compose
   end
 end
 
-ForemanMaintain::UpgradeRunner.register_version('1.14', :missing_upgrade)
+# ForemanMaintain::UpgradeRunner.register_version('1.14', :missing_upgrade)
```
kgaikwad: Feels like this leftover should have been removed

test/lib/support/definitions/scenarios/present_upgrade.rb
```diff
@@ -64,17 +80,21 @@ def compose
...
+    end
+
     def compose
       add_steps(find_checks(:post_upgrade_checks))
     end
   end
 end
 
-ForemanMaintain::UpgradeRunner.register_version('1.15', :present_upgrade)
+# ForemanMaintain::UpgradeRunner.register_version('1.15', :present_upgrade)
```
kgaikwad: ditto

kgaikwad:
+ adding @rplevka into loop.

rplevka:
I checked out this branch and tried to upgrade downstream capsule 6.6 to 6.7 - I didn't hit any issues, the upgrade went flawlessly.

test/definitions/checks/disk_performance_test.rb
```diff
@@ -1,11 +1,21 @@
...
   include UnitTestHelper
 
   let(:check_disk_performance) { described_class.new }
 
+  before do
+    assume_feature_absent(:mongo)
+    current_feature = if file_nonzero?('/var/lib/pulp')
+                        :pulp2
+                      else
+                        :pulpcore_database
```
kgaikwad: I don't like the way test failure is fixed instead of fixing the disk performance check.
@upadhyeammit, am I missing something here?

What I understood is that old pulp feature renamed to `pulp2` and confine block of this performance check 
```
confine do
   feature(:pulp)
end 
```
replaced with 
```

confine do
  feature(:instance).pulp
end
```
Inside feature(:instance), based on `pulp2` or `pulpcore` presence, pulp method returns the feature value. So I think instead of having `pulp` in default list at [line 36](https://github.com/theforeman/foreman_maintain/pull/287/files#diff-248e3e4ed2b45fee19280d48612afee0R36), we should actual add `pulp2` or `pulpcore_database` based on which pulp feature is available.

kgaikwad:
@rplevka, +1 thank you for testing.

@upadhyeammit, 
I have added one comment related to disk performance check. I know that tests related to performance check were failed because of new changes around pulp2 and pulp3 recently merged to master so don't want to block this pull-request for suggested changes as rest of changes look good to me.
So let me know if you want to fix this as part of same PR or would like to raise new PR with appropriate changes?

definitions/checks/disk/performance.rb
```diff
@@ -37,16 +32,29 @@ def run
         end
       end
 
+      def default_dirs
+        @default_dirs ||= %i[pulp2 pulpcore_database mongo foreman_database].inject({}) do |dirs, f|
+          if feature(f) && File.directory?(feature(f).data_dir)
```
kgaikwad: FYI, currently there is no `data_dir` method for feature `pulpcore_database`.

upadhyeammit: Hello as we discussed data_dir is in concern. The tests were failing because on 2.3.3, I pushed new changes, these are showing tests as green.

kgaikwad: Thanks @upadhyeammit. Not sure why job is showing pending status but tests are passed for both ruby versions.

upadhyeammit: > Thanks @upadhyeammit. Not sure why job is showing pending status but tests are passed for both ruby versions.

yep, I just forced pushed it again, and its completed now !

kgaikwad:
Thanks everyone who involved. 
@mbacovsky for your review inputs :+1: 
Tested well by @jameerpathan111  and @rplevka  
Thank you @upadhyeammit for patiently doing all back and forth changes :tada: :ok_hand:

jameerpathan111:
@kgaikwad @upadhyeammit we need support for capsule 6.8 as well.

kgaikwad:
> @kgaikwad @upadhyeammit we need support for capsule 6.8 as well.

@upadhyeammit will raise a separate PR to add scenario for 6.8.

PR #290, opened 2019-11-15
Fixes #28193 - Handle pulp3 services

johnpmitsch:
This adds the ability to handle pulp3 services. A service is used to determine if pulp3 is present as it currently does not run off rpm's in our environment. This could change to be based off an RPM later on when pulp3 is packaged and included with Katello. This also adds the ability to handle glob operators for services and optionally use the `--all` flag for systemctl

### DIFF
--- a/definitions/features/pulp3.rb
+++ b/definitions/features/pulp3.rb
@@ -1,0 +1,22 @@
+require 'foreman_maintain/utils/service/systemd'
+
+class Features::Pulp3 < ForemanMaintain::Feature
+  metadata do
+    label :pulp3
+
+    confine do
+      ForemanMaintain::Utils::Service::Systemd.new('pulpcore-api', 0).exist?
+    end
+  end
+
+  def services
+    [
+      system_service('pulpcore-api', 10),
+      system_service('pulpcore-content', 10),
+      system_service('pulpcore-resource-manager', 10),
+      system_service('pulpcore-worker@*', 20, :all => true),
+      system_service('redis', 30),
+      system_service('httpd', 30)
+    ]
+  end
+end
--- a/definitions/features/pulp3.rb
+++ b/definitions/features/pulp3.rb
@@ -14,7 +14,7 @@ class Features::Pulp3 < ForemanMaintain::Feature
       system_service('pulpcore-api', 10),
       system_service('pulpcore-content', 10),
       system_service('pulpcore-resource-manager', 10),
-      system_service('pulpcore-worker@*', 20, :all => true),
+      system_service('pulpcore-worker@*', 20, :all => true, :skip_enablement => true),
       system_service('redis', 30),
       system_service('httpd', 30)
     ]
--- a/definitions/procedures/service/base.rb
+++ b/definitions/procedures/service/base.rb
@@ -15,7 +15,7 @@ module Procedures
 
       def run_service_action(action, options)
         action_noun = feature(:service).action_noun(action).capitalize
-        puts "#{action_noun} the following service(s):\n"
+        puts "\n#{action_noun} the following service(s):"
         services = feature(:service).filtered_services(options)
         print_services(services)
 
--- a/lib/foreman_maintain/utils/service/systemd.rb
+++ b/lib/foreman_maintain/utils/service/systemd.rb
@@ -9,6 +9,12 @@ module ForemanMaintain::Utils
       def command(action, options = {})
         do_wait = options.fetch(:wait, true) # wait for service to start
         all = @options.fetch(:all, false)
+        skip_enablement = @options.fetch(:skip_enablement, false)
+
+        if skip_enablement && %w[enable disable].include?(action)
+          return skip_enablement_message(action, @name)
+        end
+
         if do_wait && File.exist?('/usr/sbin/service-wait')
           "service-wait #{@name} #{action}"
         else
@@ -55,6 +61,21 @@ module ForemanMaintain::Utils
 
       def execute(action, options = {})
         @sys.execute_with_status(command(action, options))
+      end
+
+      def skip_enablement_message(action, name)
+        # Enable and disable does not work well with globs since they treat them literally.
+        # We are skipping the pulpcore-workers@* for these actions until they are configured in
+        # a more managable way with systemd
+        # rubocop:disable Layout/IndentAssignment
+        msg =
+"
+\nWARNING: Skipping #{action} for #{name} as there are a variable amount of services to manage
+and this command will not respond to glob operators. These services have been configured by
+the installer and it is recommended to keep them enabled to prevent misconfiguration.\n
+"
+        # rubocop:enable Layout/IndentAssignment
+        puts msg
       end
     end
   end
--- a/lib/foreman_maintain/utils/service/systemd.rb
+++ b/lib/foreman_maintain/utils/service/systemd.rb
@@ -1,14 +1,14 @@
 module ForemanMaintain::Utils
   module Service
     class Systemd < Abstract
-      def initialize(name, priority, _options = {})
+      def initialize(name, priority, options = {})
         super
         @sys = SystemHelpers.new
       end
 
       def command(action, options = {})
         do_wait = options.fetch(:wait, true) # wait for service to start
-        all = options.fetch(:all, false)
+        all = @options.fetch(:all, false)
         if do_wait && File.exist?('/usr/sbin/service-wait')
           "service-wait #{@name} #{action}"
         else
--- a/lib/foreman_maintain/utils/service/systemd.rb
+++ b/lib/foreman_maintain/utils/service/systemd.rb
@@ -8,10 +8,13 @@ module ForemanMaintain::Utils
 
       def command(action, options = {})
         do_wait = options.fetch(:wait, true) # wait for service to start
+        all = options.fetch(:all, false)
         if do_wait && File.exist?('/usr/sbin/service-wait')
           "service-wait #{@name} #{action}"
         else
-          "systemctl #{action} #{@name}"
+          cmd = "systemctl #{action} #{@name}"
+          cmd += ' --all' if all
+          cmd
         end
       end
 
--- a/definitions/checks/disk/performance.rb
+++ b/definitions/checks/disk/performance.rb
@@ -8,7 +8,7 @@ module Checks
         preparation_steps { Procedures::Packages::Install.new(:packages => %w[fio]) }
 
         confine do
-          feature(:pulp)
+          feature(:instance).pulp
         end
       end
 
--- a/definitions/features/foreman_proxy.rb
+++ b/definitions/features/foreman_proxy.rb
@@ -17,7 +17,7 @@ class Features::ForemanProxy < ForemanMaintain::Feature
   end
 
   def with_content?
-    !!feature(:pulp)
+    !!feature(:instance).pulp
   end
 
   def dhcpd_conf_exist?
--- a/definitions/features/instance.rb
+++ b/definitions/features/instance.rb
@@ -64,6 +64,10 @@ class Features::Instance < ForemanMaintain::Feature
     net
   end
 
+  def pulp
+    feature(:pulp2) || feature(:pulp3)
+  end
+
   private
 
   # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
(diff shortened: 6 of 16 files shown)


definitions/features/pulp3.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def services
+    [
+      system_service('pulpcore-api', 10),
+      system_service('pulpcore-content', 10),
+      system_service('pulpcore-resource-manager', 10),
+      system_service('pulpcore-worker@*', 20, :all => true),
+      system_service('redis', 30),
+      system_service('httpd', 30)
```
johnpmitsch: @jlsherrill what do you think about the order here? Do we want to start and stop services in a particular order?

This is the order with the current values:

```
| displaying rh-mongodb34-mongod                   [OK]                         
| displaying postgresql                            [OK]                         
| displaying pulpcore-api                          [OK]                         
/ displaying pulpcore-content                      [OK]                         
/ displaying pulpcore-resource-manager             [OK]                         
/ displaying qdrouterd                             [OK]                         
/ displaying qpidd                                 [OK]                         
/ displaying squid                                 [OK]                         
/ displaying pulp_celerybeat                       [OK]                         
/ displaying pulp_resource_manager                 [OK]                         
/ displaying pulp_streamer                         [OK]                         
/ displaying pulp_workers                          [OK]                         
- displaying pulpcore-worker@*                     [OK]                         
- displaying tomcat                                [OK]                         
- displaying httpd                                 [OK]                         
- displaying puppetserver                          [OK]                         
- displaying redis                                 [OK]                         
- displaying foreman-proxy                         [OK]        
```

jlsherrill: We might ask the pulp team if it matters, i might could see it being better to shut down the api, then the resource-manager and then the workers, but in practice it may not matter.   I'd reach out to the pulp team and ask if they have a preference.

jlsherrill: also, does the order reverse itself when starting vs stopping?

mbacovsky: It does reverse

mbacovsky:
@johnpmitsch  thanks for the patch! Looks good.

There is one thing that we should consider. There are various features/procedures/checks that are conditioned by presence of pulp. We use `feature(:pulp)` to test the pulp presence. If there is going to be new feature for Pulp 3 we should revisit the code.

For Mongo 2 vs Mongo 3 we chose to have both in a single `feature(mongo)` but for Pulp it may be different and separating the two features may be useful. Would it make sense to have `feature(:pulp2)` and `feature(:pulp3)` and perhaps `feature(:instance).pulp?` defined as `feature(:pulp3) || feature(:pulp2)`?

Any other ideas?
@kgaikwad, @upadhyeammit?

johnpmitsch:
@mbacovsky  Thanks, that is a really good point and we will have to consider this as we add more pulp3 functionality to foreman-maintain such a backups, managing upgrades, and perhaps parts of the pulp 2->3 migration.

> Would it make sense to have `feature(:pulp2)` and `feature(:pulp3)` and perhaps `feature(:instance).pulp`

This sounds like a good idea, +1 to keeping them separate features as they are two very different apps and I think it would be a lot of `if X else Y` to manage them in one feature. Though we could have a single indicator if any pulp instance is present as you suggested. 

Let me see what I can do!

jlsherrill:
One thing to ask is, when would we want to know the some version of pulp is
enabled but not actually care which version?  I haven't had a chance to
look through all of the foreman maintain code yet, but are there many
examples of that?

johnpmitsch:
@jlsherrill sounds like the database should be started first:
```
dkliban | jomitsch: the database should be started first. the rest of the services can be brought up after that in 
         | any order                                                                                                
 dkliban | database and redis should be started before any other services are started                               
```

jlsherrill:
yeah, and it will in the end share a database server with foreman and candlepin, so i guess it can be handled in the same way

jlsherrill:
we may need to either hold off on supporting the pulp3 database for now, or wait to merge this until the postgresql 10/11 work is done.

kgaikwad:
> For Mongo 2 vs Mongo 3 we chose to have both in a single `feature(mongo)` but for Pulp it may be different and separating the two features may be useful. Would it make sense to have `feature(:pulp2)` and `feature(:pulp3)` and perhaps `feature(:instance).pulp?` defined as `feature(:pulp3) || feature(:pulp2)`?
> Any other ideas?
> @kgaikwad, @upadhyeammit?

@mbacovsky, sounds good to me +1

johnpmitsch:
Updated to change the pulp feature to pulp2 and use a `feature(:instance).pulp` for checking both. I've updated the codebase for this change. For the backup, I'm just assuming we are backing up pulp2, the pulp3 backup support is a different task on our list

> we may need to either hold off on supporting the pulp3 database for now, or wait to merge this until the postgresql 10/11 work is done.

@jlsherrill What will this look like, will it be another service? I think this could be a separate task as it feels like there will be another feature added for postgres 10 here, separate from the pulp feature.

jlsherrill:
Yes, i think we wait till postgresql 10 support lands, there will only be 1 instance that pulp/foreman/candlepin share, so i imagine it will look similar to: https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/candlepin_database.rb

jlsherrill:
and yes, i agree pulp3 backup is a different task.  We may need to wait till we can actually provision a box with pulp3 to start working on it

jlsherrill:
@kgaikwad @mbacovsky any more thoughts?

definitions/procedures/service/base.rb
```diff
@@ -15,7 +15,7 @@ def self.common_params(context)
 
       def run_service_action(action, options)
         action_noun = feature(:service).action_noun(action).capitalize
-        puts "#{action_noun} the following service(s):\n"
+        puts "\n#{action_noun} the following service(s):"
```
johnpmitsch: This spacing has bothered me for a while, I can do a separate PR to change, but its just a small thing so I included here! :cat:

johnpmitsch:
@kgaikwad @mbacovsky  I don't mean to be a squeaky wheel, but any chance we can look at this this week?

definitions/features/pulp3.rb
```diff
@@ -0,0 +1,22 @@
...
+      ForemanMaintain::Utils::Service::Systemd.new('pulpcore-api', 0).exist?
+    end
+  end
+
+  def services
+    [
+      system_service('pulpcore-api', 10),
+      system_service('pulpcore-content', 10),
+      system_service('pulpcore-resource-manager', 10),
+      system_service('pulpcore-worker@*', 20, :all => true),
```
jlsherrill: I tried a  'foreman-maintain service restart'  and it didn't seem to properly start the pulpcore-worker services, did it work for you?

ianballou: I see this issue as well

ianballou: looks like the `:all => true` isn't making it to the actual `systemctl` command.

johnpmitsch:
@ianballou @jlsherrill nice catch, I updated in a separate commit. I guess I was checking the pulp2 workers in my testing! The behavior makes sense because you need --all for using globs with `systemctl start` and do not need it for `systemctl stop`, so they were stopped but not started, since the globs weren't registering with `start` because `--all` wasn't being added to the command. (I'm not sure the reason for why systemctl works this way)

I'm not sure if there is any concern of having both `@options` and `options`, but using the options hash for the instance works well.

ianballou:
Latest commit fixed the Pulp 3 worker issue I was having :+1:

lib/foreman_maintain/utils/service/systemd.rb
```diff
@@ -1,17 +1,20 @@
...
       def command(action, options = {})
         do_wait = options.fetch(:wait, true) # wait for service to start
+        all = @options.fetch(:all, false)
         if do_wait && File.exist?('/usr/sbin/service-wait')
           "service-wait #{@name} #{action}"
         else
-          "systemctl #{action} #{@name}"
+          cmd = "systemctl #{action} #{@name}"
+          cmd += ' --all' if all
+          cmd
```
jlsherrill: hrmm, i just realized this means that --all will be passed to enable/disable which i'm not sure makes sense

in fact i realized that foreman-maintain even provides enable/disable support....  
which i don't think will do the right thing:

```
systemctl enable 'pulpcore-worker@*'  --all

Created symlink from /etc/systemd/system/multi-user.target.wants/pulpcore-worker@\x2a.service to /usr/lib/systemd/system/pulpcore-worker@.service.
```

so i think it might make more sense if the start/stop commands pass :all in through the options param to command.  And then we need to figure out what to do with enable/disable.  I'm not sure how its supposed to know which pulpworkers to enable if they've been disabled.

jlsherrill: the simplest thing is to not allow disabling/enabling of pulpcore-workers, but that seems kinda like a cop out.  I'm not entirely sure why this functionality is provided.

lib/foreman_maintain/utils/service/systemd.rb
```diff
@@ -1,17 +1,25 @@
...
   module Service
     class Systemd < Abstract
-      def initialize(name, priority, _options = {})
+      def initialize(name, priority, options = {})
         super
         @sys = SystemHelpers.new
       end
 
       def command(action, options = {})
+        if %w[enable disable].include?(action) && @name.include?('pulpcore-worker@*')
```
jlsherrill: I think it'd be better to pass this in as an option, :disable_enablement_control => true  or something like that, and have it be controlled here.  I'd probably handle that in the enable and disable methods rather than in this command method as well.

I'd think that this command module should not have any service-specific knowledge.

johnpmitsch:
@jlsherrill  updated to skip enable/disable for pulpcore_workers and display a message.

@mbacovsky @kgaikwad the way the pulp3 workers are set up currently is that they are enabled in systemd through the installer and there is a variable amount of them. So by disabling them, they "disappear" and have to enabled one by one, for example `pulpcore_worker@1` and `pulpcore_worker@2`. Using `pulpcore_worker@*` won't enable the correct amount of workers with the correct names, it will just literally register the `*`.

  We don't know the number of workers (its not in a config somewhere like pulp2) so we couldn't come up with a way to handle this, so we are just skipping the enable/disable commands for them and displaying a warning message. I don't think its a huge deal since enable/disable aren't probably heavily used commands, and start, stop, and status work fine. I just want to make sure you all are aware of the background of why it's handled like this in the code. Hopefully we can improve this on our end in the installer, giving a better way to manage these services (like the pulp_workers service in pulp2) and update foreman-maintain with that way.

johnpmitsch:
If we do wind up storing the number of workers somewhere, we can enable `pulpcore_worker@{1..N}`, where `N` is the number of workers, this syntax works with systemctl

(more commenting this as future reference if we come back to this later)

lib/foreman_maintain/utils/service/systemd.rb
johnpmitsch: thats a good point, let me look into that

johnpmitsch: @jlsherrill updated to take an option and use that to skip

johnpmitsch: err... hold on :)

johnpmitsch: @jlsherrill ok, its now updated to only skip on enable and disable with that option :)

jlsherrill:
This looks good, but i will have to test at a certain time.  @ianballou if you want to re-test that'd be very helpful!
So ack from a code perspective from me.  @kgaikwad you might want to take a 2nd glance at this as well

jlsherrill:
@johnpmitsch also the test failure seems unrelated, can you restart it?

johnpmitsch:
> This looks good, but i will have to test at a certain time. @ianballou if you want to re-test that'd be very helpful!
> So ack from a code perspective from me. @kgaikwad you might want to take a 2nd glance at this as well

Sounds good


> @johnpmitsch also the test failure seems unrelated, can you restart it?

They seem to be consistently failing with an possible build issue, maybe @kgaikwad has an idea of if this is a known issue?

kgaikwad:
@johnpmitsch,
Due to `mocha` gem version, now test helper will need to update `require "mocha/mini_test" ` to `"mocha/minitest"`.

@upadhyeammit is going to create a separate PR for this.

ianballou:
I've tested this and noticed that my `pulp-worker@0` is inactive after the running service restart.  Workers 1 and 2 and the rest of the services are up and running though.  If this is expected then it should be all good!

jlsherrill:
@ianballou was the 0 worker actually enabled?  
``` 
systemctl list-units -t service --full  | grep pulpcore-worker
```
the pulp installer only enables workers 1 & 2, so i wouldn't expect 0 to be started by this script unless you had enabled it manually prior.

ianballou:
@jlsherrill ah ok that was it! I'm not sure why worker 0 was even running to be honest.  ACK

kgaikwad:
@johnpmitsch,
Could you rebase this branch as I have merged #297?

PR #291, opened 2019-11-18
Fixes #28281 - no spinner on non-tty output

lzap:
When the tool is executed with redirected output (to a pipe or a file) the spinner thing makes thing terrible and unreadable. The process should detect if it's running a tty and then avoid spinning (e.g. when executed from cron).

Reproducer:

    foreman-maintain service status > /tmp/test.log

See the ^M characeters in the log file. If the action takes a lot of time (e.g. restart) there will be a lots of these.

### DIFF
--- a/lib/foreman_maintain/reporter/cli_reporter.rb
+++ b/lib/foreman_maintain/reporter/cli_reporter.rb
@@ -16,7 +16,6 @@ module ForemanMaintain
           @spinner_chars = %w[| / - \\]
           @current_line = ''
           @puts_needed = false
-          start_spinner
         end
 
         def update(line)
@@ -42,8 +41,6 @@ module ForemanMaintain
           end
         end
 
-        private
-
         def start_spinner
           @thread = Thread.new do
             loop do
@@ -52,6 +49,8 @@ module ForemanMaintain
             end
           end
         end
+
+        private
 
         def spin
           @mutex.synchronize do
@@ -80,6 +79,7 @@ module ForemanMaintain
         @line_char = '-'
         @cell_char = '|'
         @spinner = Spinner.new(self)
+        @spinner.start_spinner if @stdout.tty?
         @last_line = ''
       end
 


mbacovsky:
@lzap thanks for the patch. LGTM!
@jameerpathan111, could you please confirm the suggested change does not have any unwanted effect in our automation?

jameerpathan111:
> @lzap thanks for the patch. LGTM!
> @jameerpathan111, could you please confirm the suggested change does not have any unwanted effect in our automation?

Sorry for late reply, I will look into this.

jameerpathan111:
@mbacovsky @lzap I can still see the ^M characeters in the test.log file 
I have run `foreman-maintain service status > /tmp/test.log`.

lzap:
This patch does not remove all of them, just those which were generated via the spinner. So many of them. Can you paste the output tho?

kgaikwad:
> This patch does not remove all of them, just those which were generated via the spinner. So many of them. Can you paste the output tho?
@jameerpathan111, could you please share the output here?

jameerpathan111:
@kgaikwad 
Attaching the output of `foreman-maintain service status > /tmp/test.log` using this PR.
[fm_pr_291.log](https://github.com/theforeman/foreman_maintain/files/4521707/fm_pr_291.log)

lzap:
The text looks quite readable now, I mean there are few leftovers but it's not polluted with ^M^M^M^M.

I don't see a reason why not to merge it, unless it breaks something in your tests.

jameerpathan111:
> The text looks quite readable now, I mean there are few leftovers but it's not polluted with ^M^M^M^M.
> I don't see a reason why not to merge it, unless it breaks something in your tests.

yeah, it doesn't break anything. We can ignore few of ^M characters.

kgaikwad:
Sorry for the delay @lzap. Merging this patch.

kgaikwad:
Thank you @lzap and @jameerpathan111.

jameerpathan111:
@lzap @kgaikwad with and without pr change it is showing almost same ^M chars. Just few of them are removed. :(

kgaikwad:
@jameerpathan111 - As long as no any unwanted failure in automation or tests, I guess that's fine.

PR #292, opened 2019-11-21
Add support for dynflow_sidekiq

adamruzicka:
Still WIP, opening in current state for visibility

### DIFF
--- a/definitions/features/dynflow_sidekiq.rb
+++ b/definitions/features/dynflow_sidekiq.rb
@@ -5,32 +5,35 @@ class Features::DynflowSidekiq < ForemanMaintain::Feature
     confine do
       server? && find_package('foreman-dynflow-sidekiq')
     end
+  end
 
-    def services
-      service_names.map { |service| system_service service, instance_priority(service) }
+  def services
+    service_names.map { |service| system_service service, instance_priority(service) }
+  end
+
+  def config_files
+    # Workaround until foreman-installer can deploy scaled workers
+    service_symlinks = configured_services.map do |service|
+      "/etc/systemd/system/multi-user.target.wants/#{service}.service"
     end
+    [
+      '/etc/foreman/dynflow',
+      service_symlinks
+    ].flatten
+  end
 
-    def config_files
-      [
-        '/etc/foreman/dynflow',
-        # Workaround until foreman-installer can deploy scaled workers
-        configured_services.map { |service| "/etc/systemd/system/multi-user.target.wants/#{service}.service" }
-      ].flatten
-    end
+  private
 
-    private
+  def instance_priority(instance)
+    # Orchestrator should be started before the workers are
+    instance.end_with?('@orchestrator') ? 30 : 31
+  end
 
-    def instance_priority(instance)
-      # Orchestrator should be started before the workers are
-      instance.end_with?('@orchestrator') ? 30 : 31
-    end
+  def service_names
+    configured_instances.map { |instance| "dynflow-sidekiq@#{instance}" }
+  end
 
-    def service_names
-      configured_instances.map { |instance| "dynflow-sidekiq@#{instance}" }
-    end
-
-    def configured_instances
-      Dir["/etc/foreman/dynflow/*"].map { |config| File.basename(config).gsub(/\.yml$/, '') }
-    end
+  def configured_instances
+    Dir['/etc/foreman/dynflow/*'].map { |config| File.basename(config).gsub(/\.yml$/, '') }
   end
 end
--- a/definitions/features/dynflow_sidekiq.rb
+++ b/definitions/features/dynflow_sidekiq.rb
@@ -7,11 +7,7 @@ class Features::DynflowSidekiq < ForemanMaintain::Feature
     end
 
     def services
-      [
-        # TODO: Move redis to a separate feature when we're sure how it will be configured
-        system_service('redis', 10),
-        service_names.map { |service| system_service service, instance_priority(service) }
-      ].flatten
+      service_names.map { |service| system_service service, instance_priority(service) }
     end
 
     def config_files
--- a/definitions/features/dynflow_sidekiq.rb
+++ b/definitions/features/dynflow_sidekiq.rb
@@ -34,6 +34,6 @@ class Features::DynflowSidekiq < ForemanMaintain::Feature
   end
 
   def configured_instances
-    Dir['/etc/foreman/dynflow/*'].map { |config| File.basename(config).gsub(/\.yml$/, '') }
+    Dir['/etc/foreman/dynflow/*'].map { |config| File.basename(config, '.yml') }
   end
 end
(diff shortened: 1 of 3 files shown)


definitions/features/redis.rb
```diff
@@ -0,0 +1,22 @@
+class Features::Redis < ForemanMaintain::Feature
+  SCL_PREFIX = 'rh-redis5'.freeze
+  SERVICE_NAME = "#{SCL_PREFIX}-redis".freeze
```
ehelms: This could be fine for now, but will be an issue for Debian if ever used there. This will also break on EL8 when we land there. Unfortunately other SCLs have `syspaths` packages to allow using the classic service names. Redis does not appear to. @evgeni have you seen anything? would it be worth our shipping our own?

evgeni: We can ask the SCL team to provide one? :)

definitions/features/redis.rb
```diff
@@ -0,0 +1,22 @@
+class Features::Redis < ForemanMaintain::Feature
+  SCL_PREFIX = 'rh-redis5'.freeze
```
evgeni: IMHO this should be
```suggestion
  SCL_NAME = 'rh-redis5'.freeze
  SCL_PREFIX = '#{SCL_NAME}-'.freeze
```

adamruzicka: Isn't it overly verbose? Also, there will have to be double quotes

ehelms: A smidge, but this suggestion is based on terminology found in the SCL world. SCLs have names, and then a prefix. The prefix tends by convention to be the name + a dash. You don't have to get that verbose, its just what we in the packaging world are used to.

evgeni: It'd make EL8/Debian support easier as you'd then just can set PREFIX to `''`

ehelms:
This seems generally good to go. Can it be dropped from WIP state? We will need this both merged and released to prevent nightly breakages when the corresponding installer changes are merged.

definitions/features/redis.rb
```diff
@@ -0,0 +1,22 @@
...
+      server? && find_package(SERVICE_NAME)
+    end
+
+    def services
+      [system_service(SERVICE_NAME, 10)]
+    end
+
+    def config_files
+      ["/etc/opt/rh/#{SCL_PREFIX}/redis",
+       "/etc/opt/rh/#{SCL_PREFIX}/redis.conf"]
```
ekohl: These paths will also be different when there is no SCL. On non-SCL (EL8), it's `/etc/redis.conf`. Debian(-based) uses `/etc/redis/redis.conf`

definitions/features/redis.rb
```diff
@@ -0,0 +1,22 @@
...
+      # Luckily, the service name is the same as the package providing it
+      server? && find_package(SERVICE_NAME)
+    end
+
+    def services
+      [system_service(SERVICE_NAME, 10)]
+    end
+
+    def config_files
+      ["/etc/opt/rh/#{SCL_PREFIX}/redis",
```
ekohl: I think this is a directory puppet-redis adds to be ready for multiple instances but we don't intend to use those. AFAIK the RPM doesn't create this directory. Debian(-based) does have such a directory.

definitions/features/dynflow_sidekiq.rb
```diff
@@ -0,0 +1,39 @@
...
+    # Orchestrator should be started before the workers are
+    instance.end_with?('@orchestrator') ? 30 : 31
+  end
+
+  def service_names
+    configured_instances.map { |instance| "dynflow-sidekiq@#{instance}" }
+  end
+
+  def configured_instances
+    Dir['/etc/foreman/dynflow/*'].map { |config| File.basename(config).gsub(/\.yml$/, '') }
```
ekohl:
```suggestion
    Dir['/etc/foreman/dynflow/*'].map { |config| File.basename(config, '.yml') }
```

definitions/features/redis.rb
adamruzicka: Turned into methods to allow easier conditionals based on platform

definitions/features/redis.rb
adamruzicka: Handled it similarly as with the scl prefix. When support for el8/deb lands, the etc_prefix method will need to contain a condition checking current platform

definitions/features/redis.rb
adamruzicka: We should be safe, if a file does not exist, f-m will just skip it. I'd say it is better to try to collect everything there is

definitions/features/redis.rb
```diff
@@ -0,0 +1,34 @@
+class Features::Redis < ForemanMaintain::Feature
+  SCL_NAME = 'rh-redis5'.freeze
+
+  metadata do
+    label :redis
+
+    confine do
+      # Luckily, the service name is the same as the package providing it
+      server? && find_package(service_name)
```
kgaikwad: Given block passed as argument to `metadata` and its execution happens outside of this feature scope so it will raise an error for `service_name`.

Either you can declare one more constant or use "#{SCL_NAME}-redis" directly.

ehelms:
@kgaikwad Would a merge and release of this be possible within the next day? This would allow us to merge and get into nightly the new dynflow.

definitions/features/foreman_tasks.rb
```diff
@@ -108,7 +108,7 @@ def fetch_tasks_status(state, spinner)
   end
 
   def services
-    [system_service(service_name, 30)]
+    feature(:dynflow_sidekiq) ? [] : [system_service(service_name, 30)]
```
kgaikwad: Does that mean rest of things from feature `foreman_tasks` remain as it is and only services get changed?

adamruzicka: Exactly

ekohl: AFAIK foreman-tasks hasn't had a service ever since dynflowd was introduced. dynflowd is part of Foreman itself while foreman-tasks is an optional plugin. dynflowd.service always had the alias foreman-tasks so `systemctl status foreman-tasks` continued to work.

definitions/features/redis.rb
adamruzicka: Oh, forgot about that. Fixed in a slightly different way

adamruzicka:
We're :green_apple:

kgaikwad:
Thanks @adamruzicka!

kgaikwad:
@adamruzicka, @ehelms,
Released a new version - [foreman-maintain 0.6.0](https://rubygems.org/gems/foreman_maintain/versions/0.6.0).

My Apologies, I didn't noticed all commit messages before merging this PR so except first inital commit I have modified the remaining commit messages from `Fixes #xxxx` to `Refs #xxxx`.
Also, it would be good to have some test coverage for new changes later in a separate PR.

ehelms:
@kgaikwad thanks so much for quick reviews and the release!

adamruzicka:
Thank you @kgaikwad , I'll add the tests to my todo.

PR #293, opened 2019-11-22
Fixes #28337 - Move qpid backup to katello feature

chris1984:
@johnpmitsch here is the pr to fix the qpid issue that is causing sat-clone to fail.

@mbacovsky I moved qpid backup to katello instead of pulp. Should we put it in the Candlepin feature instead or is katello fine? Read more on the issue to see the reasoning.

### DIFF
--- a/definitions/features/katello.rb
+++ b/definitions/features/katello.rb
@@ -32,7 +32,11 @@ class Features::Katello < ForemanMaintain::Feature
       '/etc/sysconfig/tomcat*',
       '/etc/tomcat*',
       '/var/lib/candlepin',
-      '/usr/share/foreman/bundler.d/katello.rb'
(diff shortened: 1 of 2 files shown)


mbacovsky:
The change looks reasonable but wouldn't it make more sense to move also the other qpid related paths and service to the Katello feature? 
If the qpid is there for communication between components under Katello umbrella katello feature is probably the right place for it.

chris1984:
@mbacovsky updated and moved all the qpid files to the katello feature

chris1984:
@mbacovsky is this good for a merge?

chris1984:
@kgaikwad can this get a merge?

kgaikwad:
@chris1984, thank you! Merging this pull request.

PR #296, opened 2019-12-10
Fixes #28462 - Clean up qpid jrnl on restore

chris1984:
This PR addresses an issue with restoring a satellite and qpid starting but not listening due to the jrnl files being corrupted.

I reached out to @pmoravec and this is the work-around for [BZ 1781386](https://bugzilla.redhat.com/show_bug.cgi?id=1781386)

Let me know if that looks correct @pmoravec I tested it here on a 6.6 satellite and the restore worked correctly and qpid started up correctly.

### DIFF
--- a/definitions/procedures/restore/configs.rb
+++ b/definitions/procedures/restore/configs.rb
@@ -14,6 +14,7 @@ module Procedures::Restore
         spinner.update('Restoring configs')
         clean_conflicting_data
         restore_configs(backup)
+        reset_qpid_jrnls
         reload_configs
       end
     end
@@ -44,6 +45,12 @@ module Procedures::Restore
     def clean_conflicting_data
       # tar is unable to --overwrite dir with symlink
       execute('rm -rf /usr/share/foreman-proxy/.ssh')
+    end
+
+    def reset_qpid_jrnls
+      # on restore without pulp data qpid fails to start
+      # https://access.redhat.com/solutions/4645231
+      execute('rm -rf /var/lib/qpidd/.qpidd/qls/dat2/__db.00*')
     end
   end
 end


chris1984:
Test failure looks unrelated

pmoravec:
I was wrong with this workaround - we can not delete files like `queues.db` or `bindings.db` or `exchanges.db` or similar, that hold the information about queues etc. I am still checking what is convenient workaround here.

upadhyeammit:
@chris1984 if we are going to drop https://github.com/theforeman/foreman_maintain/pull/253 then I think this workaround is not required.

chris1984:
@upadhyeammit I am fine with dropping it, but lets first wait and hear back from @pmoravec to see if any more action is needed. I would hate to close it if there is more action needed, if the other pr solves the issue then we can close this one.

chris1984:
@pmoravec here is the updated pr. @upadhyeammit it looks like we still need this after talking to Pavel today in #satellite6

pmoravec:
This preventive deletion prevents https://bugzilla.redhat.com/show_bug.cgi?id=1781386 on the given reproducer. Worth confirming with @pkubatrh (a libdb maintainer) if the deletion cant have a side effect.

pmoravec:
> This preventive deletion prevents https://bugzilla.redhat.com/show_bug.cgi?id=1781386 on the given reproducer. Worth confirming with @pkubatrh (a libdb maintainer) if the deletion cant have a side effect.

OK I got reply that deleting those files is safe, as they contain just runtime environment info that will be regenerated after the start-up.

chris1984:
@upadhyeammit can this get a merge based on @pmoravec latest comment? It looks like we still need this despite the other PR you mentioned.

definitions/procedures/restore/configs.rb
```diff
@@ -45,5 +46,10 @@ def clean_conflicting_data
       # tar is unable to --overwrite dir with symlink
       execute('rm -rf /usr/share/foreman-proxy/.ssh')
     end
+
+    def reset_qpid_jrnls
+      # on restore without pulp data qpid fails to start
```
johnpmitsch: might be worth adding `https://access.redhat.com/solutions/4645231` to this comment as reference.

chris1984:
Working on testing this BZ:

https://bugzilla.redhat.com/show_bug.cgi?id=1780727

I had to use this patch and without it qpid didnt start up causing the clone to fail. So another test showing this works.

chris1984:
@kgaikwad can this get a merge please?

kgaikwad:
Thanks @chris1984!

PR #297, opened 2019-12-10
Fixes #28473 - Use minitest

upadhyeammit:
Changes:
  - As per mocha version change 'minitest' should be required instead of 'mini_test'
  - Bump bundler minimum version to 1.17


kgaikwad:
Thank you @upadhyeammit! 
Merging..

PR #298, opened 2019-12-12
Fixes #28479 - extend paused tasks count query with result=error

kgaikwad:
+ adding @jameerpathan111 into loop

### DIFF
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -55,7 +55,7 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     sql = <<-SQL
       SELECT count(*) AS count
         FROM foreman_tasks_tasks
-        WHERE state IN ('paused')
+        WHERE state IN ('paused') AND result IN ('error')
     SQL
     unless ignored_tasks.empty?
       sql << "AND label NOT IN (#{quotize(ignored_tasks)})"


definitions/features/hammer.rb
```diff
@@ -70,6 +70,10 @@ def run(args)
     execute("#{command_base} #{args}")
   end
 
+  def reload_cache
+    execute("#{command_base} --reload-cache")
```
upadhyeammit: @kgaikwad as you mentioned mostly this should not cause any issue when hammer is unable to connect to server as hammer feature takes care of it, on my system where hammer is failing with warnings it takes 1 min to complete the command; so maybe we can add the spinner but its not mandatory.

kgaikwad: > @kgaikwad as you mentioned mostly this should not cause any issue when hammer is unable to connect to server as hammer feature takes care of it

Tested this scenario. 
Instead of calling `execute` method directly, either I have to call `run` method  with args `--reload-cache` or I need to give `setup_admin_access` call before `execute` method call. This way it will handle server connection part.
Thank you for pointing this out :+1: 

> on my system where hammer is failing with warnings it takes 1 min to complete the command; so maybe we can add the spinner but its not mandatory.

Yes, I can add spinner.

kgaikwad: Done!

jameerpathan111:
@kgaikwad 
I have tried Satellite 6.6 to 6.7 upgrade using this PR and I can still see issue mentioned in [BZ1781128](https://bugzilla.redhat.com/show_bug.cgi?id=1781128).
You can see output and foreman-maintain log in [FM_PR_298.txt](https://github.com/theforeman/foreman_maintain/files/3960274/FM_PR_298.txt)

kgaikwad:
@jameerpathan111, 

I have debugged on the above mentioned issue. It seems that since from time when this check was added, it is working like this.
Using this check, it checks just for paused tasks not by considering the result value. Whereas `hammer task resume` command uses `bulk_resume` API call. By default, in this API, it takes those paused tasks having `result=error` for bulk resume.
https://github.com/theforeman/foreman-tasks/commit/99b94afab215ee079fac291df5bada335b91b060

Could you please verify `result` value for those paused tasks?
I am sure those tasks must have `result=success` not `error` and that's why it is showing paused tasks again and again in check result but doing nothing on resume procedure selection. For such scenarios, I think user has to select second option and from UI they can perform correct action for such paused tasks with result equal to success.

@upadhyeammit, what do you think?

upadhyeammit:
> @kgaikwad
> I have tried Satellite 6.6 to 6.7 upgrade using this PR and I can still see issue mentioned in [BZ1781128](https://bugzilla.redhat.com/show_bug.cgi?id=1781128).
> You can see output and foreman-maintain log in [FM_PR_298.txt](https://github.com/theforeman/foreman_maintain/files/3960274/F

> @jameerpathan111,
> 
> I have debugged on the above mentioned issue. It seems that since from time when this check was added, it is working like this.
> Using this check, it checks just for paused tasks not by considering the result value. Whereas `hammer task resume` command uses `bulk_resume` API call. By default, in this API, it takes those paused tasks having `result=error` for bulk resume.
> [theforeman/foreman-tasks@99b94af](https://github.com/theforeman/foreman-tasks/commit/99b94afab215ee079fac291df5bada335b91b060)
> 
> Could you please verify `result` value for those paused tasks?
> I am sure those tasks must have `result=success` not `error` and that's why it is showing paused tasks again and again in check result but doing nothing on resume procedure selection. For such scenarios, I think user has to select second option and from UI they can perform correct action for such paused tasks with result equal to success.
> 
> @upadhyeammit, what do you think?

@jameerpathan111 this change is not culprit to change resume task behavior.

@kgaikwad I think we should change the select query of paused task fm check and include result == 'error' ? as you mentioned and as I can see in 

https://github.com/theforeman/foreman-tasks/blob/36030277a0d8eb6b0bbfb63823f54d4f09b17b03/app/controllers/foreman_tasks/api/tasks_controller.rb#L95-L98 

I think we should be consistent on what we are checking as paused and what we will consider to resume in later stage using tasking system ?

kgaikwad:
@upadhyeammit Thank you! I have extended condition with `result=error` while getting paused task count.
@jameerpathan111, could you please test updated patch.

devendra104:
@kgaikwad hammer ping verified after post upgrade
```
Called /katello/api/ping
DEBUG -- : Response: 200,   {"status":"ok","services":{"pulp":{"status":"ok","duration_ms":"47"},"pulp_auth":{"status":"ok","duration_ms":"18"},"candlepin":{"status":"ok","du
ration_ms":"46"},"candlepin_auth":{"status":"ok","duration_ms":"14"},"foreman_tasks":{"status":"ok","duration_ms":"3"},"katello_events":{"status":"ok","message":"0 Processed, 0 Failed","duration_ms":"0"},"candle
pin_events":{"status":"ok","message":"0 Processed, 0 Failed","duration_ms":"0"}}}
```
But the provided fix failed to resume the paused task.

kgaikwad:
> @kgaikwad hammer ping verified after post upgrade

Appreciated!! Thank you @devendra104! 

> But the provided fix failed to resume the paused task.

One thing I would like to mention is using this paused task check, it identifies paused tasks with result = `error`. If there are any such task found then this check definitely will fail so this is not issue on foreman-maintain side.
Out of suggested procedures,  if user selects `resume task` then this procedure will try to resume those tasks and re-run this check again. 
a) If failure while resuming tasks, then re-running check will get those tasks and result into failure again.
b) If those tasks are successfully resumed and no new paused tasks with result `error`  then re-running check will pass.


From logs that recently shared with me, I observed that both hammer reload and paused task check are passed.

```
 out: - Reloading Hammer cache                                              [OK]
 out: --------------------------------------------------------------------------------
 out: re-enable sync plans:                                                                           
 out: \ Reloading Hammer cache                                                                             
 out: | re-enabling sync plans                                                                               
 out: | Total 0 sync plans are now enabled.                                 [OK]
 out: --------------------------------------------------------------------------------
 out: Remove maintenance_mode chain from iptables:                          [OK]
 out: --------------------------------------------------------------------------------
 out: Running Checks after upgrading to Satellite 6.7
 out: ================================================================================
 out: Check number of fact names in database:                               [OK]
 out: --------------------------------------------------------------------------------
 out: Check for verifying syntax for ISP DHCP configurations:               [SKIPPED]
 out: DHCP feature is not enabled
 out: --------------------------------------------------------------------------------
 out: Check whether all services are running:                               [OK]
 out: --------------------------------------------------------------------------------
 out: Check whether all services are running using the ping call:           [OK]
 out: --------------------------------------------------------------------------------
 out: Check for paused tasks:                                               [FAIL]
 out: There are currently 1 paused tasks in the system
 out: --------------------------------------------------------------------------------
 out: There are multiple steps to proceed:
 out: 1) Resume paused tasks
 out: 2) Investigate the tasks via UI
 out: (assuming first option)                                                                        
 out: Resume paused tasks:                                                  [OK]
 out: Warning: An error occured while loading module hammer_cli_csv.
 out: Total tasks found paused in error state: 1
 out: Total tasks resumed:                     1
 out: Resumed tasks:                           
 out:  1) Task identifier: c0608083-8187-42a4-a34c-5fc1b0341d70
 out:     Task action:     Update http Capsule details
 out:     Task errors:     The world db074132-5c3d-4ca8-a81e-9e8f1aba3987 is not active (terminating or terminated)
 out: Total tasks failed to resume:            0
 out: Failed tasks:                            
 out: Total tasks skipped:                     0
 out: Skipped tasks:
 out: --------------------------------------------------------------------------------
 out: Rerunning the check after fix procedure
 out: Check for paused tasks:                                               [OK]
 out: --------------------------------------------------------------------------------
```
 
Also, I can see that below logs - 
```
out: Upgrade finished.
out: 
Warning: run() received nonzero return code 1 while executing 'foreman-maintain upgrade check --target-version 6.7 -y'!

Fatal error: run() received nonzero return code 1 while executing!

Requested: foreman-maintain upgrade run --whitelist="disk-performance" --target-version 6.7 -y
Executed: /bin/bash -l -c "export FOREMAN_MAINTAIN_USE_BETA=\"1\" && foreman-maintain upgrade run --whitelist=\"disk-performance\" --target-version 6.7 -y"
Aborting.
```
This fatal error is irrelevant to these changes. It would be good to verify it again and share your observations to find real cause of this error.

upadhyeammit:
@kgaikwad merging this PR

PR #299, opened 2019-12-16
Fixes #28517 - Whitelist virt-who package

Rohan21Lobo:
virt-who package should be whitelisted in the foreman-protector plugin
so that hammer virt-who-config procedure does not need package unlocking
to install virt-who.

Signed-off-by: Rohan Arora <roarora@redhat.com>

### DIFF
--- a/extras/foreman_protector/foreman-protector.whitelist
+++ b/extras/foreman_protector/foreman-protector.whitelist
@@ -15,5 +15,6 @@ librdmacm
 rdma-core
 boost-random
 boost-iostreams
+virt-who
 # foreman-maintain
 rubygem-foreman_maintain


upadhyeammit:
Hello Rohan,

Thank you for the PR, if I check the virt-who spec file then I do see there are lot of requirements for this package, good thing is none of the requirements are specific to any version. This means that once package is installed it should not update any other packages on doing 'yum update virt-who'. Only concerning case is when package is not installed, at that time I see it pulls dependencies and then we end up adding more packages to whitelist else system will have dependency issues. So my preference is user should manually unlock system and update 'virt-who' package when required, what are your thoughts ?

@kgaikwad anything you like to add ?

Thank You,
Amit Upadhye.

ahumbe:
I feel that virt-who package needs to be installed on the majority of foreman instances when customers are using VDC type subscription. So it would be good to have virt-who and its dependencies whitelisted.

Rohan21Lobo:
I do not like this but about implementing into hammer deploy if we are not concerned about manual yum install virt-who 
https://github.com/Rohan21Lobo/foreman_virt_who_configure/commit/07f8aae660ea70758851cd679fb95b8fece9ca31

Rohan21Lobo:
@upadhyeammit  I didn't check the spec file but if dependecies are not version based, then seems only following dependencies need to be whitelisted : 
```
systemd-python
libvirt-python
python-suds
```
Rest of dependencies are already required by rhsm,pulp,ansible etc  so they are present. 
This may be difficult to track in future if dependencies change? 

If yes, will prefer yum --disableplugin foreman-protector or  manual unlock/lock instructions while installing in documentation

upadhyeammit:
> I do not like this but what about implementing into hammer deploy if we are not concerned about manual yum install virt-who
> [Rohan21Lobo/foreman_virt_who_configure@07f8aae](https://github.com/Rohan21Lobo/foreman_virt_who_configure/commit/07f8aae660ea70758851cd679fb95b8fece9ca31)
> 
> ```
> yum --disableplugin foreman-protector install -y virt-who 
> ```
> 
> Even if the plugin is not there, yum just says No plugin match for: foreman-protector and still moves forward. We can still check for its existence and add in the yum command if this approach is the chosen resolution

We purposely keep system lock and I don't agree to solution where we suggesting or purposely disabling the same.

upadhyeammit:
> I feel that virt-who package needs to be installed on the majority of foreman instances when customers are using VDC type subscription. So it would be good to have virt-who and its dependencies whitelisted.

I agree that its needed package but its not by default installed. This means even if we put virt-who and its dependencies there is chance that yum will still fail with dependency error, and ultimately user/support person would guide to disable the foreman-protector plugin. Because of this I prefer the standard approach of not keeping virt-who + deps in whitelist and let user update it.

With this there is also chance that dependencies are changing in future and we end up adding more dependent packages on whitelist which is after filing the bugzilla.

There could be more better solutions which I can think of now,

1. Make virt-who by default available on Satellite, and then accept this change.
2. If its not possible to have virt-who on default installation, then auto update virt-who package maybe at the time of Satellite upgrade.

If you have any better solution then request to add. 

Thank You !

upadhyeammit:
@Rohan21Lobo @ahumbe any thoughts on proposed solutions ?

kgaikwad:
Thank you @Rohan21Lobo for your contribution.
As per the offline discussion with @upadhyeammit, this should be fixed inside plugin [foreman_virt_who_configure](https://github.com/theforeman/foreman_virt_who_configure) not in foreman-maintain so closing this pull-request.

PR #300, opened 2019-12-17
Adding content prepare procedure

johnpmitsch:
This adds a procedure to run a rake task to prepare for pulp3. This is part of a larger series of commands that will be used to help prepare for pulp3, more will be added later.

It can be run with `foreman-maintain advanced procedure run content-prepare`

### DIFF
--- a/definitions/procedures/content/prepare.rb
+++ b/definitions/procedures/content/prepare.rb
@@ -1,0 +1,12 @@
+module Procedures::Content
+  class Prepare < ForemanMaintain::Procedure
+    metadata do
+      description 'Prepare content for Pulp 3'
+      for_feature :pulp3
+    end
+
+    def run
+      execute!('foreman-rake katello:pulp3_migration')
+    end
+  end
+end


definitions/procedures/content/prepare.rb
```diff
@@ -0,0 +1,11 @@
+module Procedures::Content
+  class Prepare < ForemanMaintain::Procedure
+    metadata do
+      description 'Prepare content for Pulp 3'
+    end
+
+    def run
+      execute!('foreman-rake katello:pulp3_migration')
```
jlsherrill: is it possible to require the pulp3 'feature' in order for this to run?

johnpmitsch: updated! Giving it a test now on some production machines

johnpmitsch:
I don't think there is a way to test on a pulp3 production machine, but here is the no pulp3 error on a 6.6:

```
[root@sat-6-6-qa-rhel7 foreman_maintain]# ./bin/foreman-maintain advanced procedure run content-prepare
Prepare content for Pulp 3:                                           [FAIL]
Pulp 3 is not installed on this system!

  [content-prepare]

use --whitelist="content-prepare"

```

johnpmitsch:
`foreman-installer --foreman-proxy-plugin-pulp3-enabled true` can install pulp3 on a nightly box

definitions/procedures/content/prepare.rb
```diff
@@ -0,0 +1,15 @@
+module Procedures::Content
+  class Prepare < ForemanMaintain::Procedure
+    metadata do
+      description 'Prepare content for Pulp 3'
+    end
+
+    def run
+      if feature(:pulp3)
```
kgaikwad: Instead of adding condition here, you can set `for_feature` inside metadata. Example -

```
metadata do
      description 'Prepare content for Pulp 3' 
      for_feature :pulp3
end
```

So that this procedure will be available only when `pulp3` feature exists.

johnpmitsch: Thanks, updated!

jlsherrill:
@kgaikwad this looks like its ready for re-review (and approval if everything is good).  Thanks!

kgaikwad:
Thank you @johnpmitsch

PR #301, opened 2019-12-24
Fixes #28562 - Fix fio deps issue

upadhyeammit:
If Satellite/Capsule does not have foreman-discovery-image package installed the boost-thread package is missing on system. In this case installation of fio package fails because boost-thread is not whitelisted package. To avoid this situation added boost-thread package in whitelist.

### DIFF
--- a/extras/foreman_protector/foreman-protector.whitelist
+++ b/extras/foreman_protector/foreman-protector.whitelist
@@ -15,5 +15,6 @@ librdmacm
 rdma-core
 boost-random
 boost-iostreams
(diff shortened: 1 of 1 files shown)


jameerpathan111:
> LGTM +1
> @jameerpathan111, please have one round of testing.

@kgaikwad already tested it on capsule 6.6 m/c, when I provided reproducer for this issue to @upadhyeammit .

kgaikwad:
Merging this change! Thank you @jameerpathan111

PR #303, opened 2019-12-30
Fixes #28184 - add unlink option to fio to clean benchmark file

kgaikwad:


jameerpathan111:
@kgaikwad 
Tested the PR and it seems to fix the issue.

Test steps:
1. Run `foreman-maintain health check --label disk-performance`
2. `ls -lh /var/lib/pulp/job1.0.0`

Observation:
2. While check is running it creates `/var/lib/pulp/job1.0.0` 
3. `/var/lib/pulp/job1.0.0` is not present ones check finishes execution.

kgaikwad:
Thanks @jameerpathan111, merging this PR.

PR #305, opened 2020-01-06
Fixes #28649 - self upgrade foreman-maintain

upadhyeammit:
This adds functionality to auto upgrade foreman-maintain, 

1. The auto upgrade happens only when using upgrade command with check and run options. This is to avoid calling upgrade every time foreman-maintain is being used. i.e to restart services.
2. This introduces a new method update_available? for Yum Class. This method accepts a package name as argument and returns true if update is available for same.
3. New flag for check and run command, which is '--disable-self-upgrade' . One can disable the auto upgrade using this.

### DIFF
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -140,6 +140,18 @@ module ForemanMaintain
       logger.error "Invalid Storage label i.e #{label}. Error - #{e.message}"
     end
 
+    def perform_self_upgrade
+      puts 'Checking for new version of foreman-maintain...'
+      if ForemanMaintain.package_manager.update_available?('rubygem-foreman_maintain')
+        puts "\nUpdating rubygem-foreman_maintain package."
+        ForemanMaintain.package_manager.update('rubygem-foreman_maintain', :assumeyes => true)
+        puts "\nThe rubygem-foreman_maintain package successfully updated."\
+            "\nRe-run satellite-maintain or foreman-maintain with required options!"
+        exit 0
+      end
+      puts "Nothing to update, can't find new version of rubygem-foreman_maintain!"
+    end
+
     def upgrade_in_progress
       storage[:upgrade_target_version]
     end
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -140,16 +140,26 @@ module ForemanMaintain
       logger.error "Invalid Storage label i.e #{label}. Error - #{e.message}"
     end
 
+    def pkg_and_cmd_name
+      instance_feature = ForemanMaintain.available_features(:label => :instance).first
+      if instance_feature.downstream
+        return %w[satellite-maintain satellite-maintain]
+      end
+
+      %w[rubygem-foreman_maintain foreman-maintain]
+    end
+
     def perform_self_upgrade
-      puts 'Checking for new version of foreman-maintain...'
+      package_name, command = pkg_and_cmd_name
+      puts "Checking for new version of #{package_name}..."
       if ForemanMaintain.package_manager.update_available?('rubygem-foreman_maintain')
-        puts "\nUpdating rubygem-foreman_maintain package."
+        puts "\nUpdating #{package_name} package."
         ForemanMaintain.package_manager.update('rubygem-foreman_maintain', :assumeyes => true)
-        puts "\nThe rubygem-foreman_maintain package successfully updated."\
-            "\nRe-run satellite-maintain or foreman-maintain with required options!"
+        puts "\nThe #{package_name} package successfully updated."\
+            "\nRe-run #{command} with required options!"
         exit 0
       end
-      puts "Nothing to update, can't find new version of rubygem-foreman_maintain!"
+      puts "Nothing to update, can't find new version of #{package_name}."
     end
 
     def upgrade_in_progress
--- a/lib/foreman_maintain/cli/upgrade_command.rb
+++ b/lib/foreman_maintain/cli/upgrade_command.rb
@@ -6,6 +6,11 @@ module ForemanMaintain
                :required => false
       end
 
+      def self.disable_self_upgrade_option
+        option '--disable-self-upgrade', :flag, "Don't auto update rubygem-foreman_maintain",
+               :default => false
+      end
+
       def current_target_version
         current_target_version = ForemanMaintain::UpgradeRunner.current_target_version
         if current_target_version && target_version && target_version != current_target_version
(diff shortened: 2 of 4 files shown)


upadhyeammit:
@kgaikwad request to review. Thank You !

kgaikwad:
As per design, foreman-maintain will try to do self upgrade ONLY in case of upgrade sub-commands.
But in my opinion, this action is more specific to ForemanMaintain itself so `self_upgrade` method shouldn't be a part of `ForemanMaintain::CLI::UpgradeCommand`.
I might be wrong but either `lib/foreman_maintain.rb` or a separate small module called `auto_update` could be elegant way to move this. In case, we require the same behavior at different commands later then it would be easy to call.

@upadhyeammit, @mbacovsky  any thoughts on this?

upadhyeammit:
> As per design, foreman-maintain will try to do self upgrade ONLY in case of upgrade sub-commands.
> But in my opinion, this action is more specific to ForemanMaintain itself so `self_upgrade` method shouldn't be a part of `ForemanMaintain::CLI::UpgradeCommand`.
> I might be wrong but either `lib/foreman_maintain.rb` or a separate small module called `auto_update` could be elegant way to move this. In case, we require the same behavior at different commands later then it would be easy to call.
> 
> @upadhyeammit, @mbacovsky any thoughts on this?

I agree to your thoughts ! However I am not sure apart from upgrade where we shall need it ?

upadhyeammit:
@kgaikwad moved the code to new module, request to have a look. Thank You !

jameerpathan111:
@upadhyeammit 
I have tried this PR on Satellite 6.7 and it works as expected.
Test steps:
1. Downgrade FM version 
2. Run `# foreman-maintain upgrade check --target-version 6.7.z`
It updates FM to latest version and gives message `Re-run satellite-maintain or foreman-maintain with required options!`
3. Downgrade FM version 
4. Run `# foreman-maintain upgrade check --target-version 6.7.z --disable-self-upgrade`
Skips the auto-upgrade step of FM.
5. Run `# foreman-maintain upgrade run --target-version 6.7.z --disable-self-upgrade`
Skips the auto-upgrade step of FM.
6. Run `# foreman-maintain upgrade run --target-version 6.7.z`
It updates FM to latest version and gives message `Re-run satellite-maintain or foreman-maintain with required options!`



```

[root@dhcp-3-139 foreman_maintain]# foreman-maintain upgrade run --target-version 6.7.z 
Checking for new version of foreman-maintain...

rubygem-foreman_maintain.noarch                    1:0.5.2-1.el7sat                    Sat6-CI_Red_Hat_Satellite_6_7_Composes_Satellite_Maintenance_Next_RHEL7

Updating rubygem-foreman_maintain package.
Loaded plugins: foreman-protector, product-id, search-disabled-repos, subscription-manager
Sat6-CI_Red_Hat_Satellite_6_7_Composes_Satellite_6_7_RHEL7                                                                             | 2.5 kB  00:00:00     
Sat6-CI_Red_Hat_Satellite_6_7_Composes_Satellite_Tools_6_7_RHEL7                                                                       | 2.1 kB  00:00:00     

WARNING: Excluding 4618 updates due to foreman-protector. 

Resolving Dependencies
--> Running transaction check
---> Package rubygem-foreman_maintain.noarch 1:0.4.10-1.el7sat will be updated
---> Package rubygem-foreman_maintain.noarch 1:0.5.2-1.el7sat will be an update

Dependencies Resolved

==============================================================================================================================================================
 Package                         Arch          Version                   Repository                                                                      Size
==============================================================================================================================================================
Updating:
 rubygem-foreman_maintain        noarch        1:0.5.2-1.el7sat          Sat6-CI_Red_Hat_Satellite_6_7_Composes_Satellite_Maintenance_Next_RHEL7        141 k

Transaction Summary
==============================================================================================================================================================
Upgrade  1 Package

Total download size: 141 k
Downloading packages:
No Presto metadata available for Sat6-CI_Red_Hat_Satellite_6_7_Composes_Satellite_Maintenance_Next_RHEL7
rubygem-foreman_maintain-0.5.2-1.el7sat.noarch.rpm                                                                                     | 141 kB  00:00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Updating   : 1:rubygem-foreman_maintain-0.5.2-1.el7sat.noarch                                                                                           1/2 
  Cleanup    : 1:rubygem-foreman_maintain-0.4.10-1.el7sat.noarch                                                                                          2/2 
  Verifying  : 1:rubygem-foreman_maintain-0.5.2-1.el7sat.noarch                                                                                           1/2 
  Verifying  : 1:rubygem-foreman_maintain-0.4.10-1.el7sat.noarch                                                                                          2/2 

Updated:
  rubygem-foreman_maintain.noarch 1:0.5.2-1.el7sat                                                                                                            

Complete!

The rubygem-foreman_maintain package successfully updated.
Re-run satellite-maintain or foreman-maintain with required options!
```

lib/foreman_maintain.rb
```diff
@@ -139,5 +139,17 @@ def storage(label = :default)
...
       logger.error "Invalid Storage label i.e #{label}. Error - #{e.message}"
     end
+
+    def perform_self_upgrade
+      puts 'Checking for new version of foreman-maintain...'
+      if ForemanMaintain.package_manager.update_available?('rubygem-foreman_maintain')
+        puts "\nUpdating rubygem-foreman_maintain package."
+        ForemanMaintain.package_manager.update('rubygem-foreman_maintain', :assumeyes => true)
+        puts "\nThe rubygem-foreman_maintain package successfully updated."\
+            "\nRe-run satellite-maintain or foreman-maintain with required options!"
```
kgaikwad: As offline discussion we decided to show both but after giving second thought it would be good to show only one at a time.
For upstream users, it will always show `foreman-maintain` but for downstream I think it should show only `satellite-maintain`

kgaikwad: @upadhyeammit, @jameerpathan111, what do you think?

jameerpathan111: I am OK with either.

kgaikwad:
@upadhyeammit, having conflicts due to recent merge.

lib/foreman_maintain.rb
upadhyeammit: Hello @kgaikwad I have added method to handle it as per upstream and downstream. Request to review, thank you !

lib/foreman_maintain.rb
```diff
@@ -140,6 +140,27 @@ def storage(label = :default)
       logger.error "Invalid Storage label i.e #{label}. Error - #{e.message}"
     end
 
+    def pkg_and_cmd_name
+      if package_manager.installed?(['satellite'])
+        return %w[satellite-maintain satellite-maintain]
+      end
+
+      %w[rubygem-foreman_maintain foreman-maintain]
+    end
```
kgaikwad: As we have added Capsule support, instead of just checking satellite package could you please do something like -

```
def pkg_and_cmd_name
      instance_feature = ForemanMaintain.available_features(:label => :instance).first
      if instance_feature.downstream
        return %w[satellite-maintain satellite-maintain]
      end

      %w[rubygem-foreman_maintain foreman-maintain]
    end
```

lib/foreman_maintain/cli/upgrade_command.rb
```diff
@@ -6,6 +6,11 @@ def self.target_version_option
                :required => false
       end
 
+      def self.disable_self_upgrade_option
+        option '--disable-self-upgrade', :flag, 'Disable foreman-maintain auto-upgrade',
```
kgaikwad: You are modifying package and command name but here it still shows `foreman-maintain`.

upadhyeammit:
@kgaikwad Added the changes as per suggestions. Request to review. Thank You !

sthirugn:
this is an awesome feature, thank you for doing this

PR #307, opened 2020-01-09
Fixes #28673 - Add SCL redis service name

johnpmitsch:
In production we will use redis from the Red Hat SCL. This commit adds this service to the pulp3 feature.

### DIFF
--- a/definitions/features/pulp3.rb
+++ b/definitions/features/pulp3.rb
@@ -15,7 +15,7 @@ class Features::Pulp3 < ForemanMaintain::Feature
       system_service('pulpcore-content', 10),
       system_service('pulpcore-resource-manager', 10),
       system_service('pulpcore-worker@*', 20, :all => true, :skip_enablement => true),
-      system_service('redis', 30),
+      system_service('rh-redis5-redis', 30),
       system_service('httpd', 30)
     ]
   end


johnpmitsch:
I'm not sure if we have a way to properly test this on a pulp3 production install yet, but it should match the name of the service same as any other service we list.

johnpmitsch:
We use a bit more complicated logic for mongo (which we also use an SCL package for production for), determining which package is installed.  However, it looks like that is mostly to handle the db operations we run and ensure they are executed with the same command. 

If we are just worried about having the service starting/stopping/restarting, I think having both service names will be fine. I don't think we are plan to do any redis operations in foreman-maintain.

Just want to share my thoughts in case I am assuming the wrong thing, please let me know if there is something I am missing :)

johnpmitsch:
> We use a bit more complicated logic for mongo (which we also use an SCL package in production for), determining which package is installed. However, it looks like that is mostly to handle the db operations we run and ensure they are executed with the proper command.
> 
> If we are just worried about having the service starting/stopping/restarting, I think having both service names will be fine. I don't think we plan to do any redis operations in foreman-maintain.
> 
> Just want to share my thoughts in case I am assuming the wrong thing, please let me know if there is something I am missing :)

UPDATE: After some discussion we determined that redis is not used much in the dev environment and it will be going away in dev when we move to a production pulp3 install, so I removed the `redis` service in this PR.

PR #308, opened 2020-01-13
Fixes #28500 - Pulp 3 backup/restore

ianballou:
This adds backup and restore support for Pulpcore (Pulp 3).

One question I still have is how to tell which `pgsql` folder to use for the `data_dir`.  Is it okay to assume to use the PostgreSQL 10 folder if PostgreSQL is installed?  There's a few ways I can check, either by checking the package, the existence of the postgres config file, or by reading the output of `postgres -V`.  I have the middle option commented out and the last option applied currently.

I've tested `online` and `offline` backups and restoring.  `snapshot` backups don't seem to work at the moment (`undefined method `data_dir' for nil:NilClass` during `backup-snapshot-logical-volume-confirmation`)

TO TEST:
1) Make a nightly production environment (this has been tricky lately, let me know if you're having issues).
2) Install Pulp 3 with `foreman-installer --foreman-proxy-plugin-pulp-pulpcore-enabled`
3) Checkout this PR
4) Try `foreman-maintain backup` with `online` and `offline`
5) Try `foreman-maintain restore` on all of the backups created
6) Ensure the databases restored okay, especially the Pulp 3 database.

### DIFF
--- a/definitions/checks/pulpcore/db_up.rb
+++ b/definitions/checks/pulpcore/db_up.rb
@@ -1,0 +1,29 @@
+module Checks
+  module Pulpcore
+    class DBUp < ForemanMaintain::Check
+      metadata do
+        description 'Make sure Pulpcore DB is up'
+        label :pulpcore_db_up
+        for_feature :pulpcore_database
+      end
+
+      def run
+        status = false
+        with_spinner('Checking connection to the Pulpcore DB') do
+          status = feature(:pulpcore_database).ping
+        end
+        assert(status, 'Pulpcore DB is not responding. ' \
+          'It needs to be up and running to perform the following steps',
+               :next_steps => next_steps)
+      end
+
+      def next_steps
+        if feature(:pulpcore_database).local?
+          [Procedures::Service::Start.new(:only => 'postgresql')]
+        else
+          [] # there is nothing we can do for remote db
+        end
+      end
+    end
+  end
+end
--- a/definitions/features/foreman_database.rb
+++ b/definitions/features/foreman_database.rb
@@ -16,9 +16,11 @@ class Features::ForemanDatabase < ForemanMaintain::Feature
   end
 
   def config_files
-    [
-      '/var/lib/pgsql/data/postgresql.conf'
-    ]
+    if check_min_version('foreman', '1.25')
+      ['/var/opt/rh/rh-postgresql10/lib/pgsql/data/postgresql.conf']
+    else
+      ['/var/lib/pgsql/data/postgresql.conf']
+    end
   end
 
   def services
--- a/definitions/features/instance.rb
+++ b/definitions/features/instance.rb
@@ -37,7 +37,9 @@ class Features::Instance < ForemanMaintain::Feature
   end
 
   def postgresql_local?
-    database_local?(:candlepin_database) || database_local?(:foreman_database)
+    database_local?(:candlepin_database) ||
+      database_local?(:foreman_database) ||
+      database_local?(:pulpcore_database)
   end
 
   def foreman_proxy_with_content?
@@ -65,7 +67,7 @@ class Features::Instance < ForemanMaintain::Feature
   end
 
   def pulp
-    feature(:pulp2) || feature(:pulp3)
+    feature(:pulp2) || feature(:pulpcore)
   end
 
   private
@@ -143,6 +145,7 @@ class Features::Instance < ForemanMaintain::Feature
       'candlepin' => %w[candlepin candlepin_database],
       'pulp_auth' => %w[pulp2 mongo],
       'pulp' => %w[pulp2 mongo],
+      'pulpcore' => %w[pulpcore pulpcore_database],
       'foreman_tasks' => %w[foreman_tasks]
     }
   end
--- a/definitions/features/pulpcore.rb
+++ b/definitions/features/pulpcore.rb
@@ -1,8 +1,8 @@
 require 'foreman_maintain/utils/service/systemd'
 
-class Features::Pulp3 < ForemanMaintain::Feature
+class Features::Pulpcore < ForemanMaintain::Feature
   metadata do
-    label :pulp3
+    label :pulpcore
 
     confine do
       ForemanMaintain::Utils::Service::Systemd.new('pulpcore-api', 0).exist?
@@ -17,6 +17,12 @@ class Features::Pulp3 < ForemanMaintain::Feature
       system_service('pulpcore-worker@*', 20, :all => true, :skip_enablement => true),
       system_service('rh-redis5-redis', 30),
       system_service('httpd', 30)
+    ]
+  end
+
+  def config_files
+    [
+      '/etc/pulp/settings.py'
     ]
   end
 end
--- a/definitions/features/pulpcore_database.rb
+++ b/definitions/features/pulpcore_database.rb
@@ -1,0 +1,39 @@
+class Features::PulpcoreDatabase < ForemanMaintain::Feature
+  PULPCORE_DB_CONFIG = '/etc/pulp/settings.py'.freeze
+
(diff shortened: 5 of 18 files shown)


ianballou:
Looking like it might be better to check the version to decide which PostgreSQL to choose.  I'll try using the helper here (thanks @johnpmitsch) https://github.com/theforeman/foreman_maintain/blob/25850e90a2aa1c08f5eddbe3ccd25cd6643dea5e/lib/foreman_maintain/concerns/system_helpers.rb#L35-L39

ianballou:
For now the version I'm checking Foreman 1.25, but that will exist as 2.0 at release time.  Any opinions about either checking for both to support development, or just leaving it at 2.0?

johnpmitsch:
> For now the version I'm checking Foreman 1.25, but that will exist as 2.0 at release time. Any opinions about either checking for both to support development, or just leaving it at 2.0?

If its doing actual version comparisons, which I think that method eventually does with `Gem::Version`, wouldn't a minimum version of `1.25` cover both cases?

ianballou:
@johnpmitsch that makes sense, so 1.25 should do just fine then!

johnpmitsch:
I'm sure you will eventually get the test failures, but you will need to update the files in https://github.com/theforeman/foreman_maintain/tree/master/test/files/backups adding the pulpcore_dump file (they are just blank dummy files) where appropriate

ianballou:
We're now hitting and error where `foreman-installer --reset` is running system checks when it shouldn't be.  The likely cause is this if statement here: https://github.com/theforeman/foreman_maintain/blob/488ae6726a620f94859ae4ac2be569dbc15e900b/definitions/procedures/restore/installer_reset.rb#L23

`check_min_version('katello-installer-base', '3.2.0')` doesn't seem to be relevant any more.

johnpmitsch:
> We're now hitting and error where `foreman-installer --reset` is running system checks when it shouldn't be. The likely cause is this if statement here:
> 
> https://github.com/theforeman/foreman_maintain/blob/488ae6726a620f94859ae4ac2be569dbc15e900b/definitions/procedures/restore/installer_reset.rb#L23
> 
> `check_min_version('katello-installer-base', '3.2.0')` doesn't seem to be relevant any more.

@kgaikwad @mbacovsky This katello-installer-base package does not exist on machines anymore so this is always resolving to false and the `--disable-system-checks` isn't being added on restore. I'm surprised this hasn't come up earlier (either everyone restoring has enough memory or no one is restoring!)

Can we remove this check on the katello-installer-base package? Its very old, I'm fuzzy on the versions, but I think it was to support 6.1 or 6.2, which didn't have this option. I think its ok to remove it at this point, but want to make sure i'm not missing anything.

johnpmitsch:
> Its very old, I'm fuzzy on the versions, but I think it was to support 6.1 or 6.2, which didn't have this option. 

Was able to dig up the versions:

Satellite 6.2: katello-installer-base-3.0.0.79-1.el7sat.noarch.rpm
Satellite 6.3: katello-installer-base-3.4.0.1-1.el7sat.noarch.rpm

So, this was a backwards-compatible check for 6.2, which I think still included `katello-backup` and `katello-restore`. For 6.3 and later, the installer includes the option `--disable-system-checks`

ianballou:
Looking like something relating to offline backups isn't working at the moment.  I'm checking it out.

ianballou:
I may have fixed the issue in my latest push.  I need to wait for the nightly box to work again with Pulp 3 to test.

johnpmitsch:
@ianballou I just thought of something - since foreman-maintain always has to keep backwards compatibility, it seems like checking for `pulpcore_dump` in the backup directory won't work for the backup checks in older versions. We should make sure we test out on 6.6 machine or similar too.

You could gate that particular check with a version check or maybe there is another way to maintain backwards-compatibility.

ianballou:
Ah right, yeah it will probably fail when it doesn't see the pulpcore DB. I'll block it with a check.

johnpmitsch:
@ianballou One thing I'm seeing is that its looking like the pulpcore db isn't actually dropping.

Can you double check this by doing the following?
1. Take a backup
2. Add something to pulpcore db:
For example, you could add directly to the db:
```
[vagrant@centos7-katello-nightly foreman_maintain]$ sudo su - postgres
Last login: Wed Jan 22 03:02:16 UTC 2020 on pts/0
-bash-4.2$ psql pulpcore
pulpcore=# INSERT INTO core_upload (pulp_id, pulp_created, pulp_last_updated, file, size) VALUES ('f72e9704-e7e3-4d7d-ad06-a7071c93c59b','2020-01-15 16:06:25.106069+00', '2020-01-15 16:06:25.106069+00', '1', 1);
INSERT 0 1
pulpcore=# select * from core_upload;
               pulp_id                |         pulp_created          |       pulp_last_updated       | file | size 
--------------------------------------+-------------------------------+-------------------------------+------+------
 f72e9704-e7e3-4d7d-ad06-a7071c93c59b | 2020-01-15 16:06:25.106069+00 | 2020-01-15 16:06:25.106069+00 | 1    |    1
(1 row)
```
3. Restore backup and check db for the row you added before the restore. 

In this case, it was still there. When I manually `dropdb pulpcore` before restoring, the custom row is gone and the db is properly restored.

I'm not sure why this is happening, the code seems to indicate that the db will be dropped. Let me know if you see anything different. The rest of the functionality and code looks fine and I tested on a 6.6 box ok.

ianballou:
@johnpmitsch I did see the pulpcore database revert to a previous online backup state.  However I did see this issue you mentioned after trying to revert to a previous offline backup state.  I'm testing some more to see if this is always true.

ianballou:
In the log I noticed that "Drop postgresql databases" never appeared for the offline backup restore but it does for online.  Seems like even the foreman and candlepin databases aren't being dropped either.

ianballou:
What I found so far: online backup uses the `.dump` files.  The existence of these files tells foreman-maintain to drop the databases.  Offline backup doesn't use these, and doesn't drop the databases itself.  I think `foreman-installer --reset` takes care of that, but Pulp 3 doesn't get reset as part of that yet.  For now I think I'll have to add in an extra Pulp 3 drop database.

johnpmitsch:
> What I found so far: online backup uses the .dump files. The existence of these files tells foreman-maintain to drop the databases. Offline backup doesn't use these, and doesn't drop the databases itself. I think foreman-installer --reset takes care of that, but Pulp 3 doesn't get reset as part of that yet. For now I think I'll have to add in an extra Pulp 3 drop database.

That makes sense and I was using an offline backup in those steps. 

IMO we can add to `foreman-installer --reset` separately and merge with that assumption as long as we create a task to do so.

ianballou:
@johnpmitsch that works for me.

johnpmitsch:
@kgaikwad @mbacovsky Any thoughts on this?

> > We're now hitting and error where `foreman-installer --reset` is running system checks when it shouldn't be. The likely cause is this if statement here:
> > https://github.com/theforeman/foreman_maintain/blob/488ae6726a620f94859ae4ac2be569dbc15e900b/definitions/procedures/restore/installer_reset.rb#L23
> > 
> > `check_min_version('katello-installer-base', '3.2.0')` doesn't seem to be relevant any more.
> 
> @kgaikwad @mbacovsky This katello-installer-base package does not exist on machines anymore so this is always resolving to false and the `--disable-system-checks` isn't being added on restore. I'm surprised this hasn't come up earlier (either everyone restoring has enough memory or no one is restoring!)
> 
> Can we remove this check on the katello-installer-base package? Its very old, I'm fuzzy on the versions, but I think it was to support 6.1 or 6.2, which didn't have this option. I think its ok to remove it at this point, but want to make sure i'm not missing anything.

johnpmitsch:
@ianballou all looks good and works fine, just wondering about that `--disable-system-checks` line

PR #309, opened 2020-01-14
Fixes #28700 - flatten packages list before dnf action

kgaikwad:


upadhyeammit:
@jameerpathan111 any updates on this one ?

upadhyeammit:
@jameerpathan111 we are not allowing the dnf on EL7 with this, this is to avoid issue,
~~~
E, [2019-11-08 10:54:52+0100 #6382] ERROR -- : undefined method `join' for "all":String (NoMethodError)
~~~
while running command. The dnf support with EL7 is as per the support policy of EL7 itself.

PR #310, opened 2020-01-16
Fixes #28775 - use query_csv method

upadhyeammit:
The foreman_database feature has query_csv method which can be used exclusively to get the result in csv format. The export_csv was having independent pgqsl command invocation which was causing issue when having the external database. The foreman_database already handles internal/external database and so using query_csv fixes this issue.

### DIFF
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -162,8 +162,13 @@ class Features::ForemanTasks < ForemanMaintain::Feature
   def export_csv(sql, file_name, state)
     dir = prepare_for_backup(state)
     filepath = "#{dir}/#{file_name}"
-    execute("echo \"COPY (#{sql}) TO STDOUT WITH CSV;\" \
-      | su - postgres -c '/usr/bin/psql -d foreman' | bzip2 -9 > #{filepath}.bz2")
+    csv_output = feature(:foreman_database).query_csv(sql)
+    File.open(filepath, 'w') do |f|
+      f.write(csv_output)
+      f.close
+    end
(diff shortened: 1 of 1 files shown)


upadhyeammit:
@kgaikwad This needs more work, I am changing it to WIP.

upadhyeammit:
I have to create .csv as intermediate file and supply that to bzip2 because if I echo it then it was failing with error, I tried to redirect echo output in all possible ways I know to bzip2 however it did not help, I think this is better approach considering we may have long list of tasks. With this it was also printing whole csv file to stdout, I tried to silent it and redirect but it did not have affect, 
~~~
- Backup dynflow_execution_plans [running]                            [FAIL]    
Argument list too long
~~~

I tested this on external database, its working as expected. Request @kgaikwad to review once. 
Thank You !

kgaikwad:
@lvrtelov, 
Can you help in testing this PR as it is related to backup. You will be able to reproduce this when there is external DB configured.

lvrtelov:
@kgaikwad sure I can take a look at it..but I'm not sure it will be this week..is it ok?

kgaikwad:
@lvrtelov, any updates on this?

lvrtelov:
@kgaikwad  I tested it on this set up - all DBs external with SSL connection. Apply this patch to f-m and run the check with ~30 tasks in stopped/paused state. Backups were successfully made in  /var/lib/foreman-maintain/backup-tasks. So It's ACK from my side.

kgaikwad:
Thank you @lvrtelov.

@upadhyeammit, it would be good to have more meaningful commit message. Otherwise good to go.

PR #311, opened 2020-01-24
Fixes #28040 - Unknown Puppet subcommand 'master'

patilsuraj767:
puppet master command and its subcommands have been removed from puppet 6.
I have changed the puppet path to absolute because installation of puppet 6 package does not create symbolic link /usr/bin/puppet and /opt/puppetlabs/bin/puppet is not added to the $PATH.

### DIFF
--- a/definitions/features/puppet_server.rb
+++ b/definitions/features/puppet_server.rb
@@ -29,7 +29,7 @@ class Features::PuppetServer < ForemanMaintain::Feature
   end
 
   def puppet_version
-    version(execute!('/opt/puppetlabs/bin/puppet --version'))
+    version(execute!("#{puppet_path} --version"))
   end
 
   def find_empty_cacert_request_files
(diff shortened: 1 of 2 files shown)


patilsuraj767:
Reference - https://puppet.com/docs/puppet/6.0/release_notes_puppet.html

definitions/features/puppet_server.rb
```diff
@@ -56,6 +56,10 @@ def ca_directory_path
   end
 
   def puppet_ssldir_path
-    execute!('puppet master --configprint ssldir')
+    if puppet_version.version.to_i >= 6
```
kgaikwad:
```suggestion
    if puppet_version.major >= 6
```

kgaikwad:
> Also, verify facter file which uses puppet command [here](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/utils/facter.rb#L8)

@patilsuraj767, please also check this file which uses puppet command.

patilsuraj767:
> > Also, verify facter file which uses puppet command [here](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/utils/facter.rb#L8)
> 
> @patilsuraj767, please also check this file which uses puppet command.

@kgaikwad made the required changes, Please review it again. 
Thanks.

definitions/features/puppet_server.rb
```diff
@@ -56,6 +56,14 @@ def ca_directory_path
   end
 
   def puppet_ssldir_path
-    execute!('puppet master --configprint ssldir')
+    if puppet_version.major >= 6
```
kgaikwad, line 59: @patilsuraj767 ,
while testing this PR,  observed that `puppet master` command is deprecated in latest versions but `puppet config print` command available in all versions.
So instead of conditional run, can we just use `execute!("#{puppet_path} config print ssldir")`?

patilsuraj767: Yes, agreed. 
I have made the required changes.

kgaikwad:
Thank you @patilsuraj767 for adding changes as per review comments.
Merging this pull-request.

PR #313, opened 2020-01-29
Fixes #28882: Update postgresql data location

ehelms:


kgaikwad:
@ehelms,
LGTM but I guess at one more place i.e in foreman database [feature](https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/foreman_database.rb#L20) this needs to be fix.

PR #314, opened 2020-02-06
Refs #28948 - Make pulp settings parsing more robust

johnpmitsch:
There is some more detail in the issue, but basically we are parsing a python file to get the pulp db settings. We should come up with a better solution but this change will make things more robust for now and fix the 'service' commands in the development environment.

### DIFF
--- a/definitions/features/pulpcore_database.rb
+++ b/definitions/features/pulpcore_database.rb
@@ -25,7 +25,7 @@ class Features::PulpcoreDatabase < ForemanMaintain::Feature
   private
 
   def load_configuration
-    full_config = File.read(PULPCORE_DB_CONFIG).split(/[\s,':]/).reject(&:empty?)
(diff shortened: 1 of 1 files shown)


definitions/features/pulpcore_database.rb
```diff
@@ -34,6 +34,7 @@ def load_configuration
     @configuration['database'] = full_config[full_config.index('NAME') + 1]
     @configuration['username'] = full_config[full_config.index('USER') + 1]
     @configuration['password'] = full_config[full_config.index('PASSWORD') + 1]
+    puts @configuration
```
jlsherrill: did you mean to leave this?

johnpmitsch: oops

johnpmitsch: fixed!

PR #316, opened 2020-02-17
Fixes #29002 - epel check before package install preparation step

kgaikwad:
It is just fixing the upgrade flow but in my opinion we should check EPEL repository as a part of packages command at least for downstream if we want to run epel check before package install.

To fix the upgrae behaviour, I have added EPEL check explicitly in preparation steps. As a consequence of this change, foreman-maintain will run this check twice while running foreman-maintain upgrade commands.


definitions/checks/disk/performance.rb
```diff
@@ -5,7 +5,9 @@ class Performance < ForemanMaintain::Check
         label :disk_performance
         description 'Check for recommended disk speed of pulp, mongodb, pgsql dir.'
         tags :pre_upgrade
-        preparation_steps { Procedures::Packages::Install.new(:packages => %w[fio]) }
+        preparation_steps do
+          [Checks::CheckEpelRepository.new, Procedures::Packages::Install.new(:packages => %w[fio])]
```
kgaikwad: Here, I need to check downstream condition.

definitions/checks/repositories/check_upstream_repository.rb
```diff
@@ -4,7 +4,7 @@ class Checks::CheckUpstreamRepository < ForemanMaintain::Check
     description 'Check if any upstream repositories are enabled on system'
     tags :pre_upgrade
     preparation_steps do
-      Procedures::Packages::Install.new(:packages => %w[yum-utils])
+      [Checks::CheckEpelRepository.new, Procedures::Packages::Install.new(:packages => %w[yum-utils])]
```
upadhyeammit: Few times I thought of combining EPEL check and upstream repo check. Because on downstream we dont want any of these repositories ? what do you think ?

upadhyeammit:
added one comment rest looks good to me!

definitions/checks/repositories/check_upstream_repository.rb
kgaikwad: In my opinion, for now we can keep this as it is because for remediation procedure it needs `yum-utils` package to be installed.

kgaikwad:
@jameerpathan111, could you please mind doing some testing?

jameerpathan111:
> It is just fixing the upgrade flow but in my opinion we should check EPEL repository as a part of packages command at least for downstream if we want to run epel check before package install.

yeah, I guess we should check EPEL repository as part of packages command.

kgaikwad:
> yeah, I guess we should check EPEL repository as part of packages command.

As per offline discussion on email list, the epel check is valid only for the install and upgrade processes. So it won't be part of packages command.

kgaikwad:
@upadhyeammit, I have modified check description and below messages again.

definitions/checks/check_non_rh_repository.rb
```diff
@@ -1,6 +1,6 @@
-class Checks::CheckEpelRepository < ForemanMaintain::Check
+class Checks::CheckNonRhRepository < ForemanMaintain::Check
   metadata do
-    label :check_epel_repository
+    label :check_non_redhat_repository
     description 'Check to verify no any non Red Hat repositories(Eg: EPEL) enabled'
```
upadhyeammit, line 4:
```suggestion
    description 'Check to verify there isn't any non Red Hat repositories (Eg: EPEL) enabled'
```

kgaikwad:
@upadhyeammit, 
Modified both description and spinner message. In addition to this, I have moved this check under `checks > repositories` directory

upadhyeammit:
@kgaikwad Thank You !

PR #319, opened 2020-02-27
Fixes #26017 - Skip qpidd under load

upadhyeammit:
The foreman-maintain already had logic to try taking config file backup 3 times. With this PR,

1. We try the to collect backup 2 times with qpidd data. Here 10 seconds delay between each run.
2. If backup fails then we skip qpidd data in third run,
~~~
WARNING: Attempt 1/3 to collect all config files failed!
Some files were modified during creation of the archive.
Waiting 10 seconds before re-try
/ Collecting config files to backup                                             
WARNING: Attempt 2/3 to collect all config files failed!
Some files were modified during creation of the archive.
Waiting 10 seconds before re-try
| Collecting config files to backup                                             
WARNING: Attempt 3/3 to collect all config files failed!
Some files were modified during creation of the archive.
- Collecting config files to backup, skipping /var/lib/qpidd                    
| Collecting config files to backup                                   [OK]      
~~

### DIFF
--- a/definitions/procedures/backup/config_files.rb
+++ b/definitions/procedures/backup/config_files.rb
@@ -17,12 +17,15 @@ module Procedures::Backup
             :array => true, :default => ['all']
       param :ignore_changed_files, 'Should packing tar ignore changed files',
             :flag => true, :default => false
+      param :exclude_qpidd_data, 'Exclude qpidd /var/lib/qpidd/ Directory during backup',
+            :flag => true, :default => false
     end
 
     # rubocop:disable Metrics/MethodLength
     def run
       logger.debug("Invoking tar from #{FileUtils.pwd}")
-      tar_cmd = tar_command
+      configs, to_exclude = config_files
+      tar_cmd = tar_command(configs, to_exclude)
       attempt_no = 1
       loop do
         runner = nil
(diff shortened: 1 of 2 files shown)


upadhyeammit:
need to test this in two ways,

1. Collect online backup in regular way. Its expected that we collect the qpidd directory and restore it.
2. Collect the online backup while below command is running in extra terminal,
~~~
# while true; do  echo " " >> /var/lib/qpidd/test;done
~~~
Here we expect that qpidd directory is being skipped in backup. Also check restoration after backup to confirm there are no errors.

vsedmik:
Tested the scenarios above on Sat 6.7 snap 13 (+patch and pulled Amit's branch) and seem to work well (/var/lib/qpidd/* collected and restored in case 1 and skipped after three retries in case 2), health check, UI ant the content was ok after restore as well.
Tomorrow I'm going to test 'full stack' approach with some real load via hosts and hammer, stay tuned.

definitions/procedures/backup/config_files.rb
```diff
@@ -33,7 +36,12 @@ def run
 
         puts "WARNING: Attempt #{attempt_no}/#{MAX_RETRIES} to collect all config files failed!"
         puts 'Some files were modified during creation of the archive.'
-        if attempt_no == MAX_RETRIES
+        if attempt_no == MAX_RETRIES && @exclude_qpidd_data
+          tar_cmd = tar_command(configs, to_exclude.push('/var/lib/qpidd'))
+          with_spinner('Collecting config files to backup, skipping /var/lib/qpidd') do
```
kgaikwad: As offline discussion, it would be good to have warning instead of spinner showing this information.

definitions/procedures/backup/config_files.rb
```diff
@@ -17,12 +17,15 @@ class ConfigFiles < ForemanMaintain::Procedure
...
             :flag => true, :default => false
+      param :exclude_qpidd_data, 'Exclude qpidd /var/lib/qpidd/ Directory during backup',
+            :flag => true, :default => false
     end
 
     # rubocop:disable Metrics/MethodLength
     def run
       logger.debug("Invoking tar from #{FileUtils.pwd}")
-      tar_cmd = tar_command
+      configs, to_exclude = config_files
```
kgaikwad: Instead of passing configs and to_exclude to `tar_command` method, I think it would be good to have `options` to tar_command method which will accept extra directories to exclude and append to_exclude list inside tar_command?

upadhyeammit:
I am closing this PR, shall be adding new PR for remaining work.

PR #320, opened 2020-02-28
Fixes #29227 - Add content commands to the top level

johnpmitsch:
This adds content commands to the top level so they are easier to use and find. This will help during the pulp 2 to 3 migration

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,10 @@ Subcommands:
       stop                          Stop maintenance-mode
       status                        Get maintenance-mode status
       is-enabled                    Get maintenance-mode status code
+
+    content                       Content related commands
+      prepare                       Prepare content for Pulp 3 
+      switchover                    Switch support for certain content from Pulp 2 to Pulp 3
 ```
 
 ### Upgrades
--- a/definitions/procedures/content/prepare.rb
+++ b/definitions/procedures/content/prepare.rb
@@ -6,7 +6,7 @@ module Procedures::Content
     end
 
     def run
-      execute!('foreman-rake katello:pulp3_migration')
+      puts execute!('foreman-rake katello:pulp3_migration')
     end
   end
 end
--- a/definitions/procedures/content/switchover.rb
+++ b/definitions/procedures/content/switchover.rb
@@ -6,9 +6,13 @@ module Procedures::Content
     end
 
     def run
-      execute!('foreman-rake katello:pulp3_migration')
-      execute!('foreman-rake katello:pulp3_post_migration_check')
-      execute!('foreman-rake katello:pulp3_content_switchover')
+      puts 'Performing final content migration before switching content'
+      puts execute!('foreman-rake katello:pulp3_migration')
+      puts 'Performing a check to verify everything that is needed has been migrated'
+      puts execute!('foreman-rake katello:pulp3_post_migration_check')
+      puts 'Switching specified content over to pulp 3'
+      puts execute!('foreman-rake katello:pulp3_content_switchover')
+      puts 'Re-running the installer to switch specified content over to pulp3'
       args = ['--foreman-proxy-content-proxy-pulp-isos-to-pulpcore=true',
               '--katello-use-pulp-2-for-file=false',
               '--katello-use-pulp-2-for-docker=false']
--- a/definitions/scenarios/content.rb
+++ b/definitions/scenarios/content.rb
@@ -1,0 +1,27 @@
+module ForemanMaintain::Scenarios
+  module Content
+    class Prepare < ForemanMaintain::Scenario
+      metadata do
+        label :content_prepare
+        description 'Prepare content for Pulp 3'
+        manual_detection
+      end
+
+      def compose
+        add_step(Procedures::Content::Prepare)
+      end
+    end
+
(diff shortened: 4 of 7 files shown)


johnpmitsch:
I didn't gate the command in any way, I'm not sure if we should try to gate the whole `content` subcommand or even if it is possible so it only shows for Katello.

I can look into it more though if we think that this is worthwhile.

johnpmitsch:
> f I run the task directly with `foreman-rake` I see `Content Migration completed successfully`. The output of the `content prepare` is empty however.

@mbacovsky @kgaikwad @upadhyeammit Is there a way to execute a command and print it's output? The only way I seem to able to that currently is by doing:
```ruby
result = execute("uptime")
puts result
```
Is this the best way?

The context being printing the output of [these tasks here](https://github.com/theforeman/foreman_maintain/blob/master/definitions/procedures/content/switchover.rb#L9-L11)

upadhyeammit:
> > f I run the task directly with `foreman-rake` I see `Content Migration completed successfully`. The output of the `content prepare` is empty however.
> 
> @mbacovsky @kgaikwad @upadhyeammit Is there a way to execute a command and print it's output? The only way I seem to able to that currently is by doing:
> 
> ```ruby
> result = execute("uptime")
> puts result
> ```
> 
> Is this the best way?
> 
> The context being printing the output of [these tasks here](https://github.com/theforeman/foreman_maintain/blob/master/definitions/procedures/content/switchover.rb#L9-L11)

if you just want to print output then execute() returns the output of command in string. You dont need the puts here.
if you want to print output when its successful and raise error when its not then execute!() method you should use.
if I see existing few occurrences of running foreman-rake then execute!() has been used, considering we raise error if rake tasks fails somehow.

johnpmitsch:
@upadhyeammit thanks for the response

> if you want to print output when its successful and raise error when its not then execute!() method you should use.

Its funny there is no output currently then, the tasks are using `execute!`. I'll take a look further, double checking its actual STDOUT and trying with some other commands.

johnpmitsch:
@upadhyeammit 
> if you just want to print output then execute() returns the output of command in string. You dont need the puts here.

looking at `execute!` and `execute` methods, it looks like nothing is actually being printed, just the output is being returned, which makes sense that I would have to print out the result of `execute!` or `execute` to see the output when the task is run:

https://github.com/theforeman/foreman_maintain/blob/25850e90a2aa1c08f5eddbe3ccd25cd6643dea5e/lib/foreman_maintain/concerns/system_helpers.rb#L66

Here are the paths forward I came up wtih:
- Change the lines in the tasks I have to `puts execute!('some-command')`
- Modify execute and execute! commands so they always print
- Add option to `execute` and `execute!` to print output, defaulting to false so the behavior stays the same elsewhere
- Create a new method called something like `execute_with_output` that always prints the output

Any ideas or preference on which would be best?

johnpmitsch:
> @upadhyeammit
> 
> > if you just want to print output then execute() returns the output of command in string. You dont need the puts here.
> 
> looking at `execute!` and `execute` methods, it looks like nothing is actually being printed, just the output is being returned, which makes sense that I would have to print out the result of `execute!` or `execute` to see the output when the task is run:
> 
> https://github.com/theforeman/foreman_maintain/blob/25850e90a2aa1c08f5eddbe3ccd25cd6643dea5e/lib/foreman_maintain/concerns/system_helpers.rb#L66
> 
> Here are the potential paths forward I see:
> 
> * Change the lines in the tasks I have to `puts execute!('some-command')`
> * Modify execute and execute! commands so they always print
> * Add option to `execute` and `execute!` to print output, defaulting to false so the behavior stays the same elsewhere
> * Create a new method called something like `execute_with_output` that always prints the output
> 
> Any ideas or preference on which would be best?

@kgaikwad @upadhyeammit  any thoughts on this?

kgaikwad:
> > Here are the potential paths forward I see:
> > * Change the lines in the tasks I have to `puts execute!('some-command')`
> > * Modify execute and execute! commands so they always print
> > * Add option to `execute` and `execute!` to print output, defaulting to false so the behavior stays the same elsewhere
> > * Create a new method called something like `execute_with_output` that always prints the output
> > Any ideas or preference on which would be best?
> @kgaikwad @upadhyeammit any thoughts on this?

@johnpmitsch,
In case of any task failure, `execute!` method will raise an error and don't proceed further execution. If I understood correctly then it is an expected behavior so I would suggest to change the lines in the tasks with `puts execute!('some-command')`.
@upadhyeammit, what do you think?

johnpmitsch:
> @johnpmitsch,
> In case of any task failure, `execute!` method will raise an error and don't proceed further execution. If I understood correctly then it is an expected behavior so I would suggest to change the lines in the tasks with `puts execute!('some-command')`.
> @upadhyeammit, what do you think?

@kgaikwad I updated with `puts execute!('foo')`. I agree that is the best solution for now! Please take a look at the PR and let me know what you think

@ianballou  I also added some print statements so it's more obvious what steps are running and the command hasn't hung.

kgaikwad:
Thank you @johnpmitsch, @ianballou and @upadhyeammit!

PR #321, opened 2020-03-03
Fixes #29246 - introduce new feature to handle foreman service

kgaikwad:
+ adding @sthirugn into loop.
@sthirugn, I have mentioned service priority as 30. Please correct me if any modifications required.
It would be good to have your review on overall changes.

Reference links for these changes:
https://github.com/theforeman/puppet-foreman/pull/802
https://github.com/theforeman/foreman/blob/develop/extras/systemd/foreman.service


definitions/features/puma.rb
```diff
@@ -0,0 +1,18 @@
+module ForemanMaintain
+  module Features
+    class Puma < ForemanMaintain::Feature
```
ehelms: Is it possible to name this something related to foreman-service given thats the systemd service?

upadhyeammit:
Few thoughts on this one,
1. The maintain already has foreman_server feature which has httpd as service. Here we consider passenger needs the httpd service and we dont have separate service for passenger.
2. Can we keep the foreman_server as parent class and have two children, one for httpd and other for puma ? The common things of both features goes in foreman_server and we delegate the methods, variables etc to either puma, httpd(passenger) as per what is available ?

This way we still refer to foreman_server while doing or adding changes and at same time maintain both of these ?

definitions/features/puma.rb
sthirugn: I agree here too, we don't need to call out puma.

sthirugn:
> Few thoughts on this one,
> 
>     1. The maintain already has foreman_server feature which has httpd as service. Here we consider passenger needs the httpd service and we dont have separate service for passenger.

The passenger is not a separate service but it is embedded into httpd itself. But the new `foreman.service` is a separate service outside of httpd. Also note that we will drop passenger in the near future.

>     2. Can we keep the foreman_server as parent class and have two children, one for httpd and other for puma ? The common things of both features goes in foreman_server and we delegate the methods, variables etc to either puma, httpd(passenger) as per what is available ?
> 
> 
> This way we still refer to foreman_server while doing or adding changes and at same time maintain both of these ?

I recommend keeping the new foreman.service separate as it is intended to be separate from httpd.

ehelms:
Starting with Foreman 2.1 (Sat 6.8) the architecture will be a separate foreman service with Apache as a reverse proxy. So `httpd` will no longer need to be restarted to restart Foreman with the Foreman server.

kgaikwad:
> Starting with Foreman 2.1 (Sat 6.8) the architecture will be a separate foreman service with Apache as a reverse proxy. So `httpd` will no longer need to be restarted to restart Foreman with the Foreman server.

Currently `httpd` service added in three features i.e foreman, pulp2, pulpcore. Doesn't that mean it will no longer required in all or only applicable for Foreman?

definitions/features/puma.rb
kgaikwad: Renamed feature `puma` to `foreman_service`.

kgaikwad:
> > 2. Can we keep the foreman_server as parent class and have two children, one for httpd and other for puma ? The common things of both features goes in foreman_server and we delegate the methods, variables etc to either puma, httpd(passenger) as per what is available ?
> > ```
> > 
> > 
> > This way we still refer to foreman_server while doing or adding changes and at same time maintain both of these ?
> 
> I recommend keeping the new foreman.service separate as it is intended to be separate from httpd.

@upadhyeammit, @sthirugn,
I have kept `foreman_service` as separate feature but modified services list in `foreman_server` based on `foreman_service` feature presence. 
Will it be okay?

ehelms:
Apache remains the application server for Pulp 2. For Foreman and Pulpcore
Apache is a reverse proxy to their services.

definitions/features/foreman_server.rb
```diff
@@ -9,9 +9,11 @@ class ForemanServer < ForemanMaintain::Feature
       end
 
       def services
-        [
-          system_service('httpd', 30)
-        ]
+        if feature(:foreman_service)
+          [feature(:foreman_service).service]
+        else
+          [system_service('httpd', 30)]
```
kgaikwad: Similarly, do I need to handle [line number 30](https://github.com/theforeman/foreman_maintain/pull/321/files#diff-0e91ba9aec5b0d6281275bda1ece9fb5R30)?

definitions/features/foreman_service.rb
```diff
@@ -0,0 +1,16 @@
+module ForemanMaintain
+  module Features
+    class ForemanService < ForemanMaintain::Feature
+      metadata do
+        label :puma
```
sthirugn: I think you renamed everything but left this `puma`?

sthirugn:
> @upadhyeammit, @sthirugn,
> I have kept `foreman_service` as separate feature but modified services list in `foreman_server` based on `foreman_service` feature presence.
> Will it be okay?
I am not fully aware of foreman_maintain architecture, so the changes here look sane to me in general.  I will let @upadhyeammit weigh in with respect to architecture.

definitions/features/foreman_service.rb
kgaikwad: My bad. Fixed!

kgaikwad:
Tests are failing due to current changes. I will update PR.

definitions/features/foreman_service.rb
sthirugn: no worries, thank you for the quick fix

definitions/features/foreman_server.rb
upadhyeammit: As per the comments from Eric we are going to have separate foreman_service. In this case is it better to limit the foreman_server using confine block ? This way we wont need to add any conditions of feature(:foreman_service) in foreman_server ?

ekohl:
As part of https://github.com/theforeman/foreman/pull/7536 we'll also add `foreman.socket`

kgaikwad:
For Foreman and Pulpcore Apache is a reverse proxy to their services.
@ehelms, @sthirugn, 
Does that mean `httpd` service will be there?
Only difference is foreman/pulpcore will have different services to restart.
I am asking this thing repeatitively because want to understand that along with required changes foreman-maintain should/should not show `httpd` service in service list as output of command `foreman-maintain service list`.

ehelms:
Yes, httpd should still be shown.

kgaikwad:
@ehelms, @upadhyeammit,
I have removed feature `foreman_service` and added `apache` feature just to handle `httpd` service.

> As part of theforeman/foreman#7536 we'll also add foreman.socket

@ekohl, does it mean like services we need to provide way to apply actions (start/stop/status) for socket?

definitions/features/foreman_server.rb
```diff
@@ -9,9 +9,11 @@ class ForemanServer < ForemanMaintain::Feature
       end
 
       def services
-        [
-          system_service('httpd', 30)
-        ]
+        if foreman_service_installed?
+          [system_service('foreman', 30)]
```
ehelms, line 13: We have introduced a `foreman.socket` that works like a service but does not need to be restarted. However, I think it would be helpful to ensure that the socket is running whenever we perform service operations. @ekohl what do you think regarding handling the socket?

ekohl:
If the socket is started and Apache is as well, it'll automatically start foreman.service. Probably good to ensure it's stopped as well when stopping `foreman.service`. Note that `foreman.service` requires `foreman.socket` so there's no need to start both.

Short summary:

Full stop: `systemctl stop foreman.socket foreman.service`
Start lazy (when the first request comes in): `systemctl start foreman.socket`
Start eager: `systemctl start foreman.service` (`foreman.socket` can be added but is redundant)
Restart: `systemctl restart foreman.service`
Reload: `systemctl reload foreman.service`

kgaikwad:
Thank you @ehelms, @ekohl for providing more details on it.
I have removed WIP status as I think current changes will sufficient. It would be good to have your review on this.

kgaikwad:
@sthirugn,
To test these changes, what kind of setup I will require?

ekohl:
All changes for the service are in nightly.

sthirugn:
> To test these changes, what kind of setup I will require?

Yes as @ekohl mentioned, test this in nightly.  I will test this change in my test server too.

sthirugn:
In my nightly box:
1. service restart worked great:
```console
# foreman-maintain service restart
Running Restart Services
Restart applicable services: 

rh-mongodb34-mongod, rh-redis5-redis, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, qdrouterd, qpidd, rh-redis5-redis, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, pulpcore-worker@*, tomcat, dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, dynflow-sidekiq@worker, dynflow-sidekiq@worker-hosts-queue, foreman-proxy
/ stopping foreman                                                              
Warning: Stopping foreman.service, but it can still be activated by:
  foreman.socket
/ All services stopped                                                          

rh-mongodb34-mongod, rh-redis5-redis, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, qdrouterd, qpidd, rh-redis5-redis, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, pulpcore-worker@*, tomcat, dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, dynflow-sidekiq@worker, dynflow-sidekiq@worker-hosts-queue, foreman-proxy
| All services started                                                [OK]      
```

2. service stop - worked great
```console
# foreman-maintain service stop
Running Stop Services
Stop applicable services: 

rh-mongodb34-mongod, rh-redis5-redis, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, qdrouterd, qpidd, rh-redis5-redis, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, pulpcore-worker@*, tomcat, dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, dynflow-sidekiq@worker, dynflow-sidekiq@worker-hosts-queue, foreman-proxy
/ stopping foreman                                                              
Warning: Stopping foreman.service, but it can still be activated by:
  foreman.socket
\ All services stopped                                                [OK]      

# systemctl status foreman.socket
● foreman.socket - Foreman HTTP Server Accept Sockets
   Loaded: loaded (/usr/lib/systemd/system/foreman.socket; disabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/foreman.socket.d
           └─installer.conf
   Active: active (listening) since Mon 2020-04-13 20:05:21 UTC; 2min 55s ago
   Listen: 127.0.0.1:3000 (Stream)

Apr 13 20:05:21 centos7-katello-nightly.apex.example.com systemd[1]: Listening on Foreman HTTP Server Accept Sockets.
```

sthirugn:
> ```
> # foreman-maintain service restart
>    Running Restart Services
> ================================================================================
> Check if command is run as root user:                                 [OK]
> Restart applicable services: 
> 
> Stopping the following service(s):
> rh-mongodb34-mongod, rh-redis5-redis, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, qdrouterd, qpidd, rh-redis5-redis, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, pulpcore-worker@*, tomcat, dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, dynflow-sidekiq@worker, dynflow-sidekiq@worker-hosts-queue, foreman-proxy
> / stopping foreman                                                              
> Warning: Stopping foreman.service, but it can still be activated by:
>   foreman.socket

@ekohl do you know of a way to hide this ^ warning?

> ```
> # foreman-maintain service stop
> Running Stop Services
> ================================================================================
> Check if command is run as root user:                                 [OK]
> Stop applicable services: 
> 
> Stopping the following service(s):
> rh-mongodb34-mongod, rh-redis5-redis, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, qdrouterd, qpidd, rh-redis5-redis, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, pulpcore-worker@*, tomcat, dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, dynflow-sidekiq@worker, dynflow-sidekiq@worker-hosts-queue, foreman-proxy
> / stopping foreman                                                              
> Warning: Stopping foreman.service, but it can still be activated by:
>   foreman.socket
> \ All services stopped                                                [OK]      
> 
> # systemctl status foreman.socket
> ● foreman.socket - Foreman HTTP Server Accept Sockets
>    Loaded: loaded (/usr/lib/systemd/system/foreman.socket; disabled; vendor preset: disabled)
>   Drop-In: /etc/systemd/system/foreman.socket.d
>            └─installer.conf
>    Active: active (listening) since Mon 2020-04-13 20:05:21 UTC; 2min 55s ago

@ekohl foreman.socket is still active and listening after foreman-maintain stop ^.  May be foreman-maintain stop `foreman.socket` explicitly in a stop request?

ekohl:
> @ekohl do you know of a way to hide this ^ warning?

Ensure `foreman.socket` is stopped `foreman.service`. Stopping them in the same call (`systemctl stop foreman.service foreman.socket`) also works.

> @ekohl foreman.socket is still active and listening after foreman-maintain stop ^. May be foreman-maintain stop `foreman.socket` explicitly in a stop request?

Yes, this is crucial. Otherwise Apache will start the service again on the first request.

kgaikwad:
Updated the changes.
With new changes, it will add `foreman.socket` into list in case of services stop.

```
foreman_maintain]# ./bin/foreman-maintain service stop
Running Stop Services
Stop applicable services: 

rh-mongodb34-mongod, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, qdrouterd, qpidd, rh-redis5-redis, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, pulpcore-worker@*, tomcat, 
dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, rh-redis5-redis, dynflow-sidekiq@worker, dynflow-sidekiq@worker-hosts-queue, foreman-proxy, **foreman.socket**
| All services stopped                                                [OK]    
```
and won't consider `foreman.socket` in other service actions.

@sthirugn, could you verify updated changes? please let me know if any change required.

ekohl:
Unrelated note, but why `pulpcore-worker@*` but not `dynflow-sidekiq@*`?

definitions/features/foreman_server.rb
```diff
@@ -9,9 +9,15 @@ class ForemanServer < ForemanMaintain::Feature
...
+        else
+          [system_service('httpd', 30)]
+        end
+      end
+
+      def socket_name
+        foreman_service_installed? ? 'foreman.socket' : nil
```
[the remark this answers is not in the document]
kgaikwad: In `service` feature, it verifies that particular service exists. So yes, I will return `foreman.socket` value directly.

kgaikwad:
> Unrelated note, but why `pulpcore-worker@*` but not `dynflow-sidekiq@*` 

There is change in priorities in case of `dynflow-sidekiq` services.

sthirugn:
> Updated the changes.
> With new changes, it will add `foreman.socket` into list in case of services stop.
> 
> ```
> foreman_maintain]# ./bin/foreman-maintain service stop
> Running Stop Services
> ================================================================================
> Check if command is run as root user:                                 [OK]
> Stop applicable services: 
> 
> Stopping the following service(s):
> rh-mongodb34-mongod, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, qdrouterd, qpidd, rh-redis5-redis, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, pulpcore-worker@*, tomcat, 
> dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, rh-redis5-redis, dynflow-sidekiq@worker, dynflow-sidekiq@worker-hosts-queue, foreman-proxy, **foreman.socket**
> | All services stopped                                                [OK]    
> ```
> 
> and won't consider `foreman.socket` in other service actions.
> 
> @sthirugn, could you verify updated changes? please let me know if any change required.

My test worked fine, pending @ekohl's new comments.

ekohl:
> IIRC, there is change in priorities in case of `dynflow-sidekiq` services.

Ah, that makes sense.

definitions/features/foreman_server.rb
```diff
@@ -41,6 +47,16 @@ def config_files_to_exclude
...
       end
+
+      def services_running?
+        services.all?(&:running?)
+      end
+
+      private
+
+      def foreman_service_installed?
+        package_manager.installed?(['foreman-service']) && check_min_version('foreman', '2.1')
```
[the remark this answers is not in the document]
kgaikwad: I have kept code as it is because through foreman-maintain user can perform all service actions i.e disable, enable, start, stop, restart, list. So will the use of `systemctl is-enabled foreman` be helpful here?

definitions/procedures/service/stop.rb
```diff
@@ -8,7 +8,28 @@ class Stop < Base
...
 
     def run
-      run_service_action('stop', common_options)
+      param_options = override_options
+      run_service_action('stop', param_options)
+    end
+
+    def override_options
+      param_options = common_options
+      if foreman_socket_handling_required?
```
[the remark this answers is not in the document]
kgaikwad: @ekohl, updated PR with more generic changes.
Two things that I would like to confirm from you about socket handling.
- Applied the priority same as respective service & 
- Firstly, it tries to stop service and then socket

definitions/features/foreman_server.rb
kgaikwad: Removed `foreman_service` feature and handled condition in same `foreman_server` feature.

definitions/features/foreman_server.rb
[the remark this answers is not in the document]
ekohl: The installer has supported deploying with foreman-service since 1.23 using `--foreman-passenger false`. With 2.1 we're making it the default. 2.1 also introduces `foreman.socket`. Prior to that `foreman.service` (and earlier the sysvinit version) were shipped as part of Foreman itself, dating back to as far as I can remember. IMHO `systemctl is-enabled foreman` is a good check. The installer will ensure it's either enabled or disabled. IIRC `is-enabled` also returns the correct value if a service is missing.

definitions/features/foreman_server.rb
```diff
@@ -9,9 +9,11 @@ class ForemanServer < ForemanMaintain::Feature
       end
 
       def services
-        [
-          system_service('httpd', 30)
-        ]
+        if foreman_service_installed?
```
ekohl, line 12: If `foreman-service` is installed, Apache is usually still used as a reverse proxy. The installer can set it up without Apache, but I personally haven't used that and we don't recommend it unless you know what you're doing.

kgaikwad: > If `foreman-service` is installed, Apache is usually still used as a reverse proxy.

Yes, Apache will be there. To handle this, I have added Apache feature separately which having `httpd` in services list.

kgaikwad: here -> https://github.com/theforeman/foreman_maintain/pull/321/files#diff-9c2e92406f1636a749c776ff0cd1b1ceR12

kgaikwad:
Resolved Rubocop warnings.
Also replaced package checking with `systemctl is-enabled foreman` check in `foreman_server` feature.

definitions/features/service.rb
```diff
@@ -99,9 +99,19 @@ def allowed_action?(action)
     %w[start stop restart status enable disable].include?(action)
   end
 
-  def filter_services(service_list, options)
-    service_list = include_unregistered_services(service_list, options[:include])
+  def extend_service_list_with_sockets(service_list, options)
+    return service_list unless options[:include_sockets]
 
+    if (options[:only].empty? && options[:exclude].empty?) ||
```
upadhyeammit, line 105: This if statement runs in all conditions of service command ?
~~~
service [all]
service --only [some-service]
service --exclude [some-service]
~~~
In this case is it good to remove the if statement ?

kgaikwad: Good Catch! fixed

definitions/features/service.rb
```diff
@@ -99,9 +99,15 @@ def allowed_action?(action)
...
   end
 
-  def filter_services(service_list, options)
-    service_list = include_unregistered_services(service_list, options[:include])
+  def extend_service_list_with_sockets(service_list, options)
+    return service_list unless options[:include_sockets]
+
+    socket_list = service_list.map(&:socket).compact.select(&:exist?)
+    service_list.concat(socket_list)
+    service_list
```
ekohl, line 107: Would it be cleaner not to modify `service_list` in place?
```suggestion
    service_list + socket_list
```

kgaikwad:
@upadhyeammit, @ekohl,
Updated PR with suggested changes.

sthirugn:
looks great

PR #328, opened 2020-03-31
Fixes #29443 - Add Satellite and Capsule 6.8 6.8.z scenarios

upadhyeammit:
Satellite and Capsule 6.8 6.8.z scenarios

### DIFF
--- a/definitions/features/installer.rb
+++ b/definitions/features/installer.rb
@@ -92,9 +92,7 @@ class Features::Installer < ForemanMaintain::Feature
   end
 
   def upgrade(exec_options = {})
-    arguments = '--disable-system-checks'
-    arguments += ' --upgrade' if can_upgrade?
-    run(arguments, exec_options)
+    run(installer_args, exec_options)
   end
 
   def initial_admin_username
--- a/definitions/features/installer.rb
+++ b/definitions/features/installer.rb
@@ -92,7 +92,15 @@ class Features::Installer < ForemanMaintain::Feature
   end
 
   def upgrade(exec_options = {})
-    run(installer_args, exec_options)
+    run(installer_arguments, exec_options)
+  end
+
+  def installer_arguments
+    installer_args = ' --disable-system-checks'
+    unless check_min_version('foreman', '2.1') || check_min_version('foreman-proxy', '2.1')
+      installer_args += ' --upgrade'
+    end
+    installer_args
   end
 
   def initial_admin_username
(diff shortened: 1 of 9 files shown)


kgaikwad:
> From sat 6.8+ (2.1+) onwards, it will raise out an error.

Correcting myself on this.It will not raise an error but will result in a deprecation warning message.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -192,6 +192,14 @@ def check_version(name)
           yield current_version
         end
       end
+
+      def installer_args
+        installer_args = ' --disable-system-checks'
```
kgaikwad: I don't think this is right place to add change.

definitions/features/installer.rb
```diff
@@ -92,9 +92,7 @@ def run(arguments = '', exec_options = {})
   end
 
   def upgrade(exec_options = {})
-    arguments = '--disable-system-checks'
-    arguments += ' --upgrade' if can_upgrade?
-    run(arguments, exec_options)
+    run(installer_args, exec_options)
```
kgaikwad: Two places where we have used `--upgrade` flag. One inside upgrade method of installer feature and second passed as argument to `Procedures::Installer::Run` in package scenario.  

Adding a condition for `--upgrade` flag inside `upgrade` method and changing procedure from `Procedures::Installer::Run` to `Procedures::Installer::Upgrade` inside package scenario won't be sufficient here?

upadhyeammit: I moved this as per the suggestion, changed tests accordingly. I am not sure the way I have handled the test change is correct or not, request to review. Thank you !

test/definitions/features/installer_test.rb
```diff
@@ -44,7 +45,7 @@
       it '#upgrade runs the installer with correct params' do
         assume_feature_absent(:satellite)
         installer_inst.expects(:'execute!').
-          with('LANG=en_US.utf-8 foreman-installer --disable-system-checks --upgrade',
+          with("LANG=en_US.utf-8 foreman-installer #{subject.installer_arguments}",
```
kgaikwad, line 48: Do you really need to call `installer_arguments` method here? 
In before block you are already stubbing this method to return `--disable-system-checks` & `--upgrade` flag.

test/definitions/features/installer_test.rb
```diff
@@ -11,6 +11,7 @@
     before do
       installer_config_dir(["#{data_dir}/installer/simple_config"])
       mock_installer_package('foreman-installer')
+      installer_args
```
kgaikwad, line 14: You can directly place line here than method.

```suggestion
     Features::Installer.any_instance.stubs(:installer_arguments).returns(' --disable-system-checks \
                                                                         --upgrade')
```

kgaikwad:
Thank you @upadhyeammit!

PR #332, opened 2020-04-15
Fixes #29533 - add qpidd activemq queue generation procedure

upadhyeammit:
This adds the qpidd and activemq queue generation if restore is from online backup. To know if backup is online or offline new flag 'online' has introduced in backup metadata. This flag helps to decide if restore scenario should run procedure to create queues and bindings.

### DIFF
--- a/definitions/procedures/backup/metadata.rb
+++ b/definitions/procedures/backup/metadata.rb
@@ -6,7 +6,7 @@ module Procedures::Backup
       preparation_steps { Checks::Foreman::DBUp.new if feature(:foreman_server) }
       param :backup_dir, 'Directory where to backup to', :required => true
       param :incremental_dir, 'Changes since specified backup only'
-      param :online_backup, 'Selected online backup ?'
+      param :online_backup, 'Selected online backup ?', :flag => true, :default => false
     end
 
     def run
@@ -17,7 +17,7 @@ module Procedures::Backup
         metadata['proxy_features'] = proxy_feature_list(spinner) || []
         metadata['rpms'] = rpms(spinner)
         metadata['incremental'] = @incremental_dir || false
-        metadata['online'] = @online_backup || false
+        metadata['online'] = @online_backup
         save_metadata(metadata, spinner)
       end
     end
--- a/definitions/procedures/backup/metadata.rb
+++ b/definitions/procedures/backup/metadata.rb
@@ -6,7 +6,7 @@ module Procedures::Backup
       preparation_steps { Checks::Foreman::DBUp.new if feature(:foreman_server) }
       param :backup_dir, 'Directory where to backup to', :required => true
       param :incremental_dir, 'Changes since specified backup only'
-      param :online_backup, 'Selected online backup ?', :flag => true, :default => false
+      param :online_backup, 'Select for online backup', :flag => true, :default => false
     end
 
     def run
--- a/definitions/procedures/backup/metadata.rb
+++ b/definitions/procedures/backup/metadata.rb
@@ -6,6 +6,7 @@ module Procedures::Backup
       preparation_steps { Checks::Foreman::DBUp.new if feature(:foreman_server) }
       param :backup_dir, 'Directory where to backup to', :required => true
       param :incremental_dir, 'Changes since specified backup only'
+      param :online_backup, 'Selected online backup ?'
     end
 
     def run
@@ -16,6 +17,7 @@ module Procedures::Backup
         metadata['proxy_features'] = proxy_feature_list(spinner) || []
         metadata['rpms'] = rpms(spinner)
         metadata['incremental'] = @incremental_dir || false
+        metadata['online'] = @online_backup || false
         save_metadata(metadata, spinner)
       end
     end
--- a/definitions/procedures/restore/configs.rb
+++ b/definitions/procedures/restore/configs.rb
@@ -19,10 +19,15 @@ module Procedures::Restore
       end
     end
 
+    # rubocop:disable  Metrics/MethodLength
     def restore_configs(backup)
       exclude = ForemanMaintain.available_features.each_with_object([]) do |feat, cfgs|
+        if online_backup?
+          feat.config_files_exclude_for_online.each { |f| cfgs << f.gsub(%r{^/}, '') }
+        end
         feat.config_files_to_exclude.each { |f| cfgs << f.gsub(%r{^/}, '') }
       end
+
       tar_options = {
         :overwrite => true,
         :listed_incremental => '/dev/null',
@@ -35,9 +40,15 @@ module Procedures::Restore
 
       feature(:tar).run(tar_options)
     end
+    # rubocop:enable  Metrics/MethodLength
 
     def reload_configs
       feature(:mongo).reload_db_config if feature(:mongo)
+    end
+
+    def online_backup?
+      metadata_file = YAML.load(File.read(options['backup_dir'] + 'metadata.yml'))
+      @online_flag = metadata_file['online']
     end
 
     private
--- a/definitions/procedures/restore/configs.rb
+++ b/definitions/procedures/restore/configs.rb
@@ -22,7 +22,7 @@ module Procedures::Restore
     # rubocop:disable  Metrics/MethodLength
     def restore_configs(backup)
       exclude = ForemanMaintain.available_features.each_with_object([]) do |feat, cfgs|
-        if online_backup?
+        if backup.online_backup?
           feat.config_files_exclude_for_online.each { |f| cfgs << f.gsub(%r{^/}, '') }
         end
         feat.config_files_to_exclude.each { |f| cfgs << f.gsub(%r{^/}, '') }
@@ -44,11 +44,6 @@ module Procedures::Restore
 
     def reload_configs
       feature(:mongo).reload_db_config if feature(:mongo)
-    end
-
-    def online_backup?
-      metadata_file = YAML.load(File.read(options['backup_dir'] + 'metadata.yml'))
-      @online_flag = metadata_file['online']
     end
 
     private
--- a/definitions/procedures/restore/regenerate_queues.rb
+++ b/definitions/procedures/restore/regenerate_queues.rb
@@ -1,7 +1,10 @@
 module Procedures::Restore
   class RegenerateQueues < ForemanMaintain::Procedure
     metadata do
-      description 'Regenerate required qpidd and activemq queues while restoring online backup'
+      description 'Regenerate required activemq and qpidd queues while restoring online backup'
+      confine do
+        feature(:pulp2)
+      end
     end
 
     def ssl_cert
@@ -28,6 +31,7 @@ module Procedures::Restore
       with_spinner('Resetting the queues') do |spinner|
         regenerate_activemq_queues(spinner)
         regenerate_qpidd_queues(spinner)
+        spinner.update('Queues created successfully')
       end
     end
 
@@ -50,7 +54,7 @@ module Procedures::Restore
       execute!('rm -rf /var/lib/qpidd/.qpidd/qls')
       spinner.update('Starting qpidd service')
       feature(:service).handle_services(spinner, 'start', :only => ['qpidd'])
-      spinner.update('Service qpidd started, waiting 60 seconds to start it completely.')
+      spinner.update('Service qpidd started, waiting 60 seconds to start it completely')
       sleep 60
       spinner.update('Recreating qpidd queues')
       run_qpid_command('add exchange topic event --durable')
--- a/definitions/procedures/restore/regenerate_queues.rb
+++ b/definitions/procedures/restore/regenerate_queues.rb
@@ -1,6 +1,7 @@
 module Procedures::Restore
   class RegenerateQueues < ForemanMaintain::Procedure
     metadata do
+      advanced_run false
       description 'Regenerate required activemq and qpidd queues while restoring online backup'
       confine do
         feature(:pulp2)
--- a/definitions/procedures/restore/regenerate_queues.rb
+++ b/definitions/procedures/restore/regenerate_queues.rb
@@ -1,0 +1,63 @@
+module Procedures::Restore
+  class RegenerateQueues < ForemanMaintain::Procedure
+    metadata do
+      description 'Regenerate required qpidd and activemq queues while restoring online backup'
+    end
+
+    def ssl_cert
+      "/etc/pki/katello/certs/#{hostname}-qpid-broker.crt"
+    end
+
+    def ssl_key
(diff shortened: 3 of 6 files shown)


definitions/procedures/backup/metadata.rb
```diff
@@ -6,6 +6,7 @@ class Metadata < ForemanMaintain::Procedure
       preparation_steps { Checks::Foreman::DBUp.new if feature(:foreman_server) }
       param :backup_dir, 'Directory where to backup to', :required => true
       param :incremental_dir, 'Changes since specified backup only'
+      param :online_backup, 'Selected online backup ?'
```
kgaikwad, line 9: As this param seems to be boolean type then would it be possible to declare `:online_backup` as a flag?

```suggestion
     param :online_backup, 'Select for an online backup',
            :flag => true, :default => false
```

definitions/procedures/restore/configs.rb
```diff
@@ -35,11 +40,17 @@ def restore_configs(backup)
 
       feature(:tar).run(tar_options)
     end
+    # rubocop:enable  Metrics/MethodLength
 
     def reload_configs
       feature(:mongo).reload_db_config if feature(:mongo)
     end
 
+    def online_backup?
```
kgaikwad, line 49: I think, it would be good to handle situation if file itself or `online` key doesn't exist with few logger statements.

definitions/procedures/restore/regenerate_queues.rb
```diff
@@ -0,0 +1,63 @@
+module Procedures::Restore
+  class RegenerateQueues < ForemanMaintain::Procedure
+    metadata do
+      description 'Regenerate required qpidd and activemq queues while restoring online backup'
+    end
+
+    def ssl_cert
+      "/etc/pki/katello/certs/#{hostname}-qpid-broker.crt"
```
kgaikwad, line 8: Instead of hard-coded strings for `ssl_cert` & `ssl_key`, can't we get these values from any configuration file?

definitions/procedures/restore/regenerate_queues.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def ssl_key
+      "/etc/pki/katello/private/#{hostname}-qpid-broker.key"
+    end
+
+    def amqps_url
+      'amqps://localhost:5671'
```
kgaikwad, line 16: ditto

definitions/procedures/restore/regenerate_queues.rb
```diff
@@ -0,0 +1,63 @@
+module Procedures::Restore
+  class RegenerateQueues < ForemanMaintain::Procedure
```
kgaikwad, line 2: Is this procedure valid only when `pulp2` feature enabled?

mccun934:
Are we including the hornetq directory in our backup sets during online backup? If not, we may need to include this reset routine as well (resethornetq):

https://github.com/RedHatSatellite/satellite-support/blob/master/satellite-reset#L87

if we aren't and Candlepin properly re-creates these, then we likely don't need this added to foreman-maintain

upadhyeammit:
> Are we including the hornetq directory in our backup sets during online backup? If not, we may need to include this reset routine as well (resethornetq):
> 
> https://github.com/RedHatSatellite/satellite-support/blob/master/satellite-reset#L87
> 
> if we aren't and Candlepin properly re-creates these, then we likely don't need this added to foreman-maintain

We dont backup hornetq in online/offline backup, and we dont have anyone reported hornetq problems, maybe because we dont backup and restore that ?
Is it recommended to have backup of hornetq ? if yes in which scenario ?

definitions/procedures/restore/regenerate_queues.rb
upadhyeammit: I think this should be for pulp2 only as I can see we dont have qpidd service included in pulpcore. @johnpmitsch am I correct ?

johnpmitsch: yes that is fine to gate this functionality on pulp2. Pulp 2 will still be present for a while, but it will be helpful when we do drop pulp2.

vsedmik:
Tested with several backups under load and always was able to restore them successfully. Observed the queues recreation and health check after restore.

```
Regenerate required qpidd and activemq queues while restoring online backup: 
\ Recreating qpidd queues                                             [OK]      
Start applicable services: 

rh-mongodb34-mongod, postgresql, qdrouterd, qpidd, squid, pulp_celerybeat, pulp_resource_manager, pulp_streamer, pulp_workers, smart_proxy_dynflow_core, tomcat, dynflowd, httpd, puppetserver, foreman-proxy
\ All services started                                                [OK]      
```

upadhyeammit:
@vsedmik Thank you for testing it. Request you to do one more round of testing considering below scenario,
1. We take online backup from old f-m, a version installed on system itself using rpm should be good. This ensures we dont have online flag in backup metadata.
2. Restore the online backup, we expect restore should work without any surprising errors. There could be qpidd or tomcat service failures, because in this case we wont re-generate the queues.

Thank You

upadhyeammit:
> > Are we including the hornetq directory in our backup sets during online backup? If not, we may need to include this reset routine as well (resethornetq):
> > https://github.com/RedHatSatellite/satellite-support/blob/master/satellite-reset#L87
> > if we aren't and Candlepin properly re-creates these, then we likely don't need this added to foreman-maintain
> 
> We dont backup hornetq in online/offline backup, and we dont have anyone reported hornetq problems, maybe because we dont backup and restore that ?
> Is it recommended to have backup of hornetq ? if yes in which scenario ?

@mccun934 any thoughts on above queries ?

definitions/procedures/restore/configs.rb
```diff
@@ -19,10 +19,15 @@ def run
       end
     end
 
+    # rubocop:disable  Metrics/MethodLength
     def restore_configs(backup)
       exclude = ForemanMaintain.available_features.each_with_object([]) do |feat, cfgs|
+        if online_backup?
+          feat.config_files_exclude_for_online.each { |f| cfgs << f.gsub(%r{^/}, '') }
```
kgaikwad, line 26: I think, it would be good to have a single method instead of two i.e config_files_to_exclude & config_files_exclude_for_online. 
Refactoring will be required but it won't block this PR.

definitions/procedures/backup/metadata.rb
```diff
@@ -6,6 +6,7 @@ class Metadata < ForemanMaintain::Procedure
       preparation_steps { Checks::Foreman::DBUp.new if feature(:foreman_server) }
       param :backup_dir, 'Directory where to backup to', :required => true
       param :incremental_dir, 'Changes since specified backup only'
+      param :online_backup, 'Selected online backup ?', :flag => true, :default => false
```
kgaikwad, line 9:
```suggestion
      param :online_backup, 'Select for online backup', :flag => true, :default => false
```

definitions/procedures/restore/configs.rb
upadhyeammit: Agree, we shall refactor this.

kgaikwad: Created re-factoring issue - https://projects.theforeman.org/issues/29880 just for future tracking

kgaikwad:
Code wise looks good to me.
@johnpmitsch, @mccun934, do you have any additional comments/concerns?

definitions/procedures/restore/regenerate_queues.rb
johnpmitsch: +1

definitions/procedures/restore/regenerate_queues.rb
johnpmitsch: +1

definitions/procedures/restore/configs.rb
```diff
@@ -35,11 +40,17 @@ def restore_configs(backup)
...
 
     def reload_configs
       feature(:mongo).reload_db_config if feature(:mongo)
     end
 
+    def online_backup?
+      metadata_file = YAML.load(File.read(options['backup_dir'] + 'metadata.yml'))
+      @online_flag = metadata_file['online']
+    end
+
```
johnpmitsch, line 53: This method should be in https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/utils/backup.rb - the backup instance is already created with `backup = ForemanMaintain::Utils::Backup.new(@backup_dir)`. You could either add a new method there or check the existing ones and see if the could help you.

definitions/scenarios/restore.rb
```diff
@@ -95,5 +97,10 @@ def set_context_mapping
       context.map(:incremental_backup,
                   Procedures::Selinux::SetFileSecurity => :incremental_backup)
     end
+
+    def online_backup?
+      metadata_file = YAML.load(File.read(context.get(:backup_dir) + 'metadata.yml'))
+      metadata_file['online'] || false
+    end
```
johnpmitsch, line 104: Same regarding using utils/backup class.

definitions/procedures/restore/regenerate_queues.rb
upadhyeammit: @johnpmitsch I tried to check with my limited knowledge of qpid and ssl certs, but I cant find exactly which files are keeping all of these certs path which are applicable for qpidd only, if you have any insight on same ?

definitions/procedures/restore/regenerate_queues.rb
johnpmitsch: I see the port is stored in the answers file, this is probably the thing I would worry most about changing in the future:

```
/etc/foreman-installer/scenarios.d/satellite-answers.yaml:  qpid_router_broker_port: 5671
```

We do access the answers file for other purposes https://github.com/theforeman/foreman_maintain/blob/7e1ac6d6edc0e5ad7107d7a626cefafda7b2faba/definitions/features/katello.rb#L64-L66

Perhaps you could do the same here for the port?

definitions/procedures/restore/regenerate_queues.rb
johnpmitsch: I don't see it specified anywhere with a quick search, others would know this area better than me

@ekohl does this seem ok hardcoded like this or is there a config file that these cert paths are located?

lib/foreman_maintain/utils/backup.rb
```diff
@@ -204,6 +204,10 @@ def sql_dump_files_exist?
       def incremental?
         !!metadata.fetch('incremental', false)
       end
+
+      def online_backup?
+        !!metadata.fetch('online')
+      end
```
johnpmitsch, line 210: :+1:

definitions/procedures/restore/regenerate_queues.rb
ekohl: I tried to look, but this file is not used anywhere directly. qpid itself stores it in a NSS keystore and that file is more of an intermediate file. It's unlikely to be changed since it'll be dropped in nightly soon when Pulp 2 is going away.

kgaikwad: Then I think, it would be fine to have hard-coded strings for ssl_cert & ssl_key.

Instead of having three different methods, will it be ok to have one method (example - `qpid_connection_configs`)? and only in case of `amqps_url`, as @johnpmitsch mentioned you can fetch port_number from answers file. 

@upadhyeammit, what do you say?

definitions/procedures/restore/regenerate_queues.rb
upadhyeammit: Yes, this is good suggestion. However I have never seen anyone who changes the default port of qpidd on Satellite. Or maybe I can even say I have not seen someone chaning the Satellite services ports.

johnpmitsch: > However I have never seen anyone who changes the default port of qpidd on Satellite. Or maybe I can even say I have not seen someone chaning the Satellite services ports

It keeps things flexible so that we don't have to update a ton of repos if we ever want to change that port or for some reason a customer needs to use a different port. We tend to try to use the answers file as a single source of truth

upadhyeammit:
@kgaikwad @johnpmitsch request to review. Thank You !

definitions/procedures/restore/regenerate_queues.rb
```diff
@@ -0,0 +1,69 @@
...
+        feature(:pulp2)
+      end
+    end
+
+    def qpid_router_broker_port
+      @qpid_router_broker_port ||= feature(:installer).
+                                   answers['foreman_proxy_content']['qpid_router_broker_port']
+    end
+
+    def qpid_configs
```
kgaikwad, line 16: Thank you @upadhyeammit for adding this. 
Would it be possible to declare variable like you did it for `qpid_router_broker_port`?

upadhyeammit:
> I haven't tested functionality, but the code looks good! My only suggestion as a follow-up change or something to keep in mind as things grow is to move the qpid logic to it's own feature and not in the procedure itself.
> 
> Nice work! +1

yes, I was thinking to add that qpid port method in qpid feature itself

> I haven't tested functionality, but the code looks good! My only suggestion as a follow-up change or something to keep in mind as things grow is to move the qpid logic to it's own feature and not in the procedure itself.
> 
> Nice work! +1

Thank you for all your suggestions and review !

vsedmik:
@kgaikwad yes, sure, I will prepare something for weekend run. :)

vsedmik:
So, the weekend results: 20 different online and 20 offline backups were restored. All of the online restores succeded.
However, surprisingly some of the offline backups did not succeed for the first time (but succeeded consequently in rerun). This is probably unrelated to this PR, but we will investigate it a bit more with @upadhyeammit

kgaikwad:
> So, the weekend results: 20 different online and 20 offline backups were restored. All of the online restores succeded.
> However, surprisingly some of the offline backups did not succeed for the first time (but succeeded consequently in rerun). This is probably unrelated to this PR, but we will investigate it a bit more with @upadhyeammit

Thanks @vsedmik for updates. Are you facing a issue with this PR or with master branch as well?

vsedmik:
@kgaikwad I'm facing it with default installed rubygem-foreman_maintain-0.5.3 as well, rather seems to be foreman-installer related, not f-m.

kgaikwad:
> @kgaikwad I'm facing it with default installed rubygem-foreman_maintain-0.5.3 as well, rather seems to be foreman-installer related, not f-m.

Thanks for confirmation, @vsedmik. 
This issue occurs with previous releases as well then it would be good to file a new issue against respective component with some steps or logs for the reference.
As this issue is not relevant to pull-request changes, without any delay merging this PR.

kgaikwad:
Thank you @upadhyeammit for this change!

PR #333, opened 2020-04-20
Fixes #29577 - fix undefined method server_version in downstream

kgaikwad:


jameerpathan111:
@kgaikwad tried this PR and it seem to fix the `repositories-validate` check.
Also tried Satellite 6.3 to 6.4 upgrade and it completed successfully.

PR #335, opened 2020-04-28
Fixes #29579 - added check for capsule package

patilsuraj767:


patilsuraj767:
Hello @upadhyeammit, Thank you for reviewing it. 

>     1. The check should be for both satellite and capsule, as satellite can have capsule package and vice versa. So we should not just check for the `satellite-capsule` package.
> 
>     2. Actually when you have confine block with `feature(:satellite) && feature(:capsule)` that is enough condition for this one. Because if you see both of these features we are already checking if packages are there on system ? and `feature(:satellite) && feature(:capsule)` means we have both of these.
Yes, that is correct. Confine block will take care of checking the condition if both the packages are present. 
Do you still want me to add the seperate condition in run method? 

>     3. I feel this should be in default check list instead of pre_upgrade.
Agree, will change it. 

>     4. I dont think its good idea to remove either satellite or satellite-capsule package; because we dont know exact state of system. I thought to check what scenario is there in installer answer file but in next versions of foreman-installer wont have any scenarios(capsule, satellite etc) so its again not good idea. Accordingly I feel we should just error out and let user fix this situation.

It is less likely to have satellite package on capsule server, even satellite package is not present in rhel-7-server-satellite-capsule repository. But still, I understand removing the package may create complication. I will remove the procedure and will change the assert by fail!.

>     5. The check should not be in foreman directory as its not specific to foreman feature ?

Agreed, let's move it to the parent directory "check" or you want me to add it in some specific directory?

upadhyeammit:
> Hello @upadhyeammit, Thank you for reviewing it.
> 
> > ```
> > 1. The check should be for both satellite and capsule, as satellite can have capsule package and vice versa. So we should not just check for the `satellite-capsule` package.
> > 
> > 2. Actually when you have confine block with `feature(:satellite) && feature(:capsule)` that is enough condition for this one. Because if you see both of these features we are already checking if packages are there on system ? and `feature(:satellite) && feature(:capsule)` means we have both of these.
> > ```
> 
> Yes, that is correct. Confine block will take care of checking the condition if both the packages are present.
> Do you still want me to add the seperate condition in run method?
> 
We should only keep the condition in confine block as its enough for this check.
> > ```
> > 3. I feel this should be in default check list instead of pre_upgrade.
> > ```
> 
> Agree, will change it.
> 
> > ```
> > 4. I dont think its good idea to remove either satellite or satellite-capsule package; because we dont know exact state of system. I thought to check what scenario is there in installer answer file but in next versions of foreman-installer wont have any scenarios(capsule, satellite etc) so its again not good idea. Accordingly I feel we should just error out and let user fix this situation.
> > ```
> 
> It is less likely to have satellite package on capsule server, even satellite package is not present in rhel-7-server-satellite-capsule repository. But still, I understand removing the package may create complication. I will remove the procedure and will change the assert by fail!.
> 
I still dont understand why we have satellite-capsule package in the satellite repository, we should check this internally.
> > ```
> > 5. The check should not be in foreman directory as its not specific to foreman feature ?
> > ```
> 
> Agreed, let's move it to the parent directory "check" or you want me to add it in some specific directory?
>
Maybe just in the checks directory ?

upadhyeammit:
The current code looks good, I am thinking if we can have this check run as 1st everytime, so we dont end up running all the other things and then fail ? because if this fails then others have less importance ? Last time I tried to change the check priority run, using before and after in confine but it did not work well for me, maybe @kgaikwad have better suggestion for this one ?

vsedmik:
I have tested it on a Satellite with satellite-capsule package installed and got this output:
```
[root@dhcp-2-29 foreman_maintain]# /root/foreman_maintain/bin/foreman-maintain upgrade check --target-version=6.7.z --whitelist="repositories-validate,repositories-setup,disk-performance"
Too many scenarios match {:tags=>[:upgrade_scenario, :pre_upgrade_checks]}
```

patilsuraj767:
> I have tested it on a Satellite with satellite-capsule package installed and got this output:
> 
> ```
> [root@dhcp-2-29 foreman_maintain]# /root/foreman_maintain/bin/foreman-maintain upgrade check --target-version=6.7.z --whitelist="repositories-validate,repositories-setup,disk-performance"
> Checking for new version of satellite-maintain...
> Nothing to update, can't find new version of satellite-maintain.
> Too many scenarios match {:tags=>[:upgrade_scenario, :pre_upgrade_checks]}
> ```
@vsedmik The issue does not seem caused due to this PR.  
On which version of the satellite you ran the command? Can you check `definitions/scenarios/upgrade_to_satellite_6_8.rb` maybe by mistake you might have changed `target_version` to return '6.7.z'. 

@upadhyeammit  @kgaikwad  any views on this?

patilsuraj767:
> The current code looks good, I am thinking if we can have this check run as 1st everytime, so we dont end up running all the other things and then fail ? because if this fails then others have less importance ? Last time I tried to change the check priority run, using before and after in confine but it did not work well for me, maybe @kgaikwad have better suggestion for this one ?

@upadhyeammit looking at this [check](https://github.com/theforeman/foreman_maintain/blob/1cf0796cccb67ca18900260f7b2d14caa7ae44f0/lib/foreman_maintain/concerns/metadata.rb#L36) I don't think `before` will help to run this check 1st everytime.  

Maybe @kgaikwad can suggest if we can do that.

vsedmik:
> On which version of the satellite you ran the command? Can you check `definitions/scenarios/upgrade_to_satellite_6_8.rb` maybe by mistake you might have changed `target_version` to return '6.7.z'.
@patilsuraj767 I tested it on Satellite 6.7. I get that `Too many scenarios match` only when the satellite-capsule package is installed (manually). When I remove the satellite-capsule pkg, the check succeeds.
The `upgrade_to_satellite_6_8.rb` should be in the original shape (I haven't touched it):
```
[root@dhcp-2-29 foreman_maintain]# head -n 18 definitions/scenarios/upgrade_to_satellite_6_8.rb
module Scenarios::Satellite_6_8
    def self.upgrade_metadata(&block)
      metadata do
        tags :upgrade_scenario
        confine do
          feature(:satellite) &&
            (feature(:satellite).current_minor_version == '6.7' || \
            ForemanMaintain.upgrade_in_progress == '6.8')
        end
        instance_eval(&block)
      end
    end

    def target_version
      '6.8'
    end
  end
```

patilsuraj767:
@vsedmik 
Yes, you are right. I am also facing the same issue when the capsule package is installed. 
But I think we can track this as a separate issue/bug because I am getting the same on the **master** branch. 

The **root cause** is that when the capsule package is installed upgrade-runner gets two scenarios for upgrading to 6.7.z. I think we should fix this in different PR.  


```
root@satellite ~/foreman_maintain (master) $ ./bin/foreman-maintain upgrade check --target-version=6.7.z --whitelist="repositories-validate,repositories-setup,disk-performance"

From: /root/foreman_maintain/lib/foreman_maintain/upgrade_runner.rb:52 ForemanMaintain::UpgradeRunner#scenario:

    48: def scenario(phase)
    49:   return @scenario_cache[phase] if @scenario_cache.key?(phase)
    50: 
    51:   condition = { :tags => [:upgrade_scenario, phase] }
 => 52:   binding.pry
    53:   matching_scenarios = find_scenarios(condition).select(&matching_version_test)
    54:   raise "Too many scenarios match #{condition.inspect}" if matching_scenarios.size > 1
    55: 
    56:   @scenario_cache[phase] = matching_scenarios.first
    57: end

[1] pry(#<ForemanMaintain::UpgradeRunner>)> find_scenarios(condition)
=> [Checks before upgrading to Capsule 6.7.z<Scenarios::Capsule_6_7_z::PreUpgradeCheck>,
 Checks before upgrading to Capsule 6.8<Scenarios::Capsule_6_8::PreUpgradeCheck>,
 Checks before upgrading to Satellite 6.7.z<Scenarios::Satellite_6_7_z::PreUpgradeCheck>,
 Checks before upgrading to Satellite 6.8<Scenarios::Satellite_6_8::PreUpgradeCheck>]
[2] pry(#<ForemanMaintain::UpgradeRunner>)> find_scenarios(condition).select(&matching_version_test)
=> [Checks before upgrading to Capsule 6.7.z<Scenarios::Capsule_6_7_z::PreUpgradeCheck>,
 Checks before upgrading to Satellite 6.7.z<Scenarios::Satellite_6_7_z::PreUpgradeCheck>]
[3] pry(#<ForemanMaintain::UpgradeRunner>)> exit
Too many scenarios match {:tags=>[:upgrade_scenario, :pre_upgrade_checks]}


```

patilsuraj767:
> @patilsuraj767, @upadhyeammit,
> Definitely, we can decide order on checks using before & after.
> As availability of definitions is mostly based on features, my concern is what about other scenarios/procedures where we don't run this check explicitly?
> 
> Does it make sense to load foreman-maintain itself when both packages are available?

@kgaikwad correct. Check the above issue which @vsedmik has mentioned, caused when both the packages are installed. 
I guess instead of a separate check, we should run this every time when the foreman-maintain command is executed.

upadhyeammit:
> @patilsuraj767, @upadhyeammit,
> Definitely, we can decide order on checks using before & after.
> As availability of definitions is mostly based on features, my concern is what about other scenarios/procedures where we don't run this check explicitly?
> 
> Does it make sense to load foreman-maintain itself when both packages are available?

Do you mean to not load foreman-maintain if these packages are available ?

upadhyeammit:
@patilsuraj767 @kgaikwad  yes I am think that we should not have check for this. Instead of handling this in upgrade_runner. I think we can modify satellite and capsule features to work as 'ex or' logic ? and error out when 'ex or' fails ?

kgaikwad:
> > @patilsuraj767, @upadhyeammit,
> > Definitely, we can decide order on checks using before & after.
> > As availability of definitions is mostly based on features, my concern is what about other scenarios/procedures where we don't run this check explicitly?
> > Does it make sense to load foreman-maintain itself when both packages are available?
> 
> Do you mean to not load foreman-maintain if these packages are available ?

I think, when these packages are available then it would be good to raise failure in the beginning instead of having check and other conditions.

kgaikwad:
As per offline discussion, it would be good to see a reason if any behind why `satellite-capsule` package available under satellite.

definitions/checks/check_capsule_package.rb
```diff
@@ -0,0 +1,19 @@
+module Checks
+  module Foreman
+    class Checks::CheckCapsulePackage < ForemanMaintain::Check
```
kgaikwad, line 3: @patilsuraj767, do we still need this check?

patilsuraj767: @kgaikwad It is useful in case of the satellite to check if the capsule package is installed but in case of `foreman-maintain upgrade check` command it comes useless because foreman-maintain will fail with `Too many scenarios match {:tags=>[:upgrade_scenario, :pre_upgrade_checks]} `  way before this check get execute. 
Should I remove the check?

kgaikwad: Are you still facing `Too many scenarios` error after confine block change related to satellite package checking at capsule feature?

patilsuraj767: Yeah, that confine block helped in not detecting capsule feature on satellite if satellite-capsule package is installed. Did not faced `Too many scenarios` error. I will remove the check.

vsedmik:
@kgaikwad  I tested 6.7.z and 6.8.z upgrades on Satellite server with satelllite-capsule package installed and got no errors. I mean, nor 'f-m upgrade check' nor 'f-m upgrade run' complained about anything and resulted ok.
See http://pastebin.test.redhat.com/873044

kgaikwad:
@patilsuraj767, @vsedmik thank you!

PR #337, opened 2020-05-05
Fixes #29687 - Ignore the DHCP backup if external DHCP

patilsuraj767:
DHCP backup should be only performed when DHCP provider is "dhcp_isc", that is if DHCP is configured on the satellite itself.

### DIFF
--- a/definitions/features/foreman_proxy.rb
+++ b/definitions/features/foreman_proxy.rb
@@ -66,7 +66,7 @@ class Features::ForemanProxy < ForemanMaintain::Feature
 
     configs.push('/var/lib/tftpboot') if backup_features.include?('tftp')
     configs += ['/var/named/', '/etc/named*'] if backup_features.include?('dns')
-    if backup_features.include?('dhcp') && dhcp_provider == 'dhcp_isc'
+    if backup_features.include?('dhcp') && dhcp_isc_provider?
       configs += ['/var/lib/dhcpd', File.dirname(dhcpd_config_file)]
     end
     configs.push('/usr/share/xml/scap') if backup_features.include?('openscap')
@@ -216,16 +216,15 @@ class Features::ForemanProxy < ForemanMaintain::Feature
     dhcp_path
   end
 
-  def parse_dhcp_yml
-    yaml_load(dhcp_yml_path)
+  def configs_from_dhcp_yml
+    @configs_from_dhcp_yml ||= yaml_load(dhcp_yml_path)
   end
 
-  def dhcp_provider
-    parse_dhcp_yml[:use_provider]
+  def dhcp_isc_provider?
+    configs_from_dhcp_yml[:use_provider] == 'dhcp_isc'
   end
 
   def lookup_using_dhcp_yml
-    configs_from_dhcp_yml = parse_dhcp_yml
     if configs_from_dhcp_yml.key?(:dhcp_config)
       return configs_from_dhcp_yml[:dhcp_config]
     elsif configs_from_dhcp_yml.key?(:use_provider)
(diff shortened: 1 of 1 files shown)


definitions/features/foreman_proxy.rb
```diff
@@ -209,11 +209,23 @@ def lookup_dhcpd_config_file
...
+  def dhcp_yml_path
+    dhcp_path = lookup_into(FOREMAN_PROXY_DHCP_YML_PATHS)
+    raise "Couldn't find dhcp.yml file under foreman-proxy" unless dhcp_path
+
+    dhcp_path
+  end
+
+  def parse_dhcp_yml
+    yaml_load(dhcp_yml_path)
+  end
```
kgaikwad, line 221:
```suggestion
  def configs_from_dhcp_yml
     @configs_from_dhcp_yml ||= yaml_load(dhcp_yml_path)
  end
```

definitions/features/foreman_proxy.rb
```diff
@@ -209,11 +209,23 @@ def lookup_dhcpd_config_file
...
+
+  def dhcp_provider
+    parse_dhcp_yml[:use_provider]
+  end
```
kgaikwad, line 225:
```suggestion
  def dhcp_isc_provider?
      configs_from_dhcp_yml[:use_provider] == 'dhcp_isc'
  end
```

definitions/features/foreman_proxy.rb
```diff
@@ -66,7 +66,7 @@ def config_files(for_features = ['all'])
 
     configs.push('/var/lib/tftpboot') if backup_features.include?('tftp')
     configs += ['/var/named/', '/etc/named*'] if backup_features.include?('dns')
-    if backup_features.include?('dhcp')
+    if backup_features.include?('dhcp') && dhcp_provider == 'dhcp_isc'
```
kgaikwad, line 69:
```suggestion
    if backup_features.include?('dhcp') && dhcp_isc_provider?
```

definitions/features/foreman_proxy.rb
```diff
@@ -209,11 +209,23 @@ def lookup_dhcpd_config_file
...
+    yaml_load(dhcp_yml_path)
+  end
+
+  def dhcp_provider
+    parse_dhcp_yml[:use_provider]
+  end
 
-    configs_from_dhcp_yml = yaml_load(dhcp_yml_path)
+  def lookup_using_dhcp_yml
+    configs_from_dhcp_yml = parse_dhcp_yml
```
kgaikwad, line 228: With above change, dhcp_yml configs will be accessible directly using `configs_from_dhcp_yml`

kgaikwad:
@upadhyeammit, do you have any additional comments/concerns?
@jameerpathan111, mind verifying from QE perspective?

upadhyeammit:
> @upadhyeammit, do you have any additional comments/concerns?
> @jameerpathan111, mind verifying from QE perspective?

Changes looks good, if I see the 'VerifyDhcpConfigSyntax' check I think we also need to limit that for isc_dhcp ? if yes then it will be different pr.

kgaikwad:
> if I see the 'VerifyDhcpConfigSyntax' check I think we also need to limit that for isc_dhcp ? if yes then it will be different pr.

@patilsuraj767, any thoughts on this?

patilsuraj767:
@upadhyeammit @kgaikwad Yes, `VerifyDhcpConfigSyntax` uses the DHCP [command](https://github.com/theforeman/foreman_maintain/blob/992b0a327d884c6af0641427a3ff394e3363ba16/definitions/features/foreman_proxy.rb#L187) and it should not be run in case of external DHCP. 
I have created another issue for it - [30140](https://projects.theforeman.org/issues/30140)

PR #338, opened 2020-05-07
Fixes #29645 - Make upgrade executable via REX

upadhyeammit:
There are two different variables in play when considering the non-interactive mode. One is :assumeyes and another is :interactive. If I see the :assumeyes working then we expect that if this is passed then no questions should be asked and run should be in non-interactive mode. I have handled it using the :assumeyes value.

### DIFF
--- a/definitions/procedures/installer/run.rb
+++ b/definitions/procedures/installer/run.rb
@@ -7,7 +7,6 @@ module Procedures::Installer
 
     def run
       assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
-      # If assumeyes selected we execute installer in non-interactive mode
       feature(:installer).run(@arguments, :interactive => !assumeyes_val)
     end
 
--- a/definitions/procedures/installer/run.rb
+++ b/definitions/procedures/installer/run.rb
@@ -2,10 +2,13 @@ module Procedures::Installer
   class Run < ForemanMaintain::Procedure
     metadata do
       param :arguments, 'Arguments passed to installer'
+      param :assumeyes, 'Do not ask for confirmation'
     end
 
     def run
-      feature(:installer).run(@arguments, :interactive => true)
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+      # If assumeyes selected we execute installer in non-interactive mode
+      feature(:installer).run(@arguments, :interactive => !assumeyes_val)
     end
 
     def description
--- a/definitions/procedures/installer/upgrade.rb
+++ b/definitions/procedures/installer/upgrade.rb
@@ -1,7 +1,13 @@
 module Procedures::Installer
   class Upgrade < ForemanMaintain::Procedure
+    metadata do
+      param :assumeyes, 'Do not ask for confirmation'
+    end
+
     def run
-      feature(:installer).upgrade(:interactive => true)
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+      # If assumeyes selected we execute installer in non-interactive mode
+      feature(:installer).run(@arguments, :interactive => !assumeyes_val)
     end
   end
 end
--- a/definitions/procedures/installer/upgrade.rb
+++ b/definitions/procedures/installer/upgrade.rb
@@ -6,7 +6,6 @@ module Procedures::Installer
 
     def run
       assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
-      # If assumeyes selected we execute installer in non-interactive mode
       feature(:installer).run(@arguments, :interactive => !assumeyes_val)
     end
   end
--- a/lib/foreman_maintain/package_manager/dnf.rb
+++ b/lib/foreman_maintain/package_manager/dnf.rb
@@ -10,7 +10,6 @@ module ForemanMaintain::PackageManager
     def dnf_action(action, packages, with_status: false, assumeyes: false)
       yum_options = []
       yum_options << '-y' if assumeyes
-      # If assumeyes selected we execute commands in non-interactive mode
       if with_status
         sys.execute_with_status("dnf #{yum_options.join(' ')} #{action} #{packages.join(' ')}",
                                 :interactive => !assumeyes)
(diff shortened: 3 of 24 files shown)


definitions/procedures/installer/run.rb
```diff
@@ -2,10 +2,13 @@ module Procedures::Installer
   class Run < ForemanMaintain::Procedure
     metadata do
       param :arguments, 'Arguments passed to installer'
+      param :assumeyes, 'Do not ask for confirmation'
     end
 
     def run
-      feature(:installer).run(@arguments, :interactive => true)
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+      # If assumeyes selected we execute installer in non-interactive mode
```
kgaikwad, line 10: `feature(:installer).run(@arguments, :interactive => !assumeyes_val)`

This line explains everything from code perspective about interactive/non-interactive. So in my opinion, it would be good to include this information in the documentation not here.

definitions/procedures/installer/upgrade.rb
```diff
@@ -1,7 +1,13 @@
 module Procedures::Installer
   class Upgrade < ForemanMaintain::Procedure
+    metadata do
+      param :assumeyes, 'Do not ask for confirmation'
+    end
+
     def run
-      feature(:installer).upgrade(:interactive => true)
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+      # If assumeyes selected we execute installer in non-interactive mode
```
kgaikwad, line 9: ditto

lib/foreman_maintain/package_manager/dnf.rb
```diff
@@ -1,17 +1,24 @@
...
 
     private
 
-    def dnf_action(action, packages, assumeyes: false)
+    def dnf_action(action, packages, with_status: false, assumeyes: false)
       yum_options = []
       yum_options << '-y' if assumeyes
-      sys.execute!("dnf #{yum_options.join(' ')} #{action} #{packages.join(' ')}",
-                   :interactive => true)
+      # If assumeyes selected we execute commands in non-interactive mode
```
kgaikwad, line 13: ditto

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -107,12 +107,13 @@ def yum_action(action, packages, with_status: false, assumeyes: false)
       yum_options << '-y' if assumeyes
       yum_options_s = yum_options.empty? ? '' : ' ' + yum_options.join(' ')
       packages_s = packages.empty? ? '' : ' ' + packages.join(' ')
+      # If assumeyes selected we execute commands in non-interactive mode
```
kgaikwad, line 110: ditto

upadhyeammit:
@kgaikwad thank you so much for review.
@vsedmik request to test, thank you !

kgaikwad:
@vsedmik, it would be good to test scenarios include installer run with interactive & non-interactive mode.
Example - upgrade scenario, packages install/upgrade.

vsedmik:
Hello @upadhyeammit @kgaikwad , performed these tests with these results:

1) Run upgrade via playbook:	OK, succeeded

2) Run f-m upgrade command via REX (Run command - SSH Default):
a) non-interactive:	OK, succeeded
b) interactive:		OK, required to confirm upgrade but input stream was exhausted, so exit 1

3) Run f-m upgrade manually on Capsule:
a) non-interactive:	OK, no iteraction required, succeeded
b) interactive:		OK, required to confirm upgrade, succeeded

4) Run f-m package update manually on Capsule:
a) non-interactive:	??, required to confirm with "NOTE: --assumeyes is not applicable for this check"
b) interactive:		OK, required to confirm update and run installer, succeeded

5) Run f-m package install manually on Capsule:
a) non-interactive:	NOK, required to confirm without any NOTE about --assumeyes applicability
b) interactive:		OK, required to confirm installation, succeeded

The non-interacive (a) commands were run using the --assumeyes option.

I am not sure about the 4a) and 5a) expected behaviour, but it seems to be the same as with the original f-m nightly.

kgaikwad:
>     4. Run f-m package update manually on Capsule:
>        a) non-interactive:	??, required to confirm with "NOTE: --assumeyes is not applicable for this check"
>     5. Run f-m package install manually on Capsule:
>        a) non-interactive:	NOK, required to confirm without any NOTE about --assumeyes applicability
> 
> The non-interacive (a) commands were run using the --assumeyes option.
> I am not sure about the 4a) and 5a) expected behaviour, but it seems to be the same as with the original f-m nightly.

Thank you @vsedmik for testing various scenarios. 
IIRC, scenarios 4a) and 5a) are expected. The confirmation added explicitly for installer run and here it won't check for `--assumeyes` option value. I think this is fine.

@upadhyeammit, any thoughts?

upadhyeammit:
@vsedmik below are observations and finds,

1. This command does not execute,
~~~
# bin/foreman-maintain packages update httpd --assumeyes
ERROR: too many arguments

See: 'foreman-maintain packages update --help'
~~~

2. This one executes,

As you can see even if I provide `--assumeyes` its asking for user input.

~~~
# bin/foreman-maintain packages install httpd --assumeyes
Check if tooling for package locking is installed:                    [OK]


Running install packages in unlocked session
Confirm installer run is allowed: 

WARNING: This script runs satellite-installer after the yum execution 
to ensure the Satellite is in a consistent state.
As a result some of your services may be restarted. 

Do you want to proceed?, [y(yes), q(quit)] 
~~~

Ideally 1. 2. both should not work and it should error out in both cases but its not happening, this is something of clamp and not maintain.

Here if you want to use `--assumeyes` with packages then it should be right after install/update command,

~~~
# bin/foreman-maintain packages install --help
Usage:
    foreman-maintain packages install [OPTIONS] PACKAGES ...

Parameters:
    PACKAGES ...                 packages to install

Options:
    -y, --assumeyes              Automatically answer yes for all questions
    -w, --whitelist whitelist    Comma-separated list of labels of steps to be skipped
    -f, --force                  Force steps that would be skipped as they were already run
    -h, --help                   print help
~~~

@vsedmik I request you to file separate issue for inconsistency in `--assumeyes` option.

Thank you,
Amit Upadhye.

upadhyeammit:
@kgaikwad with above description are we good to merge this one ?

vsedmik:
@upadhyeammit new issue created here: https://projects.theforeman.org/issues/29786

kgaikwad:
Thank you @upadhyeammit & @vsedmik.
Above new issue will be track into a separate PR.

PR #340, opened 2020-05-12
Fixes #29727 - Feature to clean tftp boot directory

patilsuraj767:
Reference - https://bugzilla.redhat.com/show_bug.cgi?id=1827176#c8

### DIFF
--- a/definitions/checks/foreman_proxy/check_tftp_storage.rb
+++ b/definitions/checks/foreman_proxy/check_tftp_storage.rb
@@ -11,21 +11,20 @@ module Checks::ForemanProxy
     end
 
     def run
-      tftp_boot_dir = feature(:foreman_proxy).tftp_root_directory + '/boot/'
-      if Dir.exist?(tftp_boot_dir)
-        files = old_files_from_tftp_boot(tftp_boot_dir)
+      if Dir.exist?(tftp_boot_directory)
+        files = old_files_from_tftp_boot
         assert(files.empty?,
                'There are old initrd and vmlinuz files present in tftp',
                :next_steps => Procedures::Files::Remove.new(:files => files))
       else
-        skip "TFTP root directory #{tftp_boot_dir} does not exist."
+        skip "TFTP #{tftp_boot_directory} directory doesn't exist."
       end
     end
 
-    def old_files_from_tftp_boot(tftp_boot_dir)
-      Dir.entries(tftp_boot_dir).map do |file|
+    def old_files_from_tftp_boot
+      Dir.entries(tftp_boot_directory).map do |file|
         unless File.directory?(file)
-          file_path =  tftp_boot_dir + file
+          file_path =  tftp_boot_directory + file
           file_path if File.mtime(file_path) + (token_duration * 60) < Time.now
         end
       end.compact
@@ -33,6 +32,10 @@ module Checks::ForemanProxy
 
     def self.non_zero_token_duration?
       lookup_token_duration != 0
+    end
+
+    def tftp_boot_directory
+      @tftp_boot_directory ||= "#{feature(:foreman_proxy).tftp_root_directory}/boot/"
     end
 
     def token_duration
--- a/definitions/checks/foreman_proxy/check_tftp_storage.rb
+++ b/definitions/checks/foreman_proxy/check_tftp_storage.rb
@@ -1,0 +1,50 @@
+module Checks::ForemanProxy
+  class CheckTftpStorage < ForemanMaintain::Check
+    metadata do
+      label :check_tftp_storage
+      description 'Clean old Kernel and initramfs files from tftp-boot'
+      tags :default
+      confine do
+        feature(:foreman_proxy)
+      end
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def run
+      tftp_boot_dir = feature(:foreman_proxy).tftp_root_directory + '/boot/'
+      if feature(:foreman_proxy).features.include?('tftp')
+        if Dir.exist?(tftp_boot_dir)
+          if token_duration != 0
+            files = files_to_delete(tftp_boot_dir)
+            assert(files.empty?,
+                   'There are old initrd and vmlinuz files present in tftp',
+                   :next_steps => Procedures::Files::Remove.new(:files => files))
+          else
+            skip 'Provisioning token duration is set to 0'
+          end
+        else
+          fail! "TFTP root directory #{tftp_boot_dir} does not exist."
+        end
+      else
+        skip 'TFTP feature is not enabled'
+      end
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    def files_to_delete(tftp_boot_dir)
+      files = list_files_in_directory(tftp_boot_dir). \
+              map { |file| tftp_boot_dir + file }
+      files.map do |file|
+        file if File.mtime(file) + (token_duration * 60) < Time.now
+      end.compact
+    end
+
+    def token_duration
+      data = feature(:foreman_database). \
+             query("select * from settings \
+                    where category = 'Setting::Provisioning' and name = 'token_duration'")
+      param = data[0]['value'].nil? ? 'default' : 'value'
+      YAML.load(data[0][param])
+    end
+  end
+end
--- a/definitions/checks/foreman_proxy/check_tftp_storage.rb
+++ b/definitions/checks/foreman_proxy/check_tftp_storage.rb
@@ -5,46 +5,39 @@ module Checks::ForemanProxy
       description 'Clean old Kernel and initramfs files from tftp-boot'
       tags :default
       confine do
-        feature(:foreman_proxy)
+        feature(:satellite)
       end
     end
 
-    # rubocop:disable Metrics/MethodLength
     def run
       tftp_boot_dir = feature(:foreman_proxy).tftp_root_directory + '/boot/'
-      if feature(:foreman_proxy).features.include?('tftp')
-        if Dir.exist?(tftp_boot_dir)
-          if token_duration != 0
-            files = files_to_delete(tftp_boot_dir)
-            assert(files.empty?,
-                   'There are old initrd and vmlinuz files present in tftp',
-                   :next_steps => Procedures::Files::Remove.new(:files => files))
-          else
-            skip 'Provisioning token duration is set to 0'
-          end
-        else
-          fail! "TFTP root directory #{tftp_boot_dir} does not exist."
-        end
+      if !feature(:foreman_proxy).features.include?('tftp') || token_duration == 0
+        skip 'Skipping the check as TFTP feature is disabled or token duration is zero'
+      elsif Dir.exist?(tftp_boot_dir)
+        files = files_to_delete(tftp_boot_dir)
+        assert(files.empty?,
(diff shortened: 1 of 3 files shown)


definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,50 @@
...
+          else
+            skip 'Provisioning token duration is set to 0'
+          end
+        else
+          fail! "TFTP root directory #{tftp_boot_dir} does not exist."
+        end
+      else
+        skip 'TFTP feature is not enabled'
+      end
+    end
```
upadhyeammit, line 31:
```suggestion
def run
  tftp_boot_dir = feature(:foreman_proxy).tftp_root_directory + '/boot/'
  if !feature(:foreman_proxy).features.include?('tftp') || token_duration == 0
    skip 'Skipping the check as TFTP feature is disabled or token duration is zero'
  elsif Dir.exist?(tftp_boot_dir)
     files = files_to_delete(tftp_boot_dir)
     assert(files.empty?,
            'There are old initrd and vmlinuz files present in tftp',
            :next_steps => Procedures::Files::Remove.new(:files => files))
  else
    warn! "TFTP root directory #{tftp_boot_dir} does not exist, nothing to delete"
  end
end
```

upadhyeammit:
I yet to review remaining files, I shall add comments soon.

definitions/features/foreman_proxy.rb
```diff
@@ -226,6 +233,18 @@ def lookup_using_dhcp_yml
...
 
+  def lookup_tftp_root_directory
+    tftp_yml_path = lookup_into(FOREMAN_PROXY_TFTP_YML_PATHS)
+    raise "Couldn't find tftp.yml file under foreman-proxy" unless tftp_yml_path
+
+    configs_from_tftp_yml = yaml_load(tftp_yml_path)
+    if configs_from_tftp_yml.key?(:tftproot)
+      return configs_from_tftp_yml[:tftproot]
+    else
+      raise "Couldn't find TFTP Configurations in #{tftp_yml_path}"
```
upadhyeammit, line 244: Do we need to raise error here ? I mean is it valid condition tftp feature is enabled but not configured ?

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,50 @@
+module Checks::ForemanProxy
+  class CheckTftpStorage < ForemanMaintain::Check
+    metadata do
+      label :check_tftp_storage
+      description 'Clean old Kernel and initramfs files from tftp-boot'
+      tags :default
+      confine do
+        feature(:foreman_proxy)
```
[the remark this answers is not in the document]
patilsuraj767: Yes, you are right.  
We cannot go with the 1st option because not possible to query pgsql db from the capsule. 2nd option is bit complex because there would be the need of creating separate job template and also user will need to manually copy foreman-proxy public key to the capsule.
I also tried to make use of reverse proxy `# curl -u admin:redhat https://capsule.example.com:8443/api/settings?search=name=token_duration`  but there is no way we can get satellite admin password or auth certs on the capsule server. 
I think for now we can go with the 3rd option that is limiting the check for satellite only.

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def files_to_delete(tftp_boot_dir)
+      files = list_files_in_directory(tftp_boot_dir). \
+              map { |file| tftp_boot_dir + file }
+      files.map do |file|
+        file if File.mtime(file) + (token_duration * 60) < Time.now
+      end.compact
+    end
+
+    def token_duration
```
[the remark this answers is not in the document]
patilsuraj767: Added `@token_duration` instance variable/memoization to run pgsql query only once.

definitions/checks/foreman_proxy/check_tftp_storage.rb
patilsuraj767: Done

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    def files_to_delete(tftp_boot_dir)
+      files = list_files_in_directory(tftp_boot_dir). \
+              map { |file| tftp_boot_dir + file }
+      files.map do |file|
+        file if File.mtime(file) + (token_duration * 60) < Time.now
+      end.compact
```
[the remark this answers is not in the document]
patilsuraj767: Done

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def token_duration
+      data = feature(:foreman_database). \
+             query("select * from settings \
+                    where category = 'Setting::Provisioning' and name = 'token_duration'")
+      param = data[0]['value'].nil? ? 'default' : 'value'
+      YAML.load(data[0][param])
```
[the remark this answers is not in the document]
patilsuraj767: Done

definitions/features/foreman_proxy.rb
patilsuraj767: Confirmed that in any case(external/internal TFTP) ":tftproot" will be configured, so removed the above condition.

upadhyeammit:
@patilsuraj767 thank you for changes, looks good to me now, @kgaikwad if I have missed anything ? request to check. @jameerpathan111 request to test. Thank You !

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,43 @@
...
+      tftp_boot_dir = feature(:foreman_proxy).tftp_root_directory + '/boot/'
+      if !feature(:foreman_proxy).features.include?('tftp') || token_duration == 0
+        skip 'Skipping the check as TFTP feature is disabled or token duration is zero'
+      elsif Dir.exist?(tftp_boot_dir)
+        files = files_to_delete(tftp_boot_dir)
+        assert(files.empty?,
+               'There are old initrd and vmlinuz files present in tftp',
+               :next_steps => Procedures::Files::Remove.new(:files => files))
+      else
+        warn! "TFTP root directory #{tftp_boot_dir} does not exist."
```
kgaikwad, line 22: Can't we skip this check when directory doesn't exist?
OR if we really want to show this message then I will prefer `skip` over `warn!`.

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+
+    def files_to_delete(tftp_boot_dir)
```
kgaikwad, line 26:
```suggestion
    def old_files_from_tftp_boot(tftp_boot_dir)
```

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -86,6 +86,10 @@ def file_nonzero?(filename)
         File.exist?(filename) && !File.zero?(filename)
       end
 
+      def list_files_in_directory(directory_path)
+        Dir.entries(directory_path).select { |f| !File.directory? f }
```
kgaikwad, line 90: Use cases might vary after getting the entries using directory_path so it would be good to move this logic under check itself.

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,43 @@
...
+      description 'Clean old Kernel and initramfs files from tftp-boot'
+      tags :default
+      confine do
+        feature(:satellite)
+      end
+    end
+
+    def run
+      tftp_boot_dir = feature(:foreman_proxy).tftp_root_directory + '/boot/'
+      if !feature(:foreman_proxy).features.include?('tftp') || token_duration == 0
```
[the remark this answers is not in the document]
patilsuraj767: I guess `token_duration` will not work in `confine`, So I will move the `feature(:foreman_proxy).features.include?('tftp')` to confine and will keep `token_duration` here.

definitions/checks/foreman_proxy/check_tftp_storage.rb
patilsuraj767: OK, I will replace it with `skip`.

definitions/checks/foreman_proxy/check_tftp_storage.rb
patilsuraj767: Yeah, make sense. I will change it.

definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    def token_duration
+      @token_duration ||= lookup_token_duration
+    end
+
+    def lookup_token_duration
+      data = feature(:foreman_database). \
+             query("select * from settings \
+                    where category = 'Setting::Provisioning' and name = 'token_duration'")
```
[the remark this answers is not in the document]
patilsuraj767: Ok

definitions/checks/foreman_proxy/check_tftp_storage.rb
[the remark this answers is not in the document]
kgaikwad: using class method?

lib/foreman_maintain/concerns/system_helpers.rb
patilsuraj767: @kgaikwad I actually need the same method in some other PR. I guess method name `list_files_in_directory` itself make sense, i.e. only files not directories. 
I can move this under check but then I will need to define the same method in other PR it will be code repetition.
Let me know your thoughts on this.

definitions/checks/foreman_proxy/check_tftp_storage.rb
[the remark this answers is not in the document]
patilsuraj767: Will the class method work in `confine` block? I am not sure how metadata `confine` works. I will give it a try by adding `token_duration` in confine.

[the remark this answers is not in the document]
kgaikwad: @patilsuraj767, you can achieve this using class_method `non_zero_token_duration?`.

```
module Checks::ForemanProxy
  class CheckTftpStorage < ForemanMaintain::Check
     metadata do
       label :check_tftp_storage
       description 'Clean old Kernel and initramfs files from tftp-boot'
       tags :default
       confine do
         feature(:satellite) && 
             feature(:foreman_proxy) && feature(:foreman_proxy).features.include?('tftp') &&
                non_zero_token_duration?
       end
     end

    def self.non_zero_token_duration?
      lookup_token_duration != 0
    end

    def token_duration
      @token_duration ||= CheckTftpStorage.lookup_token_duration
    end

    def self.lookup_token_duration
      data = feature(:foreman_database). \
             query("select * from settings \
                    where category = 'Setting::Provisioning' and name = 'token_duration'")
      YAML.load(data[0]['value'] || data[0]['default'])
    end
   end
end

```

PR #341, opened 2020-05-13
Fixes #29739 - Cleaning up old ruby gem directories post upgrade

patilsuraj767:


kgaikwad:
As discussed, it would be good to verify spec file once before proceed with this change.

upadhyeammit:
@patilsuraj767 did you get chance to check the spec file ?

upadhyeammit:
The issue should be fixed in packaging, hence closing this pull request.

PR #345, opened 2020-05-18
Fixes #29818 - Fixed hammer setup_admin_access

patilsuraj767:


upadhyeammit (changes requested):
It works as expected, added few comments.

PR #347, opened 2020-05-21
Fixes #29728 - fix epel check if any misconfigured repo present

patilsuraj767:


definitions/checks/repositories/check_non_rh_repository.rb
```diff
@@ -16,7 +16,8 @@ def run
     end
 
     def epel_enabled?
-      system_repos = execute("yum repolist enabled -d 6 -e 0| grep -E 'Repo-baseurl|Repo-id'")
+      cmd = "grep -ir --no-filename --include=\*.repo -E 'baseurl|metalink' /etc/yum.repos.d/"
```
upadhyeammit, line 19: If you use this grep command we end up showing epel enabled even if repository is disabled, because we are not checking for enabled flag ?
Another challenge is, one .repo file can have multiple repos, so in that case I think you end up searching the enabled flag ?

patilsuraj767: @upadhyeammit yeah, My bad. :P 
You are right. I forgot to check "is enabled" part. 
I will re-think and will submit the PR again.

upadhyeammit: @patilsuraj767 any new thoughts on this ?

upadhyeammit:
As this pull request is stale to current state of the repository I am closing it.

PR #349, opened 2020-06-02
Fixes #29760 - Check foreman tasks if running before offline backup

patilsuraj767:


upadhyeammit:
Though the problem seems valid however maybe we cant have solution for this one ?

We have this code change for task(s) which are in running state. 

a. There can be N number of capsules, N number of tasks. This N can be N+something or N-something in future or after wait_time ? like 5 tasks now, and 6 tasks after wait time ? or 5 tasks now and 4 tasks after wait_time ?

b. Now these N number of tasks may complete or may not within wait_time ? and even if it completes then there are still possibilities that we have some other tasks triggered(N+1), which may take more than wait_time ?

Considering this I think we are trying to find out the solution for the situation which is not exactly under control of wait_time ? If we want this change then I feel it should be something like below,

1. there should be user input if he wants to or wait for x amount of wait_time, and this keeps on looping until there is no tasks left in running state ? if we do this, then for sure we will need to have option of force, where we need to forcefully stop services and want the backup. With this we also need to have assumeyes implemented, which means, we keep the user in loop until there are no running tasks ? But still I doubt when we dont see the running tasks and proceed further, at the same time new task is getting triggered ? So even we dont have this option fool proof ? 

2. Other simple solution can be, we warn the user and halt backup and dont use the wait_time at all. And this is the current solution, so I think we keep it as it is ?

@kgaikwad if you have any thoughts ? Maybe I have overthink it ?

definitions/checks/foreman_tasks/not_running.rb
```diff
@@ -6,17 +6,27 @@ class NotRunning < ForemanMaintain::Check
       tags :pre_upgrade
       after :foreman_tasks_not_paused
       before :check_old_foreman_tasks
+      param :wait_time, 'Time to wait in minutes for foreman tasks to finish'
```
kgaikwad, line 9: @patilsuraj767, we should avoid having a parameterized check except checks that are used internally within other definitions.

kgaikwad:
> @kgaikwad if you have any thoughts ? Maybe I have overthink it ?

In the foreman-maintain, we already have logic of wait time & retrying in place for running task. The only difference is that user can't override those values because those are declared as constants.  Foreman Maintain have 5 minutes wait_time. 
I didn't see any request from users for this. We still really want to implement this then I vote to have configuration option than parameter to the `backup` command.

upadhyeammit:
@patilsuraj767 any progress on this one ?

ehelms:
Given the age, and activity of this PR I am going to opt to close it for now. If you would like to continue pursuing it please rebase and re-open.

PR #352, opened 2020-06-10
Fixes #30084 - Get control over interactive_option.

patilsuraj767:


lib/foreman_maintain/cli/base.rb
```diff
@@ -152,23 +152,31 @@ def self.tags_option
         end
       end
 
-      def self.interactive_option
+      # rubocop:disable  Metrics/MethodLength
+      def self.interactive_option(opts = %w[assumeyes whitelist force])
```
kgaikwad, line 156:
```suggestion
      def self.interactive_option(assumeyes: true, whitelist: true, force: true)
```

Instead of having array of strings for options, I think it would be good to have keyword arguments?
@patilsuraj767 , @upadhyeammit - what do you think?

kgaikwad: any updates on this?

patilsuraj767: I think if we change it to  `def self.interactive_option(assumeyes: true, whitelist: true, force: true)` then it would be hard from a code readability point of view. 

Example.- 
1) If we change it to `def self.interactive_option(assumeyes: true, whitelist: true, force: true)`  and suppose in someplace we do not need `--whitelist ` option then while calling the interactive_option we will need to call it like this - `interactive_option(whitelist: false)` but this does not actually tell us what we have included in interactive_option.

2) But if we keep it this way `def self.interactive_option(opts = %w[assumeyes whitelist force])` then with the above same example, we will need to call it like this - interactive_option(['assumeyes']).

The difference is of inclusion and exclusion, If we go with the 1st approach then we will need to pass the interactive_options which we go not want and if we go with the 2nd/PR suggested approach then will need to pass the options which we want to be available with the command. 

@kgaikwad @upadhyeammit Let me know your thoughts on the above.

kgaikwad: > I think if we change it to `def self.interactive_option(assumeyes: true, whitelist: true, force: true)` then it would be hard from a code readability point of view.
> 
> Example.-
> 
> 1. If we change it to `def self.interactive_option(assumeyes: true, whitelist: true, force: true)`  and suppose in someplace we do not need `--whitelist ` option then while calling the interactive_option we will need to call it like this - `interactive_option(whitelist: false)` but this does not actually tell us what we have included in interactive_option.

If readability concern is due to the default values for these options then I don't mind keeping the default values to `false`. In that way one needs to pass all applicable options to command every time. 
Main intention was to use keyword arguments over array. I have suggested `true` as default just because currently we are showing all these options for most of the commands and only few commands where one or two options are not applicable.
Secondly, these interactive options are fixed in numbers & based on arguments it only decides whether any option applicable for a command or not.  I am fine with either way!

@upadhyeammit, any suggestion on this?

upadhyeammit: Hello @patilsuraj767 as we had discussion with @gnurag can you consider about keywords or options hash and push changes ?

upadhyeammit:
@patilsuraj767 can you change interactive_options call of PackagesCommand as per changes in method implementation?

PR #353, opened 2020-06-11
Fixes #25535 - changed the config_files definition

kgaikwad:
Before this patch, the entire /var/www/html/pub directory was backed up. In some cases leading to a config_files.tar.gz of 10's og Gigs.
This change ensures that only files that begin with katello- are added during this phase.

Note that this commit is cherry-picked from original closed PR #235


kgaikwad:
Thank you @abradshaw for your contribution :+1:

PR #355, opened 2020-06-15
Fixes #30120 - Added check-update option in foreman-maintain

patilsuraj767:
```
$ yum check-update
Loaded plugins: foreman-protector, product-id, search-disabled-repos, subscription-manager

WARNING: Excluding 12529 packages due to foreman-protector. 
```

```
$ ./bin/foreman-maintain packages check-update
Running Check for available package updates
Check for available package updates: 
Loaded plugins: product-id, search-disabled-repos, subscription-manager

microcode_ctl.x86_64                 2:2.1-61.6.el7_8         rhel-7-server-rpms
tomcat.noarch                        7.0.76-12.el7_8          rhel-7-server-rpms
tomcat-el-2.2-api.noarch             7.0.76-12.el7_8          rhel-7-server-rpms
tomcat-jsp-2.2-api.noarch            7.0.76-12.el7_8          rhel-7-server-rpms
tomcat-lib.noarch                    7.0.76-12.el7_8          rhel-7-server-rpms
tomcat-servlet-3.0-api.noarch        7.0.76-12.el7_8          rhel-7-server-rpms
                                                                      [OK]

```

### DIFF
--- a/lib/foreman_maintain/package_manager/yum.rb
+++ b/lib/foreman_maintain/package_manager/yum.rb
@@ -61,6 +61,10 @@ module ForemanMaintain::PackageManager
       yum_action('clean', 'all', :assumeyes => assumeyes)
     end
 
+    def check_update
+      yum_action('check-update', nil, :assumeyes => true, :valid_exit_statuses => [100])
+    end
+
     def update_available?(package)
       cmd_output = yum_action('check-update -q', package, :with_status => true, :assumeyes => false)
       cmd_output[0] == 100
@@ -109,10 +113,11 @@ module ForemanMaintain::PackageManager
       File.open(protector_config_file, 'w') { |file| file.puts config }
     end
 
-    def yum_action(action, packages, with_status: false, assumeyes: false)
+    def yum_action(action, packages, with_status: false, assumeyes: false, valid_exit_statuses: [0])
       yum_options = []
       packages = [packages].flatten(1)
       yum_options << '-y' if assumeyes
+      yum_options << '--disableplugin=foreman-protector'
       yum_options_s = yum_options.empty? ? '' : ' ' + yum_options.join(' ')
       packages_s = packages.empty? ? '' : ' ' + packages.join(' ')
       if with_status
@@ -120,7 +125,7 @@ module ForemanMaintain::PackageManager
                                 :interactive => !assumeyes)
       else
         sys.execute!("yum#{yum_options_s} #{action}#{packages_s}",
-                     :interactive => !assumeyes)
+                     :interactive => !assumeyes, :valid_exit_statuses => valid_exit_statuses)
       end
     end
 
--- a/definitions/procedures/packages/check_update.rb
+++ b/definitions/procedures/packages/check_update.rb
@@ -1,0 +1,11 @@
+module Procedures::Packages
+  class CheckUpdate < ForemanMaintain::Procedure
+    metadata do
+      description 'Check for available package updates'
+    end
+
+    def run
+      puts package_manager.check_update
+    end
+  end
+end
--- a/definitions/scenarios/packages.rb
+++ b/definitions/scenarios/packages.rb
@@ -50,6 +50,18 @@ module ForemanMaintain::Scenarios
       end
     end
 
+    class CheckUpdate < ForemanMaintain::Scenario
+      metadata do
+        label :packages_check_update
+        description 'Check for available package updates'
+        manual_detection
+      end
+
+      def compose
+        add_step(Procedures::Packages::CheckUpdate)
+      end
+    end
+
     class Install < ForemanMaintain::Scenario
       metadata do
         description 'install packages in unlocked session'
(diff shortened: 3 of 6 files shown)


lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -101,18 +105,19 @@ def setup_protector(enabled)
       File.open(protector_config_file, 'w') { |file| file.puts config }
     end
 
-    def yum_action(action, packages, with_status: false, assumeyes: false)
+    def yum_action(action, packages, with_status: false, assumeyes: false, valid_exit_statuses: [0])
       yum_options = []
       packages = [packages].flatten(1)
       yum_options << '-y' if assumeyes
+      yum_options << '--disableplugin=foreman-protector'
```
kgaikwad, line 112: Will this option be only for `check_update`?

patilsuraj767: No, this option will for all the yum related task.
With "--disableplugin=foreman-protector" in place there is no need to unlock package while performing yum related task.
We will need to revisit all the commands/scenarios where we are doing yum related task and will need to remove the step `Procedures::Packages::UnlockVersions` For example - While installing package - [code](https://github.com/theforeman/foreman_maintain/blob/992b0a327d884c6af0641427a3ff394e3363ba16/definitions/scenarios/packages.rb#L64).
I think we should do above refactoring in separate PR.

kgaikwad: > I think we should do above refactoring in separate PR.

Could you please create one refactoring issue for tracking this change?

kgaikwad:
No additional comments from my side.
@upadhyeammit, do you have any additional comments?
@jameerpathan111, it would be good to have testing around this new command.

lib/foreman_maintain/package_manager/yum.rb
patilsuraj767: I have opened the issue for refactoring -  [#30393](https://projects.theforeman.org/issues/30393)

upadhyeammit: It works as expected. I am not liking the idea of hardcoding `--disableplugin=foreman-protector` in yum_options. I still prefer the method to unlock packages and lock it again; as this feels like hack. 
Maybe if we want to save the extra logic of disable and enable plugin which has written in setup_protector ? you can override yum_options in setup_protector itself ?
With this we should change the description of `check-update` command to,
~~~
check-update    Check for available package updates in unlocked session
~~~

patilsuraj767: I am not sure how we can override the yum_options in setup_protector, because currently yum_option variable is scoped for `def yum_action`. 

Let me tell you why I think '--disableplugin=foreman-protector' is better option. 

a) I did not found any scenario where we do yum action without unlocking packages(disabling foreman-protector) plugin. 
b) In the current code, I think there are two scenarios where foreman-maintain can leave packages in the unlocked state.
	

1. If you see the implementation here - https://github.com/theforeman/foreman_maintain/blob/c8be149a1ecb92885f8a19f882503a4209412372/definitions/scenarios/packages.rb#L67 (or any other scenario). So we are unlocking the packages but there can be the times that foreman-maintain fails or user can force terminate the process (ctrl +c). In this case, foreman-maintain will leave packages in unlocked state. 
2. We are not explicitly locking the packages in any scenario, rather we are relying on foreman-installer, thinking that it will do the job of locking the packages but what if the installer fails in between? then the package will still stay in the unlocked state. 

Using the `--disableplugin=foreman-protector` will make sure that the package will not be in unlock state ever.

patilsuraj767: Also, with this fix, we can get rid of this kind of issues - [#30414](https://projects.theforeman.org/issues/30414).

kgaikwad:
As per offline discussion, I have tested by passing invalid plugin name. 
Example - `yum install mlocate --disableplugin=xyz`
It gives output with a message `No plugin match for: xyz`. But command returns exit code based on package action,
```
[foreman_maintain]# yum install mlocate --disableplugin=xyz
Loaded plugins: product-id, search-disabled-repos, subscription-manager
No plugin match for: xyz
Package mlocate-0.26-8.el7.x86_64 already installed and latest version
Nothing to do
[foreman_maintain]# echo $?
0

[foreman_maintain]# yum install fhfh --disableplugin=xyz
Loaded plugins: product-id, search-disabled-repos, subscription-manager
No plugin match for: xyz
No package fhfh available.
Error: Nothing to do
[foreman_maintain]# echo $?
1
```

upadhyeammit:
@jameerpathan111 or @vsedmik can you help to test this change ? Most important is, with this change packages wont be unlocked while doing packages actions from foreman-maintain.

jameerpathan111:
> @jameerpathan111 or @vsedmik can you help to test this change ?

yeah, sure. will take a look at it this week.

upadhyeammit:
Tested on latest versions and works without any problem. Merging!

PR #356, opened 2020-06-16
Fixes #30132 - Add upstream upgrade support

upadhyeammit:
This change has,
1. Scenarios for Foreman 1.23,1.24,2.0 including .z scenarios.
2. Scenarios for Katello 3.14,3.15,3.16 including .z scenarios.
3. New feature, which is upstream_repositories.
4. New methods in package_manager::yum which adds the way to validate the links for upstream repos, and help to update/install repo packages.
5. I have mapped candlepin_events and katello_events components to foreman_tasks feature, I need to verify if this is correct from Katello team. This avoid the to_sym errors at post install when we verify all services running using katello/ping. This is in instance feature.
6. Added the mysql db check for 1.24 Foreman scenario, as from 2.0 we dont support mysql db only pgsql is supported.
7. Procedures to support the Foreman upgrade.


upadhyeammit:
@johnpmitsch I have mapped candlepin_events and katello_events components to foreman_tasks feature, is this correct or those should be mapped to Katello feature ? Any more suggestions and review on pr is also welcome.
Thank You!

upadhyeammit:
@kgaikwad and @patilsuraj767 request to review ! Thank You !

johnpmitsch:
> @johnpmitsch I have mapped candlepin_events and katello_events components to foreman_tasks feature, is this correct or those should be mapped to Katello feature ? Any more suggestions and review on pr is also welcome.
> Thank You!

Hey sorry for the delay, I think `candlepin_events` should be mapped to candlepin and `katello_events` mapped to Katello. @jturel may have some thoughts here as well.

jturel:
> > @johnpmitsch I have mapped candlepin_events and katello_events components to foreman_tasks feature, is this correct or those should be mapped to Katello feature ? Any more suggestions and review on pr is also welcome.
> > Thank You!
> 
> Hey sorry for the delay, I think `candlepin_events` should be mapped to candlepin and `katello_events` mapped to Katello. @jturel may have some thoughts here as well.

I think candlepin_events and katello_events would map more closely to the katello feature as they actually run directly in katello. @upadhyeammit what are the implications of the feature mappings for the `hammer ping` services?

definitions/features/katello.rb
```diff
@@ -59,6 +61,16 @@ def config_files_exclude_for_online
     ]
   end
 
+  def repos_rpm(version, os_release)
+    "https://fedorapeople.org/groups/katello/releases/yum/#{version}/katello/el#{os_release}"\
```
kgaikwad, line 65: If you can access `os_release` directly from `ForemanMaintain::Utils::Facter.os_major_release` so would it be possible to remove  method argument `os_release`?

upadhyeammit:
> > > @johnpmitsch I have mapped candlepin_events and katello_events components to foreman_tasks feature, is this correct or those should be mapped to Katello feature ? Any more suggestions and review on pr is also welcome.
> > > Thank You!
> > 
> > 
> > Hey sorry for the delay, I think `candlepin_events` should be mapped to candlepin and `katello_events` mapped to Katello. @jturel may have some thoughts here as well.
> 
> I think candlepin_events and katello_events would map more closely to the katello feature as they actually run directly in katello. @upadhyeammit what are the implications of the feature mappings for the `hammer ping` services?

I dont see any other effect than properly showing error message of which services and which associated components of that are failing.

upadhyeammit:
> > > @johnpmitsch I have mapped candlepin_events and katello_events components to foreman_tasks feature, is this correct or those should be mapped to Katello feature ? Any more suggestions and review on pr is also welcome.
> > > Thank You!
> > 
> > 
> > Hey sorry for the delay, I think `candlepin_events` should be mapped to candlepin and `katello_events` mapped to Katello. @jturel may have some thoughts here as well.
> 
> I think candlepin_events and katello_events would map more closely to the katello feature as they actually run directly in katello. @upadhyeammit what are the implications of the feature mappings for the `hammer ping` services?

Mapping it to katello feature as per your suggestion.

definitions/scenarios/upgrade_to_satellite_6_7.rb
```diff
@@ -27,6 +27,7 @@ class PreUpgradeCheck < Abstract
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:pre_upgrade))
+      add_step(Checks::RemoteExecution::VerifySettingsFileAlreadyExists.new)
```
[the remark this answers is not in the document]
upadhyeammit: old branch residue, removed.

definitions/procedures/foreman/db_seed.rb
```diff
@@ -0,0 +1,13 @@
+module Procedures::Foreman
+  class DbSeed < ForemanMaintain::Procedure
+    metadata do
+      for_feature :foreman_server
+      advanced_run false
+      description 'Update default values seed in the database'
```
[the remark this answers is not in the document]
upadhyeammit: Removed this file

definitions/features/katello.rb
upadhyeammit: The os_release is being used many times to generate repo links. So instead of calling `ForemanMaintain::Utils::Facter.os_major_release` separately and repeatedly in each feature I thought I should collect it once and pass it ?

kgaikwad: Then in my opinion it would be better to move this method inside `instance` feature than `upstream_repositories`. What do you think?

upadhyeammit:
@kgaikwad changes added as per suggestions and there were few more rough edges I found out in upstream_repositories which I have corrected. I yet to test this on Katello and Foreman when having RHEL 7 system. Apart from that if you have any more suggestions request to add. 

Thank You !

definitions/scenarios/upgrade_to_katello_3_14.rb
```diff
@@ -0,0 +1,86 @@
+module Scenarios::Katello_3_14
+  class Abstract < ForemanMaintain::Scenario
+    def self.upgrade_metadata(&block)
+      metadata do
+        tags :upgrade_scenario
+        confine do
+          feature(:instance).upstream? && feature(:foreman_server) && \
```
kgaikwad, line 7: Instead of adding `feature(:instance).upstream?`, can we just add `!feature(:instance).downstream?`

definitions/scenarios/upgrade_to_katello_3_14.rb
```diff
@@ -0,0 +1,86 @@
...
+    upgrade_metadata do
+      description 'Migration scripts to Katello 3.14'
+      tags :migrations
+    end
+
+    def compose
+      add_step(Procedures::Repositories::Setup.new(:version => '3.14'))
+      add_step(Procedures::Packages::UnlockVersions.new)
+      add_step(Procedures::Packages::Update.new(:assumeyes => true))
+      add_step(Procedures::Installer::Upgrade.new)
```
kgaikwad, line 56: Here installer will execute database migrations then the step `Procedures::Foreman::VacuumDatabase` can be run as part of katello upgrade.
What do you think?

lib/foreman_maintain/package_manager/base.rb
```diff
@@ -61,6 +61,21 @@ def files_not_owned_by_package(directory)
       raise NotImplementedError
     end
 
+    # upgrade package sources to install Foreman from
+    def upgrade_foreman_repos(to_version)
```
kgaikwad, line 65: I did find any call to this method. 
Do you really need this? or probably can be removed.

definitions/features/foreman_server.rb
```diff
@@ -47,6 +49,27 @@ def config_files_to_exclude
...
+
+      def repos_rpm(version)
+        "https://yum.theforeman.org/releases/#{version}/el#{feature(:instance).os_major_release}"\
+        "/x86_64/#{FOREMAN_RELEASE_PACKAGE}.rpm"
+      end
+
+      def update_repos(version)
+        repos_rpm = repos_rpm(version, feature(:instance).os_major_release)
+        package_manager.update_or_install(FOREMAN_RELEASE_PACKAGE, repos_rpm, :assumeyes => true)
+        %w[foreman-release-scl centos-release-scl-rh].each do |package|
```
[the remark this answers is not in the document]
upadhyeammit: correct, we prefer `foreman-release-scl` always

definitions/scenarios/upgrade_to_katello_3_14.rb
upadhyeammit: Yes, added `upstream?` for readability, however if it does not make sense then happy to use `!feature(:instance).downstream?`.

definitions/scenarios/upgrade_to_katello_3_14.rb
upadhyeammit: What I think is vacuum should not be part of upgrade as it optional step, sometimes it take hours to complete. I am thinking to take user consent before performing vacuum in post upgrade steps instead of calling it without consent ?

definitions/scenarios/upgrade_to_foreman_1_24_z.rb
```diff
@@ -0,0 +1,94 @@
...
+    upgrade_metadata do
+      description 'Procedures after migrating to Foreman 1.24'
+      tags :post_migrations
+    end
+
+    def compose
+      add_step(Procedures::Foreman::ClearCache.new)
+      add_step(Procedures::Foreman::ClearSessions.new)
+      add_step(Procedures::Foreman::ApipieCache.new)
+      add_step(Procedures::Service::Stop.new(:exclude => 'postgresql'))
```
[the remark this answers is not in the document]
upadhyeammit: The current implementation has advanced run false for vacuumdb, this means someone wont run it individually. Do you think we should keep advanced run to false or let use have option to run it when required ? I think if we keep it false then I can stop and start services in vacuumdb procedure itself.

definitions/scenarios/upgrade_to_katello_3_14.rb
kgaikwad: > What I think is vacuum should not be part of upgrade as it optional step

IIRC I have added this suggestion as I found an inconsistency where you included this step only inside foreman upgrade scenarios not inside Katello scenarios. I wanted to confirm the behavior.

> I am thinking to take user consent before performing vacuum in post upgrade steps instead of calling it without consent ?

It could be good idea or having an option for it while running sub-command may be helpful. 

My point was just that `Procedures::Foreman::VacuumDatabase` step will be useful after database migration so we can include it in Katello scenarios as well.

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,62 @@
+class Features::UpstreamRepositories < ForemanMaintain::Feature
+  metadata do
+    label :upstream_repositories
+
+    confine do
+      !feature(:instance).downstream?
+    end
+  end
+  attr_accessor :version
```
kgaikwad, line 9: Repositories setup is based on version value. Here - https://github.com/theforeman/foreman_maintain/pull/356/files#diff-a2969a5900d684668da14e17f0a8d7fcd9531ac7a3bab6ee8e2a23ded4ca08b1R17
it is expected to pass an argument i.e `version`. 
Wherever needed you already passed `version` as method argument. It is better not to declare `attr_accessor :version` and pass `version` value to nested methods.

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  VERSION_MAPPING = {
```
kgaikwad, line 11:
```suggestion
  KATELLO_VS_FOREMAN_VERSION_MAPPING = {
```

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def setup_repositories(version)
+    @version = version
+    if feature(:katello)
+      major_or_minor_repo_update(feature(:katello))
+      @version = VERSION_MAPPING[version]
+      major_or_minor_repo_update(feature(:foreman_server))
+    elsif feature(:foreman_server)
+      major_or_minor_repo_update(feature(:foreman_server))
+    end
```
kgaikwad, line 25:
```suggestion
    return unless (feature(:katello) || feature(:foreman_server))
    
    if feature(:katello)
      major_or_minor_repo_update(feature(:katello))
      @version = VERSION_MAPPING[version]
    end
    major_or_minor_repo_update(feature(:foreman_server))    
```

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,62 @@
...
+    package_manager.link_valid?(repo_rpm.fetch(feature.label))
+  end
+
+  def update_repos(feature)
+    package_manager.update_or_install(feature::RELEASE_PACKAGE,
+                                      repo_rpm.fetch(feature.label), :assumeyes => true)
+    if feature(:foreman)
+      package_manager.update('centos-release-scl-rh', :assumeyes => true)
+    end
```
kgaikwad, line 51:
```suggestion
    return unless feature(:foreman)
    
    package_manager.update('centos-release-scl-rh', :assumeyes => true)
    ```

File: definitions/scenarios/upgrade_to_katello_3_14.rb (hunk 4 above)
Comment (kgaikwad) [reply definitions/scenarios/upgrade_to_katello_3_14.rb:7]: could you please remove `upstream?` method.

**Review by kgaikwad — CHANGES_REQUESTED** [review]
@upadhyeammit, added few inline comments.

File: definitions/features/katello.rb [hunk 13]
```diff
@@ -7,6 +7,9 @@ class Features::Katello < ForemanMaintain::Feature
     end
   end
 
+  RELEASE_PACKAGE = 'katello-repos'.freeze
+  EL_REPO_URL = 'https://fedorapeople.org/groups/katello/releases/yum/'.freeze
```
evgeni, line 11:
```suggestion
  EL_REPO_URL = 'https://yum.theforeman.org/katello/'.freeze
```

definitions/features/instance.rb
```diff
@@ -63,13 +71,18 @@ def ping
   def server_connection
     net = Net::HTTP.new(ForemanMaintain.config.foreman_url, ForemanMaintain.config.foreman_port)
     net.use_ssl = true
+    net.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
evgeni, line 74: why?

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,62 @@
...
+    label :upstream_repositories
+
+    confine do
+      !feature(:instance).downstream?
+    end
+  end
+  attr_accessor :version
+
+  VERSION_MAPPING = {
+    '3.16' => '2.1',
```
evgeni, line 12: this list is obviously outdated now, given the age of the PR.

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,62 @@
...
+    '3.15' => '2.0',
+    '3.14' => '1.24'
+  }.freeze
+
+  def setup_repositories(version)
+    @version = version
+    if feature(:katello)
+      major_or_minor_repo_update(feature(:katello))
+      @version = VERSION_MAPPING[version]
```
evgeni, line 21: overriding a global variable seems dangerous.

cant we pass the right version to `major_or_minor_repo_update`?

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,62 @@
...
+    if feature.current_version == @version
+      update_repos(feature)
+    else
+      package_manager.update(repo_rpm.fetch(feature.label), :assumeyes => true)
+    end
+    package_manager.clean_cache
+  end
+
+  def available?(version)
+    @version = version
```
evgeni, line 38: a `?` method that modifies a global variable seems wrong.

definitions/procedures/knowledge_base_article.rb
```diff
@@ -26,7 +26,8 @@ def kcs_documents
       'fix_cpdb_validate_failure' => 'https://access.redhat.com/solutions/3362821',
       'fix_db_migrate_failure_on_duplicate_roles' => 'https://access.redhat.com/solutions/3998941',
       'upgrade_puppet_guide_for_sat63' => 'https://access.redhat.com/documentation/en-us/red_hat_satellite/6.3/html/upgrading_and_updating_red_hat_satellite/upgrading_puppet-1',
-      'many_fact_values' => 'https://access.redhat.com/solutions/4163891'
+      'many_fact_values' => 'https://access.redhat.com/solutions/4163891',
+      'migrate_to_postgresql' => 'https://theforeman.org/manuals/1.24/index.html#3.7MigratingtoPostgreSQL'
```
evgeni, line 30: shouldn't we have separate KCS links for up and downstream? RH KCS are not available w/o an RH account, so it seems wrong to tell users "here is the doc for you (oh and you need to register)"

people might think we're trying to force them to become RH customers (which we dont)

upadhyeammit:
I am closing this pull request. The similar work is being worked on and tracked in https://projects.theforeman.org/issues/30132

PR #360, opened 2020-06-22
Don't drop dbs for incremental restore

riffraff169:
Add confirmation message for incremental restore


riffraff169:
Need to abort/reject/deny this pr somehow.  I found deeper considerations (noted in ticket).

PR #361, opened 2020-06-24
Fixes #30204 - fix beta repositories

patilsuraj767:


lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -104,14 +104,13 @@ def common_repos(rh_version_major, full_version)
...
       def common_repos_array(rh_version_major, full_version)
-        ["rhel-#{rh_version_major}-server-satellite-maintenance-6#{use_beta}-rpms",
-         "rhel-#{rh_version_major}-server-satellite-tools-#{full_version}#{use_beta}-rpms"]
-      end
-
-      def use_beta
-        return '-beta' if ENV['FOREMAN_MAINTAIN_USE_BETA'] == '1'
-
-        nil
+        if ENV['FOREMAN_MAINTAIN_USE_BETA'] == '1'
```
kgaikwad, line 107: LGTM :+1: 
if this method only be used by method `common_repos` then it would be good to move logic inside `common_repos` method and remove this method. What do you think?

patilsuraj767: @kgaikwad 
Yes, `def use_beta` is only used in `common_repos_array` method.

kgaikwad:
Thank you @patilsuraj767 :+1:

PR #367, opened 2020-07-06
Fixs #30324 - check for http(s) proxy

sbernhard:


kgaikwad:
Thank you @sbernhard for contribution. 

I would like to know that are you running this check in any external script before `installer` run? OR
do you want to run this check as a part of any specific scenario like upgrade?

sbernhard:
I would run it before every task which uses foreman-installer, like upgrades.

definitions/checks/env_proxy.rb
```diff
@@ -0,0 +1,13 @@
+class Checks::EnvProxy < ForemanMaintain::Check
+  metadata do
+    label :env_proxy
+    tags :env_proxy
+    description 'Check if proxy is set in ENV which shouldn\'t be the case'
```
kgaikwad, line 5:
```suggestion
    description 'Check to make sure no HTTP(S) proxy set in ENV'
```

kgaikwad:
Thank you @sbernhard!

PR #372, opened 2020-07-10
Fixes #30369 - Added condition to check .snar files

patilsuraj767:


definitions/procedures/backup/prepare_directory.rb
```diff
@@ -22,8 +23,14 @@ def run
 
       FileUtils.rm(Dir.glob(File.join(@backup_dir, '.*.snar'))) if @preserve_dir
       if @incremental_dir
-        FileUtils.cp(Dir.glob(File.join(@incremental_dir, '.*.snar')), @backup_dir)
+        snar_files = Dir.glob(File.join(@incremental_dir, '.*.snar'))
+        if snar_files.empty?
```
upadhyeammit, line 27: You can combine `!snar_files.empty?` condition with parent if condition ?  As we should take incremental backup only when snar_files are available?

patilsuraj767: @upadhyeammit 
In `snar_files = Dir.glob(File.join(@incremental_dir, '.*.snar'))` we use `@incremental_dir`. So I think it is good to first check if `@incremental_dir` exist. If we shift the `!snar_files.empty?` to parent then foreman-maintain will every time try to check the snar files without checking if  `@incremental_dir` exist. 
Let me know your thoughts on above.

upadhyeammit: Agree with you !

definitions/procedures/backup/prepare_directory.rb
```diff
@@ -22,8 +23,14 @@ def run
 
       FileUtils.rm(Dir.glob(File.join(@backup_dir, '.*.snar'))) if @preserve_dir
       if @incremental_dir
```
kgaikwad, line 25:
```suggestion
      if @incremental_dir && !(snar_files = Dir.glob(File.join(@incremental_dir, '.*.snar'))).empty?
```

definitions/procedures/backup/prepare_directory.rb
```diff
@@ -21,8 +21,10 @@ def run
       end
 
       FileUtils.rm(Dir.glob(File.join(@backup_dir, '.*.snar'))) if @preserve_dir
-      if @incremental_dir
-        FileUtils.cp(Dir.glob(File.join(@incremental_dir, '.*.snar')), @backup_dir)
+      if @incremental_dir && !(snar_files = Dir.glob(File.join(@incremental_dir, '.*.snar'))).empty?
+        FileUtils.cp(snar_files, @backup_dir)
+      else
+        raise 'No .snar files found in previous backup directory'
```
kgaikwad, line 27: As there are two possibilities, I think we should change error message. Something like -
```suggestion
        raise "May be either incremental backup directory or details such as .snar files under this directory is missing. Could you please provide a valid previous backup directory."
```
@upadhyeammit, any suggestions on message wording?

upadhyeammit: Maybe:

"Either #{@incremental_dir} or #{@incremental_dir}/*.snar files unavailable. Provide valid previous backup directory"

PR #375, opened 2020-07-16
Fixes #29699 - Fix exit code for warning

patilsuraj767:


upadhyeammit:
Hello,

I was checking if there is any error code reserved for warnings but I dont find anything like. So I thought to check how python and ruby program sets the status code upon raising the warning. The status code is 0 for both, so I think we should also use 0 as status code instead of anything else ? @kgaikwad what do you think ?

~~~
$ cat warning.py 
import warnings
warnings.warn('Test')

$ python warning.py
warning.py:2: UserWarning: Test
  warnings.warn('Test')

$ echo $?
0

$ cat warning.rb 
warn('test')

$ ruby warning.rb 
test

$ echo $?
0
~~~

kgaikwad:
>  The status code is 0 for both, so I think we should also use 0 as status code instead of anything else ? @kgaikwad what do you think ?

@upadhyeammit, @patilsuraj767:  +1 for zero exit code for warning messages.

gnurag:
> ```
> $ cat warning.py 
> import warnings
> warnings.warn('Test')
> 
> $ python warning.py
> warning.py:2: UserWarning: Test
>   warnings.warn('Test')
> 
> $ echo $?
> 0
> 
> $ cat warning.rb 
> warn('test')
> 
> $ ruby warning.rb 
> test
> 
> $ echo $?
> 0
> ```

This may not be an accurate test, since `warning.warn()` or `warn()` does not cause the program to exit, and therefore does not actually return anything.

```
import warnings
warnings.warn('Test')
print 'After warning'
exit(42)
```
will return anything one wants to return, with the default exit code being 0.

While its easy to find documentation on shell exit codes for errors, I can't find anything related to warnings perhaps because commands don't actually `exit` in warning scenarios.

vsedmik:
I have tested this PR on Capsule 6.8 snap 10 and works as expected - returns 78 in case of warning.

However, I'm concerned a bit about the effect on the Capsule upgrade playbook, which expects "0" return code as the original issue #29699 proposes. In case of a warning, originally "1" was returned by foreman-maintain. With this PR "78" is returned, which actually makes no change until the result code is handled by the playbook in a different way.

kgaikwad:
As per offline discussion, @patilsuraj767 is going to update readme for this change.

patilsuraj767:
> I have tested this PR on Capsule 6.8 snap 10 and works as expected - returns 78 in case of warning.
> 
> However, I'm concerned a bit about the effect on the Capsule upgrade playbook, which expects "0" return code as the original issue #29699 proposes. In case of a warning, originally "1" was returned by foreman-maintain. With this PR "78" is returned, which actually makes no change until the result code is handled by the playbook in a different way.

Previously both error code and warning code was 1 but after this PR warning code will be 78. 
Capsule upgrade playbook will need to handle the warning code 78 separately.

kgaikwad:
To confirm whether anything such as output logs based on exit codes needs to be handle on Ansible playbook side (i.e related to  capsule upgrade through REX).
+adding @ehelms, @sthirugn into the loop.

ehelms:
This looks good.

README.md
```diff
@@ -452,6 +452,13 @@ Here, 0=ON & 1=OFF.
 If User would like to check whether maintenance-mode is ON/OFF on system in their external script then
 they can use subcommand `foreman-maintain maintenance-mode is-enabled`.
 
+## Exit Codes -
```
kgaikwad, line 455: ## Exit Codes with special meanings -

Every command returns an exit code. Any other exit status than 0 indicates a failure of some kind. Foreman Maintain uses following exit codes with special meaning.

| Exit Code  | Description                        |
| -----------| -----------------------------------|
| 75         | Temporary failure and needs re-run |
| 78         | Command executed with warning(s)   |

kgaikwad (changes requested):

kgaikwad (approved):
LGTM :+1:

kgaikwad:
Thank you @patilsuraj767

PR #376, opened 2020-07-17
Fixes #30458 - don't run steps inside content subcommands

kgaikwad:
@jameerpathan111, 
With this commit, it won't hide `content` command and its sub-commands but foreman-maintain will not run any step or procedure during their invocation.

In my opinion as this is a temporary change, we shouldn't change code base much just to hide these commands. If I get any alternative with minimal code changes which hide this content command then I will post updates.

### DIFF
--- a/definitions/procedures/content/prepare.rb
+++ b/definitions/procedures/content/prepare.rb
@@ -3,6 +3,11 @@ module Procedures::Content
     metadata do
       description 'Prepare content for Pulp 3'
       for_feature :pulpcore
+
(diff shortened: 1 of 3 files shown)


jameerpathan111:
@kgaikwad thanks for fixing this.
If this won't run/change anything on satellite then we are good.

jameerpathan111:
Tried this change on Satellite 6.8 by running `prepare` and `switchover` sub commands  and they didn't change/execute anything on satellite.

PR #377, opened 2020-07-22
Fixes #30469 - Use data_dir with tar transform option

upadhyeammit:
While introducing pgsql 12 we missed to handle tar transform option conditionally to select pgsql 12 data directory. Because of this backup is collected with transform `var/lib/pgsql/data/`. When this backup gets restored it extracts the tar file in `/var/lib/pgsql/data` instead of `/var/opt/rh/rh-postgresql12/lib/pgsql/data/` ; hence on restore the db is empty and does not match with backup.


vsedmik:
Tested with offline/online and with/without pulp data backup combinations and all worked well (content was present after restore).

PR #378, opened 2020-07-22
Fixes #30471 - Return status code 75 on successful self upgrade

upadhyeammit:
The self upgrade mechanism use to return status code 0 on successful package upgrade, however this does not help to know if fm run was actually completed or not as 0 status code is for successful command execution. To handle this scenario now self upgrade will return status code 75 which means 'Temporary failure and user intervention is required'

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,39 +116,39 @@ the *pre-upgrade check* phase. In case the upgrade failed before **migrations**
 phase made some modifying changes, the tool tries to rollback to the previous
 state of the system.
 
-The `foreman-maintain upgrade check` and `foreman-maintain upgrade run` commands execute
-the self update procedure. This tries to update `rubygem-foreman_maintain` package
-from repositories configured on system. If updade is available and successfully completed
-user is asked to re-run the upgrade to use new sources from updated package. With this
-successful upgrade also return status code as 75 which therotically means temporary
-failure and manual user intervetion is required. Below is example of self update,
+#### Self-upgrade for rubygem-foreman_maintain package
 
-```
-# foreman-maintain upgrade check --target-version TARGET_VERSION
-Checking for new version of foreman-maintain...
+**Note:** This feature is available from `rubygem-foreman_maintain` version 0.6.4 and newer.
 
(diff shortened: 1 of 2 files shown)


upadhyeammit:
> Could you please add some information about self-upgrade and exit code `75` in readme file?

Updated readme with self update details and few other corrections. Request to check. Thank You !

README.md
```diff
@@ -114,6 +116,40 @@ the *pre-upgrade check* phase. In case the upgrade failed before **migrations**
...
+
+```
+# foreman-maintain upgrade check --target-version TARGET_VERSION --disable-self-upgrade
+
+# foreman-maintain upgrade run --target-version TARGET_VERSION --disable-self-upgrade
+```
+
+If there isn't any new version available for `rubygem-foreman_maintain` package the upgrade
+continues without halt.
+
```
kgaikwad, line 152: #### Self-upgrade for rubygem-foreman_maintain package

When a user runs any foreman-maintain upgrade sub commands (i.e `foreman-maintain upgrade check` or `foreman-maintain upgrade run`) then

- if update available for `rubygem-foreman_maintain` package, the sub command first tries to update this package. After successful package update, it returns the exit code 75 and requests user to re-run with the updated source code.

- if no update available for `rubygem-foreman_maintain` package, the sub command simply executes the further steps without halt.

Here, *exit code (value 75)* is to indicate that it can not continue with further execution & needs re-run.

Example -

```
# foreman-maintain upgrade check --target-version TARGET_VERSION
Checking for new version of foreman-maintain...
rubygem-foreman_maintain.noarch   repository

Updating foreman-maintain package.

The foreman-maintain package successfully updated.

Re-run foreman-maintain with required options!

# echo $?
75

```

If user wants to skip self-update mechanism then `--disable-self-upgrade` flag can be used with upgrade sub commands. Example -

```
# foreman-maintain upgrade check --target-version TARGET_VERSION --disable-self-upgrade
# foreman-maintain upgrade run --target-version TARGET_VERSION --disable-self-upgrade
```

Note that, this self-upgrade functionality available from `#{foreman_maintain version XXXX}`.

PR #381, opened 2020-07-31
Fixes #30527 - Add check for duplicate permissions

patilsuraj767:
How to reproduce the issue- 

1) Create the duplicate filter. 
`INSERT INTO permissions (name, resource_type) VALUES ('edit_ansible_variables', 'AnsibleVariable');`

2) From foreman UI create two filters in a single role and add both the duplicate permission "edit_ansible_variables" in the filter. You might need to verify the filterings table to make sure both permissions are present in the filter. 
`select * from filterings where filter_id = <FILTER_ID> or filter_id = <FILTER_ID>;`

Run `foreman-maintain health check --label duplicate_permission` to fix the permissions.

### DIFF
--- a/definitions/procedures/foreman/remove_duplicate_permissions.rb
+++ b/definitions/procedures/foreman/remove_duplicate_permissions.rb
@@ -31,33 +31,40 @@ module Procedures::Foreman
 
     def fix_permissions(assigned_permissions)
       persist_permission = assigned_permissions.shift
-      update_filtering(assigned_permissions, persist_permission)
+      filter_ids = filters_for_permission(persist_permission).map { |filter| filter['filter_id'] }
+      update_filtering(assigned_permissions, persist_permission, filter_ids)
+      delete_filtering(assigned_permissions)
       delete_permission(assigned_permissions)
     end
 
-    def update_filtering(old_ids, new_id)
+    def filters_for_permission(permission)
+      feature(:foreman_database).query(
+        "SELECT filter_id FROM filterings WHERE permission_id = #{permission.to_i}"
+      )
+    end
+
+    def update_filtering(old_ids, new_id, filter_ids)
       sql = <<-SQL
       WITH rows AS (
-        UPDATE filterings SET permission_id = '#{new_id}' WHERE permission_id IN (#{old_ids.join(',')})
+        UPDATE filterings SET permission_id = '#{new_id}' WHERE permission_id IN (#{old_ids.join(',')}) AND filter_id NOT IN (#{filter_ids.join(',')})
         RETURNING id
       )
       SELECT id
       FROM rows
       SQL
-
       feature(:foreman_database).query(sql)
     end
 
-    def delete_permission(permission_ids)
-      sql = <<-SQL
-      WITH rows AS (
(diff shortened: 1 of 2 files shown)


definitions/procedures/foreman/remove_duplicate_permissions.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    def query_to_get_permission_ids(permission)
+      <<-SQL
+        SELECT id FROM permissions where name = '#{permission['name']}' and resource_type = '#{permission['resource_type']}'
+      SQL
+    end
+
+    def query_to_check_permission_assign_to_filter(permission_id)
+      <<-SQL
+        SELECT id, filter_id FROM filterings WHERE permission_id = '#{permission_id}'
```
upadhyeammit, line 76: I think this can be changed to permission_ids ? this should help to avoid looping over all permission_ids
~~~
SELECT id, filter_id,permission_id FROM filterings WHERE permission_id in (#{permission_ids})
~~~

definitions/procedures/foreman/remove_duplicate_permissions.rb
```diff
@@ -0,0 +1,80 @@
...
+      assigned_permissions.each do |permission|
+        update_filtering(permission, persist_permission)
+        delete_permission(permission)
+      end
+    end
+
+    def update_filtering(old_id, new_id)
+      sql = <<-SQL
+      WITH rows AS (
+        UPDATE filterings SET permission_id = '#{new_id}' WHERE permission_id = '#{old_id}'
```
upadhyeammit, line 46: As we are keeping one permission_id and replacing all other records with same, this is also eligible for `in` clause ?

definitions/procedures/foreman/remove_duplicate_permissions.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    private
+
+    def check_permissions_assign_to_filter(permission_ids)
+      feature(:foreman_database).
+        query(query_to_check_permission_assign_to_filter(permission_ids)).
+        group_by { |filtering| filtering['permission_id'] }
+    end
+
+    def fix_permissions(assigned_permissions)
```
kgaikwad, line 40: @patilsuraj767, @upadhyeammit 
Would it be good if we confirm this behavior after duplicate entries found?

kgaikwad (approved):
I haven't tested it but codewise looks good to me :+1:

vsedmik:
LGTM, tested the steps above on Sat 6.7.3:

```
[root@dhcp-3-63 foreman_maintain]# bin/foreman-maintain health check --label duplicate_permissions
Check for duplicate permissions from database:                        [FAIL]
Duplicate permissions in your database
Continue with step [Remove duplicate permissions from database]?, [y(yes), n(no), q(quit)] y
Remove duplicate permissions from database:                           [OK]      
Check for duplicate permissions from database:                        [OK]

[root@dhcp-3-63 foreman_maintain]# 
[root@dhcp-3-63 foreman_maintain]# bin/foreman-maintain health check --label duplicate_permissions
Check for duplicate permissions from database:                        [OK]
```
In DB and WebUI looks good, no duplicities.

PR #382, opened 2020-08-03
Fixes #30546 - python3 compatibility

sbernhard:


evgeni:
FWIW, I am not sure this will actually make the protector work on Python3 (at least not EL8), as it imports `yum` which is not existant on EL8:

```
from yum.plugins import PluginYumExit
from yum.plugins import TYPE_CORE
from rpmUtils.miscutils import splitFilename
from yum.packageSack import packagesNewestByName
```

But it's a start and as such we should include it, IMHO.

PR #390, opened 2020-08-28
Fixes #30729 - Check checkpoint_segments tuning parameter

patilsuraj767:


definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,50 @@
+module Checks
+  module Foreman
+    class CheckpointSegments < ForemanMaintain::Check
+      metadata do
+        label :check_checkpoint_segments
+        description 'Check if checkpoint_segments parameter persent in custom hiera'
+        tags :pre_upgrade
```
kgaikwad, line 7: Can we limit this check based on postgresql service using confine block?

patilsuraj767: Yes, that can be done. 
I will update the PR.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      HIERA_FILE = '/etc/foreman-installer/custom-hiera.yaml'.freeze
+
+      def run
+        files = []
+        files << feature(:foreman_database).config_files[0] if check_postgres_config
```
kgaikwad, line 14: It would be better if methods (i.e  check_postgres_config & check_custom_hiera) itself returns file paths when config present otherwise `nil`?

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,50 @@
...
+        unless files.empty?
+          message = ERB.new(<<-BLOCK.strip_heredoc).result(binding)
+          ERROR:
+            checkpoint_segments tuning found.
+            This tuning option is no longer valid in PostgreSQL 9.5+
+            Please remove this from the following locations and then re-run the foreman-maintain upgrade check:
+            <% files.each do |file| %>
+              - <%= file %>
+            <% end %>
+          BLOCK
```
kgaikwad, line 25:
```suggestion
failure_message = <<-MESSAGE.strip_heredoc

ERROR: Tuning option 'checkpoint_segments' found.
This option is no longer valid for PostgreSQL 9.5 and onwards.
Please remove it from following files and re-run the command.
#{files_to_modify.join("\n")}
MESSAGE
```

To maintain consistency could you make use of `MESSAGE.strip_heredoc`.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,50 @@
...
+          fail! message
+        end
+      end
+
+      def check_custom_hiera
```
kgaikwad, line 30: After defining a method inside `installer` feature, you can use a variable instead of constant.
`hiera_file = feature(:installer) ? feature(:installer).custom_hiera_file : nil`

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,50 @@
...
+          if !config.nil? && config.key?('postgresql::server::config_entries') &&
+             config['postgresql::server::config_entries'].key?('checkpoint_segments')
+            return true
+          end
+        end
+      end
+
+      def check_postgres_config
+        param = /(?<!#)checkpoint_segments/
+        config_file = feature(:foreman_database).config_files[0]
```
kgaikwad, line 42: Will it be good to make sure this is not just specific to `foreman_database` (i.e on Foreman/Satellite) but will occur in case of External Foreman Proxy/Capsule as well?

kgaikwad: In case it is applicable to both then you will need to change this logic and consider file_path based on installer/postgresql version.

kgaikwad (changes requested):
Thank you @patilsuraj767 for adding new check.
Added few inline comments.

definitions/checks/foreman/check_checkpoint_segments.rb
patilsuraj767: Yes

kgaikwad (changes requested):
Few suggestions based on scope and name changes as per return values.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,55 @@
...
+      metadata do
+        label :check_checkpoint_segments
+        description 'Check if checkpoint_segments parameter persent in custom hiera'
+      end
+
+      def run
+        files = []
+        files << check_postgres_config
+        files << check_custom_hiera
+        # Make sure array does not contain nil value
```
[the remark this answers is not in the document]
patilsuraj767: Oh Yes, forgot to remove that.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,55 @@
...
+          This option is no longer valid for PostgreSQL 9.5 and onwards.
+          Please remove it from following files and re-run the command.
+          #{files.join("\n")}
+          MESSAGE
+          fail! failure_message
+        end
+      end
+
+      def check_custom_hiera
+        hiera_file = feature(:installer) ? feature(:installer).custom_hiera_file : nil
```
ekohl, line 26: For what it's worth: the installer also throws an error if it finds this key. See https://github.com/theforeman/foreman-installer/commit/3306944761dd559d8343e1748a82d8021d1f99ad

definitions/scenarios/upgrade_to_capsule_6_8.rb
```diff
@@ -27,6 +27,7 @@ class PreUpgradeCheck < Abstract
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:pre_upgrade))
+      add_step(Checks::Foreman::CheckpointSegments)
```
ekohl, line 30: Copying from a private conversation: on Capsules this isn't really relevant since there is no PostgreSQL until we introduce Pulp 3. That's not for 6.8. Future versions that do can assume a PG 12 or newer so this check isn't needed.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,55 @@
...
+        config_file = postgres_config_files
+        if File.exist?(config_file)
+          file = File.read(config_file)
+          return config_file if file.match(param)
+        end
+      end
+
+      def postgres_config_files
+        if find_package('postgresql-server')
+          '/var/lib/pgsql/data/postgresql.conf'
```
ekohl, line 48: I'm not sure you need to check this file. This will only be used by the old version which will continue to accept it.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,55 @@
+module Checks
+  module Foreman
```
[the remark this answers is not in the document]
kgaikwad: As per [comment](https://github.com/theforeman/foreman_maintain/pull/390#discussion_r480123993), we can keep this scope and also add confine for feature `foreman`.

definitions/checks/foreman/check_checkpoint_segments.rb
patilsuraj767: @ekohl The installer has the check but foreman maintain should take care of checkpoint_segments in the pre-upgrade check. 
Scenario - Suppose the user perform # foreman-maintain upgrade run --target-version 6.8 --assumeyes  It is expected that satellite should upgrade without any error. So it is necessary to detect and remove "checkpoint_segments" in per-upgrade.

definitions/scenarios/upgrade_to_capsule_6_8.rb
patilsuraj767: @ekohl Yes, I agree. I will remove it.

definitions/checks/foreman/check_checkpoint_segments.rb
kgaikwad: @gnurag,
This will be `pre-upgrade` check so it will run before the packages update. 
In case of upgrade sat 6.7 to 6.8, while executing this check, server will only have an old PostgreSQL. 
So can we skip verificaton for this config file and just consider Hiera file?

definitions/checks/foreman/check_checkpoint_segments.rb
ekohl: As a reviewer I always like to have a complete picture with references. That makes it easy to compare when I need to.

definitions/checks/foreman/check_checkpoint_segments.rb
patilsuraj767: In the package upgrade from `postgresql-server` to `rh-postgresql12-postgresql-server` if installer is solely responsible for creating the /var/opt/rh/rh-postgresql12/lib/pgsql/data/postgresql.conf and file is not copied from older location to new location then we can remove this check.

ekohl: I don't know for sure if we copy it. If `postgresql-setup --upgrade` does, then we do but I never checked. The rest of the installer just manages a file location.

patilsuraj767: I have tired to upgrade satellite from 6.7 to 6.8 beta and confirmed that we do not need to check Postgresql configuration file. Therefore removed the configuration file check.

definitions/checks/foreman/check_checkpoint_segments.rb
ekohl: Please see that particular commit which I linked. We also warn that there's manual tuning and encourage users to convert to `--tuning`. In the past there was a tuning guide that suggested to paste options in custom-hiera.yaml which included checkpoint_segments. We want to encourage users to use `--tuning` and drop the old options. That's why the last paragraph is there.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,42 @@
...
+          if config['postgresql::server::config_entries'].nil?
+            return <<-MESSAGE.strip_heredoc
+            ERROR: 'postgresql::server::config_entries' cannot be null.
+            Please remove it from following file and re-run the command.
+            - #{hiera_file}
+            MESSAGE
+          elsif config['postgresql::server::config_entries'].key?('checkpoint_segments')
+            return <<-MESSAGE.strip_heredoc
+            ERROR: Tuning option 'checkpoint_segments' found.
+            This option is no longer valid for PostgreSQL 9.5 and onwards.
```
ekohl, line 32: Is this better?
```suggestion
            This option is no longer valid for PostgreSQL 9.5 or newer.
```

definitions/checks/foreman/check_checkpoint_segments.rb
patilsuraj767: Added/Copied the same error message from the commit which you mentioned.

definitions/checks/foreman/check_checkpoint_segments.rb
patilsuraj767: Made the change

ekohl (commented):
Overall the flow makes sense. Since I don't know the foreman-maintain internals I'll leave that part of the review to the maintainers.

patilsuraj767:
> Overall the flow makes sense. Since I don't know the foreman-maintain internals I'll leave that part of the review to the maintainers.

Thanks @ekohl 

@kgaikwad Can you please re-review this PR.

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      def run
+        failure_message = check_custom_hiera
+        fail! failure_message if failure_message
+      end
+
+      # rubocop:disable Metrics/MethodLength
+      def check_custom_hiera
+        hiera_file = feature(:installer) ? feature(:installer).custom_hiera_file : nil
```
upadhyeammit, line 20: Only `hiera_file = feature(:installer).custom_hiera_file` is sufficient as you have already guarded check with feature and existence of file ?

definitions/checks/foreman/check_checkpoint_segments.rb
```diff
@@ -0,0 +1,52 @@
...
+        return unless hiera_file
+        if File.exist?(hiera_file) && (config = YAML.load_file(hiera_file)) &&
```
upadhyeammit, line 22: Same, File.exists? not required ?

upadhyeammit (changes requested):

kgaikwad (approved):
No further comments from my side.

vsedmik:
LGTM, with checkpoint_segments in custom-hiera it shows a warning:

```
Check if checkpoint_segments configuration exists on the system:      [FAIL]
ERROR: Tuning option 'checkpoint_segments' found.
This option is no longer valid for PostgreSQL 9.5 or newer.
Please remove it from following file and re-run the command.
- /etc/foreman-installer/custom-hiera.yaml
The presence of checkpoint_segments in /etc/foreman-installer/custom-hiera.yaml indicates manual tuning.
Manual tuning can override values provided by the --tuning parameter.
Review /etc/foreman-installer/custom-hiera.yaml for values that are already provided by the built in tuning profiles.
Built in tuning profiles also provide a supported upgrade path.

/ Some repositories missing, calling `subscription-manager refresh`   [FAIL]    
Following repositories are not available on your system: rhel-7-server-satellite-6.8-rpms, rhel-7-server-satellite-maintenance-6-rpms, rhel-7-server-satellite-tools-6.8-rpms, rhel-7-server-ansible-2.8-rpms
Scenario [Checks before upgrading to Satellite 6.8] failed.

  [check-postgresql-checkpoint-segments]
  [repositories-validate]

use --whitelist="check-postgresql-checkpoint-segments,repositories-validate"
```

PR #394, opened 2020-09-10
Fixes #30809 - Adding the flag --occurrence=1 on the tar zxf command

waldirio:
Adding the flag --occurrence=1 on the `tar zxf ...` command in order to improve the FQDN check test process.

### DIFF
--- a/lib/foreman_maintain/utils/backup.rb
+++ b/lib/foreman_maintain/utils/backup.rb
@@ -163,7 +163,7 @@ module ForemanMaintain
       def validate_hostname?
         # make sure that the system hostname is the same as the backup
         config_tarball = file_map[:config_files][:path]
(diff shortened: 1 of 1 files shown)


upadhyeammit:
Hello @waldirio thanks for the PR. Can you fix the commit message, specifically `Fixes #30809 - ` is missing in commit message. Apart from that I also like to know if you have tested this change with big config files tarball ?

kgaikwad:
+ adding @jameerpathan111 into loop.

waldirio:
Hello @upadhyeammit, good morning

Fixed the commit and yea, I have tried with some huge files and the improvement/speed up is cristal clear.

Thank you!
Waldirio

upadhyeammit (approved):

PR #398, opened 2020-09-15
Fixes #30848 - Added feature foreman-cockpit

patilsuraj767:
Adding foreman-cockpit feature will help to restart the foreman-cockpit service through foreman_maintain.

### DIFF
--- a/definitions/features/foreman_cockpit.rb
+++ b/definitions/features/foreman_cockpit.rb
@@ -3,7 +3,7 @@ class Features::ForemanCockpit < ForemanMaintain::Feature
     label :foreman_cockpit
 
     confine do
-      find_package('tfm-rubygem-foreman_remote_execution-cockpit')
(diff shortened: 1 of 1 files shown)


definitions/features/foreman_cockpit.rb
```diff
@@ -0,0 +1,15 @@
+class Features::ForemanCockpit < ForemanMaintain::Feature
+  metadata do
+    label :foreman_cockpit
+
+    confine do
+      find_package('tfm-rubygem-foreman_remote_execution-cockpit')
```
upadhyeammit, line 6: We should only have this feature when foreman_server is available ?

upadhyeammit (changes requested):

definitions/features/foreman_cockpit.rb
patilsuraj767: Yeah, that would be an extra check. Added it.

upadhyeammit (approved):

PR #399, opened 2020-09-16
Fixes #30854 - Move downstream info into defs

m-bucher:
This should ensure that the package-name is definable in the downstream-instance, which might be replaced by configuring a different `definitions`-folder.

I would like to be able to specify an additional package to be upgraded in `perform_self_upgrade()` that should be specified in the upstream-instance, since I install an additional package that holds my custom-definitions folder.
Any ideas how this can be done without changing too much?

### DIFF
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -153,7 +153,7 @@ module ForemanMaintain
     def pkg_and_cmd_name
       instance_feature = ForemanMaintain.available_features(:label => :instance).first
       if instance_feature.downstream
-        return %w[satellite-maintain satellite-maintain]
+        return instance_feature_downstream.pkg_and_cmd_name
       end
 
       [main_package_name, 'foreman-maintain']
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -153,7 +153,7 @@ module ForemanMaintain
     def pkg_and_cmd_name
       instance_feature = ForemanMaintain.available_features(:label => :instance).first
       if instance_feature.downstream
-        return instance_feature_downstream.pkg_and_cmd_name
+        return instance_feature.downstream.fm_pkg_and_cmd_name
       end
 
       [main_package_name, 'foreman-maintain']
--- a/lib/foreman_maintain/concerns/downstream.rb
+++ b/lib/foreman_maintain/concerns/downstream.rb
@@ -54,6 +54,10 @@ module ForemanMaintain
         raise NotImplementedError
       end
 
+      def pkg_and_cmd_name
+        [main_package_name, 'foreman-maintain']
+      end
+
       private
 
       def rh_repos(server_version)
--- a/lib/foreman_maintain/concerns/downstream.rb
+++ b/lib/foreman_maintain/concerns/downstream.rb
@@ -54,8 +54,8 @@ module ForemanMaintain
         raise NotImplementedError
       end
 
-      def pkg_and_cmd_name
-        [main_package_name, 'foreman-maintain']
+      def fm_pkg_and_cmd_name
+        %w[satellite-maintain satellite-maintain]
       end
 
       private
--- a/definitions/features/capsule.rb
+++ b/definitions/features/capsule.rb
@@ -18,4 +18,8 @@ class Features::Capsule < ForemanMaintain::Feature
   def package_name
     'satellite-capsule'
   end
+
+  def pkg_and_cmd_name
+    %w[satellite-maintain satellite-maintain]
+  end
 end
--- a/definitions/features/capsule.rb
+++ b/definitions/features/capsule.rb
@@ -18,8 +18,4 @@ class Features::Capsule < ForemanMaintain::Feature
   def package_name
     'satellite-capsule'
   end
-
-  def pkg_and_cmd_name
-    %w[satellite-maintain satellite-maintain]
-  end
 end
--- a/definitions/features/satellite.rb
+++ b/definitions/features/satellite.rb
@@ -17,10 +17,6 @@ class Features::Satellite < ForemanMaintain::Feature
     'satellite'
   end
 
-  def pkg_and_cmd_name
-    %w[satellite-maintain satellite-maintain]
-  end
-
   private
 
   def version_from_source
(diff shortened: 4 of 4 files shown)


lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -54,6 +54,10 @@ def package_name
         raise NotImplementedError
       end
 
+      def pkg_and_cmd_name
```
upadhyeammit, line 57: You can keep pkg_and_cmd_name method in downstream concern only which should serve for both satellite and capsule features. With this I also think we should rename method to something like 'fm_pkg_and_cmd' ?

lib/foreman_maintain.rb
```diff
@@ -153,7 +153,7 @@ def upgrade_in_progress
     def pkg_and_cmd_name
       instance_feature = ForemanMaintain.available_features(:label => :instance).first
       if instance_feature.downstream
-        return %w[satellite-maintain satellite-maintain]
+        return instance_feature_downstream.pkg_and_cmd_name
```
upadhyeammit, line 156: There is typo here it should be `instance_feature.downstream`

upadhyeammit (changes requested):
As you mentioned you want to upgrade one more package with self upgrade, so in this case are you going to add that change in your fork ? Otherwise what I think is, maybe we can keep the self upgrade package list in metadata file, by default it will have the foreman-maintain package and one can add more. @kgaikwad what you think about this way ?

m-bucher:
@upadhyeammit I had not wrapped my head around how to implement the additional-packages feature so we can do this in a separate PR. I can open a separate issue for this.
If you have an idea on how to implement this nicely, please do :-)

kgaikwad:
> Otherwise what I think is, maybe we can keep the self upgrade package list in metadata file, by default it will have the foreman-maintain package and one can add more. @kgaikwad what you think about this way ?

@upadhyeammit, I didn't get which metadata file you referring here. Could you point me the exact file where you would like to have these changes?

upadhyeammit:
> > Otherwise what I think is, maybe we can keep the self upgrade package list in metadata file, by default it will have the foreman-maintain package and one can add more. @kgaikwad what you think about this way ?
> 
> @upadhyeammit, I didn't get which metadata file you referring here. Could you point me the exact file where you would like to have these changes?

sorry, I mean data.yml file ? However one can delete this file

kgaikwad:
> sorry, I mean data.yml file ? However one can delete this file

ahh okies, no problem @upadhyeammit
As per my understanding, foreman-maintain provides a configuration [option](https://github.com/theforeman/foreman_maintain/blob/master/config/foreman_maintain.yml.packaging#L11) to add custom definitions. In similar way, foreman-maintain can have another option to mention package name that provides custom definitions. Accordingly methods i.e `perform_self_upgrade` & `pkg_and_cmd_name` need to be change. 

Needs to check how it will work when a version locking enabled on the system.

m-bucher:
> As per my understanding, foreman-maintain provides a configuration [option](https://github.com/theforeman/foreman_maintain/blob/master/config/foreman_maintain.yml.packaging#L11) to add custom definitions.

Yes, that is what I am currently using for customization. The config-option `:definitions_dirs` in `/etc/foreman-maintain/foreman_maintain.yml`.

> In similar way, foreman-maintain can have another option to mention package name that provides custom definitions.

Yes, but not in the `/etc/foreman-maintain/foreman_maintain.yml` config-file, but in the class/object derived from `downstream` , which has to be included in the customized `definitions`-directory.
I am pretty sure you meant that, I just wanted to clarify that so we are all on the same page, here :smile:

kgaikwad:
> Yes, but not in the `/etc/foreman-maintain/foreman_maintain.yml` config-file, but in the class/object derived from `downstream` , which has to be included in the customized `definitions`-directory.

Yes, not using an extra option in `/etc/foreman-maintain/foreman_maintain.yml` config-file. But as @upadhyeammit mentioned `downstream` is very specific to Satellite & Capsule features so class derived from `downstream` would be just a temporary solution. 

My thinking was - if we could add an extra method on `feature` class which will allow user to mention package(s). This way, one can just override this method in his own feature class (here, in your case any custom feature file). 
By default, foreman-maintain will consider either `['rubygem-foreman_maintain', 'foreman-maintain'] or ['satellite-maintain', 'satellite-maintain']` depending upon Satellite/Capsule feature availability. After then it will take into consideration an extra package list from an available features.
But when version locking is enabled then user needs to take care of before adding any extra package(s) into their custom files as we don't have visibility of their custom directories as well as what package values they are going to add. @upadhyeammit, correct me.

BTW, if this needs to be handle in a separate pull-request then current changes looks good to me :+1: :)

@upadhyeammit, @m-bucher, let me know your thoughts on above approach or 
may be we can start discussion on [https://community.theforeman.org/](https://community.theforeman.org/) to get more suggestions?

upadhyeammit:
> > Yes, but not in the `/etc/foreman-maintain/foreman_maintain.yml` config-file, but in the class/object derived from `downstream` , which has to be included in the customized `definitions`-directory.
> 
> Yes, not using an extra option in `/etc/foreman-maintain/foreman_maintain.yml` config-file. But as @upadhyeammit mentioned `downstream` is very specific to Satellite & Capsule features so class derived from `downstream` would be just a temporary solution.
> 
> My thinking was - if we could add an extra method on `feature` class which will allow user to extend. This way, one can just override this method in his own feature class (here, in your case any custom feature file).
> By default, foreman-maintain will consider either `['rubygem-foreman_maintain', 'foreman-maintain'] or ['satellite-maintain', 'satellite-maintain']` depending upon Satellite/Capsule feature availability. After then it will take into consideration an extra package(s) from an available features to extend list.
> But when version locking is enabled then user needs to take care of before adding any extra package(s) into their custom files as we don't have visibility of their custom directories as well as what package values they are going to add. @upadhyeammit, correct me.
> 
> BTW, if this needs to be handle in a separate pull-request then current changes looks good to me +1 :)
> 
> @upadhyeammit, @m-bucher, let me know your thoughts on above approach or
> may be we can start discussion on https://community.theforeman.org/ to get more suggestions?

Agree ! So having a method in feature class is better idea as in this situation user is already using some custom definitions hence its better to add package name or list to update in that custom definition by extending a method. Apart from that, one would also need to add the required packages in allow-list or whitelist of fm, as self upgrade runs in package locked mode. The current changes looks good to me, and @m-bucher I dont see any better idea to handle this situation. Thank you for contribution !

upadhyeammit (approved):

PR #407, opened 2020-09-23
Fixes #30870 - Option to delete paused task.

patilsuraj767:


definitions/procedures/foreman_tasks/resume.rb
```diff
@@ -8,6 +8,8 @@ class Resume < ForemanMaintain::Procedure
 
     def run
       output << feature(:foreman_tasks).resume_task_using_hammer
+      puts '[ Waiting 30 seconds for resumed tasks to start. ]'
```
upadhyeammit, line 11: Do we need square braces ? if not then those can be removed ? One more thing is, when we delete tasks after this, assuming it did not resume task count may change. I think we should inform user about that ?

upadhyeammit, line 11: Do we need square braces ? if not then those can be removed ? One more thing is, when we delete tasks after this, assuming it did not resume task count may change. I think we should inform user about that ? Should we use spinner to show 30 seconds sleep ?

upadhyeammit (changes requested):
Functionality wise it works as expected. Code wise I am requesting Kavita for more views.

upadhyeammit:
@kgaikwad request to review as changes are in core.

definitions/procedures/foreman_tasks/resume.rb
patilsuraj767: Do you mean informing the user regarding the count/number of tasks which we are deleting?

lib/foreman_maintain/runner.rb
```diff
@@ -142,13 +143,16 @@ def post_step_decisions(scenario, execution)
 
     # rubocop:disable  Metrics/MethodLength
     def ask_about_offered_steps(step)
-      if assumeyes? && rerun_check?(step)
```
kgaikwad, line 145: Currently it is checking `rerun_check?` along with `assumeyes?` so what is expected when there is no any next steps?

definitions/procedures/foreman_tasks/resume.rb
```diff
@@ -8,6 +8,8 @@ class Resume < ForemanMaintain::Procedure
...
+      sleep 30
```
kgaikwad, line 12: Will it be good to define constant at either procedure level or feature level?

definitions/features/foreman_tasks.rb
```diff
@@ -176,6 +178,10 @@ def old_tasks_condition(state = "'stopped', 'paused'")
        "foreman_tasks_tasks.started_at < CURRENT_DATE - INTERVAL '#{MIN_AGE} days'"
   end
 
+  def paused_tasks_condition(state = "'paused'")
+    "foreman_tasks_tasks.state IN (#{state})"
```
kgaikwad, line 182: Here, you have taken state of tasks into consideration.
`not_paused` check calls ` feature(:foreman_tasks).paused_tasks_count(ignored_tasks)` to find specific paused tasks.  
I think, deleting only such paused tasks where result=error and not having label from  `ignored_tasks` is expected here.  Could you confirm this behavior?

kgaikwad (changes requested):
Added few inline comments.

kgaikwad:
@patilsuraj767, @upadhyeammit 
As I mentioned, changes were added to the core functionality which will change behaviour for all checks so it would be good to confirm this behaviour.

lib/foreman_maintain/runner.rb
patilsuraj767: @kgaikwad if there is no any next steps then it is expected that check will fail.

definitions/features/foreman_tasks.rb
patilsuraj767: @kgaikwad  Yes it is expected that we should not delete [ignored_tasks](https://github.com/theforeman/foreman_maintain/blob/eb981e85a7b723382f5d48f054bf402065051616/definitions/checks/foreman_tasks/not_paused.rb#L26)
These tasks are meant to be running always.

definitions/procedures/foreman_tasks/resume.rb
patilsuraj767: Yes, we can do that.

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -208,8 +210,10 @@ def filter_decision(answer)
       # rubocop:disable Metrics/MethodLength
       def ask_to_select(message, steps)
         if assumeyes?
-          puts('(assuming first option)')
-          return steps.first
+          step = steps[@select_option_counter]
+          @select_option_counter += 1
```
upadhyeammit, line 214: With this change we will end up on option 3 if somehow we fail to delete the task ? or some N no of tasks went to paused state while we are deleting previous tasks ?

patilsuraj767: So should we enable maintenance mode while deleting the tasks? Enabling maintenance mode will take care that other task will not go in paused state.

upadhyeammit: I think better way would be to conditionally have only two procedures passed to :next_steps in NotPaused check ? As example,
~~~
      def next_procedures
        if assumeyes?
          return  [Procedures::ForemanTasks::Resume.new,
                   Procedures::ForemanTasks::Delete.new(:state => :paused)]
        end
        [Procedures::ForemanTasks::Resume.new,
         Procedures::ForemanTasks::Delete.new(:state => :paused),
         Procedures::ForemanTasks::UiInvestigate.new('search_query' => scoped_search_query)]
      end

      assert(paused_tasks_count == 0,
             "There are currently #{paused_tasks_count} paused tasks in the system",
             :next_steps => next_procedures)
~~~

definitions/features/foreman_tasks.rb
jameerpathan111: @kgaikwad @patilsuraj767 Currently we delete tasks where result=error, what about the tasks where result is pending or delayed?

jameerpathan111:
@patilsuraj767 I have tried this pr and it looks good to me:

Test steps:
- have tasks with `state=paused` and `result=error`
- foreman-maintain health check --label foreman-tasks-not-paused --assumeyes

Observation:
- foreman-tasks-not-paused check offers new option i.e "Delete paused tasks"
- When check is run with --assumeyes option it will first try to resume paused tasks(`state=paused` and `result=error`) and if it fails to do so it deletes them.

Workaround for creating paused tasks:
```
[root@dhcp-2-22 ~]# foreman-rake console
Loading production environment (Rails 5.2.1)          
irb(main):018:0> t = ForemanTasks::Task.where(id: ["1575bec3-c4be-44bb-8920-22fab7a917ba"]).first
=> #<ForemanTasks::Task::DynflowTask id: "12481c54-65f2-4fa5-9e98-38f0e4b158dd", type: "ForemanTasks::Task::DynflowTask", label: "Actions::Katello::ContentView::Publish", started_at: "2020-10-28 06:36:00", ended_at: "2020-10-28 06:36:05", state: "stopped", result: "success", external_id: "9ad9ff45-7558-49b9-8e84-d486c78abc2b", parent_task_id: nil, start_at: "2020-10-28 06:36:00", start_before: nil, action: "Publish content view 'dev'; organization 'Default ...", user_id: 4, state_updated_at: "2020-10-28 06:36:05">
irb(main):019:0> t.result = "error"
=> "error"
irb(main):020:0> t.state = "paused"
=> "paused"
irb(main):021:0> t.save(:validate => false)
=> true
```

FM command output:
```
Check for paused tasks:                                               [FAIL]
There are currently 2 paused tasks in the system
1) Resume paused tasks
2) Delete paused tasks
3) Investigate the tasks via UI
(assuming option 1)
Resume paused tasks:                                                            
[ Waiting 30 seconds for resumed tasks to start. ]                    [OK]
Total tasks found paused in error state: 2
Total tasks resumed:                     0
Resumed tasks:                           

Total tasks failed to resume:            0
Failed tasks:                            

Total tasks skipped:                     2
Skipped tasks:                           
 1) Task identifier: 1575bec3-c4be-44bb-8920-22fab7a917ba
    Task action:     Publish
    Task errors:     Required lock is already taken by other running tasks.
    Please inspect their state, fix their errors and resume them.
    
    Required lock: read
    Conflicts with tasks:
    - https://dhcp-2-22.vms.sat.rdu2.redhat.com/foreman_tasks/tasks/89b39a65-ba0d-45aa-af4e-ad8201f2ad48
 2) Task identifier: 8cd065ba-9d99-4436-aa97-b788919dbe83
    Task action:     Publish
    Task errors:     Required lock is already taken by other running tasks.
    Please inspect their state, fix their errors and resume them.
    
    Required lock: read
    Conflicts with tasks:
    - https://dhcp-2-22.vms.sat.rdu2.redhat.com/foreman_tasks/tasks/89b39a65-ba0d-45aa-af4e-ad8201f2ad48
Check for paused tasks:                                               [FAIL]
There are currently 2 paused tasks in the system
1) Resume paused tasks
2) Delete paused tasks
3) Investigate the tasks via UI
(assuming option 2)
Delete tasks:                                                                   
| Deleted paused stopped and paused tasks: 2                          [OK]      
```

jameerpathan111:
There was an issue I faced while trying this pr on 6.9. I am still looking into it, so for sometime would request to hold this pr from getting merged.

definitions/features/foreman_tasks.rb
patilsuraj767: @jameerpathan111 There is another check in foreman-maintain. Check `not_running` task which is run after the check `not_paused` in upgrade workflow, it will take care of running/pending tasks.

definitions/procedures/foreman_tasks/resume.rb
upadhyeammit: Yes, number of tasks we are deleting.

patilsuraj767: Done

definitions/features/foreman_tasks.rb
```diff
@@ -97,7 +97,11 @@ def condition(state)
   end
 
   def resume_task_using_hammer
-    feature(:hammer).run('task resume')
+    if package_version('satellite').version >= '6.8'
```
upadhyeammit, line 100: how about using `check_min_version` instead of `package_version` ?

definitions/procedures/foreman_tasks/resume.rb
```diff
@@ -8,6 +8,9 @@ class Resume < ForemanMaintain::Procedure
 
     def run
       output << feature(:foreman_tasks).resume_task_using_hammer
+      with_spinner('Waiting 30 seconds for resumed tasks to start.') do
+        sleep 30
```
upadhyeammit, line 12: This still requires to be defined as constant ?

upadhyeammit:
@jameerpathan111 do you have any thoughts on latest changes ?

definitions/checks/foreman_tasks/not_paused.rb
```diff
@@ -12,9 +12,17 @@ def run
       paused_tasks_count = feature(:foreman_tasks).paused_tasks_count(ignored_tasks)
```
patilsuraj767, line 12: @upadhyeammit According to me we should also remove `ignored_tasks` parameter from here. Because even though we can ignore these tasks they should not be in the Paused state ever, especially while satellite upgrade. 
Let me know your thoughts on above.

upadhyeammit: I agree, however I have these thoughts,

- This check is tagged as default which means it runs for health checks and at the time of upgrade checks and runs. So should we only skip ignored_task when we will actually do the upgrade check and upgrade run ?
- We do this check after restarting the services and as per my limited knowledge if we restart all services these ignored tasks should not be in paused state ? So here if we delete those while doing the health check they wont be created again ? Because after deletion we are not restarting services ? However at the time of upgrade, at very end of scenario we start the services.

What you think @patilsuraj767 and @kgaikwad ?

kgaikwad: Updates on above query:

Currently, this check is tagged as default so it is running as part of health as well as upgrade. 
As per offline discussion with @upadhyeammit & @patilsuraj767, it is worth verifying few things before any solution:
- What will be the impact when it deletes paused tasks including labels "Actions::Candlepin::ListenOnCandlepinEvents", "Actions::Katello::EventQueue::Monitor" because currently it is ignoring paused tasks with those labels?
- Would it be safe to run this deletion procedure when it won't perform service restart? 
Example - foreman-maintain health check command.
- Do we really need this check as part of health command? Other tasks related checks are tagged as "pre-upgrade"

mccun934: >     * What will be the impact when it deletes paused tasks including labels "Actions::Candlepin::ListenOnCandlepinEvents", "Actions::Katello::EventQueue::Monitor" because currently it is ignoring paused tasks with those labels?
> 

delete all of these, these tasks were removed from Foreman Tasks in 6.7 and are no longer processed as a task. Definitely delete them..

>     * Would it be safe to run this deletion procedure when it won't perform service restart?
>       Example - foreman-maintain health check command.

yes, any time you delete a paused task, you do not need to restart services.

> 
>     * Do we really need this check as part of health command? Other tasks related checks are tagged as "pre-upgrade"

I'd keep this with the health check as it is now as well as during the upgrade.

upadhyeammit:
@patilsuraj767 request to do necessary changes, thank you !

upadhyeammit:
@kgaikwad any final thoughts before merging ?

upadhyeammit:
I tested this and what I can see is we are printing sort of confusing message as below. I think we should only keep 'Deleted paused tasks' as message ? its coming from Procedures::ForemanTasks::Delete procedure.

~~~
1) Resume paused tasks
2) Delete paused tasks
3) Investigate the tasks via UI
Select step to continue, [n(next)] 2
Delete tasks:                                                                   
\ Deleted paused stopped and paused tasks: 1                          [OK]      
~~~

definitions/checks/foreman_tasks/not_paused.rb
```diff
@@ -9,12 +9,20 @@ class NotPaused < ForemanMaintain::Check
     end
 
     def run
-      paused_tasks_count = feature(:foreman_tasks).paused_tasks_count(ignored_tasks)
+      paused_tasks_count = feature(:foreman_tasks).paused_tasks_count()
```
kgaikwad, line 12: As per suggestion, you have removed argument `ignored_tasks` while getting paused tasks count. Is this `ignored_tasks` method still needed? Probably leftover?

patilsuraj767: Yes, It will become dead code. Removing it.

definitions/checks/foreman_tasks/not_paused.rb
```diff
@@ -9,12 +9,20 @@ class NotPaused < ForemanMaintain::Check
...
              "There are currently #{paused_tasks_count} paused tasks in the system",
-             :next_steps =>
-               [Procedures::ForemanTasks::Resume.new,
-                Procedures::ForemanTasks::UiInvestigate.new('search_query' => scoped_search_query)])
+             :next_steps => next_procedures)
+    end
+
+    def next_procedures
+      if assumeyes?
+        [Procedures::ForemanTasks::Resume.new,
```
upadhyeammit, line 20: Can you explicitly return the array inside if block ? now even when we have assumeyes? true it returns three next steps.

PR #410, opened 2020-10-21
Fixes #31124 - Don't start pulp 3 if it's not enabled

ianballou:
Pulp 3 services should not be touched if they are not enabled.  This helps ensure that Pulp 3 is not running if the environment is not yet using it.

### DIFF
--- a/definitions/features/pulpcore.rb
+++ b/definitions/features/pulpcore.rb
@@ -5,7 +5,8 @@ class Features::Pulpcore < ForemanMaintain::Feature
     label :pulpcore
 
     confine do
-      ForemanMaintain::Utils::Service::Systemd.new('pulpcore-api', 0).exist?
+      ForemanMaintain::Utils::Service::Systemd.new('pulpcore-api', 0).exist? &&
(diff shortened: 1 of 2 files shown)


johnpmitsch:
So this is for a case when the service is installed but not enabled?

We do check for the service's presence and only perform actions on those services. So if the services do not exist, nothing is done for those services.
https://github.com/theforeman/foreman_maintain/blob/675680f031a1fe032f4df3d33c3c7657003e4faf/definitions/features/service.rb#L26

ianballou:
@johnpmitsch,
> So this is for a case when the service is installed but not enabled?

Yep! This is for the case when Pulp 3 is installed but disabled.

ianballou:
I've made an edit so that the pulpcore feature is completely disabled if the pulpcore-api service is disabled.  It doesn't seem like foreman-maintain would need to do anything with Pulp 3 before enabling its services.  @jlsherrill does that sound right to you?

johnpmitsch (approved):
Works well! pulpcore* services were started when pulpcore-api is enabled, but not when it is disabled.

PR #412, opened 2020-10-29
Fixes #31210 - add check for diskusage in cp-dir

m-bucher:
Make sure /var/lib/candlepin has enough free space.

### DIFF
--- a/definitions/features/candlepin.rb
+++ b/definitions/features/candlepin.rb
@@ -7,6 +7,10 @@ class Features::Candlepin < ForemanMaintain::Feature
     end
   end
 
+  def work_dir
+    '/var/lib/candlepin'
+  end
+
   def services
     [
       system_service('tomcat', 20),
--- a/definitions/checks/disk/available_space_candlepin.rb
+++ b/definitions/checks/disk/available_space_candlepin.rb
@@ -1,0 +1,27 @@
+module Checks
+  module Disk
+    class AvailableSpaceCandlepin < ForemanMaintain::Check
+      metadata do
+        label :available_space_cp
+        description 'Check to make sure /var/lib/candlepin has enough space'
+        tags :pre_upgrade
+        confine do
+          feature(:candlepin) && check_min_version('candlepin', '3.1')
+        end
+      end
+
+      MAX_USAGE_IN_PERCENT = 90
+
+      def run
+        assert(enough_space?, "System has more than #{MAX_USAGE_IN_PERCENT}% space used"\
+               " on #{feature(:candlepin).work_dir}.\n"\
+               'See https://bugzilla.redhat.com/show_bug.cgi?id=1898605')
+      end
+
+      def enough_space?
+        io_obj = ForemanMaintain::Utils::Disk::IODevice.new(feature(:candlepin).work_dir)
+        io_obj.space_used_in_percent < MAX_USAGE_IN_PERCENT
+      end
+    end
+  end
+end
--- a/lib/foreman_maintain/utils/disk/io_device.rb
+++ b/lib/foreman_maintain/utils/disk/io_device.rb
@@ -22,6 +22,10 @@ module ForemanMaintain
           convert_kb_to_mb(execute!("df #{dir}|awk {'print $4'}|tail -1").to_i)
         end
 
+        def space_used_in_percent
+          execute!("df #{dir}|awk {'print $5'}|tail -1").to_i
+        end
+
         private
 
         # In fio command, --direct option bypass the cache page


definitions/checks/disk/available_space_var.rb
```diff
@@ -0,0 +1,26 @@
...
+      MAX_USAGE_IN_PERCENT = 90
+
+      def run
+        assert(enough_space?, "System has more than #{MAX_USAGE_IN_PERCENT}% space used"\
+              ' on /var/lib/candlepin')
+      end
+
+      def enough_space?
+        check_dir = '/var/lib/candlepin'
+        check_dir = '/var/lib' unless File.exist? check_dir
```
upadhyeammit, line 19: Instead of checking defining the check_dir here and then confirming if it exists I think better way would be to add directory in candlepin feature ?
I have not seen anyone changing the default directory structure for candlepin, so in that case, we can just limit this check for feature(:candlepin) ?

definitions/checks/disk/available_space_var.rb
```diff
@@ -0,0 +1,26 @@
...
+      metadata do
+        label :available_space_cp
+        description 'Check to make sure /var/lib/candlepin has enough space'
+        tags :pre_upgrade
+      end
+
+      MAX_USAGE_IN_PERCENT = 90
+
+      def run
+        assert(enough_space?, "System has more than #{MAX_USAGE_IN_PERCENT}% space used"\
```
upadhyeammit, line 13: Instead of assert can we give warning ? assert will fail the check and will need the whitelist option to bypass it. 
With that can we also work with candlepin team to fix the original issue, because hanging does not seem relevant on how much free disk space is available, i.e. 200 GB free space as you mentioned in issue?
If its happening on specific versions of foreman/katello then better we limit it in confine block until original issue is fixed ?

upadhyeammit (changes requested):

definitions/checks/disk/available_space_var.rb
m-bucher: I am OK with limit it to the affected versions. I experienced it with foreman-2.1.3, but I would assume it is the case for foreman >= 2.0 (or whenever candlepin started using/switched to Artemis).
But in this case it is IMHO valid to let the check fail, because it will break the system, as long as it is not fixed in Candlepin or more disk-space is assigned.

m-bucher: Opened a bug-report for candlepin: https://bugzilla.redhat.com/show_bug.cgi?id=1898605

definitions/checks/disk/available_space_var.rb
m-bucher: Please see updated implementation.

definitions/checks/disk/available_space_var.rb
upadhyeammit: Thank you for the changes, one last thing can we scope this for upstream only as of now ? Failing the check instead of warning concerns me for downstream, because this will error out in upgrade and will need to whitelist everytime even when having enough space ?

m-bucher: I added `:warn => true` to `assert()`, is that enough?

quba42: @upadhyeammit IMHO a warning makes zero sense here, since the whole reason for having this check is that the upgrade will fail if this check fails. (Even if you have lots of space, if it is less than 10% free, Candelpin/Artemis/whatever will refuse all tasks and the `foreman-maintain upgrade` run will fail.)

upadhyeammit: Yes, I have thought over this and early failure makes sense, accordingly request to remove the `:warn =>true`.

upadhyeammit:
@m-bucher I have created separate bugzilla for foreman-maintain and updated the issue. Currently attached bugzilla is owned by me, if you wish to own it then feel free to do so.

Thank You !

definitions/checks/disk/available_space_var.rb
```diff
@@ -0,0 +1,26 @@
...
+        confine do
+          feature(:candlepin) && feature(:candlepin).current_minor_version == '3.1'
+        end
+      end
+
+      MAX_USAGE_IN_PERCENT = 90
+
+      def run
+        assert(enough_space?, "System has more than #{MAX_USAGE_IN_PERCENT}% space used"\
+              " on #{feature(:candlepin).work_dir}", :warn => true)
```
upadhyeammit, line 17: " on #{feature(:candlepin).work_dir}")

upadhyeammit (changes requested):

upadhyeammit:
@patilsuraj767 do you have any views on this ? this looks good to me !

definitions/checks/disk/available_space_var.rb
```diff
@@ -0,0 +1,26 @@
+module Checks
+  module Disk
+    class AvailableSpaceCandlepin < ForemanMaintain::Check
+      metadata do
+        label :available_space_cp
+        description 'Check to make sure /var/lib/candlepin has enough space'
+        tags :pre_upgrade
+        confine do
+          feature(:candlepin) && feature(:candlepin).current_minor_version == '3.1'
```
patilsuraj767, line 9: If we want to restrict this for only 3.1 version then it is fine, but on the candlepin BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1898605 I see that target milestone is not set. So possibly we can do. 
```suggestion
          feature(:candlepin) && check_min_version('candlepin', '3.1')'
```

patilsuraj767 (commented):
LGTM :+1:   Just added 1 small point.

m-bucher:
Integrated your change and rebased. Thanks @patilsuraj767

definitions/features/candlepin.rb
```diff
@@ -7,6 +7,22 @@ class Features::Candlepin < ForemanMaintain::Feature
     end
   end
 
+  def current_version
```
patilsuraj767, line 10: As we are now using `check_min_version` method in the check, we can remove `current_version` method as it is not required any more.

definitions/features/candlepin.rb
```diff
@@ -7,6 +7,22 @@ class Features::Candlepin < ForemanMaintain::Feature
...
+    @current_version ||= rpm_version(package_name)
+  end
+
+  def current_minor_version
```
patilsuraj767, line 14: Same for the `current_minor_version` and `package_name` method. It can also be removed.

patilsuraj767 (changes requested):

patilsuraj767 (approved):

definitions/checks/disk/available_space_var.rb
```diff
@@ -0,0 +1,26 @@
+module Checks
+  module Disk
+    class AvailableSpaceCandlepin < ForemanMaintain::Check
```
upadhyeammit, line 3: The `AvailableSpaceCandlepin` and filename `available_space_var.rb` does not match. Can you correct this ?

definitions/checks/disk/available_space_var.rb
```diff
@@ -0,0 +1,26 @@
...
+        tags :pre_upgrade
+        confine do
+          feature(:candlepin) && check_min_version('candlepin', '3.1')
+        end
+      end
+
+      MAX_USAGE_IN_PERCENT = 90
+
+      def run
+        assert(enough_space?, "System has more than #{MAX_USAGE_IN_PERCENT}% space used"\
```
upadhyeammit, line 16: Is it good idea to mention bugzilla no 1898605 in message as its public ? I think it will help user to know why this check is implemented and maybe track the bugzilla for progress ?

upadhyeammit (changes requested):
Thought to give one last round of review before I merge this. Request to check thank you !

definitions/checks/disk/available_space_var.rb
m-bucher: Nice idea, should I display the URL or is it enough to mention `Bugzilla #1898605`?

m-bucher: Do I need to add a newline into the message, because it is rather long right now?

upadhyeammit: I vote for full bugzilla URL on new line !

upadhyeammit (approved):

PR #413, opened 2020-11-03
Fixes #31246 - enable management of Pulpcore sockets

ianballou:
Previously, if all services are stopped, the Pulpcore sockets will still be running.  Now, they will stop as well.


chris1984 (approved):
ACK, looks good and works on my dev box.

johnpmitsch (approved):

PR #414, opened 2020-11-03
Fixes #31196 - add foreman-maintain content migration-stats

ianballou:
Same test steps as https://github.com/Katello/katello/pull/8995, but use `foreman-maintain content migration-stats` instead.  Testing should be done on a production install due to the use of foreman-rake.

### DIFF
--- a/definitions/procedures/content/migration_stats.rb
+++ b/definitions/procedures/content/migration_stats.rb
@@ -1,0 +1,12 @@
+module Procedures::Content
+  class MigrationStats < ForemanMaintain::Procedure
+    metadata do
+      description 'Retrieve Pulp 2 to Pulp 3 migration statistics'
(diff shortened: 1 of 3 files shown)


chris1984 (approved):
ACK, works good. Output of command:
```shell
[root@migration ~]# foreman-maintain content migration-stats
Running Retrieve Pulp 2 to Pulp 3 migration statistics
Retrieve Pulp 2 to Pulp 3 migration statistics: 
Migrated/Total RPMs: 435/1241
Migrated/Total errata: 6/10
Migrated/Total repositories: 2/4

Note: ensure there is sufficient storage space for /var/lib/pulp/published to double in size before starting the migration process.
Check the size of /var/lib/pulp/published with 'du -sh /var/lib/pulp/published/'
                                                                      [OK]
```

PR #415, opened 2020-11-04
Fixes #31253 - Validate if custom-hiera is yaml

patilsuraj767:
Foreman-maintain will not fail with below `NoMethodError` exception if `/etc/foreman-installer/custom-hiera.yaml` has a string. 

```
Check if checkpoint_segments configuration exists on the system:      [FAIL]
undefined method `key?' for "s asdfasdf":String
```

A syntax error will be only reported if `/etc/foreman-installer/custom-hiera.yaml` has a yaml content with syntax issue.

### DIFF
--- a/definitions/checks/foreman/check_checkpoint_segments.rb
+++ b/definitions/checks/foreman/check_checkpoint_segments.rb
@@ -18,31 +18,36 @@ module Checks
       # rubocop:disable Metrics/MethodLength
       def check_custom_hiera
         hiera_file = feature(:installer).custom_hiera_file
-        if (config = YAML.load_file(hiera_file)) &&
-           config.key?('postgresql::server::config_entries')
-          if config['postgresql::server::config_entries'].nil?
-            return <<-MESSAGE.strip_heredoc
-            ERROR: 'postgresql::server::config_entries' cannot be null.
-            Please remove it from following file and re-run the command.
-            - #{hiera_file}
-            MESSAGE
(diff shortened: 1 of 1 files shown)


upadhyeammit (approved):
LGTM !

upadhyeammit:
@jameerpathan111 request to test, thank you !

jameerpathan111:
@upadhyeammit @patilsuraj767 I have tried this pr and no longer see the error mentioned on BZ.
But I also don't see foreman-maintain showing any kind of syntax error for custom-hiera.yaml
The check simply says everything's ok.
on bugzilla, user is expecting it to show syntax error. Are we going to handle that?
```
Expected results:
1.) Foreman-maintain command should check for syntax issue in /etc/foreman-installer/custom-hiera.yaml should show error related to syntax and not to any other misguiding error

```

jameerpathan111:
@patilsuraj767 Tried the patch and it fixes the issue mentioned on bz.
If `/etc/foreman-installer/custom-hiera.yaml`  is not valid yaml then the check gives proper error message.
```
[root@dhcp-2-90 ~]# foreman-maintain health check --label check-postgresql-checkpoint-segments
Check if checkpoint_segments configuration exists on the system:      [FAIL]
Error: File /etc/foreman-installer/custom-hiera.yaml is not a yaml file.

  [check-postgresql-checkpoint-segments]

use --whitelist="check-postgresql-checkpoint-segments"

```

jameerpathan111 (approved):
LGTM

PR #417, opened 2020-11-05
Fixes #31194 - Add content prepare-abort command

ianballou:
This command runs a new rake task that kills all running Pulp 2 -> Pulp 3 migration tasks.

### DIFF
--- a/definitions/procedures/content/migration_stats.rb
+++ b/definitions/procedures/content/migration_stats.rb
@@ -3,6 +3,11 @@ module Procedures::Content
     metadata do
       description 'Retrieve Pulp 2 to Pulp 3 migration statistics'
       for_feature :pulpcore
+
+      confine do
+        # FIXME: remove this condition on next downstream upgrade scenario
+        !feature(:satellite) || feature(:satellite).at_least_version?('6.9')
+      end
     end
 
     def run
--- a/definitions/procedures/content/prepare_abort.rb
+++ b/definitions/procedures/content/prepare_abort.rb
@@ -3,6 +3,11 @@ module Procedures::Content
     metadata do
       description 'Abort all running Pulp 2 to Pulp 3 migration tasks'
       for_feature :pulpcore
+
+      confine do
+        # FIXME: remove this condition on next downstream upgrade scenario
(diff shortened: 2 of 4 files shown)


sjha4:
So I am seeing this https://paste.centos.org/view/16475d4e on both foreman-rake katello:pulp3_migration_abort and foreman-maintain content prepare_abort

ianballou:
hm @sjha4 can you try loading up your task in the rails console and running `execution_plan` on it? I'm not sure why the execution plan would be nil, I haven't seen that.

sjha4:
@ianballou 
I am able to get to the task in the console but strangely the same thing fails in the foreman-rake task..
https://paste.centos.org/view/fd8941ea

ianballou:
@sjha4 weird, I'll have to try it out myself.

ianballou:
@sjha4 I've fixed it, apply this PR to Katello first, then try again: https://github.com/Katello/katello/pull/9032

sjha4:
Able to cancel content prepare task. ACK :+1: 
Screenshot from 2020-11-10 16-02-07

upadhyeammit:
@ianballou I think all the commands and procedures which are exclusively required when we have 6.10 version should be guarded in confine block, else we may end up hiding those like https://github.com/theforeman/foreman_maintain/pull/376 ? @kgaikwad what you think on this and another prs which are for 6.10 migration ?

ianballou:
@upadhyeammit users can run the migration (`content prepare`, `content prepare-abort`, `content migration-stats`) during 6.9.  The only command we don't want them running ever is the one that actually puts them on Pulp 3: `content switchover`.  The `content switchover` would run during the upgrade to 6.10.

Here is what I think: I will update the `# FIXME` on `Procedures::Content::Switchover` to say to remove it for the 6.10 upgrade scenario.  I will add a block on the CLI to never show `content switchover` on a downstream system.

ianballou:
@upadhyeammit I've added this PR to address the 6.10 issues with `content switchover`.  I'm going to fix some of the other confines in this PR as well shortly

ianballou:
Hey @upadhyeammit, do the latest changes work for you?

upadhyeammit (approved):
LGTM !

PR #418, opened 2020-11-09
Fixes #31283 - add foreman-maintain prep-6.10-upgrade command

ianballou:
`foreman-maintain prep-6.10-upgrade` will run only on Satellite 6.9.  It fixes permissions for Pulp 3 to run on its own.

To test, just remove the `confine` requirement for Satellite 6.9 and run `foreman-maintain prep-6.10-upgrade`.  Double check what your permissions were before running so your environment does break.

I used `FileUtils` where I could.  Original commands:

```
chmod -R g+rwX /var/lib/pulp/content
find /var/lib/pulp/content -type d -perm -g-s -exec sudo chmod g+s {} \;
chgrp -R pulp /var/lib/pulp/content
```

### DIFF
--- a/definitions/procedures/prep_6_10_upgrade.rb
+++ b/definitions/procedures/prep_6_10_upgrade.rb
@@ -1,0 +1,31 @@
+class Procedures::Prep610Upgrade < ForemanMaintain::Procedure
+  metadata do
+    description 'Preparations for the Satellite 6.10 upgrade'
+
+    confine do
+      ::Scenarios.const_defined?('Satellite_6_10') &&
+        feature(:satellite) &&
+        feature(:satellite).current_minor_version == '6.9'
+    end
+  end
+
+  def run
+    puts time_warning
+    with_spinner('Updating filesystem permissions for Pulp 3') do |spinner|
+      spinner.update('$ chmod -R g+rwX /var/lib/pulp/content')
+      FileUtils.chmod_R 'g=rwX', '/var/lib/pulp/content'
+      spinner.update("$ find /var/lib/pulp/content -type d -perm -g-s -exec chmod g+s {} \;")
+      execute!('find /var/lib/pulp/content -type d -perm -g-s -exec chmod g+s {} \;')
+      spinner.update('$ chown -R :pulp /var/lib/pulp/content')
+      FileUtils.chown_R nil, 'pulp', '/var/lib/pulp/content'
+      # TODO: Install Pulp 3 without starting services?
+    end
+  end
+
+  private
+
+  def time_warning
+    "\e[33mprep-6.10-upgrade may take a while depending on the "\
+    "size of /var/lib/pulp/content\e[0m"
+  end
+end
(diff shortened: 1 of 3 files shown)


definitions/procedures/prep_6_10_upgrade.rb
```diff
@@ -0,0 +1,16 @@
...
+    confine do
+      feature(:satellite) && feature(:satellite).current_minor_version == '6.9'
+    end
+  end
+
+  def run
+    FileUtils.chmod_R "g=rwX", "/var/lib/pulp/content"
+    puts execute!('find /var/lib/pulp/content -type d -perm -g-s -exec sudo chmod g+s {} \;')
+    FileUtils.chown_R nil, "pulp", "/var/lib/pulp/content"
+    # TODO: Install Pulp 3 without starting services?
```
ianballou, line 14: It's not certain yet if the Pulp 3 foreman-installer command should be run here.  It can always be added later if we need it.

definitions/procedures/prep_6_10_upgrade.rb
```diff
@@ -0,0 +1,16 @@
...
+  metadata do
+    description 'Prepare for the Satellite 6.10 upgrade'
+
+    confine do
+      feature(:satellite) && feature(:satellite).current_minor_version == '6.9'
+    end
+  end
+
+  def run
+    FileUtils.chmod_R "g=rwX", "/var/lib/pulp/content"
```
jlsherrill, line 11: Each of these commands might take quite a while, it might be good to print out what they are doing prior to running them

ianballou: Ok, I've added the output.

definitions/procedures/prep_6_10_upgrade.rb
upadhyeammit: @ianballou instead of puts can you use spinner ? 
As it maybe time consuming can we put a nice banner, something like, `This may take long time to complete depending on the size of pulp content`. ?
I think we dont require sudo in find command because by default foreman-maintain is expected to run as root user ?

upadhyeammit: With current confine block the procedure will be available as soon as user upgrades to 6.9 version, should we make it available only when system is upgradable to 6.10 ?

lib/foreman_maintain/cli.rb
```diff
@@ -27,6 +27,14 @@ class MainCommand < Base
       subcommand 'content', 'Content related commands', ContentCommand
       subcommand 'maintenance-mode', 'Control maintenance-mode for application',
                  MaintenanceModeCommand
+      if ForemanMaintain.detector.feature('satellite') &&
+         ForemanMaintain.detector.feature('satellite').current_minor_version == '6.9'
```
upadhyeammit, line 31: Same, should we make it available when 6.10 upgrade is available ?

ianballou:
@upadhyeammit I think I've addressed your comments.

upadhyeammit (approved):

jlsherrill:
@upadhyeammit would you mind merging? :)

PR #419, opened 2020-11-10
Fixes #31289 - Regenerate apipie cache after pulp3 content switchover

jturel:
There are some aspects of our API that change after migrating to pulp3. So in prod environments we need to update the apipie cache to make sure hammer understands that


upadhyeammit (approved):

jlsherrill:
@upadhyeammit can we get a merge on this :)

jturel:
Thanks guys!

PR #421, opened 2020-11-18
Fixes #31327 - 6.9 sat and capsule scenarios

upadhyeammit:


upadhyeammit:
@ianballou the content migrate, prepare and abort should be available for only Satellite 6.9 version ? or it should be available from Satellite 6.9 and newer versions ?

upadhyeammit:
@ehelms `foreman-rake upgrade:run` task is needed for 6.9 and all the newer versions ?

upadhyeammit:
@jameerpathan111 request to test

ianballou:
> @ianballou the content migrate, prepare and abort should be available for only Satellite 6.9 version ? or it should be available from Satellite 6.9 and newer versions ?

@upadhyeammit `content prepare`, `content prepare-abort`, and `content migration-stats` will only be needed on 6.9.

ehelms:
> @ehelms `foreman-rake upgrade:run` task is needed for 6.9 and all the newer versions ?

Correct.

vsedmik:
Tested the happy-path upgrade on the Capsule for 6.8->6.9 and 6.9.z scenarios:
1) had 6.8 capsule registered to a satellite
2) enabled 6.9 repos from dogfood
3) f-m upgrade list-versions, upgrade check, upgrade run
4) upgrade check, upgrade run for 6.9.z scenario

During the 6.8->6.9 upgrade I bumped into a dependency issue in the `yum update` step, which does not relate to this PR and was already reported here https://bugzilla.redhat.com/show_bug.cgi?id=1898826
The workaround was to `yum remove foreman-proxy-content` package and `yum install satellite-capsule` (removed as well as a dependency) in 6.9.0 version. After that the upgrade succeeded.

Capsule 6.9.z upgrade went smoothly without any issue.

I also tried the same scenarios for Satellite. The 6.8->6.9 succeeded without any issue, but subsequent 6.9.z upgrade failed early in the EPEL check (due to ansible-runner repo enabled?) and then in ping call.
```
Check whether all services are running using the ping call:           [FAIL]
Couldn't connect to the server: undefined method `to_sym' for nil:NilClass
```
Need to investigate more...

jameerpathan111:
I also tried Satellite 6.8 -> 6.9 and it completed successfully.
Faced same issue mentioned by @vsedmik for 6.9 -> 6.9.z
filed https://bugzilla.redhat.com/show_bug.cgi?id=1901031 for the first one.
And for second issue there is https://bugzilla.redhat.com/show_bug.cgi?id=1897360
Will be trying few other scenarios.

jameerpathan111 (approved):
We can merge PR as it is for now, and will do further testing on downstream builds.

PR #427, opened 2020-12-01
Fixes #31426 - use interactive for content prepare

jlsherrill:


ianballou (approved):
Looks good to me!  Not sure what's wrong with the tests.

upadhyeammit:
I think @jlsherrill forced pushed before existing test runs completes ? Do we always need to have interactive option or there should be flag for user to enable it ?

jlsherrill:
@upadhyeammit the goal is to stream output from the rake task to the user in real time.  It appeared that the output didn't show up until the very end unless i turned on interactive.  So yes, i think we always want interactive.  I can re-push, but can you just re-trigger that test?

upadhyeammit (approved):

PR #431, opened 2020-12-16
Fixes #31520 - handle listing of systemd instantiated service

upadhyeammit:
The dynflow-sidekiq is instantiated type service. This means that it start three services as per priority. The foreman-maintain's service list command is based on 'systemctl list-unit-files --type=service' only difference is, it limits the service list to features which are available on the system, i.e dynflow, cockpit. The implementation takes introduces new attribute which helps to set parent_unit_file and detect if service is instance.

### DIFF
--- a/definitions/features/dynflow_sidekiq.rb
+++ b/definitions/features/dynflow_sidekiq.rb
@@ -20,9 +20,12 @@ class Features::DynflowSidekiq < ForemanMaintain::Feature
 
   def services
     [
-      system_service('dynflow-sidekiq@orchestrator', 30, :parent_unit => 'dynflow-sidekiq@'),
-      system_service('dynflow-sidekiq@worker-hosts-queue', 31, :parent_unit => 'dynflow-sidekiq@'),
-      system_service('dynflow-sidekiq@worker', 31, :parent_unit => 'dynflow-sidekiq@')
+      system_service('dynflow-sidekiq@orchestrator', 30,
+                     :instance_parent_unit => 'dynflow-sidekiq@'),
+      system_service('dynflow-sidekiq@worker-hosts-queue', 31,
+                     :instance_parent_unit => 'dynflow-sidekiq@'),
+      system_service('dynflow-sidekiq@worker', 31,
+                     :instance_parent_unit => 'dynflow-sidekiq@')
     ]
   end
 end
--- a/definitions/features/dynflow_sidekiq.rb
+++ b/definitions/features/dynflow_sidekiq.rb
@@ -19,13 +19,24 @@ class Features::DynflowSidekiq < ForemanMaintain::Feature
   end
 
   def services
-    [
-      system_service('dynflow-sidekiq@orchestrator', 30,
-                     :instance_parent_unit => 'dynflow-sidekiq@'),
-      system_service('dynflow-sidekiq@worker-hosts-queue', 31,
-                     :instance_parent_unit => 'dynflow-sidekiq@'),
-      system_service('dynflow-sidekiq@worker', 31,
-                     :instance_parent_unit => 'dynflow-sidekiq@')
-    ]
+    service_names.map do |service|
+      system_service service, instance_priority(service),
+                     :instance_parent_unit => 'dynflow-sidekiq@'
+    end
+  end
+
+  private
+
+  def instance_priority(instance)
+    # Orchestrator should be started before the workers are
+    instance.end_with?('@orchestrator') ? 30 : 31
+  end
+
+  def service_names
+    configured_instances.map { |instance| "dynflow-sidekiq@#{instance}" }
+  end
+
+  def configured_instances
+    Dir['/etc/foreman/dynflow/*'].map { |config| File.basename(config, '.yml') }
   end
 end
--- a/definitions/features/dynflow_sidekiq.rb
+++ b/definitions/features/dynflow_sidekiq.rb
@@ -7,10 +7,6 @@ class Features::DynflowSidekiq < ForemanMaintain::Feature
     end
   end
 
-  def services
-    service_names.map { |service| system_service service, instance_priority(service) }
-  end
-
   def config_files
     # Workaround until foreman-installer can deploy scaled workers
     service_symlinks = configured_instances.map do |service|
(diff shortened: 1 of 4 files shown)


upadhyeammit:
@patilsuraj767 request to review. Thank You !

johnpmitsch:
> @johnpmitsch I have changed few bits of dynflow feature, request to have a look.

It's been a while since I worked in the area so apologies in advance if I miss anything, but the changes look rational to me :+1:

definitions/features/dynflow_sidekiq.rb
```diff
@@ -22,18 +18,11 @@ def config_files
...
-  end
-
-  def configured_instances
-    Dir['/etc/foreman/dynflow/*'].map { |config| File.basename(config, '.yml') }
+  def services
+    [
+      system_service('dynflow-sidekiq@orchestrator', 30, :parent_unit => 'dynflow-sidekiq@'),
+      system_service('dynflow-sidekiq@worker-hosts-queue', 31, :parent_unit => 'dynflow-sidekiq@'),
+      system_service('dynflow-sidekiq@worker', 31, :parent_unit => 'dynflow-sidekiq@')
+    ]
```
kgaikwad, line 26: With new attribute `parent_unit` on `service`, just one additional option needs to be passed here. 

Example -

```
def services
   service_names.map { |service| system_service service, instance_priority(service), :parent_unit => 'dynflow-sidekiq@' }
end
```

Do you really need to remove existing private methods of this feature? Am I missing anything here?

lib/foreman_maintain/utils/service/abstract.rb
```diff
@@ -66,6 +66,14 @@ def disable
       def running?
         raise NotImplementedError
       end
+
+      def instance_parent_unit
```
kgaikwad, line 70: Here, you defined a method `instance_parent_unit` that returns `parent_unit` value from options. Instead of method, I will prefer to use `attr_reader` that ruby provides in the `Systemd` class.
 
Refer similar example from `RemoteDB` class - https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/utils/service/remote_db.rb#L4

lib/foreman_maintain/utils/service/systemd.rb
```diff
@@ -63,6 +62,14 @@ def enabled?
         end
       end
 
+      def instance?
+        instance_parent_unit ? true : false
+      end
+
+      def instance_parent_unit
+        @options.fetch(:parent_unit, nil)
```
kgaikwad, line 70: ditto

definitions/procedures/service/list.rb
```diff
@@ -16,12 +16,22 @@ def run
...
-        execute("systemctl list-unit-files | grep '#{regex}'")
+        execute("systemctl list-unit-files --type=service | \
+                 grep '#{systemd_service_names_regex(services)}'")
       else
         regex = services.map { |service| "^#{service.name} " }.join('\|')
         execute("chkconfig --list 2>&1 | grep '#{regex}'")
       end
     end
+
+    def systemd_service_names_regex(services)
```
kgaikwad, line 27:
```suggestion
    def build_regex_for_services(services)
```
Could you rename this method to something similar and not just specific to `systemd_service`?

kgaikwad (changes requested):
@upadhyeammit, added few inline comments.

definitions/features/dynflow_sidekiq.rb
upadhyeammit: We had three methods,
1. The configured_instances was there to get all instances of service. Now we are predefined those service name(3 of them) with parent_unit attribute.
2. service_names was there to add common name "dynflow-sidekiq@" to all three, here also we have predefined services now so this is also not required.
3. instance_priority was to decide priority for these three services, and we have predefined priority too. so this is not required too.

definitions/procedures/service/list.rb
upadhyeammit: I named it `systemd_service_names_regex' because only systemd service has 'instance_parent_unit'.

lib/foreman_maintain/utils/service/systemd.rb
upadhyeammit: resolved this one

lib/foreman_maintain/utils/service/abstract.rb
upadhyeammit: Resolved

definitions/features/dynflow_sidekiq.rb
kgaikwad: As long as dynflow-sidekiq has only these three & fixed configured_instances but I am not sure about number of dynflow-sidekiq service instances.

@johnpmitsch, will it be okay to remove dynamic construction for dynflow-sidekiq services and handle three instances?  any suggestion?

adamruzicka: One of the key benefits of the switch to dynflow-sidekiq from our custom executor (dynflowd) is the ability to scale up by just spawning more instances of service. If you remove the dynamic construction for dynflow-sidekiq services, then foreman-maintain will not be able to deal with scaled up installations, possibly leading to errors.

Also if I recall correctly `dynflow-sidekiq@worker-hosts-queue` is only present if katello is installed, so you would still have to make this somewhat dynamic.

kgaikwad: Thank you @adamruzicka for sharing details. very helpful!

upadhyeammit:
@kgaikwad moved back to previous dynamic instance discovery, request to review.

lib/foreman_maintain/utils/service/abstract.rb
```diff
@@ -66,6 +66,14 @@ def disable
...
+        raise NotImplementedError
+      end
+
+      def instance?
+        raise NotImplementedError
+      end
```
kgaikwad, line 76: Why do we need these methods in abstract class?

definitions/procedures/service/list.rb
```diff
@@ -16,12 +16,22 @@ def run
...
       end
     end
+
+    def build_regex_for_services(services)
+      services.map do |service|
+        if service.instance?
+          "^#{service.instance_parent_unit}.service"
+        else
+          "^#{service.name}.service"
+        end
```
kgaikwad, line 33:
```suggestion
        if service.respond_to?(:instance_parent_unit) && service.instance_parent_unit
          "^#{service.instance_parent_unit}.service"
        else
          "^#{service.name}.service"
        end
```

I think no need of `instance?` method.

lib/foreman_maintain/utils/service/systemd.rb
```diff
@@ -63,6 +64,10 @@ def enabled?
         end
       end
 
+      def instance?
```
kgaikwad, line 67: I don't see any use of `instance?` method.

kgaikwad (changes requested):
Nothing major added two nitpicks.

kgaikwad (commented):
Codewise looks good to me :+1:

kgaikwad (approved):

PR #432, opened 2020-12-21
Fixes #31535 - check for satellite and capsule packages

upadhyeammit:


lib/foreman_maintain/concerns/primary_checks.rb
```diff
@@ -0,0 +1,24 @@
...
+            raise StandardError, "Error: Important rpm package #{package} is not installed!"\
+                  "\nInstall #{package} rpm to ensure system consistency."
+          end
+        end
+      end
+
+      def package_name
+        installed_scenario = detector.feature(:installer).last_scenario
+        if installed_scenario == 'satellite'
+          'satellite'
```
kgaikwad, line 17: Features `capsule` & `satellite` have `package_name` method but those are not declared as class methods. 
If those will be class methods then I guess we can access them here and no need to hardcode them at multiple places.
BTW, this is just a thought so I don't want to block the PR due this change.

kgaikwad (changes requested):
@upadhyeammit, overall looks good to me :+1: 
Added two inline suggestions.

lib/foreman_maintain/concerns/primary_checks.rb
upadhyeammit: Modified to use class methods.

lib/foreman_maintain/concerns/primary_checks.rb
```diff
@@ -0,0 +1,24 @@
+module ForemanMaintain
+  module Concerns
+    module PrimaryChecks
+      def validate_downstream_packages
+        package = package_name
+        if !package.nil? && detector.feature(:installer).with_scenarios?
+          unless package_manager.installed?(package)
+            raise StandardError, "Error: Important rpm package #{package} is not installed!"\
+                  "\nInstall #{package} rpm to ensure system consistency."
```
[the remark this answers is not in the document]
upadhyeammit: Updated to use "ForemanMaintain::Error::Fail"

kgaikwad (approved):
@upadhyeammit, Not tested but changes look good to me!

upadhyeammit:
@jameerpathan111 request to test

upadhyeammit:
@kgaikwad as per the offline discussion we are going with not changing package_name methods of satellite and capsule features to class methods. I will raise separate pr for same with more testing coverage on same. Thank You for reviews!

PR #434, opened 2021-01-05
Fixes #31575 - Online and offline backup with --include-db-dumps fails

ianballou:
This fixes a syntax error that was failing online and offline backups.

### DIFF
--- a/definitions/procedures/backup/online/safety_confirmation.rb
+++ b/definitions/procedures/backup/online/safety_confirmation.rb
@@ -8,7 +8,7 @@ module Procedures::Backup
       end
 
       def run
-        answer = ask_decision(warning_message(@include_db_dumps), 'y(yes), q(quit)')
+        answer = ask_decision(warning_message(@include_db_dumps), actions_msg: 'y(yes), q(quit)')
         abort! unless answer == :yes
       end
 


jturel:
Works for me! Does it really relate to offline backups as well? To me it only seems like online.

ianballou:
> Works for me! Does it really relate to offline backups as well? To me it only seems like online.

Looks like `Procedures::Backup::Online::SafetyConfirmation` is used for offline backups as well if `--include-db-dumps` is present: https://github.com/theforeman/foreman_maintain/blob/master/definitions/scenarios/backup.rb#L122

jturel:
> > Works for me! Does it really relate to offline backups as well? To me it only seems like online.
> 
> Looks like `Procedures::Backup::Online::SafetyConfirmation` is used for offline backups as well if `--include-db-dumps` is present: https://github.com/theforeman/foreman_maintain/blob/master/definitions/scenarios/backup.rb#L122

Sorry, I got that the wrong way around!

```
[root@newhost certs]# foreman-maintain backup online --include-db-dumps /tmp                                            
ERROR: Unrecognised option '--include-db-dumps'
```

ACK in any case :)

jturel (approved):
ACK

upadhyeammit (approved):

PR #435, opened 2021-01-07
Fixes #31592 - various pulp migration fixes

jlsherrill:
* Move pulpcore services into class methods
  This allows referencing them even if the services
  are disabled.
* Also moves redis out of the 'migration' services,
  since its running regardless
* Do not look for a 6.10 scenario for prep_6_10_upgrade


ianballou (approved):
Worked well for me!

PR #441, opened 2021-01-27
Fixes #31617 - pulp remove procedure

jjeffers:
This change adds a procedure to remove pulp2 packages and data directories. An option is presented to execute automatically, and if declined, prompts the user for each step.


jjeffers:
Closed to reset redmine issue.

PR #446, opened 2021-02-10
Fixes #31680 - execute service actions in group

upadhyeammit:
This handles service actions in group. The services are grouped in priority and with this change with execute an action on same priority services at a time. This improves the start and stop times,
Existing sequential handling of service actions,
~~~
# time bin/foreman-maintain service restart
Running Restart Services

real	3m18.994s
user	0m2.560s
sys	0m1.176s
~~~

Improved group handling of service actions,
~~~
# time bin/foreman-maintain service restart
Running Restart Services

real	2m23.988s
user	0m2.465s
sys	0m0.997s
~~~

### DIFF
--- a/definitions/features/service.rb
+++ b/definitions/features/service.rb
@@ -8,7 +8,6 @@ class Features::Service < ForemanMaintain::Feature
     # { :only => ["httpd"] }
     # { :exclude => ["pulp-workers", "tomcat"] }
     # { :include => ["crond"] }
-
     if feature(:instance).downstream && feature(:instance).downstream.less_than_version?('6.3')
       use_katello_service(action, options)
     else
@@ -27,11 +26,12 @@ class Features::Service < ForemanMaintain::Feature
   end
 
   def filtered_services(options)
-    service_list = include_unregistered_services(existing_services, options[:include])
-    service_list = filter_services(service_list, options)
-    raise 'No services found matching your parameters' unless service_list.any?
+    services = include_unregistered_services(existing_services, options[:include])
+    services = filter_services(services, options)
+    raise 'No services found matching your parameters' unless services.any?
+    return services unless options[:reverse]
 
-    options[:reverse] ? service_list.reverse : service_list
+    Hash[services.sort_by { |k, _| k.to_i }.reverse]
   end
 
   def action_noun(action)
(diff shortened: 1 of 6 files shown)


upadhyeammit:
@Gauravtalreja1 request to review and test!

mccun934:
Tested this and replicated the above results, shaved off a minute:

before patch:
```
real	3m11.152s
user	0m1.923s
sys	0m0.813s

```

after:

```
real	2m5.128s
user	0m1.829s
sys	0m0.729s

```

Tried it with Satellite 6.8 and it made no difference, hovered around 57s:

```
real	0m57.316s
user	0m1.954s
sys	0m0.800s
```

:+1:

mccun934:
we need to improve the Rails loading time in Foreman, but this is a good patch to include in the mean time

definitions/features/service.rb
```diff
@@ -61,29 +64,27 @@ def use_system_service(action, options, spinner)
...
+    filtered_services(options).each_value do |group|
+      group.each_slice(2).to_a.each do |subgroup|
+        spinner.update("#{action_noun(action)} #{subgroup.join(', ')}")
+      end
+      result = execute_with_status('systemctl ' \
+        "#{action} #{group.map(&:name).join(' ')}")
+      if result[0] > 0 || action == 'status'
+        failed_services |= services_status(action, group, options, spinner)
       end
+      status = result[0] if result[0] > 0
```
kgaikwad, line 76:
```suggestion
      systemctl_status, _output = execute_with_status('systemctl ' \
        "#{action} #{group.map(&:name).join(' ')}")
      if systemctl_status > 0 
        status = systemctl_status
        failed_services |= services_status(action, group, options, spinner)
      end
```

definitions/features/service.rb
```diff
@@ -61,29 +64,27 @@ def use_system_service(action, options, spinner)
...
-      if item_status > 0
-        status = item_status
-        failed_services << service
+    filtered_services(options).each_value do |group|
+      group.each_slice(2).to_a.each do |subgroup|
+        spinner.update("#{action_noun(action)} #{subgroup.join(', ')}")
+      end
+      result = execute_with_status('systemctl ' \
+        "#{action} #{group.map(&:name).join(' ')}")
+      if result[0] > 0 || action == 'status'
```
kgaikwad, line 73: One query - why did you add special handling for action == `status` here and inside `format_status` method? As per my understanding it is because you clubbed two things - failed_services & formatting the message logic in a single method, right?

I found it little confusing while reading the code, will it be possible to separate out these two things

kgaikwad (changes requested):
Hello @upadhyeammit, 
I haven't tested the changes. Added few inline comments.

definitions/features/service.rb
```diff
@@ -31,7 +30,11 @@ def filtered_services(options)
     service_list = filter_services(service_list, options)
```
kgaikwad, line 30: As the return type of `filter_services` is modified from `array` to `hash` so could you rename this `service_list` variable to `services` or something else?

definitions/features/service.rb
upadhyeammit: Yes, I agree its bit confusing. Actually as services are being handled in group I need to sequentially get the status of every service if there is failure or action is status. I dont see a way to simply it more than this, at max I can remove the OR condition on line 73 and use elsif because it exactly needs same methods with different params.
With this I better I change failed_services variable to `statuses` as it resembles this implementation.

kgaikwad (changes requested):
Added a small inline comment. 
I haven't tested the changes, code wise rest of the changes look good to me.

kgaikwad (approved):
LGTM codewise :+1:

PR #447, opened 2021-02-16
Fixes #31886 - error out if psql and mongo packages missing

upadhyeammit:
When checking external db connections foreman-maintain fails to do so if psql and mongo commands are not available. This change does not try to install packages on external system as it will require auth and root access which will over complicate the restore process. Its expected to have those packages on target system. If not then we explicitly fail the checks instead of giving false warning of db connection failure.


upadhyeammit:
@patilsuraj767 @Gauravtalreja1 request to review and test. Thank You!

PR #449, opened 2021-02-17
Fixes #31906 - Add content migration reset command

ianballou:
Adds the `content migration-reset` command.

Needs https://github.com/Katello/katello/pull/9142 in Katello.

Questions:

Do we need any Satellite guards?

Do you agree with the naming and printed information?

### DIFF
--- a/definitions/procedures/content/migration_reset.rb
+++ b/definitions/procedures/content/migration_reset.rb
@@ -1,0 +1,12 @@
+module Procedures::Content
+  class MigrationReset < ForemanMaintain::Procedure
+    metadata do
+      description 'Reset the Pulp 2 to Pulp 3 migration data (pre-switchover)'
+      for_feature :pulpcore
+    end
+
+    def run
+      puts execute!('foreman-rake katello:pulp3_migration_reset')
+    end
+  end
+end
--- a/definitions/scenarios/content.rb
+++ b/definitions/scenarios/content.rb
@@ -79,6 +79,24 @@ module ForemanMaintain::Scenarios
       end
     end
 
+    class MigrationReset < ForemanMaintain::Scenario
+      metadata do
+        label :content_migration_reset
+        description 'Reset the Pulp 2 to Pulp 3 migration data (pre-switchover)'
+        manual_detection
+      end
+
+      def compose
+        if feature(:satellite) && feature(:satellite).at_least_version?('6.9')
+          enable_and_start_services
+          add_step(Procedures::Content::MigrationReset)
+          disable_and_stop_services
+        elsif !feature(:satellite)
+          add_step(Procedures::Content::MigrationReset)
+        end
+      end
+    end
+
     class RemovePulp2 < ForemanMaintain::Scenario
       metadata do
         label :content_remove_pulp2
@@ -87,7 +105,7 @@ module ForemanMaintain::Scenarios
       end
 
       def compose
-        add_step(Procedures::Pulp::Remove)
+        add_step_with_context(Procedures::Pulp::Remove)
       end
     end
   end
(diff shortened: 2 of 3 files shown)


upadhyeammit:
Hello @ianballou,

Below are my observations,

* On 6.9 I can see its failing as task is not there,

~~~
# bin/foreman-maintain content migration-reset
Running Reset the Pulp 2 to Pulp 3 migration data (pre-switchover)
Reset the Pulp 2 to Pulp 3 migration data (pre-switchover):           [FAIL]
Failed executing foreman-rake katello:pulp3_migration_reset, exit status 1:
 rake aborted!
Don't know how to build task 'katello:pulp3_migration_reset' (see --tasks)
/opt/rh/rh-ruby25/root/usr/share/gems/gems/rake-12.3.0/exe/rake:27:in `<top (required)>'
(See full trace by running task with --trace)
~~~

    When we call the procedures from scenarios I can see that its not limiting the execution as per for_feature. 
    This means that on 6.9 even when pulpcore is not availble command will execute and throw error. 

* I can see advanced run is limiting it as per for_feature,

~~~
# bin/foreman-maintain advanced procedure run --help|grep content-migration-reset
~~~

* To avoid this happening instead of using `add_step` use `add_step_with_context` method. I request you to correct below calls of content scenario. I have added `Pulp::Remove` too in list as its allowing me to remove pulp2 even when `katello-common` version is less than 4.0. Apologies for missing this in earlier changes.
~~~
          add_step(Procedures::Content::Prepare)
          add_step(Procedures::Content::Switchover)
          add_step(Procedures::Content::PrepareAbort)
          add_step(Procedures::Content::MigrationStats)
          add_step(Procedures::Content::MigrationReset)
          add_step(Procedures::Pulp::Remove)
~~~

Also like to know if we need this for 6.9 GA?

upadhyeammit (changes requested):
Requesting to do changes as per suggestions

jjeffers:
Tested this on 3.18 production, and it worked as expected.

ianballou:
@upadhyeammit I added `add_step_with_context` where you requested.  In 6.9 I noticed that the other steps still run, like restarting services.  If I add the service restart steps with context, it errors out with:
```
undefined method `present?' for #<Procedures::Service::Enable:0x000000033db2f8>
```

How do you want to handle those steps that are still running? Should we just leave them?

upadhyeammit:
> @upadhyeammit I added `add_step_with_context` where you requested. In 6.9 I noticed that the other steps still run, like restarting services. If I add the service restart steps with context, it errors out with:
> 
> ```
> undefined method `present?' for #<Procedures::Service::Enable:0x000000033db2f8>
> ```
> 
> How do you want to handle those steps that are still running? Should we just leave them?

As I understand we are calling list of services using class method and passing that to service procedure add_step_with_context wont help, I dont see any other way than limiting enable and disable service methods for pulpcore feature only?

ianballou:
@upadhyeammit so we actually want the following procedures to be runnable fully on 6.9:

```       
          add_step(Procedures::Content::Prepare)
          add_step(Procedures::Content::Switchover)
          add_step(Procedures::Content::PrepareAbort)
          add_step(Procedures::Content::MigrationStats)
          add_step(Procedures::Content::MigrationReset)
```

The pulpcore services are started by foreman maintain, then the procedure runs, then the pulpcore services are stopped.

I'm going to remove `add_step_with_context` for all of those commands except `Procedures::Pulp::Remove`.

ianballou:
@upadhyeammit how does this solution look to you in the context of my latest comments? Also that ruby 2.0.0 test failure looks unrelated.

upadhyeammit:
@ianballou this looks good to me. My only question is by default 6.9 has packages providing pulpcore-api service? Because on my 6.9 machine I dont see service available?

jlsherrill:
@upadhyeammit they should be available/installed but not enabled.  You don't see them at all under:

```systemctl list-unit-files | grep pulpcore``` ?

upadhyeammit (approved):
LGTM!

PR #450, opened 2021-02-24
Fixes #31951 - f-m online backup to skip /var/lib/pulp/cache

pmoravec:

### DIFF
--- a/definitions/features/pulp2.rb
+++ b/definitions/features/pulp2.rb
@@ -34,4 +34,10 @@ class Features::Pulp < ForemanMaintain::Feature
       '/etc/default/pulp_workers'
     ]
   end
+
+  def exclude_from_backup
+    # Exclude /var/lib/pulp/katello-export and /var/lib/pulp/cache
+    # since the tar is run from /var/lib/pulp, list subdir paths only
+    ['katello-export', 'cache']
+  end
 end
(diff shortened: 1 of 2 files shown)


upadhyeammit:
Thanks for patience!
Could you move both exclude directories in pulp2 feature and call it in the procedure? something like below,

In file : foreman_maintain/definitions/features/pulp2.rb
~~~
def exclude_for_backup
  ['var/lib/pulp/katello-export', 'var/lib/pulp/cache']
end
~~~

In file: definitions/procedures/backup/pulp.rb
~~~
:exclude => feature(:pulp2).exclude_for_backup,
~~~

upadhyeammit (changes requested):
Added small change, request to check!
thank you!

pmoravec:
Thanks for the feedback, good point the change is pulp-2 related only. PR updated.

definitions/features/pulp2.rb
```diff
@@ -34,4 +34,8 @@ def config_files
       '/etc/default/pulp_workers'
     ]
   end
+
+  def exclude_for_backup
+    ['var/lib/pulp/katello-export', 'var/lib/pulp/cache']
```
upadhyeammit, line 39: The tar command is being run from `/var/lib/pulp` and hence exclude never worked. Request to change method like below,
~~~
def exclude_from_backup
  # Exclude /var/lib/pulp/katello-export and /var/lib/pulp/cache
  %w[katello-export cache]
end
~~~

You will also need to change `feature(:pulp2).exclude_for_backup` to `feature(:pulp2).exclude_from_backup`

upadhyeammit (changes requested):

pmoravec:
Sorry for the typos :)

upadhyeammit (approved):

PR #453, opened 2021-03-08
Fixes #32033 - use mongo command instead of client.command

upadhyeammit:


vsedmik (approved):
Patch tested downstream on snap 16 (with rubygem-foreman_maintain-0.7.6-1.el7sat.noarch) and works good. :+1:

PR #457, opened 2021-03-09
Fixes #32059 - content migration-reset fails with undef error

ianballou:
`enable_and_start_services` and `disable_and_stop_services` were only defined for the Prepare class, not the MigrationReset class.  This PR fixes that by making these methods available to all content classes.


jjeffers:
Test this and it resolved the issue.

johnpmitsch (approved):

PR #460, opened 2021-03-24
Fixes #32198 - Switchover needs to proxy yum and deb too

ianballou:
Foreman maintain was never told to proxy Pulp 3 content for content types other than file. This fixes that.


jlsherrill (approved):

ianballou:
@upadhyeammit can we get a merge on this?

PR #463, opened 2021-03-25
Fixes #32211 - Add qdpidd and qdrouterd to Katello services

jturel:
Katello 4.0 inherited the pulp2 functionality of the katello-agent so it now requires qpidd+qdrouterd totally independent of pulp2!

I found this on 0.7.x and would like this backported to that line (Foreman 2.4)

### DIFF
--- a/definitions/features/katello.rb
+++ b/definitions/features/katello.rb
@@ -16,9 +16,14 @@ class Features::Katello < ForemanMaintain::Feature
   end
 
   def services
-    [
-      system_service('elasticsearch', 30)
-    ]
+    if feature(:pulp2)
+      []
+    else
+      [
+        system_service('qpidd', 10),
+        system_service('qdrouterd', 10)
+      ]
+    end
   end
 
   # rubocop:disable  Metrics/MethodLength


definitions/features/katello.rb
```diff
@@ -17,7 +17,8 @@ def current_version
 
   def services
     [
-      system_service('elasticsearch', 30)
```
jturel, line 20: Do we actually support foreman-maintain back to a version of Katello that had elasticsearch?

definitions/features/katello.rb
```diff
@@ -17,7 +17,8 @@ def current_version
...
+      system_service('qpidd', 10),
```
jturel, line 20: I know that the `foreman-maintain service` command is affected by this - but is there anything else I should consider?

definitions/features/katello.rb
ehelms: I think this is safe to drop now.

definitions/features/katello.rb
upadhyeammit: The `qpidd` and `qdrouterd` services are mapped to `pulp2` feature as per current implementation: https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/pulp2.rb#L15 
The `pulp2` feature is available when `katello` is less than 4.0. 
Considering the fact that foreman-maintain package is not tightly coupled with Foreman/Katello/Satellite releases we should avoid mapping same service(s) twice in different features(pulp2, katello). 
So I suggest to conditionally map `qpidd` service in Katello if `pulp2` feature is not available.

definitions/features/katello.rb
upadhyeammit: Yes, safe to drop, not expecting anyone on 6.1 version in downstream.

definitions/features/katello.rb
jturel: @upadhyeammit updated!

upadhyeammit (approved):

PR #464, opened 2021-03-30
Checks for katello version prior to 4.0 for pulp-server.

jjeffers:


definitions/features/pulp2.rb
```diff
@@ -5,7 +5,9 @@ class Features::Pulp < ForemanMaintain::Feature
     label :pulp2
 
     confine do
-      find_package('pulp-server')
+      if check_max_version('katello', '4.0')
```
jlsherrill, line 8: wouldn't this be true for 4.0? 

```
      def check_max_version(name, maximal_version)
        check_version(name) do |current_version|
          version(maximal_version) >= current_version
        end
      end

```
should it be 3.18?

jjeffers: Good point!

definitions/features/pulp2.rb
```diff
@@ -5,7 +5,9 @@ class Features::Pulp < ForemanMaintain::Feature
     label :pulp2
 
     confine do
-      find_package('pulp-server')
+      if check_max_version('katello-common', '3.18')
+        find_package('pulp-server')
+      end
```
ekohl, line 10: Isn't this the same?
```suggestion
      check_max_version('katello-common', '3.18') && find_package('pulp-server')
```

Or to make the final diff easier to read:
```suggestion
      find_package('pulp-server') && check_max_version('katello-common', '3.18')
```

jjeffers: I like it.

definitions/features/pulp2.rb
```diff
@@ -5,7 +5,7 @@ class Features::Pulp < ForemanMaintain::Feature
     label :pulp2
 
     confine do
-      find_package('pulp-server')
+      find_package('pulp-server') && check_max_version('katello-common', '3.18')
```
ekohl, line 8: I wonder how this works with `katello-common-3.18.1`. Sadly `check_max_version` has no API documentation and no tests.

jjeffers: We could add a test or two here for that scenario. It's a good catch.

jjeffers: It doesn't appear that check_max_version will work here.

`check_max_version('katello-common', '3.18')` returns false if the current `katello-common` version is `3.18.1`, etc.

ekohl:
I think this makes sense now but since I suggested this, I think it'd be good if someone else did a further review.

jlsherrill (approved):

test/lib/system_helpers_test.rb
```diff
@@ -14,5 +14,51 @@ module ForemanMaintain
...
+      it 'returns true if current package version is greater than supplied version' do
+        MicroSystem.any_instance.stubs(:package_version).returns(
+          MicroSystem::Version.new('4.0.1')
+        )
+        assert system.check_min_version('katello', '4.0')
+      end
+    end
+
+    describe '.check_max_version version ' do
+      it 'returns true if package version is greater current package version' do
```
upadhyeammit, line 42: it 'returns true if package version is greater than current package version' do

upadhyeammit, line 42:
```suggestion
      it 'returns true if package version is greater than current package version' do
```

test/lib/system_helpers_test.rb
```diff
@@ -14,5 +14,51 @@ module ForemanMaintain
...
+      end
+
+      it 'returns true if package version is equal to current package version with no minor' do
+        MicroSystem.any_instance.stubs(:package_version).returns(
+          MicroSystem::Version.new('3.18.1')
+        )
+        assert system.check_max_version('katello', '3.18.1')
+      end
+
+      it 'returns false if package version is greater current package version with no minor' do
```
upadhyeammit, line 56:
```suggestion
      it 'returns false if package version is greater than current package version with no minor' do
```

upadhyeammit (changes requested):

upadhyeammit (approved):

PR #468, opened 2021-04-22
Fixes #32386 - Don't regenerate activemq queues on Capsule

upadhyeammit:


vsedmik (approved):
We have tested the fix on Satellite 6.9.1 snap 2 and 6.8.6 snap 1.1 and works well, the restore succeeded.

PR #474, opened 2021-05-24
Fixes #30323 - Use pgsql scl data_dir only for EL7

upadhyeammit:


upadhyeammit:
@evgeni request to review, thank you!

evgeni:
Thanks Amit!

This works as expected!

I have a design-y question tho: Currently you check for `os_major_release == '7'` (and do the same in #475), which *technically* would also match on Debian 7 (even if that's super old and not supported) and others. I wonder if it would make sense to have a helper shortcut somewhere that is essentially `is_el7?` which checks that the OS family is RedHat (so CentOS, RHEL and friends) and major is 7.

We do something very similar in foreman-installer: https://github.com/theforeman/foreman-installer/blob/3cbb7bd7201654a491a927cbb3d6867f50523c99/hooks/boot/01-kafo-hook-extensions.rb#L92-L94

upadhyeammit:
> Thanks Amit!
> 
> This works as expected!
> 
> I have a design-y question tho: Currently you check for `os_major_release == '7'` (and do the same in #475), which _technically_ would also match on Debian 7 (even if that's super old and not supported) and others. I wonder if it would make sense to have a helper shortcut somewhere that is essentially `is_el7?` which checks that the OS family is RedHat (so CentOS, RHEL and friends) and major is 7.
> 
> We do something very similar in foreman-installer: https://github.com/theforeman/foreman-installer/blob/3cbb7bd7201654a491a927cbb3d6867f50523c99/hooks/boot/01-kafo-hook-extensions.rb#L92-L94

Done, request to review. Thank you!

evgeni (approved):
I didn't *run* the new code, but it read exactly as I have envisioned it :)

PR #475, opened 2021-05-25
Fixes #32633 - handle scl and nonscl package names dynamically

upadhyeammit:
To restrict the number of executions, different portions of the code employ package installed conditions.
This update eliminates the requirement for explicit package names for plugins by introducing helper methods to generate package names depending on OS versions EL7(scl or non scl), EL8, and Debian.

### DIFF
--- a/definitions/features/foreman_cockpit.rb
+++ b/definitions/features/foreman_cockpit.rb
@@ -3,7 +3,7 @@ class Features::ForemanCockpit < ForemanMaintain::Feature
     label :foreman_cockpit
 
     confine do
-      server? && find_package('tfm-rubygem-foreman_remote_execution-cockpit')
+      server? && find_scl_or_nonscl_package('rubygem-foreman_remote_execution-cockpit')
     end
   end
 
--- a/definitions/features/foreman_cockpit.rb
+++ b/definitions/features/foreman_cockpit.rb
@@ -3,7 +3,7 @@ class Features::ForemanCockpit < ForemanMaintain::Feature
     label :foreman_cockpit
 
     confine do
-      server? && plugin_package_name('remote_execution-cockpit', 'foreman')
+      server? && find_package(foreman_plugin_name('foreman_remote_execution-cockpit'))
     end
   end
 
--- a/definitions/features/foreman_cockpit.rb
+++ b/definitions/features/foreman_cockpit.rb
@@ -3,7 +3,7 @@ class Features::ForemanCockpit < ForemanMaintain::Feature
     label :foreman_cockpit
 
     confine do
-      server? && plugin_package(:cockpit)
+      server? && plugin_package_name('remote_execution-cockpit', 'foreman')
     end
   end
 
--- a/definitions/features/foreman_cockpit.rb
+++ b/definitions/features/foreman_cockpit.rb
@@ -3,7 +3,7 @@ class Features::ForemanCockpit < ForemanMaintain::Feature
     label :foreman_cockpit
 
     confine do
-      server? && find_scl_or_nonscl_package('rubygem-foreman_remote_execution-cockpit')
+      server? && plugin_package(:cockpit)
     end
   end
 
--- a/definitions/features/foreman_openscap.rb
+++ b/definitions/features/foreman_openscap.rb
@@ -3,7 +3,8 @@ class Features::ForemanOpenscap < ForemanMaintain::Feature
     label :foreman_openscap
 
     confine do
-      check_min_version('tfm-rubygem-foreman_openscap', '0.5.3')
+      package_name = scl_or_nonscl_package('rubygem-foreman_openscap')
+      check_min_version(package_name, '0.5.3')
     end
   end
 
--- a/definitions/features/foreman_openscap.rb
+++ b/definitions/features/foreman_openscap.rb
@@ -3,8 +3,7 @@ class Features::ForemanOpenscap < ForemanMaintain::Feature
     label :foreman_openscap
 
     confine do
-      package_name = scl_or_nonscl_package('rubygem-foreman_openscap')
-      check_min_version(package_name, '0.5.3')
+      check_min_version(plugin_package(:openscap), '0.5.3')
     end
   end
 
--- a/definitions/features/foreman_openscap.rb
+++ b/definitions/features/foreman_openscap.rb
@@ -3,7 +3,7 @@ class Features::ForemanOpenscap < ForemanMaintain::Feature
     label :foreman_openscap
 
     confine do
-      check_min_version(plugin_package(:openscap), '0.5.3')
+      plugin_package_name('openscap', 'foreman')
     end
   end
 
--- a/definitions/features/foreman_openscap.rb
+++ b/definitions/features/foreman_openscap.rb
@@ -3,7 +3,7 @@ class Features::ForemanOpenscap < ForemanMaintain::Feature
     label :foreman_openscap
 
     confine do
-      plugin_package_name('openscap', 'foreman')
+      find_package(foreman_plugin_name('foreman_openscap'))
     end
   end
 
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -23,7 +23,7 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     label :foreman_tasks
 
     confine do
-      plugin_package_name('tasks', 'foreman')
+      find_package(foreman_plugin_name('foreman-tasks'))
     end
   end
 
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -23,8 +23,9 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     label :foreman_tasks
 
     confine do
+      package_name = scl_or_nonscl_package('rubygem-foreman-tasks')
       check_min_version('ruby193-rubygem-foreman-tasks', '0.6') ||
-        check_min_version('tfm-rubygem-foreman-tasks', '0.7')
+        check_min_version(package_name, '0.7')
     end
   end
 
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -23,8 +23,7 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     label :foreman_tasks
 
     confine do
-      check_min_version('ruby193-rubygem-foreman-tasks', '0.6') ||
-        check_min_version(plugin_package(:tasks), '0.7')
+      plugin_package_name('tasks', 'foreman')
     end
   end
 
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -23,9 +23,8 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     label :foreman_tasks
 
     confine do
-      package_name = scl_or_nonscl_package('rubygem-foreman-tasks')
       check_min_version('ruby193-rubygem-foreman-tasks', '0.6') ||
-        check_min_version(package_name, '0.7')
+        check_min_version(plugin_package(:tasks), '0.7')
     end
   end
 
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -7,7 +7,7 @@ class Features::Hammer < ForemanMaintain::Feature
     label :hammer
     confine do
       # FIXME: How does this run on proxy?
-      find_scl_or_nonscl_package('rubygem-hammer_cli')
+      hammer_package
     end
   end
 
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -6,7 +6,6 @@ class Features::Hammer < ForemanMaintain::Feature
   metadata do
     label :hammer
     confine do
-      # FIXME: How does this run on proxy?
       find_package(hammer_package)
     end
   end
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -7,7 +7,7 @@ class Features::Hammer < ForemanMaintain::Feature
     label :hammer
     confine do
       # FIXME: How does this run on proxy?
-      find_package('rubygem-hammer_cli') || find_package('tfm-rubygem-hammer_cli')
+      find_scl_or_nonscl_package('rubygem-hammer_cli')
     end
   end
 
--- a/definitions/features/hammer.rb
+++ b/definitions/features/hammer.rb
@@ -7,7 +7,7 @@ class Features::Hammer < ForemanMaintain::Feature
     label :hammer
     confine do
       # FIXME: How does this run on proxy?
-      hammer_package
+      find_package(hammer_package)
     end
   end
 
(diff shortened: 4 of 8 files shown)


upadhyeammit:
@evgeni request to review, thank you!

evgeni:
Looking at the code, this looks like it would work.

But again, design question: Right now you define a helper `scl_or_nonscl_package` (and a `find_…`), which is very EL7 focused and also sort-of a lie. We do use non-SCL packages on EL7 too. Or other SCLs (Redis, PostgreSQL, MongoDB). Even tho no of these are touched in this PR (or even part of the code).

My proposal would be instead have helpers of the kind `plugin_package(plugin_name)`  and `hammer_plugin_package(plugin_name)`, which will yield the correct package for the current OS.

Like `plugin_package('discovery') would yield `rubygem-foreman_discovery` on EL8, `tfm-rubygem-foreman_discovery` on EL7 and `ruby-foreman-discovery` on Debian. See the acceptance tests of `puppet-foreman` for an example: https://github.com/theforeman/puppet-foreman/blob/b193a1465404fb867a8330a783de09d7ce7e793a/spec/classes/cli_plugins_spec.rb#L12-L18

This would make it explicit *what* kind of package we're requesting and not spill other package names with prefixes that are probably wrong for them.

definitions/features/foreman_tasks.rb
```diff
@@ -23,8 +23,9 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     label :foreman_tasks
 
     confine do
+      package_name = scl_or_nonscl_package('rubygem-foreman-tasks')
       check_min_version('ruby193-rubygem-foreman-tasks', '0.6') ||
-        check_min_version('tfm-rubygem-foreman-tasks', '0.7')
+        check_min_version(package_name, '0.7')
```
evgeni, line 28: Unrelated, but should this be "check_plugin_version"?

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -217,6 +208,37 @@ def el8?
...
+        package_names = \
+          { cockpit: 'rubygem-foreman_remote_execution-cockpit',
+            docker: 'rubygem-foreman_docker',
+            smart_proxy_dynflow_core: 'rubygem-smart_proxy_dynflow_core',
+            openscap: 'rubygem-foreman_openscap',
+            tasks: 'rubygem-foreman-tasks' }
+        if el7?
+          "tfm-#{package_names.fetch(name)}"
+        elsif el8? || debian?
+          package_names.fetch(name)
```
ekohl, line 222: The Debian bit is not correct. Debian uses `ruby-` and dashes since underscores are illegal.

So in the installer we have (check `$plugin_prefix`):
https://github.com/theforeman/puppet-foreman/blob/1734ec1774fb09fd3c4f82646cd54deb3651d14d/manifests/params.pp#L88-L104

Then we transform it here:
https://github.com/theforeman/puppet-foreman/blob/1734ec1774fb09fd3c4f82646cd54deb3651d14d/manifests/plugin.pp#L25
https://github.com/theforeman/puppet-foreman/blob/1734ec1774fb09fd3c4f82646cd54deb3651d14d/manifests/plugin.pp#L33-L40

upadhyeammit:
Hello @ekohl and @evgeni,

Added changes as per the references. With this I also removed version checks for packages because upon checking those versions those were mostly of 6.1 or 6.2 Satellite.

Request to review, thank you!

definitions/features/foreman_cockpit.rb
```diff
@@ -3,7 +3,7 @@ class Features::ForemanCockpit < ForemanMaintain::Feature
     label :foreman_cockpit
 
     confine do
-      server? && find_package('tfm-rubygem-foreman_remote_execution-cockpit')
+      server? && plugin_package_name('remote_execution-cockpit', 'foreman')
```
evgeni, line 6: same, `plugin_package_name` doesn't verify it's installed

definitions/features/foreman_openscap.rb
```diff
@@ -3,7 +3,7 @@ class Features::ForemanOpenscap < ForemanMaintain::Feature
     label :foreman_openscap
 
     confine do
-      check_min_version('tfm-rubygem-foreman_openscap', '0.5.3')
+      plugin_package_name('openscap', 'foreman')
```
evgeni, line 6: same, `plugin_package_name` doesn't verify it's installed

definitions/features/foreman_tasks.rb
```diff
@@ -23,8 +23,7 @@ class Features::ForemanTasks < ForemanMaintain::Feature
     label :foreman_tasks
 
     confine do
-      check_min_version('ruby193-rubygem-foreman-tasks', '0.6') ||
-        check_min_version('tfm-rubygem-foreman-tasks', '0.7')
+      plugin_package_name('tasks', 'foreman')
```
evgeni, line 26: same, `plugin_package_name` doesn't verify it's installed

definitions/features/hammer.rb
```diff
@@ -7,7 +7,7 @@ class Features::Hammer < ForemanMaintain::Feature
     label :hammer
     confine do
       # FIXME: How does this run on proxy?
-      find_package('rubygem-hammer_cli') || find_package('tfm-rubygem-hammer_cli')
+      hammer_package
```
evgeni, line 10: same, `hammer_package` doesn't verify it's installed

definitions/procedures/foreman_docker/remove_foreman_docker.rb
```diff
@@ -6,11 +6,12 @@ class RemoveForemanDocker < ForemanMaintain::Procedure
     end
 
     def docker_package
-      'tfm-rubygem-foreman_docker'
+      plugin_package_name('docker', 'foreman')
```
evgeni, line 9: same, `plugin_package_name` doesn't verify it's installed

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -217,6 +208,60 @@ def el8?
...
+          'tfm-rubygem-foreman_'
+        elsif el8?
+          'rubygem-foreman_'
+        elsif debian?
+          'ruby-foreman-'
+        end
+      end
+
+      def smart_proxy_plugin_prefix
+        if el7?
```
evgeni, line 223: the proxy was SCL'ed in Foreman 2.0, before that, the packages were also `rubygem-smart_proxy…`

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -217,6 +208,60 @@ def el8?
       def el_major_version
         return os_facts['os']['release']['major'] if el?
       end
+
+      def foreman_plugin_prefix
```
evgeni, line 212: I wonder if the following is more "plugable"

```ruby
def ruby_prefix(scl=true)
  if el7? && scl
    'tfm-rubygem-'
  elsif el7? || el8?
    'rubygem-'
  elsif debian?
    'ruby-'
  end
end

def foreman_plugin_name(plugin)
  if debian?
    plugin.tr!('_', '-')
  end
  ruby_prefix + plugin
end

def proxy_plugin_name(plugin)
  if debian?
    plugin.tr!('_', '-')
    proxy_plugin_prefix = 'smart-proxy-'
  else
    proxy_plugin_prefix = 'smart_proxy_'
  end
  scl = foreman_version >= '2.0'
  ruby_prefix(scl) + proxy_plugin_prefix + plugin
end
```

I intentionally left out the 'foreman' out of `foreman_plugin_name`, as we have at least one plugin that is called `foreman-tasks` (and not `foreman_tasks`)

Not insisting on the above, just giving ideas.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -217,6 +208,60 @@ def el8?
...
+        if el7?
+          'tfm-rubygem-foreman_'
```
evgeni, line 214: As mentioned in the proposal above, we have plugins that use `foreman-tasks` notation. And of course there is also katello, which doesn't use the foreman prefix at all.

evgeni (changes requested):

lib/foreman_maintain/concerns/system_helpers.rb
evgeni: Of course, if we don't need to support pre-2.0 versions, that's nice ;)

ekohl (commented):
Just a note that for RPMs there are provides:
https://github.com/theforeman/foreman-packaging/blob/2a9cbbd5baba8d4db837d1e608b32dcff6fc9bf9/packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec#L42

We have the same thing for Proxy plugins:

https://github.com/theforeman/foreman-packaging/blob/2a9cbbd5baba8d4db837d1e608b32dcff6fc9bf9/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec#L46

These are the same for both EL7 and EL8. If you can check for those, it will be more consistent and probably less code. These do not exist for Debian packages.

evgeni:
> Just a note that for RPMs there are provides:
> https://github.com/theforeman/foreman-packaging/blob/2a9cbbd5baba8d4db837d1e608b32dcff6fc9bf9/packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec#L42
> 
> We have the same thing for Proxy plugins:
> 
> https://github.com/theforeman/foreman-packaging/blob/2a9cbbd5baba8d4db837d1e608b32dcff6fc9bf9/packages/plugins/rubygem-smart_proxy_ansible/rubygem-smart_proxy_ansible.spec#L46
> 
> These are the same for both EL7 and EL8. If you can check for those, it will be more consistent and probably less code. These do not exist for Debian packages.

Is there a way to ask `rpm` if a `provides` is installed? Currently, the code uses `rpm -q` and that doesn't work for provides.

ekohl:
https://unix.stackexchange.com/questions/242092/how-to-query-which-rpm-package-provides-a-specific-version suggests `rpm -q --whatprovides` works. For example:
```console
# rpm -q --whatprovides foreman-plugin-puppetdb
tfm-rubygem-puppetdb_foreman-5.0.0-3.fm2_5.el7.noarch
# echo $?
0
# rpm -q --whatprovides foreman-plugin-doesnotexist
no package provides foreman-plugin-doesnotexist
# echo $?
1
```

upadhyeammit:
> https://unix.stackexchange.com/questions/242092/how-to-query-which-rpm-package-provides-a-specific-version suggests `rpm -q --whatprovides` works. For example:
> 
> ```
> # rpm -q --whatprovides foreman-plugin-puppetdb
> tfm-rubygem-puppetdb_foreman-5.0.0-3.fm2_5.el7.noarch
> # echo $?
> 0
> # rpm -q --whatprovides foreman-plugin-doesnotexist
> no package provides foreman-plugin-doesnotexist
> # echo $?
> 1
> ```

Hello @ekohl 

After taking path of using `--whatprovides` I can see there are at least these combination of `Provides` mentioned in plugins, accordingly I feel approach suggested by the @evgeni is simpler?

~~~
foreman-plugin-dhcp-browser
foreman-%{plugin_name}
foreman-%{plugin_name} = %{version}
foreman-proxy-plugin-%{plugin_name} = %{version}
Provides the ability to boot a host into a rescue system
rubygem(%{gem_name}) = %{version}
%{?scl_prefix}rubygem(%{gem_name}) = %{version}
%{?scl_prefix}rubygem(%{gem_name}) = %{version}-%{release}
~~~

ekohl:
For now I think @evgeni's solution is the one I'd go with. I've opened https://projects.theforeman.org/issues/33209 for a better long term solution.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -217,6 +208,53 @@ def el8?
...
+      end
+
+      def hammer_package
+        if el7?
+          'tfm-rubygem-hammer_cli'
+        elsif el8?
+          'rubygem-hammer_cli'
+        elsif debian?
+          'rubygem-hammer-cli'
+        end
```
evgeni, line 247:
```suggestion
        if debian?
          hammer_prefix = 'hammer-cli'
        else
          hammer_prefix = 'hammer_cli'
        end
        ruby_prefix + hammer_prefix
```

(the original code used `rubygem` for Debian, which is wrong, and then I was like "wait, we have a helper for that!")

evgeni (changes requested):
last nitpick!

evgeni:
I'm happy!

@ekohl, you?

evgeni (approved):

definitions/features/foreman_openscap.rb
```diff
@@ -3,7 +3,7 @@ class Features::ForemanOpenscap < ForemanMaintain::Feature
     label :foreman_openscap
 
     confine do
-      check_min_version('tfm-rubygem-foreman_openscap', '0.5.3')
```
ekohl, line 6: Looks like 0.5.3 was released in January 2016. I'm guessing that's why it's no longer relevant to check for the version.

definitions/features/hammer.rb
```diff
@@ -7,7 +7,7 @@ class Features::Hammer < ForemanMaintain::Feature
     label :hammer
     confine do
       # FIXME: How does this run on proxy?
```
ekohl, line 9: We can answer this: it can run on a proxy, but doesn't by default. Probably safe to remove.

definitions/procedures/foreman_docker/remove_foreman_docker.rb
```diff
@@ -6,11 +6,12 @@ class RemoveForemanDocker < ForemanMaintain::Procedure
     end
 
     def docker_package
-      'tfm-rubygem-foreman_docker'
+      find_package(foreman_plugin_name('foreman_docker'))
```
ekohl, line 9: Doesn't this return `nil` if it's not found? There's also already a `return unless` on line 13 so I'd think this is sufficient:
```suggestion
      foreman_plugin_name('foreman_docker')
```

definitions/procedures/foreman_docker/remove_foreman_docker.rb
```diff
@@ -6,11 +6,12 @@ class RemoveForemanDocker < ForemanMaintain::Procedure
...
     end
 
     def run
       return unless execute?("rpm -q #{docker_package}")
```
ekohl, line 13: Oh, and this should probably use the package manager's `installed?` method instead.

definitions/procedures/foreman_docker/remove_foreman_docker.rb
evgeni: Good catch

definitions/features/foreman_openscap.rb
upadhyeammit: Yes, that's what I thought

definitions/procedures/foreman_docker/remove_foreman_docker.rb
upadhyeammit: removed this and moved condition to confine block

upadhyeammit:
@ekohl request to check, thank you!

definitions/procedures/foreman_docker/remove_foreman_docker.rb
```diff
@@ -3,16 +3,14 @@ class RemoveForemanDocker < ForemanMaintain::Procedure
     metadata do
       advanced_run false
       description 'Drop foreman_docker plugin'
-    end
-
-    def docker_package
-      'tfm-rubygem-foreman_docker'
+      confine do
+        find_package(foreman_plugin_name('foreman_docker'))
+      end
```
ekohl, line 8: Oh, this looks even nicer to combine.

ekohl (approved):

PR #476, opened 2021-05-31
Fixes #32672 - Add helpers for operating system's version facts

upadhyeammit:


upadhyeammit:
@evgeni request to review thank you!

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -192,6 +192,31 @@ def check_version(name)
           yield current_version
         end
       end
+
+      def all_system_facts
+        facter = ForemanMaintain::Utils::Facter.path
+        @all_system_facts ||= JSON.parse(execute("#{facter} -j"))
```
evgeni, line 198: Given we only look at `os` facts, isn't it cheaper to call `facter -j os` and call this `os_facts` or something?

evgeni:
```
# time facter -j os
…
real	0m0.264s
user	0m0.216s
sys	0m0.048s

# time facter -j
…
real	0m0.392s
user	0m0.260s
sys	0m0.123s
```

evgeni:
One minor comment, but otherwise makes total sense!

lib/foreman_maintain/concerns/system_helpers.rb
upadhyeammit: Resolved. I thought maybe all facts can help in future however I agree that if needed we can change it later.

evgeni (approved):

PR #481, opened 2021-06-10
Refs #32424 - exclude ansible-runner packages from check

upadhyeammit:


upadhyeammit:
I am merging this one considering if we can exclude packages with None as vendor then these packages are also not having any vendor.

PR #482, opened 2021-06-10
Fixes #32778 - Add support for Pulp 3 backup and restore

upadhyeammit:


definitions/procedures/backup/pulp.rb
```diff
@@ -26,11 +29,34 @@ def run
...
 
+    def current_pulp_feature
+      if feature(:pulp2) && feature(:pulpcore)
+        return choose_pulp2_pulp3
+      end
+
+      feature(:pulp2) || feature(:pulpcore)
+    end
+
+    def choose_pulp2_pulp3
```
upadhyeammit, line 40: @jlsherrill Like to know if this makes sense? if system will be in state where user will need to decide which pulp version to backup?

vsedmik: I believe the user is advised to make a backup just before the migration. I have some additional questions: When is the backup of content migrated to pulp3 expected to be used for restore, after upgrade? What is the benefit? (Maybe a backup of a very long and successful migration?) What about backups taken during the migration?

upadhyeammit:
@jlsherrill @jjeffers request to review, thank you!

definitions/features/pulp2.rb
```diff
@@ -5,7 +5,9 @@ class Features::Pulp < ForemanMaintain::Feature
     label :pulp2
 
     confine do
-      find_package('pulp-server') && !check_min_version('katello-common', '4.0')
+      !check_min_version('katello-common', '4.0') &&
+        ForemanMaintain::Utils::Service::Systemd.new('pulp_resource_manager', 0).exist? &&
```
upadhyeammit, line 9: @jlsherrill the pulp2 should be available if service is enabled? This ensures once we complete the migration to pulpcore only pulpcore is available, specifically on 6.9.

definitions/features/pulpcore_database.rb
```diff
@@ -7,7 +7,7 @@ class Features::PulpcoreDatabase < ForemanMaintain::Feature
     label :pulpcore_database
 
     confine do
-      file_nonzero?(PULPCORE_DB_CONFIG)
+      file_nonzero?(PULPCORE_DB_CONFIG) && feature(:pulpcore)
```
upadhyeammit, line 10: @jlsherrill Similar to pulp2, the pulpcore_database should be available only when pulpcore is available? This is to avoid detection of feature when its not being used, and avoid taking backup of pulpcore_db

definitions/procedures/backup/pulp.rb
upadhyeammit: > @jlsherrill Like to know if this makes sense? if system will be in state where user will need to decide which pulp version to backup?

I have removed this choose_pulp2_pulp3 method as per latest changes in pr. As per the latest changes I want to achieve,
1. If 6.9 in freshly installed we should not take backup of pulpcore as that wont help and does not make sense.
2. Prior to migration of pulpcore we should have backup of pulp2.
3. After migration to pulpcore on 6.9, we would want user to go to 6.10 with backup. If user takes backup at this time, it should be of pulpcore.

Request to review, thank you!

definitions/features/pulp2.rb
jlsherrill: +1 this makes sense

definitions/features/pulpcore_database.rb
jlsherrill: +1 makes sense

jlsherrill:
hopefully going to test this today

definitions/features/pulpcore_database.rb
evgeni: As noted yesterday, I think this is not a good idea.

We have scenarios where people will have a Katello running with Pulp2-backed repositories, but Pulp3 is installed (and disabled) for the migration.
I read the code that in this case, the pulpcore db (which is "empty" as in: tables are there, they are fully migrated, but there is no user data) won't be backed up.
If a user will restore such a backup, they end up with no pulpcore DB at all, and then can't do the p2-to-p3 migration due to the missing db (the installer can recreate it just fine, but that's not part of the migration process by default).

jlsherrill: i thought the restore would run the installer at some point which would deploy an empty (migrated) pulp3 db.

However you bring up a good point, if this is ever run by someone running katello 3.18 i think this logic falls down, as we need to handle both pulp2 and pulp3 dbs

upadhyeammit: @jlsherrill @evgeni handled this situation, request to check

upadhyeammit:
@evgeni @jlsherrill I think tests can be written in simpler or better way, however I am unable to come out with any other way. Request to review, thank you!
@vsedmik this looks complete now, request to repull the changes and test. Thank You!

vsedmik:
Looks good, no errors spotted for online, offline, snapshot, incremental online, incremental offline (all with and without pulp) for downstream Sat 6.10, 6.9, 6.8 for backups taken by in-snap f-m (6.9, 6.8) and backups taken by this PR. Content and pulp dir (in appropriate cases) was successfully restored. :+1:

jlsherrill (approved):
ACK from me, did some light testing and worked well!

evgeni (approved):

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -2,7 +2,7 @@ module ForemanMaintain
   module Concerns
     module BaseDatabase
       def data_dir
-        if el7? && check_min_version('foreman', '2.0')
+        if el7? && check_min_version('foreman-proxy', '2.0')
```
upadhyeammit, line 5: I think instead checking version of foreman-proxy is good?

lib/foreman_maintain/concerns/pulp_common.rb
```diff
@@ -1,7 +1,7 @@
 module ForemanMaintain
   module Concerns
     module PulpCommon
-      def data_dir
+      def pulp_data_dir
```
upadhyeammit, line 4: I changed this to pulp_data_dir as this was being picked up while doing db backup of pulp2/pulpcore

lib/foreman_maintain/concerns/base_database.rb
upadhyeammit: Latest change checks for `rh-postgresql12-postgresql-server-syspaths` instead of `foreman`/`foreman-proxy`

ehelms:
@upadhyeammit Based on the approvals, this looks like you can squash and merge whenever you are ready

vsedmik:
Also tested for Capsules (same scenarios as for Satellite commented above) without errors. LGTM :+1:

upadhyeammit:
Thank you @evgeni @jlsherrill @vsedmik for reviews! Merging this now!

PR #483, opened 2021-06-14
Fixes #32777 - remove pulp 2to3 migration plugin and tables

jjeffers:
This PR changes the pulp2 remove command so that tables and migrations are removed. It is a known issue with django2 (and dbshell included in django-admin) that problem with SQL are not reflected in the error status of the pulpcore-manager dbshell command. (This is fixed in django3.)

### DIFF
--- a/definitions/procedures/pulp/remove.rb
+++ b/definitions/procedures/pulp/remove.rb
@@ -2,6 +2,7 @@ module Procedures::Pulp
   class Remove < ForemanMaintain::Procedure
     metadata do
       description 'Remove pulp2'
+      param :assumeyes, 'Do not ask for confirmation', :default => false
 
       confine do
         check_min_version('katello-common', '4.0')
@@ -43,7 +44,8 @@ module Procedures::Pulp
                     python-pulp-python-common python-pulp-repoauth python-pulp-rpm-common
                     python-pulp-streamer python-pymongo python-pymongo-gridfs python2-amqp
                     python2-billiard python2-celery python2-debpkgr python2-django python2-kombu
-                    python2-solv python2-vine pulp-katello pulp-maintenance]
+                    python2-solv python2-vine pulp-katello pulp-maintenance
+                    python3-pulp-2to3-migration]
 
       @installed_pulp_packages ||= possible.select { |pkg| find_package(pkg) }
       @installed_pulp_packages
@@ -54,9 +56,9 @@ module Procedures::Pulp
     end
 
     def ask_to_proceed(rm_cmds)
-      question = "\nWARNING: All pulp2 packages will be removed with the following commands:\n" \
-        "\n# rpm -e #{pulp_packages.join('  ')}" \
-        "\n# yum remove rh-mongodb34-*" \
+      question = "\nWARNING: All pulp2 packages will be removed with the following commands:\n"
+      question += "\n# rpm -e #{pulp_packages.join('  ')}" if pulp_packages.any?
+      question += "\n# yum remove rh-mongodb34-*" \
         "\n\nAll pulp2 data will be removed.\n"
       question += rm_cmds.collect { |cmd| "\n# #{cmd}" }.join
       question += "\n\nDo you want to proceed?"
@@ -66,11 +68,18 @@ module Procedures::Pulp
 
     def run
       rm_cmds = data_dir_removal_cmds
-      ask_to_proceed(rm_cmds) if rm_cmds.any? && !@assume_yes
 
-      remove_pulp
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+
+      ask_to_proceed(rm_cmds) unless assumeyes_val
+
+      remove_pulp if pulp_packages.any?
 
       remove_mongo
+
+      drop_migration_tables
+
+      drop_migrations
 
       delete_pulp_data(rm_cmds) if rm_cmds.any?
     end
@@ -86,6 +95,60 @@ module Procedures::Pulp
         packages_action(:remove, ['rh-mongodb34-*'], :assumeyes => true)
       end
     end
+
+    def drop_migration_tables
+      with_spinner('Dropping migration tables') do
+        feature(:pulpcore_database).psql(<<-SQL)
+          BEGIN;
+          DROP TABLE IF EXISTS pulp_2to3_migration_migrationplan,
+            pulp_2to3_migration_pulp2blob,
+            pulp_2to3_migration_pulp2content,
+            pulp_2to3_migration_pulp2debcomponent,
+            pulp_2to3_migration_pulp2debcomponentpackage,
+            pulp_2to3_migration_pulp2debpackage,
+            pulp_2to3_migration_pulp2debrelease,
+            pulp_2to3_migration_pulp2debreleasearchitecture,
+            pulp_2to3_migration_pulp2distribution,
+            pulp_2to3_migration_pulp2distributor,
+            pulp_2to3_migration_pulp2distributor_pulp2_repos,
+            pulp_2to3_migration_pulp2erratum,
+            pulp_2to3_migration_pulp2importer,
+            pulp_2to3_migration_pulp2iso,
+            pulp_2to3_migration_pulp2lazycatalog,
+            pulp_2to3_migration_pulp2manifest,
+            pulp_2to3_migration_pulp2manifestlist,
+            pulp_2to3_migration_pulp2modulemd,
+            pulp_2to3_migration_pulp2modulemddefaults,
+            pulp_2to3_migration_pulp2packagecategory,
+            pulp_2to3_migration_pulp2packageenvironment,
+            pulp_2to3_migration_pulp2packagegroup,
+            pulp_2to3_migration_pulp2packagelangpacks,
+            pulp_2to3_migration_pulp2repocontent,
+            pulp_2to3_migration_pulp2repository,
+            pulp_2to3_migration_pulp2rpm,
+            pulp_2to3_migration_pulp2srpm,
+            pulp_2to3_migration_pulp2tag,
+            pulp_2to3_migration_pulp2yumrepometadatafile,
+            pulp_2to3_migration_reposetup CASCADE;
+          DROP SEQUENCE IF EXISTS pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq;
+          END;
+        SQL
+      end
+    end
+
+    def drop_migrations
+      with_spinner('Dropping migrations') do
+        feature(:pulpcore_database).psql(<<-SQL)
+          BEGIN;
+          DELETE FROM django_migrations WHERE app = 'pulp_2to3_migration';
+          DELETE FROM auth_permission WHERE content_type_id in (SELECT id FROM django_content_type WHERE app_label = 'pulp_2to3_migration');
+          DELETE FROM django_admin_log WHERE content_type_id in (SELECT id FROM django_content_type WHERE app_label = 'pulp_2to3_migration');
+          DELETE FROM django_content_type WHERE app_label = 'pulp_2to3_migration';
+          END;
+        SQL
+      end
+    end
+    # rubocop:enable Metrics/BlockLength
 
     def delete_pulp_data(rm_cmds)
       with_spinner('Deleting pulp2 data directories') do |spinner|
--- a/lib/foreman_maintain/cli/content_command.rb
+++ b/lib/foreman_maintain/cli/content_command.rb
@@ -35,8 +35,11 @@ module ForemanMaintain
       end
 
       subcommand 'remove-pulp2', 'Remove pulp2 and mongodb packages and data' do
+        interactive_option(%w[assumeyes plaintext])
         def execute
-          run_scenarios_and_exit(Scenarios::Content::RemovePulp2.new)
+          run_scenarios_and_exit(
+            Scenarios::Content::RemovePulp2.new(:assumeyes => assumeyes?)
+          )
         end
       end
     end
--- a/definitions/scenarios/content.rb
+++ b/definitions/scenarios/content.rb
@@ -101,7 +101,12 @@ module ForemanMaintain::Scenarios
       metadata do
         label :content_remove_pulp2
         description 'Remove Pulp2 and mongodb packages and data'
+        param :assumeyes, 'Do not ask for confirmation'
         manual_detection
+      end
+
+      def set_context_mapping
+        context.map(:assumeyes, Procedures::Pulp::Remove => :assumeyes)
       end
 
       def compose


definitions/procedures/pulp/remove.rb
```diff
@@ -73,6 +75,53 @@ def remove_mongo
...
+          pulp_2to3_migration_pulp2repository,
+          pulp_2to3_migration_pulp2rpm,
+          pulp_2to3_migration_pulp2srpm,
+          pulp_2to3_migration_pulp2tag,
+          pulp_2to3_migration_pulp2yumrepometadatafile,
+          pulp_2to3_migration_reposetup CASCADE;
+        END;
+        DBCOMMAND
+
+        db_shell_cmd = 'PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager dbshell'
```
upadhyeammit, line 117: Can we add method(s) in PulpcoreDatabase to execute queries for pulpcoredb?

upadhyeammit (changes requested):

definitions/procedures/pulp/remove.rb
jjeffers: @upadhyeammit Great idea!

jturel:
I might be testing this incorrectly (PR doesn't have any pointers) but after running `foreman-maintain content remove-pulp2` I see all of the DB tables are still there, and one seemingly unnecessary package is left over:

```
[root@centos7-katello-3-18 foreman_maintain-0.7.10]# foreman-maintain content remove-pulp2
Running Remove Pulp2 and mongodb packages and data
Remove pulp2: 

WARNING: All pulp2 packages will be removed with the following commands:

# rpm -e pulp-deb-plugins  pulp-docker-plugins  pulp-puppet-plugins  pulp-puppet-tools  pulp-rpm-plugins  pulp-selinux  pulp-server  python-bson  python-crane  python-isodate  python-mongoengine  python-nectar  python-pulp-common  python-pulp-deb-common  python-pulp-docker-common  python-pulp-oid_validation  python-pulp-puppet-common  python-pulp-repoauth  python-pulp-rpm-common  python-pulp-streamer  python-pymongo  python-pymongo-gridfs  python2-amqp  python2-billiard  python2-celery  python2-debpkgr  python2-django  python2-kombu  python2-solv  python2-vine  pulp-katello  python3-pulp-2to3-migration
# yum remove rh-mongodb34-*

All pulp2 data will be removed.

# rm -rf /var/lib/pulp/published
# rm -rf /var/lib/pulp/uploads
# rm -rf /var/lib/mongodb/

Do you want to proceed?, [y(yes), q(quit)] y
| Removing pulp2 packages                                                       
- Removing mongo packages                                                       
| Dropping migration tables                                                     
/ Done deleting pulp2 data directories                                [OK]      
```

Katello is dependent on this RPM. Has it been dropped as a dependency in packaging?
```
[root@centos7-katello-3-18 foreman_maintain-0.7.10]# rpm -qa | grep migration
tfm-rubygem-pulp_2to3_migration_client-0.11.0-1.el7.noarch 
```

```
 public | pulp_2to3_migration_migrationplan                       | table    | pulp
 public | pulp_2to3_migration_pulp2blob                           | table    | pulp
 public | pulp_2to3_migration_pulp2content                        | table    | pulp
 public | pulp_2to3_migration_pulp2debcomponent                   | table    | pulp
 public | pulp_2to3_migration_pulp2debcomponentpackage            | table    | pulp
 public | pulp_2to3_migration_pulp2debpackage                     | table    | pulp
 public | pulp_2to3_migration_pulp2debrelease                     | table    | pulp
 public | pulp_2to3_migration_pulp2debreleasearchitecture         | table    | pulp
 public | pulp_2to3_migration_pulp2distribution                   | table    | pulp
 public | pulp_2to3_migration_pulp2distributor                    | table    | pulp
 public | pulp_2to3_migration_pulp2distributor_pulp2_repos        | table    | pulp
 public | pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq | sequence | pulp
 public | pulp_2to3_migration_pulp2erratum                        | table    | pulp
 public | pulp_2to3_migration_pulp2importer                       | table    | pulp
 public | pulp_2to3_migration_pulp2iso                            | table    | pulp
 public | pulp_2to3_migration_pulp2lazycatalog                    | table    | pulp
 public | pulp_2to3_migration_pulp2manifest                       | table    | pulp
 public | pulp_2to3_migration_pulp2manifestlist                   | table    | pulp
 public | pulp_2to3_migration_pulp2modulemd                       | table    | pulp
 public | pulp_2to3_migration_pulp2modulemddefaults               | table    | pulp
 public | pulp_2to3_migration_pulp2packagecategory                | table    | pulp
 public | pulp_2to3_migration_pulp2packageenvironment             | table    | pulp
 public | pulp_2to3_migration_pulp2packagegroup                   | table    | pulp
 public | pulp_2to3_migration_pulp2packagelangpacks               | table    | pulp
 public | pulp_2to3_migration_pulp2repocontent                    | table    | pulp
 public | pulp_2to3_migration_pulp2repository                     | table    | pulp
 public | pulp_2to3_migration_pulp2rpm                            | table    | pulp
 public | pulp_2to3_migration_pulp2srpm                           | table    | pulp
 public | pulp_2to3_migration_pulp2tag                            | table    | pulp
 public | pulp_2to3_migration_pulp2yumrepometadatafile            | table    | pulp
 public | pulp_2to3_migration_reposetup                           | table    | pulp
```

jlsherrill:
the client binding should still exist, thats expected.  Strange that the db tables are still there though.

ekohl (changes requested):
Is anyone planning on working on the installer? Otherwise it'll happily install the plugin again.

Another thing is that this removal is incomplete. You also need to drop the migrations it ran. From https://patrick.cloke.us/posts/2020/01/23/cleanly-removing-a-django-app-with-models/
```sql
# Remove Django migrations.
DELETE FROM `django_migrations` WHERE `app` = 'my_app';
# Get rid of permissions and content types.
DELETE FROM `auth_permission` WHERE `content_type_id` in (SELECT `id` FROM `django_content_type` WHERE `app_label` = 'my_app');
# Get rid of admin changes.
DELETE FROM `django_admin_log` WHERE `content_type_id` in (SELECT `id` FROM `django_content_type` WHERE `app_label` = 'my_app');
# Finally, delete the content type.
DELETE FROM `django_content_type` WHERE `app_label` = 'my_app';
```

jlsherrill:
@goosemania i think you mentioned we could leave those migrations alone, but do you think its worth cleaning up?

ekohl:
Oh good point. In https://github.com/theforeman/puppet-foreman_proxy_content/commit/b8a699ff9b3c2e3457c3165d33217d3657bee091 we did indeed drop `pulpcore::plugin::migration`.

goosemania:
> @goosemania i think you mentioned we could leave those migrations alone, but do you think its worth cleaning up?

It's cleaner to remove the bits from all the tables, I agree.
However, it's not important/critical for the migration plugin specifically because it has one-time use and will never be installed again.

We'll do similar to what @ekohl pointed to in the plugin removal command but it's with the mindset for re-installs of the plugins.

FWIW, `my_app` in the example above should be substituted with `pulp_2to3_migration` (underscores, not dashes), if you decide to cleanup everything. The queries look correct to me.

upadhyeammit:
@jjeffers codewise LGTM! request to fix rubocop errors.

ekohl (commented):
The code itself looks good. I think my only concern with this is that we're adding it to the Pulp 2 removal. In the upstream community people may have already ran this. Can they run the same procedure again?

upadhyeammit:
> The code itself looks good. I think my only concern with this is that we're adding it to the Pulp 2 removal. In the upstream community people may have already ran this. Can they run the same procedure again?

I dont see any problem to rerun this if any or some of the pulp2 resources are already removed as,
1. We are checking if packages are installed and then removing it using rpm -e
2. The mongo packages are being removed using `packages_action` which actually executes action using yum and if yum does not find package to remove then it does not fail.
3. The tables are being deleted if those exist as per the query.
4. The migrations deletion will only happen if it finds any records as per where clause.

jturel:
```
D, [2021-06-29 14:13:26+0000 #4310] DEBUG -- : Running command echo '        BEGIN;
        DROP TABLE IF EXISTS pulp_2to3_migration_migrationplan,
          pulp_2to3_migration_pulp2blob,
          pulp_2to3_migration_pulp2content,
          pulp_2to3_migration_pulp2debcomponent,
          pulp_2to3_migration_pulp2debcomponentpackage,
          pulp_2to3_migration_pulp2debpackage,
          pulp_2to3_migration_pulp2debrelease,
          pulp_2to3_migration_pulp2debreleasearchitecture,
          pulp_2to3_migration_pulp2distribution,
          pulp_2to3_migration_pulp2distributor,
          pulp_2to3_migration_pulp2distributor_pulp2_repos,
          pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq,
          pulp_2to3_migration_pulp2erratum,
          pulp_2to3_migration_pulp2importer,
          pulp_2to3_migration_pulp2iso,
          pulp_2to3_migration_pulp2lazycatalog,
          pulp_2to3_migration_pulp2manifest,
          pulp_2to3_migration_pulp2manifestlist,
          pulp_2to3_migration_pulp2modulemd,
          pulp_2to3_migration_pulp2modulemddefaults,
          pulp_2to3_migration_pulp2packagecategory,
          pulp_2to3_migration_pulp2packageenvironment,
          pulp_2to3_migration_pulp2packagegroup,
          pulp_2to3_migration_pulp2packagelangpacks,
          pulp_2to3_migration_pulp2repocontent,
          pulp_2to3_migration_pulp2repository,
          pulp_2to3_migration_pulp2rpm,
          pulp_2to3_migration_pulp2srpm,
          pulp_2to3_migration_pulp2tag,
          pulp_2to3_migration_pulp2yumrepometadatafile,
          pulp_2to3_migration_reposetup CASCADE;
        DELETE FROM `django_migrations` WHERE `app` = 'pulp_2to3_migration';
        DELETE FROM `auth_permission` WHERE `content_type_id` in (SELECT `id` FROM `django_content_type` WHERE `app_label` = 'pulp_2to3_migration');
        DELETE FROM `django_admin_log` WHERE `content_type_id` in (SELECT `id` FROM `django_content_type` WHERE `app_label` = 'pulp_2to3_migration');
        DELETE FROM `django_content_type` WHERE `app_label` = 'pulp_2to3_migration';
        END;
' | PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager dbshell with stdin nil
D, [2021-06-29 14:13:28+0000 #4310] DEBUG -- : output of the command:
 BEGIN
ERROR:  "pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq" is not a table
HINT:  Use DROP SEQUENCE to remove a sequence.
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `django_migrations` WHERE `app` = pulp_2to3_migr...
                    ^
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `auth_permission` WHERE `content_type_id` in (SE...
                    ^
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `django_admin_log` WHERE `content_type_id` in (S...
                    ^
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `django_content_type` WHERE `app_label` = pulp_2...
                    ^
ROLLBACK
D, [2021-06-29 14:13:28+0000 #4310] DEBUG -- : BEGIN
ERROR:  "pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq" is not a table
HINT:  Use DROP SEQUENCE to remove a sequence.
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `django_migrations` WHERE `app` = pulp_2to3_migr...
                    ^
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `auth_permission` WHERE `content_type_id` in (SE...
                    ^
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `django_admin_log` WHERE `content_type_id` in (S...
                    ^
ERROR:  syntax error at or near "`"
LINE 1: DELETE FROM `django_content_type` WHERE `app_label` = pulp_2...
                    ^
ROLLBACK
I, [2021-06-29 14:13:28+0000 #4310]  INFO -- : --- Execution step 'Remove pulp2' finished ---
I, [2021-06-29 14:13:28+0000 #4310]  INFO -- : === Scenario 'Remove Pulp2 and mongodb packages and data' finished ===
I, [2021-06-29 14:13:28+0000 #4310]  INFO -- : foreman-maintain command finished with 
```

ekohl:
I think backticks are MySQL-only.

jturel:
```
D, [2021-06-29 15:10:46+0000 #5690] DEBUG -- : Running command echo '        BEGIN;
        DROP TABLE IF EXISTS pulp_2to3_migration_migrationplan,
          pulp_2to3_migration_pulp2blob,
          pulp_2to3_migration_pulp2content,
          pulp_2to3_migration_pulp2debcomponent,
          pulp_2to3_migration_pulp2debcomponentpackage,
          pulp_2to3_migration_pulp2debpackage,
          pulp_2to3_migration_pulp2debrelease,
          pulp_2to3_migration_pulp2debreleasearchitecture,
          pulp_2to3_migration_pulp2distribution,
          pulp_2to3_migration_pulp2distributor,
          pulp_2to3_migration_pulp2distributor_pulp2_repos,
          pulp_2to3_migration_pulp2erratum,
          pulp_2to3_migration_pulp2importer,
          pulp_2to3_migration_pulp2iso,
          pulp_2to3_migration_pulp2lazycatalog,
          pulp_2to3_migration_pulp2manifest,
          pulp_2to3_migration_pulp2manifestlist,
          pulp_2to3_migration_pulp2modulemd,
          pulp_2to3_migration_pulp2modulemddefaults,
          pulp_2to3_migration_pulp2packagecategory,
          pulp_2to3_migration_pulp2packageenvironment,
          pulp_2to3_migration_pulp2packagegroup,
          pulp_2to3_migration_pulp2packagelangpacks,
          pulp_2to3_migration_pulp2repocontent,
          pulp_2to3_migration_pulp2repository,
          pulp_2to3_migration_pulp2rpm,
          pulp_2to3_migration_pulp2srpm,
          pulp_2to3_migration_pulp2tag,
          pulp_2to3_migration_pulp2yumrepometadatafile,
          pulp_2to3_migration_reposetup CASCADE;
        DROP SEQUENCE IF EXISTS pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq;
        DELETE FROM django_migrations WHERE app = "pulp_2to3_migration";
        DELETE FROM auth_permission WHERE content_type_id in (SELECT id FROM django_content_type WHERE app_label = 'pulp_2to3_migration');
        DELETE FROM django_admin_log WHERE content_type_id in (SELECT id FROM django_content_type WHERE app_label = 'pulp_2to3_migration');
        DELETE FROM django_content_type WHERE app_label = pulp_2to3_migration;
        END;
' | PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager dbshell with stdin nil
D, [2021-06-29 15:10:48+0000 #5690] DEBUG -- : output of the command:
 BEGIN
DROP TABLE
NOTICE:  sequence "pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq" does not exist, skipping
DROP SEQUENCE
ERROR:  column "pulp_2to3_migration" does not exist
LINE 1: DELETE FROM django_migrations WHERE app = "pulp_2to3_migrati...
                                                  ^
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ROLLBACK
D, [2021-06-29 15:10:48+0000 #5690] DEBUG -- : BEGIN
DROP TABLE
NOTICE:  sequence "pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq" does not exist, skipping
DROP SEQUENCE
ERROR:  column "pulp_2to3_migration" does not exist
LINE 1: DELETE FROM django_migrations WHERE app = "pulp_2to3_migrati...
                                                  ^
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ROLLBACK
I, [2021-06-29 15:10:48+0000 #5690]  INFO -- : --- Execution step 'Remove pulp2' finished ---
I, [2021-06-29 15:10:48+0000 #5690]  INFO -- : === Scenario 'Remove Pulp2 and mongodb packages and data' finished ===
I, [2021-06-29 15:10:48+0000 #5690]  INFO -- : foreman-maintain command finished with 
```

```
ulpcore=# \l django_migrations;
                       List of databases
 Name | Owner | Encoding | Collate | Ctype | Access privileges 
------+-------+----------+---------+-------+-------------------
```

jjeffers:
> ERROR:  column "pulp_2to3_migration" does not exist
> LINE 1: DELETE FROM django_migrations WHERE app = "pulp_2to3_migrati...

I haven't yet nailed down why dbshell throws this error, where using psql and the same syntax does not.

jturel:
> > ERROR:  column "pulp_2to3_migration" does not exist
> > LINE 1: DELETE FROM django_migrations WHERE app = "pulp_2to3_migrati...
> 
> I haven't yet nailed down why dbshell throws this error, where using psql and the same syntax does not.

single quotes maybe

jjeffers:
> > I haven't yet nailed down why dbshell throws this error, where using psql and the same syntax does not.
> 
> single quotes maybe

I've tried both single and doublequotes - but still the same issue. Still looking.

jturel:
> > > I haven't yet nailed down why dbshell throws this error, where using psql and the same syntax does not.
> > 
> > 
> > single quotes maybe
> 
> I've tried both single and doublequotes - but still the same issue. Still looking.

Interesting!

```
pulpcore=# select * from django_migrations where app = "pulp_2to3_migration";
ERROR:  column "pulp_2to3_migration" does not exist
LINE 1: select * from django_migrations where app = "pulp_2to3_migrat...
                                                    ^
```

```
pulpcore=# select * from django_migrations where app = 'pulp_2to3_migration';
 id  |         app         |                      name                       |            applied            
-----+---------------------+-------------------------------------------------+-------------------------------
  95 | pulp_2to3_migration | 0001_initial                                    | 2021-06-15 13:52:58.042614+00
  96 | pulp_2to3_migration | 0002_remove_pulp2distributor_pulp2_auto_publish | 2021-06-15 13:52:58.224397+00

```

definitions/procedures/pulp/remove.rb
```diff
@@ -85,6 +87,55 @@ def remove_mongo
...
+          pulp_2to3_migration_pulp2rpm,
+          pulp_2to3_migration_pulp2srpm,
+          pulp_2to3_migration_pulp2tag,
+          pulp_2to3_migration_pulp2yumrepometadatafile,
+          pulp_2to3_migration_reposetup CASCADE;
+        DROP SEQUENCE IF EXISTS pulp_2to3_migration_pulp2distributor_pulp2_repos_id_seq;
+        DELETE FROM django_migrations WHERE app = "pulp_2to3_migration";
+        DELETE FROM auth_permission WHERE content_type_id in (SELECT id FROM django_content_type WHERE app_label = 'pulp_2to3_migration');
+        DELETE FROM django_admin_log WHERE content_type_id in (SELECT id FROM django_content_type WHERE app_label = 'pulp_2to3_migration');
+        DELETE FROM django_content_type WHERE app_label = pulp_2to3_migration;
```
ekohl, line 129: Doesn't this line also need quotes?
```suggestion
        DELETE FROM django_content_type WHERE 'app_label = pulp_2to3_migration';
```

jjeffers: Quoting the condition does not work, at least not for sending STDIN to the `pulpcore-manager dbshell` command. Perhaps we need to rethink using dbshell? I'm open to suggestions.

definitions/features/pulpcore_database.rb
```diff
@@ -22,6 +23,12 @@ def services
     ]
   end
 
+  def execute_with_db_shell(command_input)
+    cmd = "echo '#{command_input}' | PULP_SETTINGS=#{PULPCORE_DB_CONFIG}"
```
jturel, line 27: What if you replace the single quotes in the echo part with double quotes?

ekohl: Maybe the helper should wrap [open3](https://docs.ruby-lang.org/en/2.0.0/Open3.html) (I'm linking to Ruby 2.0.0 since that's what we need to support) and use `capture2` or `capture3`. That allows you to pass in the SQL via stdin directly rather than via `echo`.

Another option
```suggestion
    cmd = "cat <<SQL | PULP_SETTINGS=#{PULPCORE_DB_CONFIG} PULPCORE_DB_SHELL_COMMAND\n"
    cmd += command_input
    cmd += "\nSQL"
```

jjeffers: @jturel the `echo` using single quotes was causing the problem, switching to a double quote works.

@ekohl at some point, should the existing helper functions be looked at for improvements as you suggested? I feel like a big refactor of the `execute` functions might be worth a separate PR.

ekohl: @jturel I still think using `cat` over `echo` is better because it avoids quoting altogether. It should look something like this in pure shell (without the added Ruby on top of it):
```bash
cat <<SQL | PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager dbshell
SELECT 1
SQL
```

jjeffers: @ekohl Let's try the `cat` approach.

jjeffers:
@upadhyeammit this change uses the pulpcore-manager command to issue SQL commands. Is there an advantage to just using the psql methods already present in foreman-maintain?

ekohl:
> @upadhyeammit this change uses the pulpcore-manager command to issue SQL commands. Is there an advantage to just using the psql methods already present in foreman-maintain?

Using `pulpcore-manager dbshell` means you don't need to parse the Django config nor the answers file and it's guaranteed to be in sync, whether it's a local DB or a remote DB. It's IMHO a really elegant way. Retrieving them from the answers file is something I would love to get rid of.

jjeffers:
> Using `pulpcore-manager dbshell` means you don't need to parse the Django config nor the answers file and it's guaranteed to be in sync, whether it's a local DB or a remote DB. It's IMHO a really elegant way. Retrieving them from the answers file is something I would love to get rid of.

Ok, thank you for that clarification. I've noticed a potential problem with using pulpcore-manager - errors in the SQL commands are not bubbled back up through as an error return code. Users wouldn't be aware there was a problem running this remove command unless they happened to look in the foreman-maintain log. The output of the foreman-maintain command doesn't give any indication that anything has gone wrong.

definitions/features/pulpcore_database.rb
jjeffers: @ekohl the `cat` strategy works pretty well.

ekohl: Glad to hear that.

ekohl:
> I've noticed a potential problem with using pulpcore-manager - errors in the SQL commands are not bubbled back up through as an error return code.

That is an ugly problem. Looking at the source (https://github.com/django/django/blob/main/django/core/management/commands/dbshell.py) you're right that only failures to start the shell return result in an exit code. I'm not sure how we can solve that. Perhaps it's something we should contribute to upstream Django, but we can't rely on that for a long time (given Pulp 3 is still on Django 2.2 LTS). That may be an effective blocker to use dbshell.

jjeffers:
> > I've noticed a potential problem with using pulpcore-manager - errors in the SQL commands are not bubbled back up through as an error return code.
> 
> That is an ugly problem. Looking at the source (https://github.com/django/django/blob/main/django/core/management/commands/dbshell.py) you're right that only failures to start the shell return result in an exit code. I'm not sure how we can solve that. Perhaps it's something we should contribute to upstream Django, but we can't rely on that for a long time (given Pulp 3 is still on Django 2.2 LTS). That may be an effective blocker to use dbshell.

Against dbshell going forward or in particular on this PR? @upadhyeammit thoughts on if we should continue to go down the dbshell road?

upadhyeammit:
> Against



> > > I've noticed a potential problem with using pulpcore-manager - errors in the SQL commands are not bubbled back up through as an error return code.
> > 
> > 
> > That is an ugly problem. Looking at the source (https://github.com/django/django/blob/main/django/core/management/commands/dbshell.py) you're right that only failures to start the shell return result in an exit code. I'm not sure how we can solve that. Perhaps it's something we should contribute to upstream Django, but we can't rely on that for a long time (given Pulp 3 is still on Django 2.2 LTS). That may be an effective blocker to use dbshell.
> 
> Against dbshell going forward or in particular on this PR? @upadhyeammit thoughts on if we should continue to go down the dbshell road?

Considering this is post upgrade removal procedure and the fact that dbshell wont be fixed soon I vote to not use dbshell from this pr itself. Better to use psql commands in foreman-maintain and add missing bits for pulpcore database if any to support it.

ekohl:
I agree with that. While I still like the concept of something built in that uses the right credentials automatically, the exit codes are more important.

jjeffers:
@jturel this is ready for a re-test.

definitions/features/pulpcore_database.rb
```diff
@@ -1,5 +1,6 @@
 class Features::PulpcoreDatabase < ForemanMaintain::Feature
   PULPCORE_DB_CONFIG = '/etc/pulp/settings.py'.freeze
+  PULPCORE_DB_SHELL_COMMAND = 'pulpcore-manager dbshell'.freeze
```
jturel, line 3: This isn't needed anymore :)

jturel:
This change seems functionally complete - I don't see any of the migration plugin tables left after running the command and the output looks good:

```
[root@centos7-katello-3-18 foreman_maintain-0.8.2]# foreman-maintain content remove-pulp2                                               
Running Remove Pulp2 and mongodb packages and data                                                                                      
Remove pulp2:                                                                                                                           
                                                                                                                                        
WARNING: All pulp2 packages will be removed with the following commands:                                                                
                                                                                                                                        
# rpm -e pulp-deb-plugins  pulp-docker-plugins  pulp-puppet-plugins  pulp-puppet-tools  pulp-rpm-plugins  pulp-selinux  pulp-server  python-bson  python-crane  python-isodate  python-mongoengine  python-nectar  python-pulp-common  python-pulp-deb-common  python-pulp-docke
r-common  python-pulp-oid_validation  python-pulp-puppet-common  python-pulp-repoauth  python-pulp-rpm-common  python-pulp-streamer  python-pymongo  python-pymongo-gridfs  python2-amqp  python2-billiard  python2-celery  python2-debpkgr  python2-django  python2-kombu  pyth
on2-solv  python2-vine  pulp-katello  python3-pulp-2to3-migration                                                                       
# yum remove rh-mongodb34-*                                                                                                             
                                                                                                                                        
All pulp2 data will be removed.                                                                                                         
                                                                                                                                        
# rm -rf /var/lib/pulp/published                                                                                                        
# rm -rf /var/lib/pulp/uploads                                                                                                          
# rm -rf /var/lib/mongodb/                                                                                                              
                                                                                                                                        
Do you want to proceed?, [y(yes), q(quit)] y                        
| Removing pulp2 packages                                                                                                               
| Removing mongo packages                                                                                                               
\ Dropping migration tables                                                                                                             
- Dropping migrations                                                                                                                   
| Done deleting pulp2 data directories                                [OK]                                                              

[root@centos7-katello-3-18 foreman_maintain-0.8.2]# rpm -qa | grep pulp | sort                                                          
pulp-client-1.0-1.noarch                                            
pulpcore-selinux-1.2.4-1.el7.x86_64                                 
python3-pulp-certguard-1.1.0-1.el7.noarch                           
python3-pulp-container-2.2.2-1.el7.noarch                           
python3-pulpcore-3.9.1-2.el7.noarch                                 
python3-pulp-deb-2.9.2-1.el7.noarch                                 
python3-pulp-file-1.5.0-1.el7.noarch                                
python3-pulp-rpm-3.10.0-1.el7.noarch                                
tfm-rubygem-pulp_2to3_migration_client-0.11.0-1.el7.noarch                                                                              
tfm-rubygem-pulp_ansible_client-0.6.0-1.el7.noarch                                                                                      
tfm-rubygem-pulp_certguard_client-1.0.3-1.el7.noarch                                                                                    
tfm-rubygem-pulp_container_client-2.2.0-1.el7.noarch                                                                                    
tfm-rubygem-pulpcore_client-3.9.0-1.el7.noarch                      
tfm-rubygem-pulp_deb_client-2.9.1-1.el7.noarch                      
tfm-rubygem-pulp_file_client-1.5.0-1.el7.noarch                                                                                         
tfm-rubygem-pulp_rpm_client-3.10.0-1.el7.noarch                                                                                         
tfm-rubygem-smart_proxy_pulp-2.1.0-3.fm2_2.el7.noarch 
```

jturel (approved):

jjeffers:
@ekohl @upadhyeammit can this be re-reviewed? Thanks!

ekohl (approved):
I do wonder why this can't be a single transaction but otherwise :+1:

jjeffers:
> I do wonder why this can't be a single transaction but otherwise

IIRC I split them while trying to debug the dbshell failures. There is no reason for them to be split anymore.

upadhyeammit:
Hello @jjeffers maybe I should have thought these things earlier, apologies for that. I have below questions,
* It calls `ask_to_proceed` if there are pulp2 directories on system, and hence this fails the `remove_pulp` method if there aren't any packages pulp2 packages.
* The `ask_to_proceed` has question about whole pulp2 stuff removal but the current implementation only asks the question when pulp2 directories exist?
* I am not aware about any other method or document about wiping out pulp2 stuff, if there is any then is there a chance to have partial residue? if yes then `ask_to_proceed` and whole question should be dynamic at least for directories and pulp2 packages?

jjeffers:
> Hello @jjeffers maybe I should have thought these things earlier, apologies for that. I have below questions,

All good questions, but I am wondering if the need to ask to proceed is more for information purposes - that is, to warn the user that destructive processes are about to start. Instead of making the question dependent on the state of the packages, dirs, etc, perhaps we could simply ask to proceed ALWAYS, simply skipping uneeded steps, UNLESS the user indicated a "--yes" option?

I think @evgeni requested this "do not ask" option to ease scripting, so that might be desirable.

evgeni:
Yeah, an --assume-yes as other options have it would be awesome for automation.

jturel:
I'll re-test when this has full acks from maintainers

jjeffers:
@upadhyeammit I modified the PR to observe the --assumeyes option, which is `false` by default. If set, the command will not ask to proceed. All commands will be tried, and if there is nothing to do it will not generate a failure. Otherwise, the command will still ask to proceed regardless if any pulp data dirs exist, packages are present, etc.

lib/foreman_maintain/cli/content_command.rb
```diff
@@ -35,8 +35,11 @@ def execute
       end
 
       subcommand 'remove-pulp2', 'Remove pulp2 and mongodb packages and data' do
+        interactive_option(['assumeyes'])
```
upadhyeammit, line 38: There is addition of `plaintext` flag recently, though we dont have lots of spinner I feel it will help to get automation friendly output if required?

definitions/procedures/pulp/remove.rb
```diff
@@ -64,18 +65,25 @@ def ask_to_proceed(rm_cmds)
 
     def run
       rm_cmds = data_dir_removal_cmds
-      ask_to_proceed(rm_cmds) if rm_cmds.any?
+
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+
+      ask_to_proceed(rm_cmds) unless assumeyes_val
 
       remove_pulp
```
upadhyeammit, line 73: Better to not run `remove_pulp` itself if `pulp_packages` is empty? because the question wont include `rpm -e` command we will still display removal message?

upadhyeammit (changes requested):
Few inline comments, otherwise looks good!

upadhyeammit:
@jjeffers should we add `squid` package in list or separate pr for that : https://projects.theforeman.org/issues/32945

definitions/procedures/pulp/remove.rb
upadhyeammit: @jjeffers I wished to have `if pulp_packages.any?` here and remove it from `remove_pulp`,

~~~
   def remove_pulp
      with_spinner('Removing pulp2 packages') do
        execute!("rpm -e #{pulp_packages.join('  ')}")
      end
    end
~~~

upadhyeammit (approved):

PR #484, opened 2021-06-15
Fixes #32818 - run pulp3 switchover as part of 6.10 upgrade

jlsherrill:


jlsherrill:
I still need to do some testing on this

upadhyeammit:
@jlsherrill with this I think its also good to add
1. Pulp2 removal as part of post upgrade?
2. An extra `automatic` switch added to upgrade command on Capsule?

jlsherrill:
we had only planned to do pulp2 removal automatically as part of capsule upgrade, but i'm not 100% sure why we don't do it on the satellite too.  I'll discuss it

jlsherrill:
i discussed with @ehelms  a bit and we came to the conclusion to just do it automatically in both places.  One question is however, if we add this 'option' to automatically remove, it will be on the upgrade command forever, correct?

jlsherrill:
testing is going well so far, a couple of notes:

1) i think we need to run prep-6.10-upgrade as part of this (its included here), but probably want to skip it if its already been run.  I think i can simply check the ownership of /var/lib/pulp/content and skip if its owned by pulp.  Will work on that

2) This requires this commit: https://github.com/Katello/katello/commit/78f56831ad217a446a8034c4c151afdf7ea3f0f7  to allow the migration/swtichover tasks to be run even if the switchover is complete.  (thats linked int he same bz)

jlsherrill:
This is now ready for review/testing.  Tested myself alongside https://github.com/theforeman/foreman_maintain/pull/486 and got clean upgrades.

definitions/scenarios/upgrade_to_satellite_6_10.rb
```diff
@@ -72,6 +87,7 @@ class PostMigrations < Abstract
     def compose
       add_step(Procedures::Service::Start.new)
       add_steps(find_procedures(:post_migrations))
+      add_step(Procedures::Pulp::Remove.new(:assume_yes => true))
```
jlsherrill, line 90: Here's the removal.  I'm still kinda 50/50 on this.  This completely wipes their pulp2 content dir.  So imagine a situation where something went really wrong, the user likely has db backups, but may not have a full /var/lib/pulp/ backup, they would be hosed.  

Curious thoughts about just printing a statement telling the user to run the command once they verifeid full operation?   To me that seems much safer.

definitions/scenarios/upgrade_to_satellite_6_10.rb
```diff
@@ -54,7 +54,22 @@ def set_context_mapping
...
+
     def compose
+      check_var_lib_pulp
+
+      add_step(Procedures::Service::Start)
+      add_step(Procedures::Service::Enable.
+          new(:only => Features::Pulpcore.pulpcore_migration_services))
+      add_step(Procedures::Service::Start.
+          new(:only => Features::Pulpcore.pulpcore_migration_services))
+      add_step(Procedures::Content::Switchover.new(:skip_deb => true))
```
jlsherrill, line 72: @upadhyeammit if this step fails, i need the entire operation to stop.  I don't think that happens today.  Is there an easy way to accomplish this?   Would moving this portion to the pre-migrations compose accomplish this?

definitions/scenarios/upgrade_to_satellite_6_10.rb
jlsherrill: @ehelms curious your thoughts as well

definitions/scenarios/upgrade_to_satellite_6_10.rb
upadhyeammit: Can you try adding `run_strategy :fail_fast` for `Migrations` class in `upgrade_metadata`?

upadhyeammit: Can you try adding `run_strategy :fail_fast` for `Migrations` class in `upgrade_metadata`?

definitions/scenarios/upgrade_to_satellite_6_10.rb
upadhyeammit: @jlsherrill agree, better to let users take a call to wipe out pulp2 content dir once migration is successful.

upadhyeammit:
> i discussed with @ehelms a bit and we came to the conclusion to just do it automatically in both places. One question is however, if we add this 'option' to automatically remove, it will be on the upgrade command forever, correct?

Yes it will be there forever. If we conditionally want to load the option for `6.10` then the current upgrade command is utilizing `ForemanMaintain::UpgradeRunner` to validate the version. My thought is we have a method which adds `automatic` or similar option if we see `ForemanMaintain::UpgradeRunner.available_targets.include?('6.10')`. We call the method for `run` subcommand? When we strictly check for `6.10` version `automatic` option wont be available when `6.9.z` upgrade is in progress, which I think is correct.

definitions/scenarios/upgrade_to_satellite_6_10.rb
jlsherrill: Yes, okay so i've updated to print instructions, here's the output:

```
SNIP
Running Checks after upgrading to Satellite 6.10
Check if only installed assets are present on the system: 
/ Checking for presence of non-original assets...                     [OK]      
Print pulp 2 removal instructions: 
======================================================
Migration of content from Pulp 2 to Pulp3 is complete
After verifying accessibility of content from clients,
it is highly recommend to run 'foreman-maintain content remove-pulp2'
This will remove Pulp 2, MongoDB, and all pulp2 content in /var/lib/pulp/content/
======================================================                [OK]


Upgrade finished.

```

definitions/scenarios/upgrade_to_satellite_6_10.rb
jlsherrill: that worked, thank you!

jlsherrill:
This is now ready, since we're NOT doing the pulp2-removal here, there's no need to do a --automatic flag on this pr.  I will work on capsule upgrades next, which i guess will need that flag

definitions/procedures/pulp/print_remove_instructions.rb
```diff
@@ -0,0 +1,16 @@
...
+  class PrintRemoveInstructions < ForemanMaintain::Procedure
+    metadata do
+      description 'Print pulp 2 removal instructions'
+    end
+
+    def run
+      puts '======================================================'
+      puts 'Migration of content from Pulp 2 to Pulp3 is complete'
+      puts 'After verifying accessibility of content from clients,'
+      puts 'it is highly recommend to run "foreman-maintain content remove-pulp2"'
```
upadhyeammit, line 11: `it is recommended` / `it is strongly recommended` feels better?

definitions/scenarios/upgrade_to_satellite_6_10.rb
upadhyeammit: Instead of passing `:skip_deb` as `true` I think we better keep it true by default something like,
~~~
param :skip_deb, 'Do not run debian options in installer.', :flag => true, :default => true
~~~

upadhyeammit (changes requested):
Added few comments, otherwise looks good!

definitions/scenarios/upgrade_to_satellite_6_10.rb
jlsherrill: defining it true by default means that the normal switchover command needs to set it to false, which seems not ideal?   curious why it should be true by default?

upadhyeammit: ah yes for upstream switchover we need to have the deb options and those are not required for this specific scenario! Agree to keep current implementation.

jlsherrill:
Anyone testing this may want to apply https://github.com/Katello/katello/pull/9412  locally

jturel:
I believe I got the expected output:

```
Switch support for certain content from Pulp 2 to Pulp 3: 
Performing final content migration before switching content
Starting task.
2021-07-15 20:03:30 +0000: Content migration starting. 
Performing a check to verify everything that is needed has been migrated

Switching specified content over to pulp 3
Content Switchover completed successfully
Re-running the installer to switch specified content over to pulp3    [OK]
```

jturel:
When I re-ran it after the switchover was already completed:

```
Switch support for certain content from Pulp 2 to Pulp 3: 
Performing final content migration before switching content
Switchover is already complete, skipping migration work.
Performing a check to verify everything that is needed has been migrated

Switching specified content over to pulp 3
Switchover is already complete, skipping switchover task.
Re-running the installer to switch specified content over to pulp3    [OK]
```

(looks good)

jturel:
And when it's all said and done:

```
Print pulp 2 removal instructions: 
======================================================
Migration of content from Pulp 2 to Pulp3 is complete
After verifying accessibility of content from clients,
it is strongly recommend to run "foreman-maintain content remove-pulp2"
This will remove Pulp 2, MongoDB, and all pulp2 content in /var/lib/pulp/content/
======================================================                [OK]


Upgrade finished.
```

jturel (approved):

upadhyeammit (approved):

upadhyeammit:
@jturel thanks for testing! Merging now!

PR #486, opened 2021-06-23
Fixes #32870 - refresh features after 6.10 upgrade

jlsherrill:


jlsherrill:
@upadhyeammit i haven't actually tested this, but i'm curious if i'm on the right track

jlsherrill:
This seems to have worked!  

```
Running Procedures after migrating to Satellite 6.10
Refresh detected features:                                            [OK]
Start applicable services: 

rh-mongodb34-mongod, rh-redis5-redis, postgresql, pulpcore-api, pulpcore-content, pulpcore-resource-manager, rh-redis5-redis, tomcat, dynflow-sidekiq@orchestrator, foreman, httpd, puppetserver, dynflow-sidekiq@worker-1, dynflow-sidekiq@worker-hosts-queue-1, foreman-proxy
```

jlsherrill:
tested in conjunction with https://github.com/theforeman/foreman_maintain/pull/484

upadhyeammit (commented):
LGTM! We also need to refresh features in Capsule upgrade scenario as services are changing on that too?

ehelms:
Yes, we would need to refresh them on a Capsule upgrade as well. @upadhyeammit do you think we could merge this and do a follow up with Capsule or if you could take this PR and do the Capsule update? @jlsherrill is away for a few days which would otherwise delay this change.

upadhyeammit (approved):

PR #487, opened 2021-06-29
Fixes #32909 - use keys.openpgp.org instead of dead sks-keyservers

evgeni:


evgeni:
look ma, it only took me 5 tries to get a simple ruby 2.0 to work!

upadhyeammit (approved):

PR #488, opened 2021-06-29
Refs #32909 - don't install default ruby with rvm

evgeni:
when you call rvms installer with `--ruby`, it installs the default ruby
(3.0 at the time of writing), but we don't need that as we will
explicitly install 2.0 two lines later

saves some setup time :)


evgeni:
@upadhyeammit this was something that confused me when I did the other pr… "why does it install ruby 3.0 first!?"

upadhyeammit:
> @upadhyeammit this was something that confused me when I did the other pr… "why does it install ruby 3.0 first!?"

Thank you, some more minutes saved!

upadhyeammit (approved):

PR #490, opened 2021-06-30
Fixes #30094 - drop unwanted interactive options from packages cmd

upadhyeammit:
The package sub-commands by default use to have 'assumeyes' and
'whitelist' options. These options are not required for all
sub-commands.

### DIFF
--- a/lib/foreman_maintain/cli/packages_command.rb
+++ b/lib/foreman_maintain/cli/packages_command.rb
@@ -2,28 +2,28 @@ module ForemanMaintain
   module Cli
     class PackagesCommand < Base
       subcommand 'lock', 'Prevent packages from automatic update' do
-        interactive_option
+        interactive_option(['assumeyes'])
         def execute
           run_scenarios_and_exit(Scenarios::Packages::Lock.new)
         end
       end
 
       subcommand 'unlock', 'Enable packages for automatic update' do
-        interactive_option
+        interactive_option(['assumeyes'])
         def execute
           run_scenarios_and_exit(Scenarios::Packages::Unlock.new)
         end
       end
 
       subcommand 'status', 'Check if packages are protected against update' do
-        interactive_option
+        interactive_option(['assumeyes'])
         def execute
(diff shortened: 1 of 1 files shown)


lib/foreman_maintain/cli/packages_command.rb
```diff
@@ -51,7 +51,7 @@ def execute
       end
 
       subcommand 'is-locked', 'Check if update of packages is allowed' do
-        interactive_option
+        interactive_option([])
```
ekohl, line 54: The installer uses this:
https://github.com/theforeman/foreman-installer/blob/bc575c2d662dad4779623b89d24ca7750dcec3bd/hooks/pre_commit/09-version_locking.rb#L2

I'd say we should patch the installer (https://github.com/theforeman/foreman-installer/pull/701) but considering downstream shipping perhaps this should also accept `assumeyes`?

rplevka: for the record - [the issue](https://projects.theforeman.org/issues/30094) linked to this PR also expects the `assumeyes` to be available for this subcommand (see "expected result")

upadhyeammit: @ekohl technically `assumeyes` is useless for `is-locked` command as there isn't any user interaction happening in the process. But if we drop the `--assumeyes` from the foremain-installer then it will just increase the one more `if else` block in maintain because there will be older versions of `foreman-installer` which are still using the `--assumeyes`. Hence I propose not to remove `--assumeyes` from the installer, and I readd it for the `is-locked`.

ekohl: Isn't `--assumeyes` also useless in older versions? So the installer could safely drop it without creating an `else` clause? The behavior should be the same, right?

upadhyeammit: Yes its useless for older versions. If you can cherry pick the installer changes in older versions too then we can keep this as is and change will only happen to installer.

PR #493, opened 2021-07-06
Fixes #32946 - use pulpcore_resource_manager if unit file is exist

upadhyeammit:


upadhyeammit:
@ekohl @evgeni request to review, thank you!

definitions/features/service.rb
```diff
@@ -50,6 +50,11 @@ def filter_disabled_services!(action, service_list)
     service_list
   end
 
+  def unit_file_available?(name)
+    exit_status, = execute_with_status("systemctl list-unit-files --type=service | grep -w #{name}")
```
evgeni, line 54:
```suggestion
    exit_status, = execute_with_status("systemctl --no-legend --no-pager list-unit-files --type=service #{name} | grep -qw #{name}")
```

I think this is cleaner:
* `--no-legend` will suppress additional output from systemctl, which should not hurt, but you never know.
* `--no-pager` will disable piping through `less` and friends. `systemctl` *does* autodetect this, but I always prefer being explicit
* if you add the `#{name}` to the systemctl call, it will only list that one unit
* we don't care for the *output* of `grep`, so we can `-q` (quiet) it. and while we're at it, it's probably better to write `--word-regexp --quiet` instead.

And now after I said this, I wonder if it would make sense (not for the usecase at hand, but in future?) to actually store the whole list of unit files in f-maintain and match *all* services against that list?

ekohl: Do we even need `grep` here?
```console
$ systemctl --no-legend --no-pager list-unit-files --type=service doesnotexist
$ echo $?
1
$ systemctl --no-legend --no-pager list-unit-files --type=service systemd-resolved.service
systemd-resolved.service enabled enabled
$ echo $?
0
```

definitions/features/pulpcore.rb
```diff
@@ -34,10 +34,21 @@ def config_files
...
-      ForemanMaintain::Utils.system_service('pulpcore-resource-manager', 10)
+      ForemanMaintain::Utils.system_service('pulpcore-content', 10, :socket => 'pulpcore-content')
     ]
+    common_services + pulpcore_resource_manager_service
+  end
+
+  def pulpcore_resource_manager_service
+    # The pulpcore_resource_manager is only required on 3.14+ if new tasking system is being used
+    # The foreman-installer does not create unit file for this service,
+    # if old tasking system is being used
```
ekohl, line 47:
```suggestion
    # The pulpcore_resource_manager is only required on 3.14+ if the old tasking system is being used
    # The foreman-installer does not create unit file for this service,
    # if the new tasking system is being used
```

definitions/features/service.rb
upadhyeammit: @evgeni Thanks for description, changes are added accordingly. I like to know how keeping list of all unit files will help? As foreman-maintain cares about services which are defined under features.
@ekohl as we have discussed the systemctl does not return exit code 1 on failing to find the unit file on EL7 and Fedora 31. The Fedora 34 fixed this, however we are keeping the grep command for older version compatibility.

ekohl: Yep, given Fedora 31 doesn't have it, it's also safe to assume EL8 doesn't so it'll be a long time before we can count on it.

evgeni (approved):

ekohl (approved):

PR #346, opened 2020-05-19
Fixes #25943 - Added --disable-spinner flag

patilsuraj767:
Added new flag in interactive_option. "**--disable-spinner**" will stop spinner from activating and will also not print the carriage return char "^M" if the output is redirected to the file. 

`# foreman-maintain backup offline  -y --skip-pulp-content --disable-spinner /tmp/backup/ > ~/after_spinner_fix/logs_without_spinner.txt`

Below is output of logs_without_spinner.txt

https://gist.github.com/patilsuraj767/0b31584c032b1eb7277b091aac07d869

PR #494, opened 2021-07-06
Fixes #25943 - Add support in spinner for plaintext output

ianballou:
This PR adds an option to foreman-maintain backup to output in plaintext.  Simply add --plaintext as an argument to try it out.

Based on this original PR: https://github.com/theforeman/foreman_maintain/pull/346

### DIFF
--- a/lib/foreman_maintain/reporter/cli_reporter.rb
+++ b/lib/foreman_maintain/reporter/cli_reporter.rb
@@ -30,7 +30,7 @@ module ForemanMaintain
         end
 
         def activate
-          @mutex.synchronize { @active = true }
+          @mutex.synchronize { @active = true } unless @reporter.plaintext?
           spin
         end
 
@@ -62,8 +62,11 @@ module ForemanMaintain
 
         def print_current_line
           @reporter.clear_line
-          line = "#{@spinner_chars[@spinner_index]} #{@current_line}"
-          @reporter.print(line)
+          if @reporter.plaintext?
+            @reporter.print(@current_line.to_s)
+          else
+            @reporter.print("#{@spinner_chars[@spinner_index]} #{@current_line}")
+          end
         end
       end
 
@@ -75,6 +78,7 @@ module ForemanMaintain
         @stdin = stdin
         options.validate_options!(:assumeyes)
         @assumeyes = options.fetch(:assumeyes, false)
+        @plaintext = options.fetch(:plaintext, false)
         @hl = HighLine.new(@stdin, @stdout)
         @max_length = 80
         @line_char = '-'
@@ -162,7 +166,11 @@ module ForemanMaintain
       end
 
       def clear_line
-        print "\r" + ' ' * @max_length + "\r"
+        if plaintext?
+          print "\n"
+        else
+          print "\r" + ' ' * @max_length + "\r"
+        end
       end
 
       def assumeyes?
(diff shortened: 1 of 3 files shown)


lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -271,7 +276,11 @@ def status_label(status)
                     :already_run => { :label => '[ALREADY RUN]', :color => :yellow },
                     :warning => { :label => '[WARNING]', :color => :yellow } }
         properties = mapping[status]
-        @hl.color(properties[:label], properties[:color], :bold)
+        if @plaintext
+          @hl.color(properties[:label], :bold)
```
upadhyeammit, line 280: I think plaintext can still have color?

upadhyeammit (changes requested):
I think its better to unset `@spinner_chars` if we have `@plaintext` set or modify `print_current_line` as below?
~~~
        def print_current_line
          @reporter.clear_line
          if @reporter.plaintext?
            line = "#{@current_line}"
          else
            line = "#{@spinner_chars[@spinner_index]} #{@current_line}"
          end
          @reporter.print(line)
        end
~~~

The current output adds the spinner_chars like,
~~~
| starting pulpcore-resource-manager
| starting rh-redis5-redis
Warning: postgresql.service changed on disk. Run 'systemctl daemon-reload' to reload units.
~~~

Removing these looks like,
~~~
starting pulpcore-resource-manager
starting rh-redis5-redis
Warning: postgresql.service changed on disk. Run 'systemctl daemon-reload' to reload units.
~~~

lib/foreman_maintain/reporter/cli_reporter.rb
ianballou: The issue with color is that, depending on how you view the log file, you might see the "extra" coloring characters.

upadhyeammit: > The issue with color is that, depending on how you view the log file, you might see the ANSI coloring characters instead of color. I hit that with VIM I think.

then better to keep it colorless!

ianballou: I realized the bold was causing troubles too so I removed it in my latest push.

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -62,8 +62,11 @@ def spin
 
         def print_current_line
           @reporter.clear_line
-          line = "#{@spinner_chars[@spinner_index]} #{@current_line}"
-          @reporter.print(line)
+          if @reporter.plaintext?
+            @reporter.print(@current_line.to_s)
+          else
+            @reporter.print("#{@spinner_chars[@spinner_index]} #{@current_line}")
```
upadhyeammit, line 68: Should we also call `.to_s` on `@current_line` in else block? I can't find any occurrences where `update` method called with non string param, however its safe to do string typecasting to avoid obvious issues?

upadhyeammit (changes requested):
Added one comment, otherwise LGTM!

lib/foreman_maintain/reporter/cli_reporter.rb
ianballou: @upadhyeammit RuboCop complaints: " W: Redundant use of Object#to_s in interpolation."

I think it's fine since the interpolation bit is calling to_s automatically in the background.  I had to change the above line because RuboCop didn't want me using string interpolation on a single variable.

upadhyeammit (approved):

PR #496, opened 2021-07-09
Refs #32946 - properly reference pulpcore_resource_manager_service

evgeni:
Otherwise foreman-maintain fails with:

    undefined local variable or method `pulpcore_resource_manager_service'
    for Feature Class pulpcore<Features::Pulpcore>:Class


evgeni:
@upadhyeammit could you please have a look, and if happy merge and release a 0.8.5 with that change asap?

upadhyeammit (approved):

PR #501, opened 2021-07-15
Fixes #32319 - skip external databases while filtering services

upadhyeammit:
The external databases can't be enabled or disabled hence
not considering pgsql databases while filtering disabled
services

### DIFF
--- a/definitions/features/service.rb
+++ b/definitions/features/service.rb
@@ -45,7 +45,7 @@ class Features::Service < ForemanMaintain::Feature
 
   def filter_disabled_services!(action, service_list)
     if %w[start stop restart status].include?(action)
-      service_list.select!(&:enabled?)
+      service_list.select! { |service| !service.respond_to?(:enabled?) || service.enabled? }
     end
     service_list
   end


upadhyeammit:
@ekohl while this works, I would love to know if this can be written in better way? Thanks!

evgeni:
I don't have a setup to test, but doesn't the else block mean the external DB *will* be added to the service list and thus considered for later operations?

upadhyeammit:
> I don't have a setup to test, but doesn't the else block mean the external DB _will_ be added to the service list and thus considered for later operations?

For start/stop/restart/enable/disable external dbs are skiped and only connection status is disaplyed,
~~~
# bin/foreman-maintain service stop
Warning: Stopping foreman.service, but it can still be activated by:
  foreman.socket
| stopping squid                                                                
postgresql (candlepin) is remote and is UP.

postgresql (foreman) is remote and is UP.

postgresql (pulpcore) is remote and is UP.
/ stopping rh-mongodb34-mongod                                                  
rh-mongodb34-mongod is remote and is UP.

# bin/foreman-maintain service enable --only rh-mongodb34-mongod
Running Enable Services
Enable applicable services: 

Enabling the following service(s):
rh-mongodb34-mongod
/ enabling rh-mongodb34-mongod                                                  
rh-mongodb34-mongod is remote and is UP. It can't be enabled.
/ All services enabled                                                [OK]      

~~~

evgeni:
Ah, cool, thanks for the explanation!

definitions/features/service.rb
```diff
@@ -44,8 +44,18 @@ def action_past_tense(action)
   end
 
   def filter_disabled_services!(action, service_list)
```
ekohl, line 46: Now that you no longer modify `service_list` in place, this no longer needs the `!`:
```suggestion
  def filter_disabled_services(action, service_list)
```

definitions/features/service.rb
```diff
@@ -44,8 +44,18 @@ def action_past_tense(action)
...
+        # Hence running the filteration on local services only
+        if service.respond_to?(:enabled?)
+          enabled_services << service if service.enabled?
+        else
+          enabled_services << service
+        end
+      end
+      return enabled_services
     end
     service_list
```
ekohl, line 60: This does change the behavior. If the action isn't in the included list, it now returns an empty list. A possible smaller diff that maintains the current behavior:
```ruby
   def filter_disabled_services!(action, service_list) 
    if %w[start stop restart status].include?(action)
      service_list.select! { |service| !service.respond_to?(:enabled?) || service.enabled? }
    end
    service_list
  end
```

upadhyeammit: Yes to avoid returning empty list I explicitly returned the `enabled_services` from if block. Your approach if far cleaner, thank you!

ekohl (approved):

PR #513, opened 2021-07-29
Fixes #33183 - adjusts the interactive command runner for nohup

jjeffers:
Example of problem before, running a migration command in the background:
```
Running Prepare content for Pulp 3
Prepare content for Pulp 3: 
                                                                      [FAIL]
Failed executing foreman-rake katello:pulp3_migration, exit status 256
Scenario [Prepare content for Pulp 3] failed.

  [content-prepare]

use --whitelist="content-prepare"
```

And after:
```
Running Prepare content for Pulp 3
Prepare content for Pulp 3: 
Starting task.

2021-07-29 15:59:30 +0000: Content migration starting. 

Starting task.

2021-07-29 15:59:30 +0000: Content migration starting. 
                             [OK]
```

### DIFF
--- a/lib/foreman_maintain/utils/command_runner.rb
+++ b/lib/foreman_maintain/utils/command_runner.rb
@@ -59,12 +59,15 @@ module ForemanMaintain
 
       private
 
+      # rubocop:disable Metrics/MethodLength
       def run_interactively
         # use tmp files to capture output and exit status of the command when
         # running interactively
         log_file = Tempfile.open('captured-output')
         exit_file = Tempfile.open('captured-exit-code')
-        Kernel.system("script -qc '#{full_command}; echo $? > #{exit_file.path}' #{log_file.path}")
+        Kernel.system(
+          "bash -c '#{full_command}; echo $? > #{exit_file.path}' | tee -i #{log_file.path}"
+        )
         File.open(log_file.path) { |f| @output = f.read }
         File.open(exit_file.path) do |f|
           exit_status = f.read.strip
@@ -78,6 +81,7 @@ module ForemanMaintain
         log_file.close
         exit_file.close
       end
+      # rubocop:enable Metrics/MethodLength
 
       def run_non_interactively
         IO.popen(full_command, 'r+') do |f|


lib/foreman_maintain/utils/command_runner.rb
```diff
@@ -59,12 +59,15 @@ def execution_error
...
 
+      # rubocop:disable Metrics/MethodLength
       def run_interactively
         # use tmp files to capture output and exit status of the command when
         # running interactively
         log_file = Tempfile.open('captured-output')
         exit_file = Tempfile.open('captured-exit-code')
-        Kernel.system("script -qc '#{full_command}; echo $? > #{exit_file.path}' #{log_file.path}")
+        Kernel.system(
+          "bash -c '#{full_command}; echo $? > #{exit_file.path}' | tee -i #{log_file.path}"
```
upadhyeammit, line 69: @jjeffers I am unable to find the difference between these commands apart from the piped `tee`. Because `script -c` runs the command instead of interactive shell and the `bash -c` is also doing the same. The `script` records the logs and exit status in given files. Bit more explanation of command change should help, thank you!

jjeffers: @upadhyeammit hey I made this PR based on feedback from a user. I am not sure about why they changed the command from `script` to `bash`, etc. I didn't notice that before. I was looking at the `tee` near the end.

Let me see if I can figure out if the change is meaningful. Thanks!

jjeffers: @upadhyeammit Testing with `bash` showed the change was not helpful. So I restored `script -qc`.

definitions/procedures/content/prepare.rb
```diff
@@ -8,7 +8,10 @@ class Prepare < ForemanMaintain::Procedure
     def run
       sleep(20) # in satellite 6.9 the services are still coming up
       # use interactive to get realtime output
-      puts execute!('foreman-rake katello:pulp3_migration', :interactive => true)
+      puts execute!(
+        'foreman-rake katello:pulp3_migration preserve_output=true',
```
ianballou, line 12: You won't need this with my PR here: https://github.com/theforeman/foreman_maintain/pull/516

ianballou (commented):
Worked fine for me! Just one comment about the `preserve_output` bit.  I just put out a PR to make that optional.

ianballou (approved):
Looks fine to me!

jjeffers:
@upadhyeammit I think this is ready for review.

lib/foreman_maintain/utils/command_runner.rb
```diff
@@ -59,12 +59,15 @@ def execution_error
...
 
+      # rubocop:disable Metrics/MethodLength
       def run_interactively
         # use tmp files to capture output and exit status of the command when
         # running interactively
         log_file = Tempfile.open('captured-output')
         exit_file = Tempfile.open('captured-exit-code')
-        Kernel.system("script -qc '#{full_command}; echo $? > #{exit_file.path}' #{log_file.path}")
+        Kernel.system(
+          "script -qc '#{full_command}; echo $? > #{exit_file.path}' | tee -i #{log_file.path}"
```
upadhyeammit, line 69: @jjeffers I think we don't need this pull request at all? because its just extra tee command which I think is not required, because original `script` command redirects the logs in file?

jjeffers: If we truly don't need it, then let's close this.

ianballou:
@upadhyeammit I believe this PR is needed. Without it I see this error in the logs before the migration proceeds:
```
tail: content-prepare.log: file truncated
nohup: ignoring input
Running Prepare content for Pulp 3
Prepare content for Pulp 3: 
                                                                      [FAIL]
Failed executing foreman-rake katello:pulp3_migration, exit status 256
Scenario [Prepare content for Pulp 3] failed.

  [content-prepare]

use --whitelist="content-prepare"
```

lib/foreman_maintain/utils/command_runner.rb
jjeffers: Peter Vreeman made an observation about the need for "bash" vs "script":
```
The problem is that 'script' does not work nice togeher with nohup (or any
other standard shell backgrounding from shell) because 'script' changes the
'setsid' and detaches from the parent process when going in the background,
then the 'nohup' does not have a sub process to follow anymore
```

upadhyeammit:
@jjeffers and @ianballou thanks for explanation! I will do few checks with this code on existing functionality and will share the feedback.

upadhyeammit (commented):
@jjeffers works without any problem. Can you limit the commit message to 50 chars?

upadhyeammit (approved):

PR #516, opened 2021-08-03
Fixes #33202 - add append-output mode to content prepare

ianballou:
With `--append-output` or `-a`, `foreman-maintain content prepare` will now show all output instead of overwriting the last line.

### DIFF
--- a/definitions/procedures/content/prepare.rb
+++ b/definitions/procedures/content/prepare.rb
@@ -3,12 +3,14 @@ module Procedures::Content
     metadata do
       description 'Prepare content for Pulp 3'
       for_feature :pulpcore
+      param :quiet, 'Keep the output on a single line', :flag => true, :default => false
     end
 
     def run
       sleep(20) # in satellite 6.9 the services are still coming up
       # use interactive to get realtime output
-      puts execute!('foreman-rake katello:pulp3_migration', :interactive => true)
+      env_vars = @quiet ? '' : 'preserve_output=true '
+      puts execute!("#{env_vars}foreman-rake katello:pulp3_migration", :interactive => true)
     end
   end
 end
--- a/definitions/scenarios/content.rb
+++ b/definitions/scenarios/content.rb
@@ -2,11 +2,10 @@ module ForemanMaintain::Scenarios
   module Content
     class ContentBase < ForemanMaintain::Scenario
       def enable_and_start_services
-        add_step(Procedures::Service::Start)
         add_step(Procedures::Service::Enable.
                  new(:only => Features::Pulpcore.pulpcore_migration_services))
         add_step(Procedures::Service::Start.
-                 new(:only => Features::Pulpcore.pulpcore_migration_services))
+                 new(:include => Features::Pulpcore.pulpcore_migration_services))
       end
 
       def disable_and_stop_services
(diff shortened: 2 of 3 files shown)


definitions/procedures/content/prepare.rb
```diff
@@ -3,12 +3,14 @@ class Prepare < ForemanMaintain::Procedure
     metadata do
       description 'Prepare content for Pulp 3'
       for_feature :pulpcore
+      param :append_output, 'Preserve all output lines', :flag => true, :default => false
```
upadhyeammit, line 6: As per the redmine details and considering overall user exp I feel we should invert this flag? By default it will show the progress details and if required one can silent it? If you agree then I also think the param can be something like `quite`?

upadhyeammit (commented):
Works well, added one comment. Request to check thank you!

ianballou:
@upadhyeammit I've added your suggestion.

ianballou:
@upadhyeammit is this looking okay now?

upadhyeammit:
I tried this and I dont see any difference between quite and verbose mode, maybe I think its because my 6.9 Satellite has one repository,

Output is below in both cases,
~~~
Prepare content for Pulp 3: 
Starting task.
2021-08-18 23:07:28 +0530: Content migration starting. 

Script started on Wed 18 Aug 2021 11:06:39 PM IST
Starting task.
2021-08-18 23:07:28 +0530: Content migration starting. 
~~~

While testing this I noticed another issue. The content abort fails to find services,
~~~
Stopping the following service(s):                                    [FAIL]
Scenario [Abort all running Pulp 2 to Pulp 3 migration tasks] failed.
~~~

This is happening because in abort we are trying to stop services which are disabled and foreman-maintain doesn't stop disabled services. Request to fix this as well : https://github.com/theforeman/foreman_maintain/blob/master/definitions/scenarios/content.rb#L64

I suggest to enable services before stopping those for now, and I will think if we want to have some override flag kind of implementation to skip from disabled services method.

ianballou:
> I tried this and I dont see any difference between quite and verbose mode, maybe I think its because my 6.9 Satellite has one repository

@upadhyeammit you might need to reset your migration or try a larger synced repository.

ianballou:
@upadhyeammit I've added the fix for enabling and starting services.

upadhyeammit:
@ianballou apologies for back and forth! One unrelated question,
The `enable_and_start_services` starts all services, enables required `pulpcore` services and then at the end start only `pulpcore` services. I think this means we want all services running including `pulpcore` services? so I think we can just enable required `pulpcore` services and start all of the services at once?

ianballou:
@upadhyeammit that sounds right to me, but I'll need to check if there's a particular reason why we didn't just start all services from the beginning.

definitions/scenarios/content.rb
```diff
@@ -60,6 +66,7 @@ class PrepareAbort < ContentBase
 
       def compose
         if !feature(:satellite) || feature(:satellite).at_least_version?('6.9')
+          enable_and_start_services
```
ianballou, line 69: I missed `if feature(:satellite)` here.

ianballou:
@upadhyeammit `enable_and_start_services` now starts all services at once.

upadhyeammit (approved):
Request you to change commit message. I can see mostly commits are having about 68 chars message however I think we should follow 50 chars as that's common guideline. Maybe I will add check for this. Rest looks good to me!

ianballou:
@upadhyeammit great, I've updated the commit message.

upadhyeammit:
@ianballou thanks a lot for contributions! Merging this one!

PR #518, opened 2021-08-11
Fixes #32634 - account for non-SCL redis services on el8

evgeni:


definitions/features/pulpcore.rb
```diff
@@ -12,7 +12,7 @@ class Features::Pulpcore < ForemanMaintain::Feature
 
   def services
     self.class.pulpcore_common_services + configured_workers + [
-      system_service('rh-redis5-redis', 5),
+      system_service(feature(:redis).service_name, 5),
```
evgeni, line 15: should this instead be
```
      feature(:redis).services
```

from https://github.com/theforeman/foreman_maintain/blob/238cbf6de91814df3b316724d221e184d73282de/definitions/features/redis.rb#L11-L13

but then the priority is different?!

definitions/features/redis.rb
```diff
@@ -20,11 +20,19 @@ class << self
...
+      else
+        '/etc'
+      end
     end
 
     def scl_prefix
-      "#{SCL_NAME}-"
+      if el7?
+        "#{SCL_NAME}-"
+      else
```
upadhyeammit, line 33: Do we require else block? because the string interpolation with `nil` should work?
~~~
2.6.6 :001 > a = nil
 => nil 
2.6.6 :003 > puts "#{a}amit"
amit
 => nil 
~~~

definitions/features/pulpcore.rb
upadhyeammit: The redis is required for pulpcore and hence I think it should be priority 5 in redis feature as well

definitions/features/redis.rb
evgeni: Probably not, was just thinking of being explicit.

definitions/features/pulpcore.rb
evgeni: Okay, will update then. Thanks

definitions/features/pulpcore.rb
```diff
@@ -11,10 +11,8 @@ class Features::Pulpcore < ForemanMaintain::Feature
   end
 
   def services
-    self.class.pulpcore_common_services + configured_workers + [
-      system_service('rh-redis5-redis', 5),
-      system_service('httpd', 30)
-    ]
+    self.class.pulpcore_common_services + configured_workers +
+      feature(:redis).services + feature(:apache).services
```
evgeni, line 15: Makes me wonder if features should have some kind of "dependent" features (in this case pulpcore depends on redis and apache), thus automatically inheriting the services etc?

evgeni: I heard you yesterday saying "it's too complicated", and that sounds fair :)

upadhyeammit (approved):

definitions/features/pulpcore.rb
upadhyeammit: Yes, I can see there is for_feature but its not something we can use for this purpose. Maybe this is something for future

upadhyeammit:
LGTM! thanks @evgeni for contribution!

PR #521, opened 2021-08-24
Allow options and parameters in any order

ehelms:


upadhyeammit:
The latest version of clamp https://rubygems.org/gems/clamp/versions/1.3.2 is not available in non scl maintenance repository. I can see `rubygem-clamp-0.6.2-2.` is doing duties now. So this means we need to bump clamp and required deps in maintenance repository too.

ehelms:
Good catch, I can look at if we could upgrade clamp. Or wrap this in a version check (for EL8 foreman-maintain) or wait to see if we SCL foreman-maintain if we drop older Ruby.

ehelms:
We now use the right version of Clamp! This allows for things that do not work to work.

Prior to change:
```
satellite-maintain packages install -y ansible-collection-redhat-satellite
```

After change:
```
satellite-maintain packages install ansible-collection-redhat-satellite -y
```

evgeni (approved):

PR #522, opened 2021-08-25
Fixes #33346 - limit installer run of switchover for Katello 4.0

upadhyeammit:


definitions/procedures/content/switchover.rb
```diff
@@ -5,8 +5,7 @@ class Switchover < ForemanMaintain::Procedure
       for_feature :pulpcore
 
       confine do
-        # FIXME: remove this condition for the 6.10 upgrade scenario
-        !feature(:instance).downstream
+        !check_min_version(foreman_plugin_name('katello'), '4.0')
```
upadhyeammit, line 8: I tried to depend on `katello-common` but removal of this package does not remove `tfm-rubygem-katello`, so instead used exact package name which has the rake task.

definitions/scenarios/content.rb
```diff
@@ -43,11 +43,8 @@ class Switchover < ContentBase
       end
 
       def compose
-        # FIXME: remove this condition for the 6.10 upgrade scenario
-        if Procedures::Content::Switchover.present?
-          add_step(Procedures::Content::Switchover)
+          add_step_with_context(Procedures::Content::Switchover)
```
upadhyeammit, line 46: Removed the present? condition as procedure is present now, and added step using context so confine block will be used before running the procedure.

upadhyeammit:
@jlsherrill request to review, thank you!

definitions/procedures/content/switchover.rb
jlsherrill: i think we need to limit the migration procedure too, thats what actually failed in the associated bz

jlsherrill: nevermind!

jlsherrill (approved):

PR #526, opened 2021-08-26
Fixes #33360 - use base database to get transform and config

evgeni:


upadhyeammit (approved):
Works well! thanks for the quick fixes!

PR #528, opened 2021-08-31
Fixes #33384 - fix pulp dirs of snapshot backup

upadhyeammit:


vsedmik:
Just tested on Sat 6.10.0 snap 15 using this PR to take the snapshot backup and restore via 0.8.13 from koji and worked well.

upadhyeammit:
@vsedmik thanks for the review! Merging this one!

PR #529, opened 2021-09-02
Fixes #33406 - add latest z-stream check

jlsherrill:
this will check that the katello and pulp migration plugin
packages are updated and if not will fail the upgrade.

Even if they are upgraded, we pause and ask the user to
confirm that they are on the latest z-stream, as requested
by CEE

### DIFF
--- a/definitions/checks/check_for_newer_packages.rb
+++ b/definitions/checks/check_for_newer_packages.rb
@@ -1,0 +1,26 @@
+class Checks::CheckForNewerPackages < ForemanMaintain::Check
+  metadata do
+    label :check_for_newer_packages
+    description 'Check for newer packages and optionally ask for confirmation if not found.'
+
+    param :packages,
+          'package names to validate',
+          :required => true
+    param :manual_confirmation_version,
+          'Version of satellite (6.9) to ask the user if they are on the latest minor release of.',
+          :required => false
+  end
+
+  def run
+    exit_status, = ForemanMaintain.package_manager.check_update(packages: @packages,
+                                                                with_status: true)
+    assert(exit_status == 0, 'An update is available for one of these packages: '\
+                "#{@packages.join(',')}. Please update before proceeding.")
+    if @manual_confirmation_version
+      question = 'Confirm that you are running the latest minor release of Satellite '\
+                "#{@manual_confirmation_version}"
+      answer = ask_decision(question, actions_msg: 'y(yes), q(quit)')
+      abort! if answer != :yes
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


definitions/checks/check_for_newer_packages.rb
```diff
@@ -0,0 +1,25 @@
...
+    param :packages,
+          'package names to validate',
+          :required => true
+    param :manual_confirmation_version,
+          'Version of satellite (6.9) to ask the user if they are on the latest z-stream of.',
+          :required => false
+  end
+
+  def run
+    result = ForemanMaintain.package_manager.check_update(packages: @packages, with_status: true)
```
upadhyeammit, line 15: Maybe you can use vars like:
```suggestion
    exit_status, _ = ForemanMaintain.package_manager.check_update(packages: @packages, with_status: true)
```

definitions/checks/check_for_newer_packages.rb
```diff
@@ -0,0 +1,25 @@
...
+    assert(result[0] == 0, 'An update is available for one of these packages: '\
+                "#{@packages.join(',')}. Please update before proceeding.")
+    if @manual_confirmation_version
+      question = 'Confirm that you are running the latest z-stream of Satellite '\
```
upadhyeammit, line 19: Maybe z-stream can confuse the user? instead can we use term minor version?

definitions/scenarios/upgrade_to_satellite_6_10.rb
```diff
@@ -26,8 +26,12 @@ class PreUpgradeCheck < Abstract
 
     def compose
       add_step(Checks::Puppet::WarnAboutPuppetRemoval)
+      add_step(Checks::CheckForNewerPackages.new(:packages => ['tfm-rubygem-katello',
```
upadhyeammit, line 29: I dont think this will be executed on EL8, however there is new method added `foreman_plugin_name` in the system_helpers which can be used to take care of scl/nonscl package naming: https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/concerns/system_helpers.rb#L222

upadhyeammit (changes requested):

jlsherrill:
updated!

ianballou:
This is looking fine to me. I saw the check fail when I needed updates, and I also saw the question after I updated and re-ran foreman-maintain.

jlsherrill:
@upadhyeammit updated based on your comments, @ianballou did some testing, any other comments?  Thanks!

upadhyeammit (approved):
LGTM!

upadhyeammit:
Works as expected!
@ianballou thanks for review, @jlsherrill thanks for contribution!

PR #530, opened 2021-09-03
Fixes #33411 - disable and stop pulp3 services

upadhyeammit:
The pulp3 workers should be stopped and disabled after migration
to pulp3


upadhyeammit:
@jlsherrill request to review!

jlsherrill (approved):

PR #531, opened 2021-09-07
Refs #33346 - limit switchover run below Katello 4.0 version

upadhyeammit:


upadhyeammit:
as switchover procedure wont be available because of confine block the add_step_with_context wont work from scenario. I added unless condition there too.
@jlsherrill  request to check, thank you!

upadhyeammit:
and I also think we can skip whole `pulp3_switchover_steps` as per the unless condition?

jlsherrill:
i agree, we could put the whole section in that unless block that starts/stops the services too

upadhyeammit:
@jlsherrill thanks for the feedback, merging this!

PR #532, opened 2021-09-08
Fixes #33436 - drop facter dependency

upadhyeammit:


upadhyeammit:
@ekohl @evgeni I yet to fully test it however basic tests worked fine. Request to review, thank you!

lib/foreman_maintain/utils/os_facts.rb
```diff
@@ -0,0 +1,22 @@
+module ForemanMaintain::Utils
+  module OsFacts
+    include ForemanMaintain::Concerns::SystemHelpers
+    OS_RELEASE = '/etc/os-release'.freeze
+
+    def self.grep_cmd(str)
+      execute!("grep -w #{str} #{OS_RELEASE}|cut -d'=' -f 2").tr('"', '')
```
ekohl, line 7: Can't you just read the file without grep and parse it using pure Ruby?

lib/foreman_maintain/utils/os_facts.rb
```diff
@@ -0,0 +1,22 @@
...
+    def self.version_id
+      grep_cmd('VERSION_ID')
+    end
+
+    def self.id
+      grep_cmd('ID')
+    end
+
+    def self.id_like
+      grep_cmd('ID_LIKE')
```
ekohl, line 19: I'd present this as a list. See https://www.freedesktop.org/software/systemd/man/os-release.html#ID_LIKE= as well.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -185,28 +185,31 @@ def package_manager
...
       def debian?
-        os_facts['os']['family'] == 'Debian'
+        ids = %w[debian ubuntu]
+        ids.include?(os_facts.id_like) ||
+          ids.include?(os_facts.id)
       end
 
       def el7?
-        os_facts['os']['release']['major'] == '7' && el?
+        os_facts.version_id.to_i == 7 && el?
```
ekohl, line 204: Why not implement this as 
```suggestion
        el_major_version == 7
```

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -185,28 +185,31 @@ def package_manager
...
+        ids = ['rhel', 'rhel fedora']
+        ids.include?(os_facts.id_like) ||
+          ids.include?(os_facts.id)
       end
 
       def debian?
-        os_facts['os']['family'] == 'Debian'
+        ids = %w[debian ubuntu]
+        ids.include?(os_facts.id_like) ||
+          ids.include?(os_facts.id)
```
ekohl, line 200: It's probably find to use this (provided `id_like` becomes an array)
```suggestion
        os_facts.id == 'debian'  || os_facts.id_like.include?('debian')
```

It may also be sufficient to check if `/etc/debian_version` exists.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -185,28 +185,31 @@ def package_manager
...
-        facter = ForemanMaintain::Utils::Facter.path
-        @os_facts ||= JSON.parse(execute("#{facter} -j os"))
+        @os_facts ||= ForemanMaintain::Utils::OsFacts
       end
 
       def el?
-        os_facts['os']['family'] == 'RedHat'
+        ids = ['rhel', 'rhel fedora']
+        ids.include?(os_facts.id_like) ||
+          ids.include?(os_facts.id)
```
ekohl, line 194: This won't work on  some EL-derivatives. https://github.com/stejskalleos/os_release is a nice collection of the various os-release files if you want examples.

Another way is to check for the existence of `/etc/redhat-release`.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -185,28 +185,31 @@ def package_manager
       end
 
       def os_facts
-        facter = ForemanMaintain::Utils::Facter.path
-        @os_facts ||= JSON.parse(execute("#{facter} -j os"))
+        @os_facts ||= ForemanMaintain::Utils::OsFacts
```
ekohl, line 188: `OsFacts` is a class. I'm not sure if there is a benefit to keeping `@os_facts`
```suggestion
        ForemanMaintain::Utils::OsFacts
```

lib/foreman_maintain/utils/os_facts.rb
evgeni: It's also not present on all OSes (like Fedora)

lib/foreman_maintain/concerns/system_helpers.rb
evgeni: I think checking for `/etc/redhat-release` is the most trivial and correct way.

This *will* flag Fedora as being EL, but so does the current code?
And we don't really care about Fedora right now.

lib/foreman_maintain/concerns/system_helpers.rb
evgeni: +1 on `/etc/debian_version`

upadhyeammit: In general I dont think `/etc/debian_version` can be missing on system. However I felt to keep `/etc/os-release` if in user error case that is missing.

lib/foreman_maintain/concerns/system_helpers.rb
upadhyeammit: and same for this one, if in  case redhat-release file is missing we check for os-release.

upadhyeammit:
@ekohl @evgeni I felt concern makes more sense instead of util. Considering this I push separate commit to compare with existing changes. Request to review, thank you!

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -0,0 +1,58 @@
+module ForemanMaintain
+  module Concerns
+    module OsFacts
+      OS_RELEASE_FILE = '/etc/os-release'.freeze
+
+      def facts
+        facts = {}
+        tr_map = { '"' => '', "\n" => '' }
+        rexp = Regexp.union(tr_map.keys)
+        file = File.open(OS_RELEASE_FILE)
```
ekohl, line 10: I'd always prefer a block with open so you're sure it's closed. While you're unlikely to leak file descriptors in maintain, for daemons it's a real risk.

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -0,0 +1,58 @@
...
+        file.readlines.each do |line|
+          key, value = line.gsub(rexp, tr_map).split('=')
```
ekohl, line 12: I think lines can have comments which should be skipped.

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+
+      def os_id
+        facts.fetch('ID')
+      end
+
+      def os_id_like_list
+        return facts.fetch('ID_LIKE').split(' ') if facts.key?('ID_LIKE')
+
+        []
```
ekohl, line 29: I think this does the same
```suggestion
        facts.fetch('ID_LIKE', '').split
```

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -0,0 +1,58 @@
+module ForemanMaintain
+  module Concerns
+    module OsFacts
+      OS_RELEASE_FILE = '/etc/os-release'.freeze
+
+      def facts
```
ekohl, line 6: Should this method cache the result to avoid reading it over and over again?

ekohl (commented):
I'm not sure if you need to check more than the release/version file for RH/Debian. They're pretty guaranteed to exist I think.

lib/foreman_maintain/concerns/os_facts.rb
upadhyeammit: Added one more regex which filters the comments, tested it with adding multiple comments throughout the file and works as expected.

lib/foreman_maintain/concerns/os_facts.rb
upadhyeammit: Added block

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -0,0 +1,54 @@
...
+        unless defined?(@facts)
+          @facts = {}
+          tr_map = { /^#.*/ => '', '"' => '', "\n" => '' }
+          rexp = Regexp.union(tr_map.keys)
+          File.open(OS_RELEASE_FILE) do |file|
+            file.readlines.each do |line|
+              key, value = line.gsub(rexp, tr_map).split('=')
+              @facts[key] = value unless key.nil?
+            end
+          end
```
evgeni, line 16: I must admit I find the Rexexp solution hard to read, and would propose the following alternative

```suggestion
          File.open(OS_RELEASE_FILE) do |file|
            file.readlines.each do |line|
              line.strip! # drop any whitespace, including newlines from start and end of the line
              next if line.start_with?('#') # ignore comments
              key, value = line.split('=', 2) # split at most into 2 items, if the value ever contains an =
              @facts[key] = value.gsub(/"/, '') unless key.nil?
            end
          end
```

(not blocking tho, if you prefer your solution, I'm good with that)

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -0,0 +1,54 @@
+module ForemanMaintain
+  module Concerns
+    module OsFacts
+      OS_RELEASE_FILE = '/etc/os-release'.freeze
```
evgeni, line 4: According to https://www.freedesktop.org/software/systemd/man/os-release.html, there is a fallback location at `/usr/lib/os-release` - should we add support for that too?

upadhyeammit: For sure, pushed the change for same!

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -0,0 +1,68 @@
...
+      def facts
+        unless defined?(@facts)
+          @facts = {}
+          File.open(os_release_file) do |file|
+            file.readlines.each do |line|
+              line.strip! # drop any whitespace, including newlines from start and end of the line
+              next if line.start_with?('#') # ignore comments
+              # split at most into 2 items, if the value ever contains an =
+              key, value = line.split('=', 2)
+              @facts[key] = value.delete('"') unless key.nil?
```
ekohl, line 28: Untested, but perhaps you can use this to support both single and double quotes. I also wondered if we should handle no value at all or imply skip it:
```suggestion
              next unless key && value
              @facts[key] = value.gsub(/^(["'])(.*)(\1)$/, '\2')
```

evgeni: Technically, you'd also need to unquote other chars:
"Variable assignment values must be enclosed in double or single quotes if they include spaces, semicolons or other special characters outside of A–Z, a–z, 0–9. Shell special characters ("$", quotes, backslash, backtick) must be escaped with backslashes, following shell style." (via https://www.freedesktop.org/software/systemd/man/os-release.html)

evgeni: In https://www.freedesktop.org/software/systemd/man/os-release.html#id-1.7.5 they use `ast.literal_eval` in python to parse this properly. No idea if something like this exists in Ruby? (plain eval sounds bad)

upadhyeammit: With the special char and backslash the key value pair comes as `\"\\$bar\"` . The @ekohl regex plucks the string and it comes as `"\\$bar"`, so I think its correct to just do `string.delete("\\"), pushed the change request to check.

ekohl (approved):
LGTM. @evgeni?

PR #533, opened 2021-09-09
Fixes #33437 - skip installer run after switchover 

upadhyeammit:
The installer run stops and disables pulpcore-workers*
We need to stop the services but not disable those

### DIFF
--- a/definitions/procedures/content/switchover.rb
+++ b/definitions/procedures/content/switchover.rb
@@ -11,7 +11,6 @@ module Procedures::Content
       param :skip_deb, 'Do not run debian options in installer.'
     end
 
-    # rubocop:disable Metrics/MethodLength
     def run
       puts 'Performing final content migration before switching content'
       puts execute!('foreman-rake katello:pulp3_migration')
@@ -19,6 +18,10 @@ module Procedures::Content
       puts execute!('foreman-rake katello:pulp3_post_migration_check')
       puts 'Switching specified content over to pulp 3'
       puts execute!('foreman-rake katello:pulp3_content_switchover')
+      run_installer if feature(:upstream)
+    end
+
+    def run_installer
       puts 'Re-running the installer to switch specified content over to pulp3'
       args = ['--foreman-proxy-content-proxy-pulp-isos-to-pulpcore=true',
               '--foreman-proxy-content-proxy-pulp-yum-to-pulpcore=true',
--- a/definitions/scenarios/upgrade_to_satellite_6_10.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_10.rb
@@ -64,15 +64,11 @@ module Scenarios::Satellite_6_10
     end
 
     def pulp3_switchover_steps
-      add_step(Procedures::Service::Start)
       add_step(Procedures::Service::Enable.
           new(:only => Features::Pulpcore.pulpcore_migration_services))
-      add_step(Procedures::Service::Start.
-          new(:only => Features::Pulpcore.pulpcore_migration_services))
-      add_step(Procedures::Content::Switchover.new(:skip_deb => true))
+      add_step(Procedures::Service::Start)
+      add_step(Procedures::Content::Switchover.new)
       add_step(Procedures::Service::Stop.
-                 new(:only => Features::Pulpcore.pulpcore_migration_services))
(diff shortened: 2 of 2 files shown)


ianballou (approved):
Looks good to me!

jlsherrill:
i'm a bit confused,  if we're removing this, won't we hit https://projects.theforeman.org/issues/33411?

jlsherrill:
would it make sense to skip the disable step but keep the stop step?

definitions/scenarios/upgrade_to_satellite_6_10.rb
```diff
@@ -64,16 +64,12 @@ def check_var_lib_pulp
     end
 
     def pulp3_switchover_steps
-      add_step(Procedures::Service::Start)
```
upadhyeammit, line 67: Instead of starting all services, then enabling pulpcore services and then starting pulpcore services only. I did,
Enable pulpcore services
Start all services

definitions/scenarios/upgrade_to_satellite_6_10.rb
```diff
@@ -64,16 +64,12 @@ def check_var_lib_pulp
...
       add_step(Procedures::Service::Enable.
           new(:only => Features::Pulpcore.pulpcore_migration_services))
-      add_step(Procedures::Service::Start.
-          new(:only => Features::Pulpcore.pulpcore_migration_services))
-      add_step(Procedures::Content::Switchover.new(:skip_deb => true))
+      add_step(Procedures::Service::Start)
+      add_step(Procedures::Content::Switchover.new)
       add_step(Procedures::Service::Stop.
```
upadhyeammit, line 71: I kept the service stop, if in case rake tasks fail then we stop the services.

definitions/scenarios/upgrade_to_satellite_6_10.rb
```diff
@@ -64,16 +64,12 @@ def check_var_lib_pulp
...
                  new(:only => Features::Pulpcore.pulpcore_migration_services))
-      add_step(Procedures::Service::Disable.
```
upadhyeammit, line 75: Removed as services are expected to be enabled.

definitions/procedures/content/switchover.rb
```diff
@@ -11,14 +11,17 @@ class Switchover < ForemanMaintain::Procedure
...
 
-    # rubocop:disable Metrics/MethodLength
     def run
       puts 'Performing final content migration before switching content'
       puts execute!('foreman-rake katello:pulp3_migration')
       puts 'Performing a check to verify everything that is needed has been migrated'
       puts execute!('foreman-rake katello:pulp3_post_migration_check')
       puts 'Switching specified content over to pulp 3'
       puts execute!('foreman-rake katello:pulp3_content_switchover')
+      run_installer if feature(:upstream)
```
upadhyeammit, line 21: refactored a bit!

jlsherrill (approved):
makes sense for me, were we able to test it on one of the reproducers?

definitions/scenarios/upgrade_to_satellite_6_10.rb
```diff
@@ -64,16 +64,12 @@ def check_var_lib_pulp
...
 
     def pulp3_switchover_steps
-      add_step(Procedures::Service::Start)
       add_step(Procedures::Service::Enable.
           new(:only => Features::Pulpcore.pulpcore_migration_services))
-      add_step(Procedures::Service::Start.
-          new(:only => Features::Pulpcore.pulpcore_migration_services))
-      add_step(Procedures::Content::Switchover.new(:skip_deb => true))
+      add_step(Procedures::Service::Start)
+      add_step(Procedures::Content::Switchover.new)
```
ianballou, line 70: Why did `:skip_deb => true` go away?

jlsherrill: the installer isn't actually getting run here downstream as part of switchover, so the parameter isn't needed

mccun934:
> makes sense for me, were we able to test it on one of the reproducers?

yup, passed/worked

upadhyeammit:
Thanks @mccun934 @jlsherrill @ianballou for reviews! Merging now!

PR #535, opened 2021-09-15
Fixes #33488 - wipe journals for old (1.36) and new (1.39) qpidd

evgeni:


definitions/procedures/restore/configs.rb
```diff
@@ -56,7 +56,11 @@ def clean_conflicting_data
     def reset_qpid_jrnls
       # on restore without pulp data qpid fails to start
       # https://access.redhat.com/solutions/4645231
-      execute('rm -rf /var/lib/qpidd/.qpidd/qls/dat2/__db.00*')
+      ['/var/lib/qpidd/.qpidd/', '/var/lib/qpidd/'].each do |qpidd_path|
+        if File.exist?("#{qpidd_path}/qls/dat2/")
```
evgeni, line 60: I never know what's better, `File.exist?` or `Dir.exist?`

jlsherrill: if its a directory, it probably doesn't matter.  Dir.exists? might show better intent, but in code it likely doesn't matter

upadhyeammit (approved):

PR #536, opened 2021-09-20
Fixes #33273 - hostname and DNS/DHCP proxy metadata

evgeni:


evgeni:
I don't think single commit is valid here. The one commit adds the feature, the other implements the use of it.

upadhyeammit (approved):

PR #537, opened 2021-09-20
Fixes #33271 - check DHCP and DNS interfaces before the restore

evgeni:


lib/foreman_maintain/utils/backup.rb
```diff
@@ -241,6 +241,24 @@ def validate_hostname?
         end
       end
 
+      def validate_interfaces?
```
upadhyeammit, line 244: Few thoughts,
1. The invalid interfaces can be dhcp or dns or both. So if user fixes one of the interface maybe in next restore run he will need to fix another one?
2. The method does not return which interface is invalid, or what is expected. I think if we show that, it can help? because we are failing on invalid interfaces which means it needs to be fixed to successfully restore? or skip the check.
3. Should we provide a way to fix it by either suggesting change the name of interface or change the dns or dhcp config on successful restore. This means user can whitelist or skip this check and proceed with restore.

lib/foreman_maintain/utils/backup.rb
```diff
@@ -241,6 +241,24 @@ def validate_hostname?
...
+        # I wanted to do `Socket.getifaddrs.map(&:name).uniq`,
+        # but this has to work with Ruby 2.0, and Socket.getifaddrs is 2.1+
+        system_interfaces = Dir.children('/sys/class/net')
```
upadhyeammit, line 247: Maybe we can do something like below, return invalid interfaces and actual check will confirm if hash is empty?
~~~
        system_interfaces = Dir.children('/sys/class/net')
        invalid_interfaces = {}
        proxy_config = metadata.fetch('proxy_config', {})
        dns_dhcp_config = ( ['dns', 'dhcp'] & proxy_config.keys )
        if dns_dhcp_config.any?
          dns_dhcp_config.each do |interface|
            unless system_interfaces.include?(proxy_config.fetch(interface, 'lo'))
              invalid_interfaces[interface] = proxy_config.fetch(interface, 'lo')
            end
          end
        end
        invalid_interfaces
~~~

upadhyeammit (commented):
Added few comments, request to check

lib/foreman_maintain/utils/backup.rb
```diff
@@ -241,6 +241,24 @@ def validate_hostname?
...
+
+        proxy_config = metadata.fetch('proxy_config', {})
+
+        if proxy_config.fetch('dns', false) &&
+           !system_interfaces.include?(proxy_config.fetch('dns_interface', 'lo'))
+          return false
+        elsif proxy_config.fetch('dhcp', false) &&
+              !system_interfaces.include?(proxy_config.fetch('dhcp_interface', 'lo'))
+          return false
```
ekohl, line 256: DNS and DHCP could run on separate interfaces. I'd prefer to check both (so 2 if statements instead of elsif).

lib/foreman_maintain/utils/backup.rb
evgeni: Yeah, I do agree. I think I was mostly not sure about the interface/contract (if there is any) between `ForemanMaintain::Utils` and `Checks`. The other checks seemed to use a boolean `validate_XXX?`, which I copied over.
If there is no such hard requirement, I'd probably return a `Hash` of the form `{feature: {'configured': …, 'available': …}, …}` and let the check look whether the hash is empty (no errors) or construct a message from the hash based on features and configured/available.

evgeni:
@upadhyeammit is there a way to write tests for the Check part?

evgeni:
well, this now has tests for the check too, and is reviewable, I think

definitions/checks/restore/validate_interfaces.rb
```diff
@@ -0,0 +1,24 @@
+require 'foreman_maintain/utils/backup'
+
+module Checks::Restore
+  class ValidateInterfaces < ForemanMaintain::Check
+    metadata do
+      description 'Validate interfaces are the same as backup'
```
ekohl, line 6:
```suggestion
      description 'Validate network interfaces match the backup'
```

definitions/scenarios/restore.rb
```diff
@@ -18,6 +18,7 @@ def compose
       add_steps_with_context(Checks::Restore::ValidateBackup,
                              Procedures::Restore::Confirmation,
                              Checks::Restore::ValidateHostname,
+                             Checks::Restore::ValidateInterfaces,
```
ekohl, line 21: How does the error reporting work? If both the hostname and the interfaces are incorrect, does report both error or does it quit after the first check fails?

I'm asking because I want to have a check later on for the Smart Proxy Realm feature. Ideally we would report all errors at the same time so the user doesn't get in a loop of fix problem, next restore, fix next problem, etc. That means we may need to have a check `ValidateAnswersMatchTargetSystem` or something like that.

evgeni: My understanding that today it quits after the first error, but I might be wrong.

upadhyeammit: @evgeni yes it quits after first error as default run_strategy is fail_fast. I think the restore scenario needs implementation like upgrade scenario. So in that case the all the pre-restore checks will be in one seperate scenario and that will be having fail_slow run strategy.

test/lib/utils/backup_test.rb
```diff
@@ -281,5 +281,46 @@ def assume_feature_absent(label)
...
+      Dir.stubs(:children).returns(%w[lo eth0])
+      backup.stubs(:metadata).returns('proxy_config' =>
+                                      { 'dhcp' => true,
+                                        'dhcp_interface' => 'eth0',
+                                        'dns' => true,
+                                        'dns_interface' => 'eth0' })
+      assert backup.validate_interfaces.empty?
+    end
+
+    it 'rejects backup with proxy config and enabled DHCP/DNS in the metadata' do
```
upadhyeammit, line 313: this should be something like below as we dont reject every backup with DHCP/DNS?
```suggestion
    it 'rejects backup when DHCP/DNS configured interfaces are not found on system' do
```

upadhyeammit:
Added few comments otherwise looks good to me :)

test/lib/utils/backup_test.rb
evgeni: agreed, also updated the accept one before to match

definitions/scenarios/restore.rb
evgeni: Yepp, sounds like a good follow-up!

definitions/checks/restore/validate_interfaces.rb
```diff
@@ -0,0 +1,24 @@
...
+      param :backup_dir,
+            'Path to backup directory',
+            :required => true
+      manual_detection
+    end
+
+    def run
+      backup = ForemanMaintain::Utils::Backup.new(@backup_dir)
+      invalid_interfaces = backup.validate_interfaces
+      msg = 'The following features are enabled in the backup, '\
```
upadhyeammit, line 17: Before merging I just felt we are not telling where interfaces are missing, and I have added newline as message looks long for a line,
```suggestion
      msg = 'The following features are enabled in the backup, '\
                 "\nbut the system does not have the interfaces used by these features: "
```

upadhyeammit:
@ekohl looks good to me, like to know your thoughts.

PR #540, opened 2021-09-21
update ruby versions to match destination OSes

evgeni:
EL7 uses 2.0.0
EL8 uses 2.7.4 (and used 2.5.9 for a while)
EL9 uses 3.0.2 (at least today)
Debian/Ubuntu have 2.5/2.7

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: [2.3.3, 2.4.5, 2.5.3, 2.6.5]
+        ruby-version: ['2.5.3', '2.7.4', '3.0.2']
     steps:
       - uses: actions/checkout@v2
       - name: Install facter package
--- a/Gemfile
+++ b/Gemfile
@@ -7,3 +7,7 @@ if File.exist? "#{__FILE__}.local"
   eval(File.read("#{__FILE__}.local"), nil, "#{__FILE__}.local")
 end
 # rubocop:enable  Security/Eval
+
+if RUBY_VERSION >= '3.0'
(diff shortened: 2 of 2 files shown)


upadhyeammit (approved):

foreman_maintain.gemspec
```diff
@@ -24,6 +24,10 @@ the Foreman/Satellite up and running."
   s.add_dependency 'clamp'
   s.add_dependency 'highline'
 
+  if RUBY_VERSION >= '3.0'
+    s.add_dependency 'rexml' # rexml was a default gem in ruby < 3.0, and is a bundled gem in >= 3.0
+  end
```
ekohl, line 29: Sadly you can't do this in a gemspec. AFAIK the gemspec is evaluated during `gem build` but not after. So that's what's then uploaded to Rubygems. I'm not sure how you're supposed to deal with it.

ekohl (changes requested):

foreman_maintain.gemspec
evgeni: ewww.

well, we could add the same logic to `Gemfile`?

evgeni: opn the other hand, adding the rexml dependency on pre 3.0 doesn't hurt anyone

ekohl: I don't know if it doesn't hurt anyone. That said, Ruby 3 support is still a long way out for Foreman itself. It needs some serious work on fog which is a deep rabbit hole.

evgeni: it's bundled in < 3.0, so depending on it shouldn't change anything.

I can remove ruby3 from this PR, but this also means I can't run FM tests on my Fedora :(

ekohl: I'd recommend adding it to Gemfile for now then.

evgeni: done

upadhyeammit (approved):

PR #541, opened 2021-09-24
Fixes #33568 - validate external db backup explicitly

upadhyeammit:
If some of the db(s) are configured externally instead
of all then restore fails while validating files present
in backup. This adds separate method to validate this kind
of hybrid setups.

### DIFF
--- a/lib/foreman_maintain/utils/backup.rb
+++ b/lib/foreman_maintain/utils/backup.rb
@@ -77,11 +77,15 @@ module ForemanMaintain
       end
 
       def valid_fpc_backup?
-        fpc_online_backup? || fpc_standard_backup? || fpc_logical_backup?
+        fpc_online_backup? || fpc_standard_backup? || fpc_logical_backup? || \
+          # fpc can have setup where mongo or pulpcore db is external but not both
+          fpc_hybrid_db_backup?
       end
 
       def valid_katello_backup?
-        katello_online_backup? || katello_standard_backup? || katello_logical_backup?
+        katello_online_backup? || katello_standard_backup? || katello_logical_backup? || \
+          # Katello can have setup where some of dbs are external but not all
+          katello_hybrid_db_backup?
       end
 
       def valid_foreman_backup?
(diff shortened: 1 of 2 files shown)


test/lib/utils/backup_test.rb
```diff
@@ -38,6 +38,10 @@ module ForemanMaintain
       file_path = '../../files/backups/katello_logical_pulp2_pulpcore_database'
       File.expand_path(file_path, File.dirname(__FILE__))
     end
+    let(:katello_hybrid_db_backup) do
+      file_path = '../../files/backups/katello_hybrid_db_backup/'
```
ekohl, line 42: Probably best for a separate PR, but have you thought about a `fixture()` helper method so it would be `let(:katello_hybrid_db_backup) { fixture('backups/katello_hybrid_db_backup') }`

lib/foreman_maintain/utils/backup.rb
```diff
@@ -224,6 +244,27 @@ def foreman_logical_backup?
...
+      def dumps_for_hybrid_db_setup(dbs_hash)
+        present = []
+        absent = []
+        dbs_hash.each do |data_file, dbs|
+          dbs.each do |db|
+            feature_label = db == 'mongo' ? db : "#{db}_database"
+            dump_file = "#{db}_dump"
+            if feature(feature_label.to_sym).local?
+              present |= [data_file]
+              absent.concat [dump_file.to_sym]
```
ekohl, line 257: Isn't this equal?
```suggestion
              absent << dump_file.to_sym
```

ekohl (commented):
Why is there a need for a separate `katello_` vs `fpc_` method? Can't you unify them and check for the `foreman` and `candlepin` features to be present?

vsedmik:
Tested on 6.10 and 6.9 with external DBs:
1) Taken the offline backup (with pulp) using this PR and another offline backup (with pulp) using the in-snap f-m. 
2) Successfully restored both backups, checked the services were and content present.

upadhyeammit:
> Why is there a need for a separate `katello_` vs `fpc_` method? Can't you unify them and check for the `foreman` and `candlepin` features to be present?

Makes sense, however I am not confident to change it now, created redmine for it: https://projects.theforeman.org/issues/33587

test/lib/utils/backup_test.rb
upadhyeammit: +1 https://projects.theforeman.org/issues/33588

ekohl (approved):

PR #543, opened 2021-09-29
Fixes #33583 - dont add quotes while doing shellescape

upadhyeammit:


upadhyeammit:
When the password has the single quotes special chars works without preceding escape character,
~~~
# /usr/share/candlepin/cpdb --validate --verbose -u 'candlepin' -p password\! -d 'candlepin' --dbhost 'abc.example.com' --dbport '5432'
Configuring PostgreSQL with JDBC URL: jdbc:postgresql://abc.example.com:5432/candlepin
Validating Candlepin database
--driver=org.postgresql.Driver --classpath=/var/lib/tomcat/webapps/candlepin/WEB-INF/lib/postgresql-42.2.2.jar:/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/ --changeLogFile=db/changelog/changelog-validate.xml --url="jdbc:postgresql://abc.example.com:5432/candlepin" --username=$DBUSERNAME --password=$DBPASSWORD --logLevel=debug
DEBUG 9/29/21 1:12 PM:liquibase: Connected to candlepin@jdbc:postgresql://abc.example.com:5432/candlepin
DEBUG 9/29/21 1:12 PM:liquibase: Setting auto commit to false from true
~~~

However if escape chars are added with quotes then command fails,
~~~
# /usr/share/candlepin/cpdb --validate --verbose -u 'candlepin' -p 'password\!' -d 'candlepin' --dbhost 'abc.example.com' --dbport '5432'
Configuring PostgreSQL with JDBC URL: jdbc:postgresql://abc.example.com:5432/candlepin
Validating Candlepin database
--driver=org.postgresql.Driver --classpath=/var/lib/tomcat/webapps/candlepin/WEB-INF/lib/postgresql-42.2.2.jar:/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/ --changeLogFile=db/changelog/changelog-validate.xml --url="jdbc:postgresql://abc.example.com:5432/candlepin" --username=$DBUSERNAME --password=$DBPASSWORD --logLevel=debug

########## ERROR ############
Error running command: /usr/share/candlepin/liquibase.sh --driver=org.postgresql.Driver --classpath=/var/lib/tomcat/webapps/candlepin/WEB-INF/lib/postgresql-42.2.2.jar:/var/lib/tomcat/webapps/candlepin/WEB-INF/classes/ --changeLogFile=db/changelog/changelog-validate.xml --url="jdbc:postgresql://abc.example.com:5432/candlepin" --username=$DBUSERNAME --password=$DBPASSWORD --logLevel=debug migrate -Dcommunity=False
Status code: 65280
Command output: Sep 29, 2021 1:15:21 PM org.postgresql.core.v3.ConnectionFactoryImpl log
WARNING: SQLException occurred while connecting to abc.example.com:5432
org.postgresql.util.PSQLException: FATAL: password authentication failed for user "candlepin"
~~~

definitions/features/candlepin_database.rb
```diff
@@ -35,9 +35,7 @@ def check_option_using_cpdb_help(option_name, parent_cmd = '')
   def execute_cpdb_validate_cmd
     main_cmd = cpdb_validate_cmd
     return [true, nil] if main_cmd.empty?
-    main_cmd += format_shell_args(
-      '-u' => configuration['username'], '-p' => configuration[%(password)]
-    )
+    main_cmd += " -u #{configuration['username']} -p #{configuration[%(password)]} "
```
evgeni, line 38: I think the real problem is in 
https://github.com/theforeman/foreman_maintain/blob/494a3f12955bbb1b9f453d463a3b50d88c969175/lib/foreman_maintain/concerns/system_helpers.rb#L149-L151

You use `Shellwords.shellescape`, but then add single quotes around the command.
However, the [docs for `shellescape` say](https://ruby-doc.org/stdlib-2.5.1/libdoc/shellwords/rdoc/Shellwords.html#method-c-shellescape): "Note that a resulted string should be used unquoted and is not intended for use in double quotes nor in single quotes."

evgeni:
I *think* this is correct now, but have no way to test it :(

pipopopo:
I saw that the fix changed so I tested that one also.

`# cat ext_cdp.patch
````
diff -up /usr/share/gems/gems/foreman_maintain-0.8.10/lib/foreman_maintain/concerns/system_helpers.rb.bkp /usr/share/gems/gems/foreman_maintain-0.8.10/lib/foreman_maintain/concerns/system_helpers.rb
--- /usr/share/gems/gems/foreman_maintain-0.8.10/lib/foreman_maintain/concerns/system_helpers.rb.bkp    2021-09-29 14:41:48.219373044 +0200
+++ /usr/share/gems/gems/foreman_maintain-0.8.10/lib/foreman_maintain/concerns/system_helpers.rb        2021-09-29 14:42:27.672888450 +0200
@@ -147,7 +147,7 @@ module ForemanMaintain
       end

       def format_shell_args(options = {})
-        options.map { |shell_optn, val| " #{shell_optn} '#{shellescape(val)}'" }.join
+      options.map { |shell_optn, val| " #{shell_optn} #{shellescape(val)}" }.join
       end

       def find_symlinks(dir_path)

```

Was able to pass the upgrade also with this fix.

TEST: PASSED

`# foreman-maintain upgrade check --target-version=6.10
Check whether system has any non Red Hat repositories (e.g.: EPEL) enabled:
| Checking repositories enabled on the system                         [OK]


Running Checks before upgrading to Satellite 6.10
Warn about Puppet content removal prior to 6.10 upgrade:
Puppet repositories found!
Upgrading to 6.10 will delete the Puppet repositories
if any of the following conditions are met,
1. Any Puppet repositories in Library lifecycle environment.
2. Unpublished content view's Puppet repositories.

Note: Any Puppet content that is in use by hosts will not be deleted.
After the upgrade, hosts will continue to use the Puppet modules,
but future Puppet content management must be handled outside of the Satellite!

Do you want to proceed?, [y(yes), n(no)] y
                                                                      [OK]
Clean old Kernel and initramfs files from tftp-boot:                  [OK]
Check to make sure root(/) partition has enough space:                [OK]
Check to make sure /var/lib/candlepin has enough space:               [OK]
Check to validate candlepin database:                                 [OK]
Check for running tasks:                                              [OK]
Check for old tasks in paused/stopped state:                          [FAIL]
Found 34 paused or stopped task(s) older than 30 days
--------------------------------------------------------------------------------`

test/lib/system_helpers_test.rb
```diff
@@ -60,5 +60,25 @@ module ForemanMaintain
         refute system.check_max_version('katello', '3.18')
       end
     end
+
+    describe 'format_shell_args options' do
+      it 'returns the string without single or double quotes' do
+        config = { 'user' => 'foo', 'password' => 'foopassword1!' }
+        escaped_string = ' user foo password foopassword1\\!'
```
evgeni, line 67: I wonder if we also should include `'` and/or `"` in the to-be-escaped strings somewhere, but on the other hand we're not writing tests for `Shellwords.escape` here…

evgeni (approved):
Two comments about the tests, but not blocking on them.

test/lib/system_helpers_test.rb
```diff
@@ -60,5 +60,22 @@ module ForemanMaintain
...
+        assert_match escaped_string, system.format_shell_args(config)
+      end
+    end
+
+    describe 'shellescape string' do
+      it 'escapes quotes' do
+        password_one = "foo\'bar"
```
evgeni, line 74: shouldn't this be 
```suggestion
        password_one = "foo'bar"
```

evgeni: or are you expecting it to escape the `\` too?

upadhyeammit: nope, my mistake, changed this!

evgeni (approved):

PR #544, opened 2021-09-29
Fixes #33593 - only look for katello 3.18 package updates

upadhyeammit:


upadhyeammit:
@ianballou and @jlsherrill let me know your thoughts, thank you!

upadhyeammit:
@evgeni request your review too, thank you!

definitions/checks/check_for_newer_packages.rb
```diff
@@ -12,15 +12,54 @@ class Checks::CheckForNewerPackages < ForemanMaintain::Check
...
+                        "#{packages_with_updates.join(',')}. Please update before proceeding!"
+      fail! failure_message
+    end
+  end
+
+  def packages_and_versions(output)
+    pkgs_versions = {}
+    pkg_details = output.split("\n\n")[1].split
+    iterator = 0
+    while iterator < 6
```
evgeni, line 45: I don't really understand this iterator and the magic numbers 6 and 3

definitions/checks/check_for_newer_packages.rb
```diff
@@ -12,15 +12,54 @@ class Checks::CheckForNewerPackages < ForemanMaintain::Check
...
+    pkgs_versions
+  end
+
+  def compare_pkg_versions(output)
+    compare_pkg_versions = {}
+    packages_and_versions = packages_and_versions(output)
+    pkg_versions610 = { 'python3-pulp-2to3-migration' => version('0.12'),
+                        'tfm-rubygem-katello' => version('4.1') }
+    packages_and_versions.each do |name, version|
+      compare_pkg_versions[name] = version.<=>(pkg_versions610[name])
```
evgeni, line 61: Should this use `Gem::Version` and it's compare functionality? Comparing versions as strings does not always yield the expected results.

definitions/checks/check_for_newer_packages.rb
upadhyeammit: The check is for two packages `yum check-update` command,
1. We collect the packages and versions when exit code is 100 which is when updates available.
2. So at max it has 6 entries in array as below. Though I expect exact entries, better I should use pkg_details.length, 
~~~
['package1',
'version',
'repo',
'package1',
'version',
'repo']
~~~

definitions/checks/check_for_newer_packages.rb
upadhyeammit: The `packages_and_versions` forms the hash using `version` method which uses Gem::Version

evgeni: Ah, great, wasn't obvious to me. Thanks.

definitions/checks/check_for_newer_packages.rb
```diff
@@ -12,15 +12,54 @@ class Checks::CheckForNewerPackages < ForemanMaintain::Check
...
+    pkgs_versions = {}
+    pkg_details = output.split("\n\n")[1].split
+    iterator = 0
+    while iterator < pkg_details.length
+      break if pkg_details[iterator].nil?
+      pkg_name = pkg_details[iterator].split('.').first
+      pkg_version = version(pkg_details[iterator + 1].split('-').first)
+      pkgs_versions[pkg_name] = pkg_version
+      iterator += 3
+    end
```
evgeni, line 51:
```suggestion
  @packages.each do |pkg|
    output.lines.each do |line|
      line.strip!
      next unless line.start_with?(pkg)
      pkg_name, pkg_version, repo = line.split
      pkgs_versions[pkg_name] = version(pkg_version.split('-').first)
    end
  end
  
  how about this
```

upadhyeammit: Thats better however looping with `.lines` does not work as its splitting just after the package `python3-pulp-2to3-migration`,
~~~
"Loaded plugins: product-id, search-disabled-repos, subscription-manager\n\npython3-pulp-2to3-migration.noarch\n                             0.11.4-1.el7pc     rhel-7-server-satellite-6.9-rpms\ntfm-rubygem-katello.noarch   3.18.1.40-1.el7sat rhel-7-server-satellite-6.9-rpms"
~~~
The same logic with array works, pushing this change now!
~~~
    @packages.each do |pkg|
      pkg_details.each_with_index do |ele, index|
        next unless ele.start_with?(pkg)
        pkgs_versions[pkg] = version(pkg_details[index+1].split('-').first)
        break
      end
    end
~~~

evgeni: Eww, yeah, yum does this sometimes with too long lines :/

Thanks!

evgeni:
I'm happy!

@ianballou @jlsherrill, you?

jlsherrill:
code looks good, testing!

jlsherrill (approved):

PR #545, opened 2021-10-04
Fixes #33622 - manually detect the check_for_newer_packages

upadhyeammit:


Gauravtalreja1 (approved):
Tested this PR on 6.10 Sat/Cap, health list seems to work.

PR #546, opened 2021-10-05
Fixes #33620 - Command for upgrading foreman-maintain to major version

upadhyeammit:


definitions/procedures/repositories/backup_enabled_repos.rb
```diff
@@ -0,0 +1,18 @@
...
+      description 'Stores enabled repositories in yaml file'
+    end
+
+    def run
+      enabled_repos_ids = feature(:system_repos).enabled_repos_ids
+      backup_dir = File.expand_path(ForemanMaintain.config.backup_dir)
+      unless enabled_repos_ids.empty?
+        File.open(File.join(backup_dir, 'enabled_repos.yml'), 'w') do |repos_file|
+          repos_file.puts enabled_repos_ids.to_yaml
+        end
```
ekohl, line 14: Nit: this can be simplified:
```suggestion
        File.write(File.join(backup_dir, 'enabled_repos.yml'), enabled_repos_ids.to_yaml)
```

definitions/scenarios/self_upgrade.rb
```diff
@@ -0,0 +1,90 @@
...
+      @maintenance_repo ||= "rhel-#{el_major_version}-server-satellite-maintenance-#{version}-rpms"
+    end
+
+    def stored_enabled_repos_ids
+      unless defined?(@stored_enabled_repos_ids)
+        @stored_enabled_repos_ids = []
+        backup_dir = File.expand_path(ForemanMaintain.config.backup_dir)
+        File.open(File.join(backup_dir, 'enabled_repos.yml'), 'r') do |repos_file|
+          @stored_enabled_repos_ids = YAML.load(repos_file.read)
+        end
```
ekohl, line 33: This file doesn't have to exist. `BackupEnabledRepos` only writes it if it's non-empty. I think this would fail then. It also won't be in older backups. You probably need to deal with that to prevent exceptions. You could also use `File.read` instead of `File.open` + `.read`, but that's a small nit.
```suggestion
        path = File.expand_path('enabled_repos.yml', ForemanMaintain.config.backup_dir)
        @stored_enabled_repos_ids = File.file?(path) ? YAML.load(File.read(path)) : []
```

definitions/scenarios/self_upgrade.rb
```diff
@@ -0,0 +1,90 @@
...
+      @target_version ||= context.get(:target_version)
+    end
+
+    def current_version
+      feature(:instance).downstream.current_minor_version
+    end
+
+    def maintenance_repo(version)
+      @maintenance_repo ||= "rhel-#{el_major_version}-server-satellite-maintenance-#{version}-rpms"
+    end
```
ekohl, line 25: This is an odd caching because it has a parameter. Consider this:
```ruby
self_upgrade_base.maintenance_repo('6.9')
self_upgrade_base.maintenance_repo('6.10')
```

The first invocation will return 6.9 (as you'd expect, but the second will also return 6.9. That's surprising. It looks like this isn't used elsewhere either so I'd drop it.
```suggestion
    def maintenance_repo(version)
      "rhel-#{el_major_version}-server-satellite-maintenance-#{version}-rpms"
    end
```

lib/foreman_maintain/cli/self_upgrade_command.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      def allow_major_version_upgrade_only
+        begin
+          next_version = Gem::Version.new(target_version)
+        rescue ArgumentError => err
+          raise Error::UsageError, "Invalid version! #{err}"
+        end
+        output = current_downstream_version.<=>(next_version)
+        unless output.negative?
```
ekohl, line 30: Can't you write this:
```suggestion
        if current_downstream_version <= next_version
```

definitions/scenarios/self_upgrade.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    def all_maintenance_repos
+      search_id = "rhel-#{el_major_version}-server-satellite-maintenance-"
+      all_maintenance_repos = []
+      stored_enabled_repos_ids.each do |id|
+        next unless id.start_with?(search_id)
+        all_maintenance_repos << id
+      end
+      all_maintenance_repos
```
ekohl, line 48: I'd write it this way:
```suggestion
      repo_prefix = "rhel-#{el_major_version}-server-satellite-maintenance-"
      stored_enabled_repos_ids.filter { |id| id.start_with?(repo_prefix) }
```

You can even extract the `repo_prefix` definition to a separate method and reuse it in `def maintenance_repo()`. That said, you should consider that (AFAIK) `el_major_version` isn't cached and then it can become expensive.

definitions/scenarios/self_upgrade.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    # Need to remove this before merging
+    def skip_repo_enablement?
+      !!context.get(:skip_repo_enablement)
+    end
+
+    def stored_enabled_repos_ids
+      unless defined?(@stored_enabled_repos_ids)
+        @stored_enabled_repos_ids = []
```
ekohl, line 34: There's no need to define this if you're going to redefine it 2 lines below:
```suggestion
```

definitions/scenarios/self_upgrade.rb
```diff
@@ -0,0 +1,96 @@
...
+      "rhel-#{el_major_version}-server-satellite-maintenance-#{version}-rpms"
+    end
+
+    # Need to remove this before merging
+    def skip_repo_enablement?
+      !!context.get(:skip_repo_enablement)
+    end
+
+    def stored_enabled_repos_ids
+      unless defined?(@stored_enabled_repos_ids)
```
ekohl, line 33: I think you can also use this pattern but do check that Ruby 2.0 allows it.
```ruby
@stored_enabled_repos_ids ||= begin
  path = ...
  File.file?(path) ? YAML.load(File.read(path)) : []
end
```

definitions/scenarios/self_upgrade.rb
```diff
@@ -0,0 +1,96 @@
...
+      all_maintenance_repos
+    end
+
+    def repos_ids_to_reenable
+      repos_ids_to_reenable = stored_enabled_repos_ids - all_maintenance_repos
+      repos_ids_to_reenable << if current_version == '6.10'
+                                 maintenance_repo(6)
+                               else
+                                 maintenance_repo(current_version)
+                               end
```
ekohl, line 57: Given `current_version` is only used once, you could also write a method that's `maintenance_repo_version` and use that here so you end up with:
```suggestion
      repos_ids_to_reenable << maintenance_repo(maintenance_repo_version)
```

ehelms:
@upadhyeammit looks good! I think this can be squashed and merged.

upadhyeammit:
@ehelms there were few things I have handled in latest commit. The confine block is not working for the scenario out of the box, need to check what is missing for that. Accordingly as of now this scenario is limited using if condition in compose. And I will raise another pr to change it for PuppetDisable command as well.

definitions/scenarios/self_upgrade.rb
```diff
@@ -21,7 +21,9 @@ def current_version
     end
 
     def maintenance_repo_id(version)
-      return ENV['maintenance_repo'] unless ENV['maintenance_repo'].empty?
+      unless ENV['maintenance_repo'].nil?
+        return ENV['maintenance_repo'] unless ENV['maintenance_repo'].empty?
+      end
```
ehelms, line 26: Could you do this as:

```
return ENV['maintenance_repo'] if ENV['maintenance_repo']
```

ekohl: No because `.empty?` it's still good to check for a non-empty string. In Rails and Ruby 2.5 you can do `var&.present?` but this isn't rails (so `.present?` doesn't exist) and we don't have the safe operator due to Ruby 2.0.

I'd write this as:
```ruby
if (repo = ENV['maintenance_repo'])
  return repo unless repo.empty?
end
```

upadhyeammit:
@ehelms changes have been pushed, if there are no more comments then I will squash and merge.

ehelms (approved):

PR #547, opened 2021-10-12
Fixes #33686 - 6.10 upgrade check & warn if user has HTTPS proxy

ianballou:
This PR adds a check to see if users have any HTTPS proxies before continuing with the 6.10 upgrade. They are given an option about whether or not to continue the upgrade.

Example output:

```
Check for HTTPS proxies from DB: 
Warning: Syncing repositories via HTTPS proxies is temporarily unsupported.
The following proxies use HTTPS: https proxy.
Continue upgrade?, [y(yes), q(quit)] q
                                                                      [ABORTED] 
```

### DIFF
--- a/definitions/checks/foreman/check_https_proxies.rb
+++ b/definitions/checks/foreman/check_https_proxies.rb
@@ -1,0 +1,34 @@
+module Checks
+  class CheckHttpsProxies < ForemanMaintain::Check
+    metadata do
+      label :https_proxies
+      for_feature :foreman_database
+      description 'Check for HTTPS proxies from the database'
+      manual_detection
+    end
+
+    def run
+      https_proxies = find_https_proxies
+      unless https_proxies.empty?
+        https_proxy_names = https_proxies.map { |proxy| proxy['name'] }
+        question = "Syncing repositories through an 'HTTP Proxy' that uses the HTTPS\n"\
+                    "protocol is not supported directly with Satellite 6.10.\n"\
+                    "The following proxies use HTTPS: #{https_proxy_names.join(', ')}.\n"\
+                    "For a suggested solution see https://access.redhat.com/solutions/6414991\n"\
+                    'Continue upgrade?'
+        answer = ask_decision(question, actions_msg: 'y(yes), q(quit)')
+        abort! if answer != :yes
+      end
+    end
+
+    def find_https_proxies
+      feature(:foreman_database).query(self.class.query_to_get_https_proxies)
+    end
+
+    def self.query_to_get_https_proxies
+      <<-SQL
+        SELECT \"http_proxies\".* FROM \"http_proxies\" WHERE (http_proxies.url ilike 'https://%')
+      SQL
+    end
+  end
+end
--- a/definitions/scenarios/upgrade_to_satellite_6_10.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_10.rb
@@ -29,6 +29,7 @@ module Scenarios::Satellite_6_10
       add_step(Checks::CheckForNewerPackages.new(:packages => [foreman_plugin_name('katello'),
                                                                'python3-pulp-2to3-migration'],
                                                  :manual_confirmation_version => '6.9'))
+      add_step(Checks::CheckHttpsProxies)
       add_steps(find_checks(:default))
       add_steps(find_checks(:pre_upgrade))
 


definitions/checks/foreman/check_https_proxies.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def run
+        https_proxies = find_https_proxies
+        unless https_proxies.empty?
+          https_proxy_names = https_proxies.map { |proxy| proxy['name'] }
+          question = 'Warning: Syncing repositories via HTTPS proxies is temporarily '\
+                      "unsupported.\n"\
+                      "The following proxies use HTTPS: #{https_proxy_names.join(',')}.\n"\
+                      'Continue upgrade?'
```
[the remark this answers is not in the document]
upadhyeammit: and for kcs you can use `KnowledgeBaseArticle` procedure

definitions/checks/foreman/check_https_proxies.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def run
+      https_proxies = find_https_proxies
+      unless https_proxies.empty?
+        https_proxy_names = https_proxies.map { |proxy| proxy['name'] }
+        question = "Syncing repositories through an \'HTTP Proxy\' that uses the https:// "\
+                    "protocol is not supported directly with Satellite 6.10.\n"\
+                    "The following proxies use HTTPS: #{https_proxy_names.join(',')}.\n"\
+                    "For a workaround see https://access.redhat.com/solutions/6414991\n"\
```
ianballou, line 17: I put the KCS here too because the article procedure doesn't seem to run if the user aborts the upgrade.  I feel like they might want to look at the KCS article even if they abort.

definitions/checks/foreman/check_https_proxies.rb
```diff
@@ -0,0 +1,35 @@
...
+                    'Continue upgrade?'
+        answer = ask_decision(question, actions_msg: 'y(yes), q(quit)')
+        abort! if answer != :yes
```
upadhyeammit, line 20: The whole pre_upgrade steps are running in fail_slow strategy so the abort does not help here. As it wont actually abort the upgrade instead it will just show [ABORTED] as status for the check. 
I like to know if aborting upgrade is required if https proxies are there? I think warning is enough? and if warning is enough then do we need question for the user?

definitions/checks/foreman/check_https_proxies.rb
```diff
@@ -0,0 +1,35 @@
...
+        next_steps.concat([Procedures::KnowledgeBaseArticle.new(:doc => 'use_https_proxy_sat610')])
```
upadhyeammit, line 21: Just adding next_steps wont work as I dont see its getting executed. However even if next_steps would execute it means we will show the same message twice if user proceeds with `yes` option?

definitions/checks/foreman/check_https_proxies.rb
```diff
@@ -0,0 +1,35 @@
+module Checks
+  class CheckHttpsProxies < ForemanMaintain::Check
+    metadata do
+      label :https_proxies
+      for_feature :foreman_database
+      description 'Check for HTTPS proxies from DB'
```
upadhyeammit, line 6:
```suggestion
      description 'Check for HTTPS proxies from database'
```

upadhyeammit (changes requested):

definitions/checks/foreman/check_https_proxies.rb
ianballou: When I tested it the upgrade aborted after all of the steps finished running.  The abort is necessary because users with HTTPS proxies will be unable to sync any content until either the proxies are removed or the steps in the KCS article are accomplished

```

The processing was aborted by user during the following steps:

  [https-proxies]
```

definitions/checks/foreman/check_https_proxies.rb
```diff
@@ -0,0 +1,34 @@
...
+      for_feature :foreman_database
+      description 'Check for HTTPS proxies from the database'
+      manual_detection
+    end
+
+    def run
+      https_proxies = find_https_proxies
+      unless https_proxies.empty?
+        https_proxy_names = https_proxies.map { |proxy| proxy['name'] }
+        question = "Syncing repositories through an \'HTTP Proxy\' that uses the https:// "\
```
evgeni, line 14: You shouldn't need to escape the single quotes here?

Also, I'd say HTTPS protocol, not https:// (that's more of an URI format)

PR #550, opened 2021-10-25
Fixes #33734 - add db key for pulpcore backup

jlsherrill:


jlsherrill:
merging after discussion with @upadhyeammit

PR #553, opened 2021-10-28
Fixes #33281 - run installer with upgrade flags after restore

evgeni:


jlsherrill:
the rake task can also be useful for zstreams (if you're restoring on a newer z-stream sometimes upgrade steps are added to resolve issues).  But it should be run with this change. 

I don't know that comparing versions is worth the headache to save a single installer run

evgeni:
@jlsherrill `upgrade:run` is executed for Katello as part of an installer run, but not Satellite.

jlsherrill:
@evgeni does anything run it as part of satellite?

evgeni:
`foreman-maintain upgrade …` does run it, yes.

jlsherrill:
looks good to me!

jlsherrill (approved):

PR #555, opened 2021-11-02
Fixes #33829 - add disable Puppet command

ezr-ondrej:
Adds command to disable Puppet plugin and remove all the Puppet data.
Adds also command to re-enable Puppet.

### DIFF
--- a/definitions/checks/foreman/check_puppet_capsules.rb
+++ b/definitions/checks/foreman/check_puppet_capsules.rb
@@ -1,0 +1,43 @@
+module Checks
+  class CheckPuppetCapsules < ForemanMaintain::Check
+    metadata do
+      label :puppet_capsules
+      for_feature :foreman_database
+      description 'Check for Puppet capsules from the database'
+      manual_detection
+    end
+
+    def run
+      puppet_proxies = find_puppet_proxies.reject { |proxy| local_proxy?(proxy['url']) }
+      unless puppet_proxies.empty?
+        names = puppet_proxies.map { |proxy| proxy['name'] }
+        print('You have proxies with Puppet feature enabled, '\
+              "please disable Puppet on all proxies first.\n"\
+              "The following proxies have Puppet feature: #{names.join(', ')}.")
+        abort!
+      end
+    end
+
+    def find_puppet_proxies
+      feature(:foreman_database).query(puppet_proxies_query)
+    end
+
+    private
+
+    def local_proxy?(url)
+      URI.parse(url).hostname.casecmp(hostname) == 0
+    end
+
+    def puppet_proxies_query
+      <<-SQL
+        SELECT \"smart_proxies\".*
+          FROM \"smart_proxies\"
+            INNER JOIN \"smart_proxy_features\"
+              ON \"smart_proxies\".\"id\" = \"smart_proxy_features\".\"smart_proxy_id\"
+            INNER JOIN \"features\"
+              ON \"features\".\"id\" = \"smart_proxy_features\".\"feature_id\"
+          WHERE \"features\".\"name\" = 'Puppet'
+      SQL
+    end
+  end
+end
--- a/definitions/procedures/puppet/remove_puppet_data.rb
+++ b/definitions/procedures/puppet/remove_puppet_data.rb
@@ -1,0 +1,21 @@
+module Procedures::Puppet
+  class RemovePuppetData < ForemanMaintain::Procedure
+    metadata do
+      description 'Remove Puppet data'
+    end
+
+    def run
+      execute!('foreman-rake purge:puppet')
+      execute!('rm -r ' + files_to_purge.join(' '))
+    end
+
+    private
+
+    def files_to_purge
+      %w[
+        /etc/puppetlabs
+        /opt/puppetlabs/server/data
+      ]
+    end
+  end
+end
--- a/definitions/scenarios/puppet.rb
+++ b/definitions/scenarios/puppet.rb
@@ -1,0 +1,22 @@
+module ForemanMaintain::Scenarios
+  module Puppet
+    class RemovePuppet < ForemanMaintain::Scenario
+      metadata do
+        description 'Remove Puppet feature'
+        tags :puppet_disable
+        label :puppet_disable
+        param :remove_data, 'Purge the Puppet data after disabling plugin'
+        confine do
+          check_min_version('foreman', '3.0') || check_min_version('foreman-proxy', '3.0')
+        end
+        manual_detection
+      end
+
+      def compose
+        add_step(Checks::CheckPuppetCapsules) if server?
+        add_step(Procedures::Puppet::RemovePuppet)
+        add_step(Procedures::Puppet::RemovePuppetData) if context.get(:remove_data)
+      end
+    end
+  end
+end
--- a/lib/foreman_maintain/cli.rb
+++ b/lib/foreman_maintain/cli.rb
@@ -11,6 +11,7 @@ require 'foreman_maintain/cli/restore_command'
 require 'foreman_maintain/cli/maintenance_mode_command'
 require 'foreman_maintain/cli/packages_command'
 require 'foreman_maintain/cli/content_command'
+require 'foreman_maintain/cli/plugin_command'
 
 module ForemanMaintain
   module Cli
@@ -25,6 +26,7 @@ module ForemanMaintain
       subcommand 'packages', 'Lock/Unlock package protection, install, update', PackagesCommand
       subcommand 'advanced', 'Advanced tools for server maintenance', AdvancedCommand
       subcommand 'content', 'Content related commands', ContentCommand
+      subcommand 'plugin', 'Manage optional plugins on your server', PluginCommand
       subcommand 'maintenance-mode', 'Control maintenance-mode for application',
                  MaintenanceModeCommand
       if ForemanMaintain.detector.feature(:satellite) &&
--- a/lib/foreman_maintain/concerns/system_helpers.rb
+++ b/lib/foreman_maintain/concerns/system_helpers.rb
@@ -196,21 +196,24 @@ module ForemanMaintain
       end
 
       def foreman_plugin_name(plugin)
-        if debian?
-          plugin.tr!('_', '-')
-        end
+        plugin = plugin.tr('_', '-') if debian?
         ruby_prefix + plugin
       end
 
       def proxy_plugin_name(plugin)
         if debian?
-          plugin.tr!('_', '-')
+          plugin = plugin.tr('_', '-')
           proxy_plugin_prefix = 'smart-proxy-'
         else
           proxy_plugin_prefix = 'smart_proxy_'
         end
         scl = check_min_version('foreman-proxy', '2.0')
         ruby_prefix(scl) + proxy_plugin_prefix + plugin
+      end
+
+      def hammer_plugin_name(plugin)
+        plugin = plugin.tr('_', '-') if debian?
+        [hammer_package, plugin].join(debian? ? '-' : '_')
       end
 
       def hammer_package
--- a/test/lib/support/definitions/scenarios/puppet.rb
+++ b/test/lib/support/definitions/scenarios/puppet.rb
@@ -1,0 +1,15 @@
+module Scenarios::Puppet
+  class RemovePuppet < ForemanMaintain::Scenario
+    metadata do
+      description 'Remove Puppet feature'
+      tags :puppet_disable
+      label :puppet_disable
+      param :remove_data, 'Purge the Puppet data after disabling plugin'
+      manual_detection
+    end
+
+    def compose
+      add_steps([Checks::Dummy::Success])
+    end
+  end
+end
--- a/definitions/procedures/puppet/remove_puppet.rb
+++ b/definitions/procedures/puppet/remove_puppet.rb
@@ -1,0 +1,50 @@
+module Procedures::Puppet
+  class RemovePuppet < ForemanMaintain::Procedure
+    metadata do
+      description 'Remove Puppet feature'
+      confine do
+        feature(:puppet_server) &&
+          (check_min_version('foreman', '3.0') || check_min_version('foreman-proxy', '3.0'))
+      end
+      advanced_run false
+    end
+
+    def run
+      services = feature(:foreman_server).services + feature(:dynflow_sidekiq).services
+      Procedures::Service::Stop.new(:only => services)
+      execute!('foreman-rake db:migrate VERSION=0 SCOPE=foreman_puppet') if server_with_puppet?
+      feature(:installer).run(installer_arguments_disabling_puppet.join(' '), :interactive => false)
+      packages_action(:remove, packages_to_remove, :assumeyes => true)
+    end
+
+    private
+
+    def server_with_puppet?
+      find_package(foreman_plugin_name('foreman_puppet'))
+    end
+
+    def installer_arguments_disabling_puppet
+      answers = feature(:installer).answers
+
+      options = []
+      options << '--no-enable-puppet' if answers.key?('puppet')
+      options << '--no-enable-foreman-plugin-puppet' if answers['foreman::plugin::puppet']
+      options << '--no-enable-foreman-cli-puppet' if answers['foreman::cli::puppet']
+      if answers['foreman_proxy']
+        options << '--foreman-proxy-puppet false'
+        options << '--foreman-proxy-puppetca false'
+      end
+      options << '--foreman-proxy-content-puppet false' if answers.key?('foreman_proxy_content')
+      options
+    end
+
+    def packages_to_remove
+      packages = ['puppetserver']
+      if server_with_puppet?
+        packages << foreman_plugin_name('foreman_puppet')
+        packages << hammer_plugin_name('foreman_puppet')
+      end
+      packages
+    end
+  end
+end
--- a/lib/foreman_maintain/cli/plugin_command.rb
+++ b/lib/foreman_maintain/cli/plugin_command.rb
@@ -1,0 +1,14 @@
+module ForemanMaintain
+  module Cli
+    class PluginCommand < Base
+      subcommand 'purge-puppet', 'Remove the Puppet feature' do
+        option ['-f', '--remove-all-data'], :flag, 'Purge all the Puppet data',
+               :attribute_name => :remove_data
+
+        def execute
+          run_scenarios_and_exit(Scenarios::Puppet::RemovePuppet.new(:remove_data => remove_data?))
+        end
+      end
+    end
+  end
+end
--- a/test/data/puppet/foreman-answers-default.yaml
+++ b/test/data/puppet/foreman-answers-default.yaml
@@ -1,0 +1,11 @@
+# foreman-answers-default.yaml
+# scenario: foreman
+# This is what foreman-installer ships out of the box, stripped of unrelated entries
+---
+foreman: {}
+foreman::cli: true
+foreman::cli::puppet: true
+foreman::plugin::puppet: true
+foreman_proxy: {}
+puppet:
+  server: true
--- a/test/data/puppet/foreman-answers-no-foreman-with-puppet.yaml
+++ b/test/data/puppet/foreman-answers-no-foreman-with-puppet.yaml
@@ -1,0 +1,11 @@
+# foreman-answers-no-foreman-with-puppet.yaml
+# scenario: foreman
+# This is only a foreman-proxy with Puppet enabled
+---
+foreman: false
+foreman::cli: false
+foreman::cli::puppet: false
+foreman::plugin::puppet: false
+foreman_proxy: {}
+puppet:
+  server: true
--- a/test/data/puppet/foreman-proxy-content-with-puppet.yaml
+++ b/test/data/puppet/foreman-proxy-content-with-puppet.yaml
@@ -1,0 +1,14 @@
+# foreman-proxy-content-answers-with-puppet.yaml
+# scenario: foreman-proxy-content
+# This is where a user enabled puppet
+---
+foreman_proxy:
+  puppet: true
+  puppetca: true
+foreman_proxy_content:
+  puppet: true
+puppet:
+  server: true
+  puppet_server_foreman_ssl_ca: /etc/pki/katello/puppet/puppet_client_ca.crt
+  puppet_server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
+  puppet_server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
--- a/test/data/puppet/katello-answers-with-puppet.yaml
+++ b/test/data/puppet/katello-answers-with-puppet.yaml
@@ -1,0 +1,18 @@
+# katello-answers-with-puppet.yaml
+# scenario: katello
+# This is where a user enabled puppet
+---
+foreman: {}
+foreman::cli: true
+foreman::cli::puppet: true
+foreman::plugin::puppet: true
+foreman_proxy:
+  puppet: true
+  puppetca: true
+foreman_proxy_content:
+  puppet: true
+puppet:
+  server: true
+  puppet_server_foreman_ssl_ca: /etc/pki/katello/puppet/puppet_client_ca.crt
+  puppet_server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
+  puppet_server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
--- a/test/definitions/procedures/puppet/remove_puppet_test.rb
+++ b/test/definitions/procedures/puppet/remove_puppet_test.rb
@@ -1,0 +1,66 @@
+require 'test_helper'
+
+describe Procedures::Puppet::RemovePuppet do
+  include DefinitionsTestHelper
+
+  subject do
+    Procedures::Puppet::RemovePuppet.new
+  end
+
+  before do
+    mock_installer_package('foreman-installer')
+  end
+
+  describe '#installer_arguments_disabling_puppet' do
+    it 'works for foreman default' do
+      stub_installer_answers('foreman-answers-default.yaml')
+      expected_arguments = [
+        '--no-enable-foreman-cli-puppet',
+        '--no-enable-foreman-plugin-puppet',
+        '--foreman-proxy-puppet false',
+        '--foreman-proxy-puppetca false',
+        '--no-enable-puppet'
+      ]
+      _(subject.send(:installer_arguments_disabling_puppet).sort).must_equal expected_arguments.sort
+    end
+
+    it 'works for proxy with puppet' do
+      stub_installer_answers('foreman-answers-no-foreman-with-puppet.yaml')
+      expected_arguments = [
+        '--foreman-proxy-puppet false',
+        '--foreman-proxy-puppetca false',
+        '--no-enable-puppet'
+      ]
+      _(subject.send(:installer_arguments_disabling_puppet).sort).must_equal expected_arguments.sort
+    end
+
+    it 'works for katello' do
+      stub_installer_answers('katello-answers-with-puppet.yaml')
+      expected_arguments = [
+        '--no-enable-foreman-cli-puppet',
+        '--no-enable-foreman-plugin-puppet',
+        '--foreman-proxy-puppet false',
+        '--foreman-proxy-puppetca false',
+        '--foreman-proxy-content-puppet false',
+        '--no-enable-puppet'
+      ]
+      _(subject.send(:installer_arguments_disabling_puppet).sort).must_equal expected_arguments.sort
+    end
+
+    it 'works for proxy-content' do
+      stub_installer_answers('foreman-proxy-content-with-puppet.yaml')
+      expected_arguments = [
+        '--foreman-proxy-puppet false',
+        '--foreman-proxy-puppetca false',
+        '--foreman-proxy-content-puppet false',
+        '--no-enable-puppet'
+      ]
+      _(subject.send(:installer_arguments_disabling_puppet).sort).must_equal expected_arguments.sort
+    end
+  end
+
+  def stub_installer_answers(filename)
+    file = "test/data/puppet/#{filename}"
+    assume_feature_present(:installer, configuration: { answer_file: file })
+  end
+end
--- a/test/lib/cli/plugin_command_test.rb
+++ b/test/lib/cli/plugin_command_test.rb
@@ -1,0 +1,51 @@
+require 'test_helper'
+
+require 'foreman_maintain/cli'
+
+module ForemanMaintain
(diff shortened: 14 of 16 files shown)


definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,35 @@
...
+      ]
+    end
+
+    def packages_to_remove
+      [
+        foreman_plugin_name('foreman_puppet'),
+        hammer_plugin_name('foreman_puppet'),
+        'puppetserver'
+      ]
+    end
```
ezr-ondrej, line 33: @ekohl could you please check if this list of options and packages is correct?

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,35 @@
...
+      packages_action(:remove, packages_to_remove, :assumeyes => true)
+    end
+
+    private
+
+    def installer_arguments_disabling_puppet
+      [
+        '--no-enable-puppet',
+        '--no-enable-foreman-plugin-puppet',
+        '--no-enable-foreman-cli-puppet',
```
ekohl, line 22: This will fail on a scenario where foreman isn't exposed, such as the `foreman-proxy-content` scenario. You need to account for that

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,35 @@
...
+        '--foreman-proxy-content-puppet false'
```
ekohl, line 23: This at least deserves a TODO that this is only valid in a katello-based scenario. It means this can't be used in the `foreman` installer scenario. Ideally speaking it would skip this if the scenario is `foreman`.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -213,6 +213,13 @@ def proxy_plugin_name(plugin)
         ruby_prefix(scl) + proxy_plugin_prefix + plugin
       end
 
+      def hammer_plugin_name(plugin)
+        if debian?
+          plugin.tr!('_', '-')
+        end
```
ekohl, line 219: I'm not a fan of in place replacements. How about:
```suggestion
        plugin = plugin.tr('_', '-') if debian?
```

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def run
+      feature(:installer).run(installer_arguments_enabling_puppet.join(' '), :interactive => false)
+    end
+
+    private
+
+    def installer_arguments_enabling_puppet
+      [
+        '--enable-puppet',
```
ekohl, line 15: For vanilla foreman you do want the server enabled it is simply:
```suggestion
        '--enable-puppet --puppet-server true',
```

Katello additionally needs you need to specify the certificates. See https://github.com/theforeman/foreman-documentation/pull/799 or https://github.com/theforeman/forklift/blob/64ae4e81d0ae2cad13a4c925b0b4ccb0c923057e/bats/fb-test-puppet.bats#L18 for the horrible mess.

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,22 @@
...
+        '--enable-foreman-plugin-puppet',
+        '--enable-foreman-cli-puppet',
+        '--foreman-proxy-content-puppet true'
```
ekohl, line 18: Same comments as in the disable part about scenarios.

definitions/checks/foreman/check_puppet_capsules.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+    end
+
+    def find_puppet_proxies
+      feature(:foreman_database).query(self.class.puppet_proxies_query)
+    end
+
+    def self.puppet_proxies_query
+      <<-SQL
+        SELECT \"smart_proxies\".*
```
ekohl, line 28: I wonder why these need to be escaped.

ekohl (changes requested):

definitions/procedures/puppet/disable_puppet.rb
ezr-ondrej: You mean the proxy scenario will fail? I hope I've accounted for that now hopefully.

lib/foreman_maintain/concerns/system_helpers.rb
ezr-ondrej: I've changed it everywhere.

definitions/checks/foreman/check_puppet_capsules.rb
ezr-ondrej: I've copied it over from http_proxies and it works so I didn't question it :shrug:

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,45 @@
...
+    private
+
+    def server_with_puppet_plugin?
+      find_package(foreman_plugin_name('foreman_puppet'))
+    end
+
+    def installer_arguments_disabling_puppet
+      options = []
+      options << '--no-enable-puppet'
+      options << '--puppet-server false'
```
ekohl: This is implied by `--no-enable-puppet`. That it now works is a bug (https://projects.theforeman.org/issues/33821)

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,46 @@
...
+        '--puppet-server false',
+        '--foreman-proxy-puppet false',
+        '--foreman-proxy-puppetca false',
+        '--foreman-proxy-content-puppet false'
+      ]
+      if server_with_puppet_plugin?
+        options << '--no-enable-foreman-plugin-puppet'
+        options << '--no-enable-foreman-cli-puppet'
+      end
+      options
```
ekohl: This is where I'm a bit limited on my knowledge, but going by `definitions/features/katello.rb` we can have use `feature(:installer).answers` to get the installer answers.

From there on we can detect what's available. So this would make it fully dynamic. Technically the check if `puppet` is present is redundant since it's always present:

```suggestion
    def installer_arguments_disabling_puppet
      answers = feature(:installer).answers

      options = []
      options << '--no-enable-puppet' if answers.key?('puppet')
      options << '--no-enable-foreman-plugin-puppet' if answers.key?('foreman::plugin::puppet')
      options << '--no-enable-foreman-cli-puppet' if answers.key?('foreman::cli::puppet')
      if answers['foreman_proxy']
        options << '--foreman-proxy-puppet false'
        options << '--foreman-proxy-puppetca false'
      end
      options << '--foreman-proxy-content-puppet false' if answers.key?('foreman_proxy_content')
      options
    end
```

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,32 @@
...
+    def installer_arguments_enabling_puppet
+      options = [
+        '--enable-puppet',
+        '--puppet-server true',
+        '--foreman-proxy-puppet true',
+        '--foreman-proxy-puppetca true',
+        '--foreman-proxy-content-puppet true',
+        '--puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt',
+        '--puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt',
+        '--puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key'
```
ekohl, line 23: Can we hide this behind an if `answers.key?('foreman_proxy_content')` similar to the disable part? That way it works for vanilla Foreman too.

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,32 @@
...
+      ]
+      if server?
+        options << '--enable-foreman-plugin-puppet'
+        options << '--enable-foreman-cli-puppet'
```
ekohl, line 27: Please check `if answers['foreman::cli']` because if the CLI is disabled, you can't enable the plugin.

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,32 @@
...
+        '--puppet-server true',
+        '--foreman-proxy-puppet true',
+        '--foreman-proxy-puppetca true',
+        '--foreman-proxy-content-puppet true',
+        '--puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt',
+        '--puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt',
+        '--puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key'
+      ]
+      if server?
+        options << '--enable-foreman-plugin-puppet'
```
ekohl, line 26: Please check `if answers['foreman']`. Then you can also drop `if server?`.

That way it's compatible with upstream Foreman users who install a proxy (without Foreman) can use it there too. This is because we don't have a proxy-only scenario in vanilla Foreman.

definitions/procedures/puppet/remove_puppet_data.rb
```diff
@@ -0,0 +1,21 @@
...
+    def run
+      execute!('foreman-rake purge:puppet')
+      execute!('rm -r ' + files_to_purge.join(' '))
+    end
+
+    private
+
+    def files_to_purge
+      %w[
+        /etc/puppetlabs
```
ekohl, line 16: My biggest issue with this is that files which belong to puppet-agent are also removed. For now this is probably OK, but I wonder what the output of `rpm -qV puppet-agent` is after running the scenario.

lib/foreman_maintain/concerns/system_helpers.rb
```diff
@@ -213,6 +211,11 @@ def proxy_plugin_name(plugin)
         ruby_prefix(scl) + proxy_plugin_prefix + plugin
       end
 
+      def hammer_plugin_name(plugin)
+        plugin = plugin.tr('_', '-') if debian?
+        [hammer_package, plugin].join(debian? ? '-' : '_')
```
ekohl, line 216: Given the commonality between these, it feels like there could be a helper that runs `result.tr('_', '-')`.

lib/foreman_maintain/cli.rb
```diff
@@ -25,6 +26,7 @@ class MainCommand < Base
       subcommand 'packages', 'Lock/Unlock package protection, install, update', PackagesCommand
       subcommand 'advanced', 'Advanced tools for server maintenance', AdvancedCommand
       subcommand 'content', 'Content related commands', ContentCommand
+      subcommand 'plugin', 'Control features of your server', PluginCommand
```
ezr-ondrej, line 29: Is this actually a good name?
Should we call it something else? `feature`?

ekohl: Ah, the naming problem. It always comes up. This is technically the first time we deal with such a thing in downstream. Before this it was always a "fixed set".

What came close was the foreman_docker removal and optional installation of Kubevirt. However, that wasn't handled via explicit maintain commands.

`feature` may conflict in naming and I think `plugin` feels natural. However, I would suggest
```suggestion
      subcommand 'plugin', 'Manage optional plugins on your server', PluginCommand
```

ezr-ondrej:
Looks much cleaner now :+1:

definitions/checks/foreman/check_puppet_capsules.rb
```diff
@@ -0,0 +1,38 @@
...
+    metadata do
+      label :puppet_capsules
+      for_feature :foreman_database
+      description 'Check for Puppet capsules from the database'
+      manual_detection
+    end
+
+    def run
+      puppet_proxies = find_puppet_proxies.reject { |proxy| proxy['url'].include?(hostname) }
+      unless puppet_proxies.empty?
```
ekohl, line 12: Can we at least parse the URL and do a proper hostname comparisons? Hostnames are not simple strings: hostnames are case-insensitive. I'd be OK with ignoring IDN (International Domain Names, which can be recognized by starting with `xn--`)

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,47 @@
...
+        options << '--foreman-proxy-puppet false'
+        options << '--foreman-proxy-puppetca false'
+      end
+      options << '--foreman-proxy-content-puppet false' if answers.key?('foreman_proxy_content')
+      options
+    end
+
+    def packages_to_remove
+      packages = ['puppetserver']
+      if server_with_puppet?
```
ekohl, line 40: Open question here to @upadhyeammit: is it safe to simply always remove all packages? I'd think so but not sure.

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,47 @@
...
+      packages_action(:remove, packages_to_remove, :assumeyes => true)
+    end
+
+    private
+
+    def server_with_puppet?
+      find_package(foreman_plugin_name('foreman_puppet'))
+    end
+
+    # TODO: This works only for a katello scenario, we should consider vanilla foreman
```
ekohl, line 22: Due to the modifications this is now solved
```suggestion
```

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,33 @@
...
+      if answers.key?('foreman_proxy')
+        options << '--foreman-proxy-puppet true'
+        options << '--foreman-proxy-puppetca true'
+      end
+      if answers.key?('foreman_proxy_content')
+        options << '--foreman-proxy-content-puppet true'
+        options << '--puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt'
+        options << '--puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt'
+        options << '--puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key'
+      end
```
ekohl, line 27: I think it's safe to assume this key exists (it should in every scenario). The question is what we do if it's disabled. In that case, enabling these two options don't make sense. If you don't enable these two options, are not passed in, passing in `--enable-puppet` (and friends) don't make sense either. So it becomes:

```suggestion
      options = []
      
      if answers['foreman_proxy']
        options += [
          '--enable-puppet',
          '--puppet-server true',
          '--foreman-proxy-puppet true',
          '--foreman-proxy-puppetca true'
        ]
        if answers.key?('foreman_proxy_content')
          options << '--foreman-proxy-content-puppet true'
          options << '--puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt'
          options << '--puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt'
          options << '--puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key'
        end
      end
```

One other thing I'm wondering about is how to deal with a Puppet setup without a CA. This can't deal with compile only servers since we're explicitly enabling the Foreman Proxy PuppetCA plugin. Should we also pass `--puppet-server-ca true`? Can a scenario support parameters like `--with-ca` or `--without-ca`?

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,33 @@
...
+      description 'Enable Puppet feature'
+    end
+
+    def run
+      feature(:installer).run(installer_arguments_enabling_puppet.join(' '), :interactive => false)
+    end
+
+    private
+
+    # TODO: This works only for a katello scenario, we should consider vanilla foreman
```
ekohl, line 13: This should be dealt with now
```suggestion
```

ezr-ondrej:
I've added `--no-ca` option for the enable command to enable deciding wheter users want to enable also CA and improved the `local_proxy?` check to parse the URI

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,43 @@
...
+      answers = feature(:installer).answers
+
+      options = []
+      options.concat(proxy_options) if answers['foreman_proxy']
+      if answers.key?('foreman_proxy_content')
+        options << '--foreman-proxy-content-puppet true'
+        options << '--puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt'
+        options << '--puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt'
+        options << '--puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key'
+      end
```
ekohl, line 24: This part should only happen `if answers['foreman_proxy']`.

ekohl (changes requested):

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,46 @@
...
+      description 'Disable Puppet feature'
+      confine do
+        feature(:puppet_server)
+      end
+    end
+
+    def run
+      execute!('foreman-rake db:migrate VERSION=0 SCOPE=foreman_puppet') if server_with_puppet?
+      feature(:installer).run(installer_arguments_disabling_puppet.join(' '), :interactive => false)
+      packages_action(:remove, packages_to_remove, :assumeyes => true)
```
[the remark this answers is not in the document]
ezr-ondrej: Are we sure? This could be restarted after the installer run and that would be less safe but much less downtime

ekohl:
I wrote out some test scenarios. I hope I covered them all and they're correct. I may have missed things. I certainly didn't capture the non-CA case but they should cover the most important cases.

# foreman

## foreman-answers-default.yaml

```yaml
# foreman-answers-default.yaml
# scenario: foreman
# This is what foreman-installer ships out of the box, stripped of unrelated entries
---
foreman: {}
foreman::cli: true
foreman::cli::puppet: true
foreman::plugin::puppet: true
foreman_proxy: {}
puppet:
  server: true
```

I would expect the following arguments to disable it:

```
--no-enable-foreman-cli-puppet
--no-enable-foreman-plugin-puppet
--foreman-proxy-puppet false
--foreman-proxy-puppetca false
--no-enable-puppet
```

## foreman-answers-without-puppet.yaml

```yaml
# foreman-answers-without-puppet.yaml
# scenario: foreman
# Here a user disabled Puppet
---
foreman: {}
foreman::cli: true
foreman::cli::puppet: false
foreman::plugin::puppet: false
foreman_proxy:
  puppet: false
  puppetca: false
puppet: false
```

I would expect the following arguments to enable it:

```
--enable-foreman-cli-puppet
--enable-foreman-plugin-puppet
--foreman-proxy-puppet true
--foreman-proxy-puppetca true
--enable-puppet
--puppet-server true
--puppet-server-ca true
```

## foreman-answers-no-foreman-with-puppet.yaml

```yaml
# foreman-answers-no-foreman-with-puppet.yaml
# scenario: foreman
# This is only a foreman-proxy with Puppet enabled
---
foreman: false
foreman::cli: false
foreman::cli::puppet: false
foreman::plugin::puppet: false
foreman_proxy: {}
puppet:
  server: true
```

I would expect the following arguments to disable it:
```
--foreman-proxy-puppet false
--foreman-proxy-puppetca false
--no-enable-puppet
```

## foreman-answers-no-foreman-without-puppet.yaml

```yaml
# foreman-answers-no-foreman-without-puppet.yaml
# scenario: foreman
# This is only a foreman-proxy with Puppet disabled
---
foreman: false
foreman::cli: false
foreman::cli::puppet: false
foreman::plugin::puppet: false
foreman_proxy:
  puppet: false
  puppetca: false
puppet: false
```

I would expect the following arguments to enable it:
```
--foreman-proxy-puppet true
--foreman-proxy-puppetca true
--enable-puppet
--puppet-server true
--puppet-server-ca true
```

# Katello

## katello-answers-default.yaml

```yaml
# katello-answers-default.yaml
# scenario: katello
# This is what foreman-installer ships out of the box, stripped of unrelated entries
---
foreman: {}
foreman::cli::true
foreman::cli::puppet: false
foreman::plugin::puppet: false
foreman_proxy: {}
foreman_proxy_content: true
puppet: false
```

I would expect the following arguments to enable it:
```
--enable-foreman-cli-puppet
--enable-foreman-plugin-puppet
--foreman-proxy-puppet true
--foreman-proxy-puppetca true
--foreman-proxy-content-puppet true
--enable-puppet
--puppet-server true
--puppet-server-ca true
--puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt
--puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt
--puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key
```

## katello-answers-with-puppet.yaml

```yaml
# katello-answers-with-puppet.yaml
# scenario: katello
# This is where a user enabled puppet
---
foreman: {}
foreman::cli::true
foreman::cli::puppet: true
foreman::plugin::puppet: true
foreman_proxy:
  puppet: true
  puppetca: true
foreman_proxy_content:
  puppet: true
puppet:
  server: true
  puppet_server_foreman_ssl_ca: /etc/pki/katello/puppet/puppet_client_ca.crt
  puppet_server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
  puppet_server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
```

I would expect the following arguments to disable it:
```
--no-enable-foreman-cli-puppet
--no-enable-foreman-plugin-puppet
--foreman-proxy-puppet false
--foreman-proxy-puppetca false
--foreman-proxy-content-puppet false
--no-enable-puppet
```

# foreman-proxy-content

## foreman-proxy-content-default.yaml

```yaml
# foreman-proxy-content-answers-default.yaml
# scenario: foreman-proxy-content
# This is what foreman-installer ships out of the box, stripped of unrelated entries
---
foreman_proxy: {}
foreman_proxy_content: true
puppet: false
```

I would expect the following arguments to enable it:
```
--foreman-proxy-puppet true
--foreman-proxy-puppetca true
--foreman-proxy-content-puppet true
--enable-puppet
--puppet-server true
--puppet-server-ca true
--puppet-server-foreman-ssl-ca /etc/pki/katello/puppet/puppet_client_ca.crt
--puppet-server-foreman-ssl-cert /etc/pki/katello/puppet/puppet_client.crt
--puppet-server-foreman-ssl-key /etc/pki/katello/puppet/puppet_client.key
```

## foreman-proxy-content-answers-with-puppet.yaml

```yaml
# foreman-proxy-content-answers-with-puppet.yaml
# scenario: foreman-proxy-content
# This is where a user enabled puppet
---
foreman_proxy:
  puppet: true
  puppetca: true
foreman_proxy_content:
  puppet: true
puppet:
  server: true
  puppet_server_foreman_ssl_ca: /etc/pki/katello/puppet/puppet_client_ca.crt
  puppet_server_foreman_ssl_cert: /etc/pki/katello/puppet/puppet_client.crt
  puppet_server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
```

I would expect the following arguments to disable it:
```
--foreman-proxy-puppet false
--foreman-proxy-puppetca false
--foreman-proxy-content-puppet false
--no-enable-puppet
```

definitions/procedures/puppet/disable_puppet.rb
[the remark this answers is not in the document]
ekohl: We 100% know for sure they need to be restarted. So then the question is, do we restart it at least once but possibly twice? Or do we accept a bit more downtime to only restart it once. I'd feel safer with stopping it prior to it. It should also be faster in most cases.

@upadhyeammit thoughts?

definitions/scenarios/puppet.rb
```diff
@@ -0,0 +1,33 @@
+module ForemanMaintain::Scenarios
+  module Puppet
+    class DisablePuppet < ForemanMaintain::Scenario
+      metadata do
+        description 'Disable Puppet feature'
+        tags :puppet_disable
+        label :puppet_disable
+        param :remove_data, 'Purge the Puppet data after disabling plugin'
+        manual_detection
```
upadhyeammit, line 9: This should be available on Satellite 7.0 only and not on capsules as well?

```suggestion
      metadata do
        description 'Disable Puppet feature'
        tags :puppet_disable
        label :puppet_disable
        param :remove_data, 'Purge the Puppet data after disabling plugin'
        confine do
          feature(:satellite) && feature(:satellite).current_minor_version == '7.0'
        end
        manual_detection
```

ezr-ondrej: Well actually no, we've ended up with a version, that works on capsules as well, so you can actually disable a puppet on capsules with this.

upadhyeammit (commented):
Adding few comments, will revisit this for tests

ezr-ondrej:
@upadhyeammit the whole enable/disable is designed to work on foreman/katello/proxy/proxy-content scenarios so I do not think we need the checks for sat. Tho foreman >= 3.0 would be appropriate, but not sure if that's necessary

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,49 @@
+module Procedures::Puppet
+  class EnablePuppet < ForemanMaintain::Procedure
+    metadata do
+      description 'Enable Puppet feature'
```
ehelms, line 4: Just saying outloud that this starts to feel like scope creep and almost more like something the installer should have a native option for through hooks. Disable given it touches many things on the system feels better in foreman-maintain, this feels like just wrapping installer options. I don't know where to draw the line and this is not a blocker but I think it's worth pausing a moment to ensure this is what we intend.

Also, I am assuming this is being added just as a convenience other documenting how to do it?

ekohl: I had the same thought. I can't find the comment now but in the past I have thought about an `on_module_enable` / `on_module_disable` hook. IMHO there's at least a few options we have. But I agree it's more of a workaround in the wrong tool due to deadlines.

definitions/scenarios/puppet.rb
upadhyeammit: > @upadhyeammit the whole enable/disable is designed to work on foreman/katello/proxy/proxy-content scenarios so I do not think we need the checks for sat. Tho foreman >= 3.0 would be appropriate, but not sure if that's necessary

The foreman >= 3.0 should be required as these commands will be available on 6.10 as well.

definitions/checks/foreman/check_puppet_capsules.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    def find_puppet_proxies
+      feature(:foreman_database).query(puppet_proxies_query)
+    end
+
+    private
+
+    def local_proxy?(url)
+      URI.parse(url).hostname.casecmp(hostname) == 0
```
ekohl, line 28: Isn't this the same?
```suggestion
      URI.parse(url).hostname.casecmp?(hostname)
```

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,46 @@
+module Procedures::Puppet
+  class DisablePuppet < ForemanMaintain::Procedure
+    metadata do
+      description 'Disable Puppet feature'
+      confine do
+        feature(:puppet_server)
+      end
```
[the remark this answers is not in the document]
ekohl: That means it can no longer be used on setups without Foreman, such as (content) proxies.

definitions/procedures/puppet/disable_puppet.rb
ekohl: I mean that if you return a package here that isn't installed, will that be properly handled?

definitions/procedures/puppet/enable_puppet.rb
```diff
@@ -0,0 +1,49 @@
...
+      param :without_ca, 'Do not enable Puppet CA'
+      advanced_run false
+    end
+
+    def run
+      services = feature(:foreman_server).services + feature(:dynflow_sidekiq).services
+      Procedures::Service::Stop.new(:only => services)
```
ekohl, line 11: Stopping services isn't needed on installation, I'd feel safe to relying on the installer to do the right thing.

definitions/checks/foreman/check_puppet_capsules.rb
ezr-ondrej: It is apart of the version support... The `casecmp?` is not part of 2.0 apparently

definitions/procedures/puppet/enable_puppet.rb
ezr-ondrej: Enable is no more :)

ezr-ondrej:
Enable is out of scope now and I've added condition for foreman `>= 3.0`

test/lib/support/definitions/scenarios/puppet.rb
```diff
@@ -0,0 +1,29 @@
...
+      param :remove_data, 'Purge the Puppet data after disabling plugin'
+      manual_detection
+    end
+
+    def compose
+      add_steps([Checks::Dummy::Success])
+    end
+  end
+
+  class EnablePuppet < ForemanMaintain::Scenario
```
upadhyeammit, line 16: This should be removed as EnablePuppet is no more?

definitions/procedures/puppet/disable_puppet.rb
```diff
@@ -0,0 +1,50 @@
+module Procedures::Puppet
+  class DisablePuppet < ForemanMaintain::Procedure
+    metadata do
+      description 'Disable Puppet feature'
+      confine do
+        feature(:puppet_server) &&
+          check_min_version('foreman', '3.0') || check_min_version('foreman-proxy', '3.0')
```
ekohl, line 7:
```suggestion
          (check_min_version('foreman', '3.0') || check_min_version('foreman-proxy', '3.0'))
```

upadhyeammit (approved):

definitions/procedures/puppet/disable_puppet.rb
upadhyeammit: Yes, it does not fail if package is not installed.

definitions/procedures/puppet/enable_puppet.rb
upadhyeammit: Marking this resolved as enable is out of scope.

definitions/procedures/puppet/enable_puppet.rb
upadhyeammit: Marking this resolved as enable is out of scope.

upadhyeammit:
There are two unresolved comments,
- Removal of puppet-agent stuff[1]: as per comment its okay for now, we will revisit this if required.
- The refactoring of system_helpers[2]: will revisit this in future changes.

[1] https://github.com/theforeman/foreman_maintain/pull/555#discussion_r742048008
[2] https://github.com/theforeman/foreman_maintain/pull/555#discussion_r742049922

Accordingly I am merging this pull request. Thank you @ezr-ondrej @ekohl and @ehelms!

PR #558, opened 2021-11-11
Fixes #33788 - introduce repos_manager and refactor code

upadhyeammit:


upadhyeammit:
@evgeni request to review

lib/foreman_maintain/repos_manager/el.rb
```diff
@@ -0,0 +1,35 @@
...
+  class El
+    include ForemanMaintain::Concerns::OsFacts
+    include ForemanMaintain::Concerns::SystemHelpers
+
+    def disable_repos(repo_ids)
+      if el8?
+        execute!("dnf config-manager --set-disabled #{repo_ids.join(',')}")
+      elsif el7?
+        execute!("yum-config-manager --disable #{repo_ids.join(',')}")
+      end
```
evgeni, line 11: given EL9 will also have dnf, should this read more like

```suggestion
      if el7?
        execute!("yum-config-manager --disable #{repo_ids.join(',')}")
      else
        execute!("dnf config-manager --set-disabled #{repo_ids.join(',')}")
      end
```

lib/foreman_maintain/repos_manager/el.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def enable_repos(repo_ids)
+      if el8?
+        execute!("dnf config-manager --set-enabled #{repo_ids.join(',')}")
+      elsif el7?
+        execute!("yum-config-manager --enable #{repo_ids.join(',')}")
+      end
```
evgeni, line 19: same as above ("if el7, else")

lib/foreman_maintain/repos_manager/el.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def enabled_repos_hash
+      yum_cmd = 'yum repolist enabled -d 6 -e 0 2> /dev/null'\
```
evgeni, line 23: should this use `dnf` on el8+? right now there is an alias, so calling yum instead of dnf works, but at some point it might be gone, and it seems nicer to log that we called "dnf" on modern OSes :)

lib/foreman_maintain/repos_manager/el.rb
```diff
@@ -0,0 +1,35 @@
+module ForemanMaintain::ReposManager
```
evgeni, line 1: I kinda dislike the name, why not "RepoManager"? or "RepositoryManager"? The plural "repos" sounds weird to me.

definitions/features/upstream_repos.rb
```diff
@@ -1,40 +1,25 @@
...
 
   private
 
-  def trim_repoids(repos)
-    repos.map { |r| r.gsub(%r{Repo-id:|\/+\w*}, '') }
-  end
-
   def upstream_repo_urls
     repo_urls = { :Foreman => %r{yum.theforeman.org\/},
                   :Katello => %r{fedorapeople.org\/groups\/katello\/releases\/yum\/[\/|\w|.]*} }
```
evgeni, line 25: this needs updating btw, we are on yum.theforeman.org now too :)

definitions/features/upstream_repos.rb
```diff
@@ -1,40 +1,25 @@
-class Features::SystemRepos < ForemanMaintain::Feature
+class Features::UpstreamRepos < ForemanMaintain::Feature
   metadata do
-    label :system_repos
-    description 'Feature for operations on yum repositories of system'
+    label :upstream_repos
+    description 'Feature for upstream repositories'
```
evgeni, line 4: Feature for what? This somehow is missing a verb or something. Maybe "for managing upstream repositories" or the old "operations on"?

lib/foreman_maintain/repos_manager/el.rb
upadhyeammit: Yes, miss from my side, shall update!

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,24 @@
...
+        repoids_and_urls[repo] = url if url =~ regex
+      end
+    end
+    repoids_and_urls
+  end
+
+  private
+
+  def repo_urls
+    [%r{yum.theforeman.org\/*},
```
upadhyeammit, line 20: I think `yum.theforeman.org` is enough to know if repo is upstream? Instead of trying to check separate paths of Foreman, Katello, plugins, pulpcore etc?

lib/foreman_maintain/repository_manager/el.rb
```diff
@@ -0,0 +1,43 @@
...
+    def enable_repos(repo_ids)
+      if el7?
+        execute!("yum-config-manager --enable #{repo_ids.join(',')}")
+      else
+        execute!("dnf config-manager --set-enabled #{repo_ids.join(',')}")
+      end
+    end
+
+    def enabled_repos_hash
+      yum_cmd = 'yum repolist enabled -d 6 -e 0 2> /dev/null'
```
upadhyeammit, line 23: Removed the grep and used ruby's grep instead.
Hmm, I am somewhat confused on if I should put `repolist` in package manager or in repository manager?

lib/foreman_maintain/repository_manager/el.rb
```diff
@@ -0,0 +1,43 @@
...
+      repos = execute(yum_cmd)
+      return {} if repos.empty?
+
+      repo_hash = Hash[*repos.delete!(' ').split("\n").grep(/Repo-id|Repo-baseurl/)]
```
upadhyeammit, line 27: This use to include `Repo-id` and `Repo-baseurl` and then later on when required we called `trim_repo_ids`. As these labels are not required I thought to remove those.

upadhyeammit (commented):
@evgeni added few comments and pushed new changes. Request to check thank you!

definitions/features/upstream_repositories.rb
evgeni: yeah, for *current* upstream releases, this is always on `yum.theforeman.org` , different subfolders, but that shouldn't matter for your matcher

definitions/features/upstream_repos.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  private
+
+  def upstream_repo_urls
+    repo_urls = { :Foreman => %r{yum.theforeman.org\/},
+                  :Katello => %r{fedorapeople.org\/groups\/katello\/releases\/yum\/[\/|\w|.]*} }
+    [/#{repo_urls[:Foreman]}+releases/,
+     /#{repo_urls[:Foreman]}+plugins/,
+     /#{repo_urls[:Katello]}+katello/,
+     /#{repo_urls[:Katello]}+client/,
```
ehelms, line 25: These exist at `repo_urls[:Foreman]`

definitions/features/upstream_repos.rb
```diff
@@ -0,0 +1,31 @@
...
+     /#{repo_urls[:Katello]}+candlepin/,
+     /#{repo_urls[:Katello]}+pulp/,
+     %r{yum.puppetlabs.com\/el\/[\w|\/|\.]*\/x86_64},
+     %r{repos.fedorapeople.org\/repos\/pulp\/[\/|\w|.]*\/x86_64}]
```
ehelms, line 29: Pulp 3 repositories are not hosted here and we can likely build this withouth having to worry about Pulp 2.

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,24 @@
+class Features::UpstreamRepositories < ForemanMaintain::Feature
```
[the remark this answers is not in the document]
upadhyeammit: I simply missed to delete upstream_repos.rb file.

lib/foreman_maintain/repository_manager/el.rb
upadhyeammit: To solve this confusion, I used package_manager to get the yum/dnf and then use that in cmd. Marking solved accordingly.

upadhyeammit:
@evgeni request to review

lib/foreman_maintain/repository_manager/el.rb
```diff
@@ -0,0 +1,47 @@
...
+    def pkg_manager
+      package_manager.class.name.split('::').last.downcase
+    end
+
+    def enabled_repos_hash
+      cmd = "#{pkg_manager} repolist enabled -d 6 -e 0 2> /dev/null"
+      repos = execute(cmd)
+      return {} if repos.empty?
+
+      repo_hash = Hash[*repos.delete!(' ').split("\n").grep(/Repo-id|Repo-baseurl/)]
```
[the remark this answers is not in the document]
evgeni: Actually, you can replace this (and the whole `trim_repos`) with:

```ruby
Hash[*repos.split("\n").grep(/Repo-id|Repo-baseurl/).map{|entry| entry.split(':', 2).last.strip}]
```

To explain (as IMHO both versions are not too readable):
* `repos` is the output of `dnf repolist enabled` as a string, especially, thanks to `-d6`, it contains the repository information in a mostly machine readable way ("Repo-id      : sat6tools7", "Repo-baseurl : http://dogfood.sat.engineering.redhat.com/", "Repo-revision: 1636129236" etc).
* We can split by `\n` to get it into an Array
* Then we filter out all the lines that we don't care about and only keep "id" and "baseurl"

Now the old code remained at that, so we got (after `Hash` conversion) something like:
```ruby
{"Repo-id            : copr:copr.fedorainfracloud.org:evgeni:bugwarrior"=>"Repo-baseurl       : https://download.copr.fedorainfracloud.org/results/evgeni/bugwarrior/fedora-35-x86_64/",
 "Repo-id            : copr:copr.fedorainfracloud.org:flatcap:neomutt"=>"Repo-baseurl       : https://download.copr.fedorainfracloud.org/results/flatcap/neomutt/fedora-35-x86_64/"}
```

The new code now splits each entry (of the Array) at the first `:` and takes the `last` entry of the resulting array, `strip`ing whitespace as we go. If we now convert THAT to a Hash, we get

```ruby
{"copr:copr.fedorainfracloud.org:evgeni:bugwarrior"=>"https://download.copr.fedorainfracloud.org/results/evgeni/bugwarrior/fedora-35-x86_64/",
 "copr:copr.fedorainfracloud.org:flatcap:neomutt"=>"https://download.copr.fedorainfracloud.org/results/flatcap/neomutt/fedora-35-x86_64/"}
```

Which is exactly what we wanted.

lib/foreman_maintain/repository_manager.rb
```diff
@@ -0,0 +1,13 @@
+require 'foreman_maintain/repository_manager/el'
+include ForemanMaintain::Concerns::OsFacts
+module ForemanMaintain
```
evgeni, line 3: the `include` looks weird outside the `module`, is that intentional?

definitions/features/upstream_repositories.rb
```diff
@@ -0,0 +1,23 @@
...
+  metadata do
+    label :upstream_repositories
+    description 'Feature for operations on upstream repositories'
+  end
+
+  def repoids_and_urls
```
evgeni, line 7: I would have called this `enabled_upstream_repos` to match the `enabled_repos` of the `repository_manager`, but this might be just me ;)

evgeni (commented):
I like this a lot. Left a few comments how *I* would have approached this, but wouldn't block on them :)

lib/foreman_maintain/repository_manager.rb
upadhyeammit: I could be wrong, but the concern OsFacts is module and we are calling that in another module. If I include that in ForemanMaintain module then methods are not available.

upadhyeammit:
Request to review @evgeni

lib/foreman_maintain/repository_manager.rb
evgeni: Not saying this is wrong (I don't actually know), just that it looked weird to me.

lib/foreman_maintain/repository_manager/el.rb
```diff
@@ -0,0 +1,88 @@
...
+        "--#{option} #{repo_ids.join(',')}"
+      end
+    end
+
+    def rhsm_available?
+      @rhsm_available ||= find_package('subscription-manager')
+    end
+
+    def rhsm_list_all_repos
+      repos = execute(%(LANG=en_US.utf-8 subscription-manager repos --list 2>&1))
```
evgeni, line 63: I was about to suggest using `--list-enabled` here, and then say that `sub-man repos` is wrong to detect *all* repos, but this is actually not used for "which repos do you have enabled" but for "is repo X available for enablement", which pretty much confused me.

Wonder if mixing those two is a really good idea.

lib/foreman_maintain/repository_manager/el.rb
```diff
@@ -0,0 +1,88 @@
...
+    def config_manager_options(repo_ids, option)
+      if repo_ids.is_a?(String)
+        if repo_ids == '*'
+          return format_shell_args("--#{option}" => repo_ids.to_s)
+        end
+
+        "--#{option}=#{repo_ids}"
+      elsif repo_ids.is_a?(Array)
+        "--#{option} #{repo_ids.join(',')}"
+      end
```
evgeni, line 55: why not

```ruby
repo_ids_string = if repo_ids.is_a?(Array)
                    repo_ids.join(',')
                  else
                    repo_ids
                  end
format_shell_args("--#{option}" => repo_ids_string)
```

While other repo ids shouldn't contain special shell chars, it doesn't hurt to call `foreman_shell_args` on these too, and makes the code more compact.

lib/foreman_maintain/repository_manager/el.rb
upadhyeammit: I added a default option which can be used to get enabled/disabled repositories.

ehelms (approved):
Looks good to me, I'll rely on @evgeni to give final approval

evgeni (approved):
small nit, but it's also present in the old version of the code, so it's not a regression or anything

upadhyeammit:
@evgeni @ehelms thanks a lot for reviews!

PR #559, opened 2021-11-17
Fixes #33272 - implement dry-run for restore

evgeni:


definitions/scenarios/restore.rb
```diff
@@ -6,19 +6,27 @@ class Restore < ForemanMaintain::Scenario
...
 
     # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
+    # rubocop:disable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
     def compose
       backup = ForemanMaintain::Utils::Backup.new(context.get(:backup_dir))
 
       add_steps(find_checks(:root_user))
       supported_version_check
       add_steps_with_context(Checks::Restore::ValidateBackup,
-                             Procedures::Restore::Confirmation,
```
evgeni, line 19: The confirmation should really happen after *all* checks, this is unrelated to the `dry-run` case I think (can split out in a separate PR, if you want to?)

definitions/scenarios/restore.rb
```diff
@@ -6,19 +6,27 @@ class Restore < ForemanMaintain::Scenario
...
                              Checks::Restore::ValidateHostname,
-                             Checks::Restore::ValidateInterfaces,
+                             Checks::Restore::ValidateInterfaces)
+
+      if context.get(:dry_run)
+        self.class.metadata[:run_strategy] = :fail_slow
+        return
+      end
```
evgeni, line 27: Happy to see a "nicer" implementation of "if dry run, just add checks and fail_slow", but this one works, so :woman_shrugging:

evgeni:
Example output with dry run failing:

```console
# foreman-maintain restore --dry-run /backup
Running Restore backup
Validate backup has appropriate files:                                [OK]
Validate hostname is the same as backup:                              [FAIL]
The hostname in the backup does not match the hostname of the system
Validate network interfaces match the backup:                         [OK]
Scenario [Restore backup] failed.

  [restore-validate-hostname]

use --whitelist="restore-validate-hostname"
```

And non-dry-run failing:

```console
# foreman-maintain restore /backup
Running Restore backup
Validate backup has appropriate files:                                [OK]
Validate hostname is the same as backup:                              [FAIL]
The hostname in the backup does not match the hostname of the system
Scenario [Restore backup] failed.

  [restore-validate-hostname]

use --whitelist="restore-validate-hostname"



Running Rescue Restore backup

```

ehelms (approved):

evgeni:
added some very basic tests :)

definitions/scenarios/restore.rb
upadhyeammit: Yep, confirmation after _all checks_ would be nice, +1 for separate pr.

definitions/scenarios/restore.rb
upadhyeammit: It works as expected so good to go from my side.

upadhyeammit (approved):
Works as expected, maybe we can have separate pr for the confirmation procedure but thats not the blocker.

evgeni:
rebased!

PR #561, opened 2021-11-18
Fixes #33245 - use correct repo names on EL8

evgeni:


evgeni:
this freaks out the :police_officer:, but we can discuss the layout before I try to defund them?

lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -101,22 +122,42 @@ def puppet4_repo(full_version)
...
+          else
+            repos = ["satellite-maintenance-#{sat_maint_version}-for-rhel-#{el_major_version}-x86_64-rpms",
+                     "satellite-client-#{sat_client_version}-for-rhel-#{el_major_version}-x86_64-rpms"]
+          end
+        end
 
-        return repos_arrary.first(1) if feature(:satellite)
 
-        repos_arrary
+        return repos.first(1) if feature(:satellite)
```
evgeni, line 148: why do we even have the client repos here? are they used on capsules (I thought not)

ehelms:
@upadhyeammit per our conversations, here is another good example of being able to define and store the repositories in a single place to reduce chance to typo them or get it wrong, both within here and https://github.com/theforeman/foreman_maintain/pull/546 we are defining repo names

upadhyeammit:
> @upadhyeammit per our conversations, here is another good example of being able to define and store the repositories in a single place to reduce chance to typo them or get it wrong, both within here and #546 we are defining repo names

Ack, I will have a separate pr for that: https://projects.theforeman.org/issues/33995

lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -74,21 +74,42 @@ def rh_repos(server_version)
...
+            repos << "rhel-server-#{el_major_version}-#{package_name}-6-beta-rpms"
+          else
+            repos << "rhel-#{el_major_version}-server-#{package_name}-#{full_version}-rpms"
+          end
+        else
+          if use_beta_repos?
+            # FIXME: this is for sure wrong
+            repos << "satellite-#{package_name}-for-rhel-#{el_major_version}-x86_64-rpms"
+          else
+            repos << "satellite-#{package_name}-for-rhel-#{el_major_version}-x86_64-rpms"
```
ehelms, line 109: What would package_name be here? I think this would match capsule/maintenance/client pattern but not the primary repository.

upadhyeammit: Yeah, its as per the feature, Satellite/Capsule. So will be resolved to `satellite`/`capsule`.

lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -101,22 +122,42 @@ def puppet4_repo(full_version)
...
 
       def common_repos(full_version)
-        repos_arrary = if ENV['FOREMAN_MAINTAIN_USE_BETA'] == '1'
-                         ["rhel-#{el_major_version}-server-satellite-maintenance-6-beta-rpms",
-                          "rhel-#{el_major_version}-server-satellite-tools-6-beta-rpms"]
-                       else
-                         ["rhel-#{el_major_version}-server-satellite-maintenance-6-rpms",
-                          "rhel-#{el_major_version}-server-satellite-tools-#{full_version}-rpms"]
-                       end
+        sat_maint_version = '6'
```
upadhyeammit, line 125: The sat_maint_version should be 6 upto Satellite or Capsule version 6.10. Since 7.0 it should be `full_version`.

lib/foreman_maintain/concerns/downstream.rb
upadhyeammit: Correct, I checked the old changes but cant see a reason or discussion on why it was added. Safe to remove!

upadhyeammit (changes requested):

evgeni:
Updated, now with passing :police_officer: and simple tests.

Beta names still borked.

lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -74,21 +74,36 @@ def rh_repos(server_version)
...
-                 else
-                   "rhel-#{el_major_version}-server-#{package_name}-#{full_version}-rpms"
-                 end
+        maj_version = full_version[0]
+        repos = if el7? && use_beta_repos?
+                  ["rhel-server-#{el_major_version}-#{package_name}-#{maj_version}-beta-rpms"]
+                elsif el7?
+                  ["rhel-#{el_major_version}-server-#{package_name}-#{full_version}-rpms"]
+                elsif use_beta_repos?
+                  ["#{package_name}-#{maj_version}-beta-for-rhel-#{el_major_version}-x86_64-rpms"]
```
evgeni, line 102: this follows the style of `satellite-tools-6-beta-for-rhel-8-x86_64-rpms` which we already have today.

lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -101,22 +116,35 @@ def puppet4_repo(full_version)
...
+                              '6'
+                            end
+
+        # rubocop:disable Metrics/LineLength
+        repos = if el7? && use_beta_repos?
+                  ["rhel-#{el_major_version}-server-satellite-maintenance-#{sat_maint_version}-beta-rpms"]
+                elsif el7?
+                  ["rhel-#{el_major_version}-server-satellite-maintenance-#{sat_maint_version}-rpms"]
+                elsif use_beta_repos?
+                  ["satellite-maintenance-#{sat_maint_version}-beta-for-rhel-#{el_major_version}-x86_64-rpms"]
```
evgeni, line 131: will it be `satellite-maintenance-7-beta-for-rhel-8-x86_64-rpms` or `satellite-maintenance-7.0-beta-for-rhel-8-x86_64-rpms`? I would guess the former?

ehelms: I think the latter since the repository for maintenance is now versioned per Satellite version

evgeni: sure, but also for *satellite* itself, the beta repo was always "6-beta", whichever beta we shipped in there (I don't mind either way, just pointing out history)

ehelms: Ohh you are saying it was a one size fits all beta repository. That is indeed a question I do not know the answer to.

test/lib/concerns/downstream_test.rb
```diff
@@ -0,0 +1,120 @@
...
+        system.stubs(:el_major_version).returns(7)
+        expected_repos = ['rhel-7-server-satellite-6.10-rpms',
+                          'rhel-7-server-satellite-maintenance-6-rpms']
+        assert_equal expected_repos,
+                     system.send(:product_specific_repos, '6.10')
+      end
+
+      it 'returns correct repos for 7.0 on el8' do
+        system.stubs(:el_major_version).returns(8)
+        expected_repos = ['satellite-7.0-for-rhel-8-x86_64-rpms',
```
ehelms, line 51: Should we also add a test for the `satellite-capsule` repository to ensure `package_name` resolves correctly in both situations?

evgeni: oh yeah, I can add this

lib/foreman_maintain/concerns/downstream.rb
evgeni: I've updated it now to use `-7-beta-` to match the other names, but can adjust later, of course.

test/lib/concerns/downstream_test.rb
evgeni: done

upadhyeammit (approved):

PR #562, opened 2021-11-18
Refs #33829 - confine does not work to limit the execution

upadhyeammit:


upadhyeammit:
The confine does not work and I can see it works for upgrade_scenarios which are not manually detected. Accordingly I am using if condition to execute the steps.

ehelms (approved):

PR #565, opened 2021-11-26
Fixes #34017 - add Satellite and Capsule 7.0, 7.0.z upgrade scenarios

upadhyeammit:


upadhyeammit:
Most of this is copy paste, however would like few eyes to go through it once @evgeni @ehelms request to review.

evgeni:
While I still think we should have those in a more declarative way, this would require significant refactoring and nothing we have the bandwidth for today.

Looking at the existing code this makes sense and I am not aware of any special cases we need to take care during this update.

upadhyeammit:
> While I still think we should have those in a more declarative way, this would require significant refactoring and nothing we have the bandwidth for today.
> 
> Looking at the existing code this makes sense and I am not aware of any special cases we need to take care during this update.

Agree, forked redmine for refactoring https://projects.theforeman.org/issues/34050 will try to visit it in near future.

PR #566, opened 2021-11-26
Fixes #34018 - Remove ancient stuff from foreman-maintain

upadhyeammit:


ehelms:
I've taken some of these items and broken them into dedicated PRs and also caused this to need a massive rebase, so I'll opt to close it and handle it through those other PRs.

PR #567, opened 2021-11-30
Fixes #34039 - create check for pulp content's group ownership

upadhyeammit:


upadhyeammit:
Looks like when we init the scenarios we call compose method and then ultimately it actually calls the method which can raise the error. So learning from this is scenarios are placeholders and not to actually raise errors or something that sort.

ehelms (approved):

PR #569, opened 2021-12-07
Fixes #34101 - Drop locks and links when deleting tasks

adamruzicka:


adamruzicka:
Please note this is a quick and dirty fix.

Does anyone recall why we do it like this instead of running `foreman-rake foreman_tasks:cleanup` which should be the go-to tool for removing tasks?

upadhyeammit:
> Please note this is a quick and dirty fix.
> 
> Does anyone recall why we do it like this instead of running `foreman-rake foreman_tasks:cleanup` which should be the go-to tool for removing tasks?

If the question is about why we run delete queries from foreman-maintain instead of calling foreman_tasks:cleanup then I think foreman-maintain didn't change, the delete method and queries were introduced 5 years ago. So I think its time to revisit the foreman_tasks feature and surrounding operations to align it to current implementations.

adamruzicka:
Makes sense. So how about we get merge it as-is now, and then I'll open a followup?

definitions/features/foreman_tasks.rb
```diff
@@ -88,6 +88,9 @@ def delete(state)
        DELETE FROM dynflow_actions USING foreman_tasks_tasks WHERE (foreman_tasks_tasks.external_id = dynflow_actions.execution_plan_uuid::varchar) AND #{tasks_condition};
        DELETE FROM dynflow_execution_plans USING foreman_tasks_tasks WHERE (foreman_tasks_tasks.external_id = dynflow_execution_plans.uuid::varchar) AND #{tasks_condition};
        DELETE FROM foreman_tasks_tasks WHERE #{tasks_condition};
+       -- Delete locks and links which may now be orphaned
+       DELETE FROM foreman_tasks_locks as ftl LEFT JOIN foreman_tasks_tasks as ftt ON ftl.task_id = ftt.id WHERE ftt.id IS NULL;
+       DELETE FROM foreman_tasks_links as ftl LEFT JOIN foreman_tasks_tasks as ftt ON ftl.task_id = ftt.id WHERE ftt.id IS NULL;
```
upadhyeammit, line 93: The Postgresql is weird, it does not allow to use joins with delete. Request to change queries to use `using`,
```suggestion
       DELETE FROM foreman_tasks_locks ftl using foreman_tasks_tasks ftt where ftl.task_id = ftt.id and ftt.id IS NULL;
       DELETE FROM foreman_tasks_links ftl using foreman_tasks_tasks ftt where ftl.task_id = ftt.id and ftt.id IS NULL;
```

upadhyeammit (changes requested):

definitions/features/foreman_tasks.rb
adamruzicka:
```
foreman=# select * from foreman_tasks_links as ftl left join foreman_tasks_tasks as ftt on ftl.task_id = ftt.id where ftt.id IS NULL;
  id  |               task_id                | resource_type | resource_id | id | type | label | started_at | ended_at | state | result | external_id | parent_task_id | start_at | start_before | action | user_id | state_updated_at 
------+--------------------------------------+---------------+-------------+----+------+-------+------------+----------+-------+--------+-------------+----------------+----------+--------------+--------+---------+------------------
 5305 | 417541f1-cde4-46d9-8644-a5efe1815a44 | JobInvocation |         908 |    |      |       |            |          |       |        |             |                |          |              |        |         | 
(1 row)

foreman=# delete from foreman_tasks_links as ftl left join foreman_tasks_tasks as ftt on ftl.task_id = ftt.id where ftt.id IS NULL;
ERROR:  syntax error at or near "left"
LINE 1: delete from foreman_tasks_links as ftl left join foreman_tas...

foreman=# DELETE FROM foreman_tasks_links ftl using foreman_tasks_tasks ftt where ftl.task_id = ftt.id and ftt.id IS NULL;
DELETE 0
```

Good catch, although using `using` doesn't seem to help us here. Guess we'll have to go with a subquery

```
foreman=# select * from foreman_tasks_links as ftl where ftl.task_id not in (select id from foreman_tasks_tasks);
  id  |               task_id                | resource_type | resource_id 
------+--------------------------------------+---------------+-------------
 5305 | 417541f1-cde4-46d9-8644-a5efe1815a44 | JobInvocation |         908
(1 row)

foreman=# delete from foreman_tasks_links as ftl where ftl.task_id not in (select id from foreman_tasks_tasks);
DELETE 1

foreman=# select * from foreman_tasks_links as ftl where ftl.task_id not in (select id from foreman_tasks_tasks);
 id | task_id | resource_type | resource_id 
----+---------+---------------+-------------
(0 rows)
```

upadhyeammit (approved):

PR #570, opened 2021-12-07
Fixes #34045 - adds a restart step after removing pulp2

jjeffers:
After the removal of pulp2 items, restart pulpcore services. Without this step, users were often surprised by syncs that failed post -migration and pulp2 removal.


jlsherrill:
code looks good, i'll run a test

jlsherrill (approved):
thanks @jjeffers !

PR #572, opened 2021-12-14
Fixes #34136 - correct message for remote db if psql is unavailable

upadhyeammit:


upadhyeammit:
Unfortunately this change does not work, I doubt its because of using Threads in the service feature.

upadhyeammit:
Closing this pull request as its not required now. The foreman-maintain rpm instead depends on psql now.

PR #573, opened 2021-12-17
drop support for service-wait

evgeni:
this was a relict from EL6 which we don't support for a long time now

### DIFF
--- a/lib/foreman_maintain/utils/service/systemd.rb
+++ b/lib/foreman_maintain/utils/service/systemd.rb
@@ -8,21 +8,16 @@ module ForemanMaintain::Utils
         @instance_parent_unit = options.fetch(:instance_parent_unit, nil)
       end
 
-      def command(action, options = {})
-        do_wait = options.fetch(:wait, true) # wait for service to start
+      def command(action)
         all = @options.fetch(:all, false)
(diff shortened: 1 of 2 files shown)


lib/foreman_maintain/utils/service/systemd.rb
```diff
@@ -9,20 +9,15 @@ def initialize(name, priority, options = {})
       end
 
       def command(action, options = {})
-        do_wait = options.fetch(:wait, true) # wait for service to start
         all = @options.fetch(:all, false)
```
evgeni, line 12: eh, why is this using `@options` and not `options`?!

evgeni: answering myself, it wants the options of the object (as passed to `new`), not the ones to `command`…

ekohl (approved):
:+1: since we dropped it a long time ago.

Interestingly enough, it appears to have been added to foreman-maintain around the time we were actively removing it from the other tooling. Compare https://projects.theforeman.org/issues/23271 and https://projects.theforeman.org/issues/23407.

It was actually removed in https://github.com/theforeman/foreman-packaging/commit/46243c12bb57c7a1eb44a4ade07015914489a789 (so https://projects.theforeman.org/issues/23397), so Katello 3.7.0.

PR #575, opened 2021-12-23
Fixes #34197 - reuse find_installed_package instead of rpm_version

upadhyeammit:


upadhyeammit:
While including apt support I felt `rpm_version` method can be refactored to use `find_installed_package` . Once apt is available package_manager can selectively query respective `find_installed_package` method as per the OS.

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -38,8 +38,12 @@ def installed?(packages)
       sys.execute?(%(rpm -q #{packages_list}))
     end
 
-    def find_installed_package(name)
-      status, result = sys.execute_with_status(%(rpm -q '#{name}'))
+    def find_installed_package(name, queryfm = '')
```
evgeni, line 41: I'd call this param `queryfmt` or `queryformat`, `fmt` is a common abbreviation for "format" and is also used in `man rpm` to describe the *value* passed after `--queryformat`. `fm` on the other hand can mean `foreman-maintain` and that initially confused me.

evgeni (commented):
small naming nit, otherwise I like it!

lib/foreman_maintain/package_manager/yum.rb
upadhyeammit: Thanks for the review, as I am at changing param name, I changed it for list_installed_packages as well.

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -76,12 +80,12 @@ def files_not_owned_by_package(directory)
       sys.execute(find_cmd).split("\n")
     end
 
-    def list_installed_packages(queryfm = '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n')
-      # The queryfm should only include valid tag(s) as per `rpm --querytag` list.
+    def list_installed_packages(queryformat = '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n')
+      # The queryformat should only include valid tag(s) as per `rpm --querytag` list.
       # If any special formatting is required with querytag then it should be provided with tag i.e,
       # querytag = "--%{VENDOR}"
```
evgeni, line 86: what's `querytag` here?

evgeni (approved):

lib/foreman_maintain/package_manager/yum.rb
upadhyeammit: Thanks for the catch, corrected it

test/test_helper.rb
```diff
@@ -63,7 +63,7 @@ def installed?(packages)
     [packages].flatten(1).all? { |p| @packages.include?(p) }
   end
 
-  def find_installed_package(name)
+  def find_installed_package(name, _queryfm = '')
```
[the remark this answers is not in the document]
upadhyeammit: Agree, done!

evgeni (approved):

upadhyeammit:
@evgeni thanks for review! Merging now!

PR #577, opened 2022-01-06
Fixes #34231 - Fix to check pulp group ownership to "/var/lib/pulp/content"

shubhamsg199:
During upgrade check we are checking group ownership of "/var/lib/pulp" (pulp2 dir)
And, during "foreman-maintain prep-6.10-upgrade" we are checking group ownership "/var/lib/pulp/content" (pulp3) dir

Added a small fix pointing to correct "/var/lib/pulp/content" directory
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2037648


jlsherrill (approved):
Thanks @shubhamsg199

PR #580, opened 2022-01-31
Fixes #34282 - introduce nftables for maintenance mode

upadhyeammit:


upadhyeammit:
@ekohl and @evgeni request to review, thank you!

definitions/procedures/maintenance_mode/enable_maintenance_mode.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def run
+      if feature(:nftables)
+        nftables_enable_maintenance_mode
+      elsif feature(:iptables)
+        feature(:iptables).add_maintenance_mode_chain
+      else
+        warn! 'Unable to find iptables or nftables!'
+      end
```
evgeni, line 20: same comment as above, I think both features should implement the same interface, and we should call that interface from the procedure

definitions/checks/maintenance_mode/check_consistency.rb
```diff
@@ -25,8 +25,8 @@ def run
     def verify_with_features
       procedure_arr = []
       feature_status_msgs = []
-      is_mode_on = feature(:iptables).maintenance_mode_chain_exist?
-      [:iptables, :sync_plans, :cron].each do |feature_name|
+      is_mode_on = feature(:instance).firewall.maintenance_mode_status?
+      [feature(:instance).firewall.label, :sync_plans, :cron].each do |feature_name|
```
ekohl, line 29: How expensive is it to determine `feature(:instance).firewall`? Should this be stored in a separate variable and reused?

ekohl (commented):
I don't see where in the logic you would run it for IPv4 and IPv6. Is that a follow up feature since today it doesn't exist with IPv6 either?

upadhyeammit:
> I don't see where in the logic you would run it for IPv4 and IPv6. Is that a follow up feature since today it doesn't exist with IPv6 either?

for iptables its not there.. the nft rule should be applicable to both ipv4 and ipv6. I am checking regarding iptables ipv6, if its not much then I can add it in same pr

ekohl:
Oh right, my bad. I forgot that in nftables you only have just `inet` and it was in iptables that it was separate.

definitions/procedures/maintenance_mode/enable_maintenance_mode.rb
evgeni: much more readable,thanks!

upadhyeammit:
> Oh right, my bad. I forgot that in nftables you only have just `inet` and it was in iptables that it was separate.

added ip6tables as well :)

evgeni:
I have not much knowledge about nftables, but overall the code looks sound to me.

One thing that crept in my mind tho: the PR is called "nftables for maintenance mode", but it *also* adds ip6tables support which wasn't there before. Don't get me wrong, I am all for getting v6 Support also for legacy OSes, but I do see a risk here: we do have users who run with v6 completely disabled (https://access.redhat.com/solutions/8709) and depending on *how* they've chosen to do so, I'd expect `ip6tables` operations to just fail.

I think if we drop the add/drop/exists calls for `v6` (but leave the rest of the "generate the right command" code in place) we can get the nftables feature out and then add v6 for iptables easily afterwards.

upadhyeammit:
> I have not much knowledge about nftables, but overall the code looks sound to me.
> 
> One thing that crept in my mind tho: the PR is called "nftables for maintenance mode", but it _also_ adds ip6tables support which wasn't there before. Don't get me wrong, I am all for getting v6 Support also for legacy OSes, but I do see a risk here: we do have users who run with v6 completely disabled (https://access.redhat.com/solutions/8709) and depending on _how_ they've chosen to do so, I'd expect `ip6tables` operations to just fail.
> 
> I think if we drop the add/drop/exists calls for `v6` (but leave the rest of the "generate the right command" code in place) we can get the nftables feature out and then add v6 for iptables easily afterwards.

I have removed the ipv6 commit will create separate pr for it.

evgeni (approved):

ekohl (approved):
I don't have the setup to verify this, but from reading the code this looks OK to me.

PR #581, opened 2022-02-03
Fixes #34386 - don't validate the token in confine

upadhyeammit:


definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -29,19 +29,19 @@ def old_files_from_tftp_boot
       end.compact
     end
 
-    def self.non_zero_token_duration?
-      lookup_token_duration != 0
+    def non_zero_token_duration?
+      lookup_token_duration == 0
```
evgeni, line 33: this doesn't look right? because now you check if the token duration is zero?

evgeni (approved):

PR #582, opened 2022-02-04
Fixes #34393 - use repoquery format that works on EL7 and EL8

evgeni:
On EL8 repoquery doesn't understand the `nvra` format, leading to the
following error:

    Error: 'Package' object has no attribute 'nvra'

And thus no data reported from that query.
Let's use `%{name}-%{evr}.%{arch}` instead, which is understood by
repoquery on EL7 and EL8.

### DIFF
--- a/definitions/checks/check_hotfix_installed.rb
+++ b/definitions/checks/check_hotfix_installed.rb
@@ -46,9 +46,11 @@ class Checks::CheckHotfixInstalled < ForemanMaintain::Check
   def installed_packages
     packages = []
     repoquery_cmd = execute!('which repoquery')
-    IO.popen([repoquery_cmd, '-a', '--installed', '--qf', '%{ui_from_repo} %{nvra}']) do |io|
+    query_format = '%{ui_from_repo} %{name}-%{evr}.%{arch}'
+    IO.popen([repoquery_cmd, '-a', '--installed', '--qf', query_format]) do |io|
       io.each do |line|
         repo, pkg = line.chomp.split
+        next if repo.nil? || pkg.nil?
         packages << pkg if /satellite|rhscl/ =~ repo[1..-1].downcase
       end
     end


definitions/checks/check_hotfix_installed.rb
```diff
@@ -46,7 +46,7 @@ def modified_files(package)
   def installed_packages
     packages = []
     repoquery_cmd = execute!('which repoquery')
```
evgeni, line 48: Why are we using `which repoquery` here? I don't see any benefit over calling `repoquery` directly, without any path?

definitions/checks/check_hotfix_installed.rb
```diff
@@ -46,7 +46,7 @@ def modified_files(package)
...
-    IO.popen([repoquery_cmd, '-a', '--installed', '--qf', '%{ui_from_repo} %{nvra}']) do |io|
+    IO.popen([repoquery_cmd, '-a', '--installed', '--qf', '%{ui_from_repo} %{name}-%{evr}.%{arch}']) do |io|
       io.each do |line|
         repo, pkg = line.chomp.split
         packages << pkg if /satellite|rhscl/ =~ repo[1..-1].downcase
```
evgeni, line 52: why are we using `repo[1..-1]` here? that means that  if the repo *starts* with `satellite` (like the new EL8 repos do), this gets mangled to `atellite` and thus won't match?

definitions/checks/check_hotfix_installed.rb
upadhyeammit: Yep, does not make sense. Instead maybe we should check if repoquery cmd exist? Which I believe should be there in minimal installation as it comes from yum/dnf-utils

evgeni: the foreman-maintain RPM depends on yum-utils, so it's guaranteed to be there

definitions/checks/check_hotfix_installed.rb
upadhyeammit: Not able to understand logic behind [1..1], the original pr does not say anything about same as well. So we can consider this as extra fix.

upadhyeammit:
@evgeni while we are at it, do you think repoquery should be refactored into repository_manager?
And the same check fails when system is not registered to CDN however subscription-manager is enabled(bz 2048913) do you think we should handle the fix for this as well in same pr?

evgeni:
Added a safeguard for `repo` or `pkg` being `nil`.

Refactoring… Probably, but I'd prefer not to do it in the fix PR here.

upadhyeammit (approved):

PR #586, opened 2022-02-10
Fixes #34435 - update repoquery query format to support EL8

upadhyeammit:
The repoquery command on EL8 shows @System as the repo from which packages are installed. This does not help to filter out packages which are from Satellite repository.

### DIFF
--- a/definitions/checks/check_hotfix_installed.rb
+++ b/definitions/checks/check_hotfix_installed.rb
@@ -45,7 +45,6 @@ class Checks::CheckHotfixInstalled < ForemanMaintain::Check
 
   def installed_packages
     packages = []
-    query_format = '%{ui_from_repo} %{name}-%{evr}.%{arch}'
     IO.popen(['repoquery', '-a', '--installed', '--qf', query_format]) do |io|
       io.each do |line|
         repo, pkg = line.chomp.split
@@ -54,6 +53,14 @@ class Checks::CheckHotfixInstalled < ForemanMaintain::Check
       end
     end
     packages
+  end
+
+  def query_format
+    if el7?
+      return '%{ui_from_repo} %{name}-%{evr}.%{arch}'
+    end
+
+    '%{from_repo} %{name}-%{evr}.%{arch}'
   end
 
   def find_hotfix_packages


evgeni:
This still shows `@System` for me?

```console
[root@9a533c64e2f3 /]# LC_ALL=de_DE.utf8 yum info --installed vim-enhanced
Installed Packages
Name         : vim-enhanced
Epoch        : 2
Version      : 8.0.1763
Release      : 16.el8_5.7
Architecture : x86_64
Size         : 2.9 M
Source       : vim-8.0.1763-16.el8_5.7.src.rpm
Repository   : @System
From repo    : appstream
Summary      : A version of the VIM editor which includes recent enhancements
URL          : http://www.vim.org/
License      : Vim and MIT
Description  : VIM (VIsual editor iMproved) is an updated and improved version of the
             : vi editor.  Vi was the first real screen-based editor for UNIX, and is
             : still very popular.  VIM improves on vi by adding new features:
             : multiple windows, multi-level undo, block highlighting and more.  The
             : vim-enhanced package contains a version of VIM with extra, recently
             : introduced features like Python and Perl interpreters.
             : 
             : Install the vim-enhanced package if you'd like to use a version of the
             : VIM editor which includes recently added enhancements like
             : interpreters for the Python and Perl scripting languages.  You'll also
             : need to install the vim-common package.
```

(Yes, I installed `vim` from the repo, instead having it pre-installed in the container)

evgeni:
Ah, you're using the `from repo` field…
You can get that on `repoquery` too: `repoquery --installed -a --qf '%{from_repo} %{name}-%{evr}.%{arch}'`

evgeni:
And now I also know *why* there was this weird `repo[1..-1]` thing:

```console
[root@e94615e40845 /]# repoquery --installed -a --qf '%{ui_from_repo} %{name}-%{evr}.%{arch}'
@CentOS acl-0:2.2.51-15.el7.x86_64
@CentOS audit-libs-0:2.8.5-4.el7.x86_64
```

To strip away the `@`, but it doesn't hurt at all…

evgeni:
(`from_repo` only works on EL8, `ui_from_repo` only on EL7, but hey, at least it's a 2 line change)

evgeni (approved):

PR #587, opened 2022-02-11
Fixes #34438 - don't try to find services for `nil`

evgeni:
when there is no mapping from a component to a feature,
`component_features_map[component]` will return `nil` and `nil.to_sym`
ends badly.

ideally, each component would be present in the map, but this makes the
code more robust

### DIFF
--- a/definitions/features/instance.rb
+++ b/definitions/features/instance.rb
@@ -142,11 +142,15 @@ class Features::Instance < ForemanMaintain::Feature
   def component_features_map
     {
       'candlepin_auth' => %w[candlepin candlepin_database],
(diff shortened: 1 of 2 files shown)


evgeni:
the underlying issue was that the `katello_agent` feature was broken, but there is no `katello_agent` in the `component_features_map`

evgeni:
@upadhyeammit should we also add `'katello_agent' => %w[katello]` to the `c_f_m`? As the katello feature correctly describes the `qpidd` and `qdrouterd` services that are broken in the reproducer setup

upadhyeammit:
@evgeni thanks for the fix! Works perfectly!

PR #592, opened 2022-03-11
Refs #34101 - Properly drop locks and links when deleting tasks

adamruzicka:


definitions/features/foreman_tasks.rb
```diff
@@ -89,8 +89,8 @@ def delete(state)
        DELETE FROM dynflow_execution_plans USING foreman_tasks_tasks WHERE (foreman_tasks_tasks.external_id = dynflow_execution_plans.uuid::varchar) AND #{tasks_condition};
        DELETE FROM foreman_tasks_tasks WHERE #{tasks_condition};
        -- Delete locks and links which may now be orphaned
-       DELETE FROM foreman_tasks_locks as ftl where ftl.id NOT IN (SELECT id FROM foreman_tasks_tasks);
-       DELETE FROM foreman_tasks_links as ftl where ftl.id NOT IN (SELECT id FROM foreman_tasks_tasks);
+       DELETE FROM foreman_tasks_locks as ftl where ftl.task_id NOT IN (SELECT id FROM foreman_tasks_tasks);
+       DELETE FROM foreman_tasks_links as ftl where ftl.task_id NOT IN (SELECT id FROM foreman_tasks_tasks);
```
upadhyeammit, line 93: I am having Satellite 6.9.8 and on that I don't see table foreman_tasks_links , if its expected to not have it on older versions then the query should be run conditionally on or above Satellite 6.10?
Problem is any latest version of foreman-maintain can be used on 6.9 and older versions. Can you conditionally run delete on foreman_tasks_links maybe using `check_min_version` for foreman 2.5?

upadhyeammit (changes requested):

definitions/features/foreman_tasks.rb
adamruzicka: Based the condition on version of foreman-tasks in which the table first appear

PR #593, opened 2022-03-11
Fixes #34717 - The foreman-protector plugin for the dnf

upadhyeammit:


extras/foreman_protector/dnf-foreman-protector.py
```diff
@@ -0,0 +1,86 @@
...
+import configparser
+
+class ForemanProtector(dnf.Plugin):
+    name = 'dnf-foreman-protector'
+
+    def __init__(self,base,cli):
+        self.base = base
+        self.cli = cli
+
+    def _config(self):
```
m-blaha, line 20: The hook is called `config`, not `_config`. Therefore currently this code is not executed.

m-blaha: You can also utilize read_config() method of the plugin:
`parser = self.read_config(self.base.conf)`

upadhyeammit:
@m-blaha thanks a lot for all the help, I request to review this plugin if in case you find code can be improved anywhere.
Apart from that the output ends with Error which I don't see in yum, like to know if its possible to not show that Error?,
~~~
# dnf update yum-utils
Updating Subscription Management repositories.
Red Hat Ansible Engine 2 for RHEL 8 x86_64 (RPMs)                                                       1.0 kB/s | 2.3 kB     00:02    
Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)                                                   1.1 kB/s | 2.4 kB     00:02    
Satellite 7.0 RHEL8                                                                                     1.0 kB/s | 2.5 kB     00:02    
Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)                                                1.2 kB/s | 2.8 kB     00:02    
Satellite Maintenance 7.0 RHEL8                                                                         1.0 kB/s | 2.1 kB     00:02    
Satellite Utils 7.0 RHEL8                                                                               768  B/s | 2.1 kB     00:02    
Reading Foreman protector configuration
*** Excluded total: 9427

WARNING: Excluding 9427 packages due to foreman-protector. 

Error: 
 Problem: package yum-utils-4.0.21-4.el8_5.noarch requires dnf-plugins-core = 4.0.21-4.el8_5, but none of the providers can be installed
  - cannot install the best update candidate for package yum-utils-4.0.21-3.el8.noarch
  - package dnf-plugins-core-4.0.21-4.el8_5.noarch is filtered out by exclude filtering
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
~~~

m-blaha:
The thing is, that whitelisted packages (let's say yum-utils) have dependencies (dnf-plugins-core) and sometimes the updated version the whitelisted package requires newer version of dependency. So the question is what should happen in such case?

1. you can either error out (like in your implementation) because dnf-plugins-core is not whitelisted and so the update is not allowed
2. you can also recursively whitelist dependencies of whitelisted packages. But that would allow upgrading of much broader package set, it will basically go down to the most basic libraries
3. you can also tune the whitelist a bit. For example yum-utils package is now only a thin wrapper around dnf-plugins-core and those two packages are tightly tied together. It's basically not possible to update yum-utils without updating dnf-plugins-core. You will also need python3-dnf-plugins-core which is in turn dependency of dnf-plugins-core. To explore what dependencies might need to be added you can use `dnf repoquery --requires <package>` command and if you see in the output dependency on the concrete version, you probably need to also add this (this is usually the case for packages originating in the same source rpm).

For example (this comes from fedora so the version do not match to those on rhel):

❯ dnf repoquery --requires dnf-plugins-core-0:4.1.0-1.fc35.noarch
python3-dnf-plugins-core = 4.1.0-1.fc35

On the other hand python3-dnf-plugins-core is more benevolent in terms of dependencies versions:
❯ dnf repoquery --requires python3-dnf-plugins-core-0:4.1.0-1.fc35.noarch
python(abi) = 3.10
python3-dateutil
python3-dbus
python3-distro
python3-dnf >= 4.11.0
python3-hawkey >= 0.64.0

upadhyeammit:
Hello @evgeni the plugin is working as expected Request to review, any suggestions will be helpful, I understand this can be something not familiar for you, however more eyes on code always help :)

extras/foreman_protector/dnf/foreman-protector.py
```diff
@@ -0,0 +1,76 @@
...
+            raise dnf.exceptions('Unable to read Foreman protector"s configuration: %s' % e)
+
+    def _add_obsoletes(self):
+        global package_whitelist
+        final_query = self.base.sack.query()
+        if package_whitelist:
+        #  If anything obsoletes something that we have whitelisted ... then
+        # whitelist that too.
+            whitelist_query = self.base.sack.query().filter(name=package_whitelist)
+            obsoletes_query = self.base.sack.query().filterm(obsoletes=list(whitelist_query))
```
evgeni, line 51: why `filterM` here? According to docs, it is "Similar to [dnf.query.Query.filter()](https://dnf.readthedocs.io/en/latest/api_queries.html#dnf.query.Query.filter) but it modifies the query in place." but I don't see a reason to do this in place here?

extras/foreman_protector/dnf/foreman-protector.py
```diff
@@ -0,0 +1,76 @@
...
+            if total > 1:
+                suffix = 's'
+            else:
+                suffix = ''
+        logger.info(_('\n'
+                        'WARNING: Excluding %d package%s due to foreman-protector. \n'
+                        'Use foreman-maintain packages install/update <package> \n'
+                        'to safely install packages without restrictions.\n'
+                        'Use foreman-maintain upgrade run for full upgrade.\n'
+                        % (total, suffix)))
```
evgeni, line 76: When `total` is 0, it means nothing was excluded, right? But the two `logger.info` calls still happen, which seems odd to me? Especially the second one.

extras/foreman_protector/dnf/foreman-protector.py
```diff
@@ -0,0 +1,76 @@
...
+class ForemanProtector(dnf.Plugin):
+    name = 'foreman-protector'
+    config_name = 'foreman-protector'
+
+    def __init__(self,base,cli):
+        self.base = base
+        self.cli = cli
+
+    def config(self):
+        global fileurl
```
evgeni, line 17: I don't like globals, but I guess this is some kind of `dnf` convention when it comes to passing data between things?

evgeni:
Added a few questions, but nothing that jumps at me as really being a blocker.

extras/foreman_protector/dnf/foreman-protector.py
upadhyeammit: Programmatically yes. However we are excluding all available packages always and whitelist is tiny, so we wont ever have 0 in total. Yeah but it wont hurt to make it correct, hence doing it :)

upadhyeammit: Added extra condition and message to handle the situation if we ever get it.

extras/foreman_protector/dnf/foreman-protector.py
m-blaha: I would recommend using `filterm` on both places, it saves some copying. `.filter` returns a new copy of the query with filter applied, `filterm` avoids copying and applies the filter in place.

extras/foreman_protector/dnf/foreman-protector.py
m-blaha: There are some use cases for `global` in dnf plugins. There are two "flavours" of dnf plugins:

1. plugins that change behaviour of dnf globally in some sense (for example this foreman-protector)
2. plugins that provide new command

And sometimes (like in dnf-plugins-core/plugins/versionlock.py) you need to share some data between both types (they are represented by separate classes). Then the global may be useful.

But here I do not see the reason for global, fileurl could be member of ForemanProtector class.

extras/foreman_protector/dnf/foreman-protector.py
```diff
@@ -0,0 +1,77 @@
...
+    def _load_whitelist(self):
+        package_whitelist = set()
+        try:
+            if fileurl:
+                llfile = open(fileurl, 'r')
+                for line in llfile.readlines():
+                    if line.startswith('#') or line.strip() == '':
+                        continue
+
+                    package_whitelist.add(line.rstrip().lower())
```
m-blaha, line 37: Not sure about `.lower()`. Package names are case sensitive.

extras/foreman_protector/dnf/foreman-protector.py
```diff
@@ -0,0 +1,77 @@
...
+        # whitelist that too.
+            whitelist_query = self.base.sack.query().filterm(name=package_whitelist)
+            obsoletes_query = self.base.sack.query().filterm(obsoletes=list(whitelist_query))
+
+            final_query = whitelist_query.union(obsoletes_query)
+        return final_query
+
+    def sack(self):
+        whitelist_and_obsoletes = self._add_obsoletes()
+        all_available_updates = self.base.sack.query().available()
```
m-blaha, line 57: Minor - the variable name `all_available_updates` is misleading. It contains all packages available in repositories, not only updates.

m-blaha (changes requested):
I've left some comments. Only the `.lower()` think I consider important.

extras/foreman_protector/dnf/foreman-protector.py
upadhyeammit: corrected!

extras/foreman_protector/dnf/foreman-protector.py
upadhyeammit: removed!

m-blaha (approved):

evgeni (approved):

PR #596, opened 2022-03-16
Fixes #34631 - add command & procedure to update pulpcore artifact ownership

ianballou:
This PR adds a command to fix the pulpcore artifact ownership.  When it's apache:pulp, there are sync errors. It's likely that users have artifacts with these permissions due to old Pulp 2 to Pulp 3 migration bugs.

This new procedure also automatically runs when the user removes Pulp 2.

To test:

1) `chown -hR apache.pulp /var/lib/pulp/media/artifact` on any Katello machine running Pulp 3.
2) Run `foreman-maintain content fix-pulpcore-artifact-ownership`.
3) Check that the files under `/var/lib/pulp/media/artifact` are all owned by the pulp user.
4) Set up a Satellite 6.10 or Katello 4.x machine that still has Pulp 2.
5) Run the Pulp 2 removal procedure and see that the artifact ownership procedure is also run.

### DIFF
--- a/definitions/procedures/content/fix_pulpcore_artifact_permissions.rb
+++ b/definitions/procedures/content/fix_pulpcore_artifact_permissions.rb
@@ -1,0 +1,30 @@
+module Procedures::Content
+  class FixPulpcoreArtifactOwnership < ForemanMaintain::Procedure
+    metadata do
+      description 'Fix Pulpcore artifact ownership to be pulp:pulp'
+      param :assumeyes, 'Do not ask for confirmation', :default => false
+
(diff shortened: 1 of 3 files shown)


jlsherrill:
test failure here seems relevant for ruby 3.0.2

jlsherrill:
tested both scenarios and worked great!  Also tested and made sure that --assumeyes worked passing through remove-pulp2.  Just the one test failure and this is good to go

ianballou:
Just need some help merging... @upadhyeammit ?

PR #597, opened 2022-03-28
Refs #34685 - modify repository versions to 6.11

upadhyeammit:


evgeni (approved):
Assuming this is only addressing the needed changes to the feature from https://projects.theforeman.org/issues/33245, ACK

PR #600, opened 2022-03-30
Fixes #34699 - don't try to delete pulp mountpoints

upadhyeammit:


definitions/procedures/pulp/remove.rb
```diff
@@ -24,7 +24,18 @@ def pulp_data_dirs
...
-      ]
+      ].select { |dir| File.directory?(dir) }
+    end
+
+    def pulp_data_dirs_mountpoints
+      pulp_data_dirs.map do |pulp_dir|
+        dir_path = Pathname.new(pulp_dir)
+        pulp_dir if dir_path.mountpoint?
+      end.compact
+    end
```
evgeni, line 35:
```suggestion
    def pulp_data_dirs_mountpoints
      pulp_data_dirs.select{ |d| Pathname(d).mountpoint? }.compact
    end
```

evgeni: And I think you need to `require 'pathname'` for this to work?

definitions/procedures/pulp/remove.rb
```diff
@@ -70,7 +81,7 @@ def ask_to_proceed(rm_cmds)
     end
 
     def run
-      rm_cmds = data_dir_removal_cmds
+      rm_cmds = data_dir_removal_cmds(pulp_data_dirs)
```
evgeni, line 84: you now call `data_dir_removal_cmds` inside `delete_pulp_data`, so this is not really needed anymore, if you also move the `if rm_cmds.any` *inside* `delete_pulp_data` on line 100?

definitions/procedures/pulp/remove.rb
```diff
@@ -163,12 +174,33 @@ def drop_migrations
...
     # rubocop:enable Metrics/BlockLength
 
-    def delete_pulp_data(rm_cmds)
+    def delete_pulp_data
+      non_mountpoints = data_dir_removal_cmds(deletable_pulp_dirs)
+      mountpoints = pulp_data_dirs_mountpoints
       with_spinner('Deleting pulp2 data directories') do |spinner|
-        rm_cmds.each do |cmd|
-          execute!(cmd)
+        if non_mountpoints.any?
```
evgeni, line 181: I'd put the `any?` outside the `with_spinner` block, as otherwise you'd possibly end up with displaying "Deleting pulp2 data" without doing anything if *everything* is a mointpoint (I don't think those setups exist, but we still can be correct ;))

definitions/procedures/pulp/remove.rb
```diff
@@ -163,12 +174,33 @@ def drop_migrations
...
+          spinner.update('Done deleting pulp2 data directories')
+        end
+      end
+    end
+
+    def msg_for_del_mountpoints
+      mountpoints = pulp_data_dirs_mountpoints
+      if mountpoints.count > 1
+        puts "\nThe directories #{mountpoints.join(',')} are individual mountpoints."\
+              "\nThe satellite-maintain won't delete these directories.\n"\
```
evgeni, line 198: It's foreman-maintain, please ;)

definitions/procedures/pulp/remove.rb
upadhyeammit: Moved it to `ask_to_proceed` as its not being shared with `delete_pulp_data` anymore.

definitions/procedures/pulp/remove.rb
upadhyeammit: done!

definitions/procedures/pulp/remove.rb
upadhyeammit: Yeah, was about the require it, however I can see installer feature is using it already, I can't find explicit require anywhere. But yes it works without require.

definitions/procedures/pulp/remove.rb
upadhyeammit: Now calling `ForemanMaintain.pkg_and_cmd_name` which should render the cmd as per features :)

definitions/procedures/pulp/remove.rb
```diff
@@ -163,12 +174,33 @@ def drop_migrations
...
+        if mountpoints.any?
+          msg_for_del_mountpoints
+        else
+          spinner.update('Done deleting pulp2 data directories')
+        end
+      end
+    end
+
+    def msg_for_del_mountpoints
+      mountpoints = pulp_data_dirs_mountpoints
```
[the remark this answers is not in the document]
upadhyeammit: done!

definitions/procedures/pulp/remove.rb
```diff
@@ -24,7 +24,15 @@ def pulp_data_dirs
         '/var/lib/pulp/uploads',
         '/var/lib/mongodb/',
         '/var/cache/pulp'
-      ]
+      ].select { |dir| File.directory?(dir) }
```
evgeni, line 27: What is this `directory?` for? I know it was there before, just not understanding the point.

definitions/procedures/pulp/remove.rb
```diff
@@ -163,12 +170,34 @@ def drop_migrations
...
+          spinner.update('Deleting pulp2 data directories')
+          non_mountpoints.each do |cmd|
+            execute!(cmd)
+          end
         end
-        spinner.update('Done deleting pulp2 data directories')
+        if mountpoints.any?
+          msg_for_del_mountpoints(mountpoints)
+        else
+          spinner.update('Done deleting pulp2 data directories')
```
evgeni, line 186: So the spinner is never updated if there are mountpoints?

I must admit I am not too well versed with the spinner stuff, but as a user I would have thought I'd get something like:
* am am deleting all your old stuff
* ok, I am done deleting your stuff
* (optional) BTW, I am done, but couldn't get *everything* done, here is some homework: …

evgeni:
@ekohl @ehelms would one of you please double check us on the `Pathname(d).mountpoint?` idea (it was originally mine, and it kinda feels wrong to ACK the PR, even if Amit wrote the actual code)?

definitions/procedures/pulp/remove.rb
```diff
@@ -24,7 +24,15 @@ def pulp_data_dirs
         '/var/lib/pulp/uploads',
         '/var/lib/mongodb/',
         '/var/cache/pulp'
-      ]
+      ].select { |dir| File.directory?(dir) }
+    end
+
+    def pulp_data_dirs_mountpoints
+      pulp_data_dirs.select { |d| Pathname(d).mountpoint? }.compact
```
ekohl, line 31: Why is compact needed? I don't see how there would be nil values in there.

definitions/procedures/pulp/remove.rb
ekohl: > (I don't think those setups exist, but we still can be correct ;))

One possible scenario is that no directories exist and only a mountpoint remains. Still unlikely but a lot more likely than the scenario where there are only mountpoints :)

definitions/procedures/pulp/remove.rb
upadhyeammit: Below is what I felt should be the states, however if you think we should handle it in any better way then happy to do the changes,

* am am deleting all your old stuff -> we should show this only when we are actually deleting something?
* ok, I am done deleting your stuff -> we should show this only when we can delete all the stuff but not in partial state?
* (optional) BTW, I am done, but couldn't get everything done, here is some homework: -> for this there are explicit messages.

definitions/procedures/pulp/remove.rb
ekohl: Btw, I think I still agree with @evgeni that it should be (roughly)
```ruby
if non_mountpoints.any?
  with_spinner('') do |spinner|
    # ...
  end
end
```

definitions/procedures/pulp/remove.rb
ekohl: Shouldn't it be `elsif non_mountpoints.any?` instead of `else` here?

upadhyeammit: @ekohl and @evgeni improved the messaging in recent commit, let me know your thoughts.

definitions/procedures/pulp/remove.rb
```diff
@@ -163,12 +170,41 @@ def drop_migrations
...
         end
-        spinner.update('Done deleting pulp2 data directories')
+      end
+    end
+
+    def msg_for_del_non_moutpoints(mountpoints, spinner)
+      if mountpoints.empty?
+        spinner.update('Done deleting all pulp2 data directories.')
+      else
+        spinner.update("Deleted only: #{deletable_pulp_dirs.join("\n")}")
```
ekohl, line 194: I think `only` is a bit redundant and can be left out:
```suggestion
        spinner.update("Deleted: #{deletable_pulp_dirs.join("\n")}")
```

ekohl (approved):

PR #601, opened 2022-03-31
Fixes #34706 - introduce an option to exclude step from whitelist

upadhyeammit:


evgeni:
This doesn't prevent users from whitelisting those steps, right? It just doesn't suggest that it's possible/desireable to do so?

If the above is correct, then the patch should do what you say, but the commit message is misleading ;-)

jlsherrill:
tested this and this is what i got.  

```
Tasks: TOP => katello:pulp3_content_switchover
(See full trace by running task with --trace)
Scenario [Migration scripts to Satellite 6.10] failed.

  [content-switchover]

```

I kinda think we should make this step never skippable.  Its asking for too many problems, but i'm fine with whatever support says

upadhyeammit:
> This doesn't prevent users from whitelisting those steps, right? It just doesn't suggest that it's possible/desireable to do so?
> 
> If the above is correct, then the patch should do what you say, but the commit message is misleading ;-)

Yeah, I feel there could be still a situation where failures are false positives. In that case users should have option to skip it? Unless we are very sure whatever problems reported by the switchover are always legit.

evgeni:
I think that there are situations/steps that should not be disable-able, at all.

upadhyeammit:
> > This doesn't prevent users from whitelisting those steps, right? It just doesn't suggest that it's possible/desireable to do so?
> > If the above is correct, then the patch should do what you say, but the commit message is misleading ;-)
> 
> Yeah, I feel there could be still a situation where failures are false positives. In that case users should have option to skip it? Unless we are very sure whatever problems reported by the switchover are always legit.



> tested this and this is what i got.
> 
> ```
> Tasks: TOP => katello:pulp3_content_switchover
> (See full trace by running task with --trace)
> Checking for valid Katello configuraton.
> Scenario [Migration scripts to Satellite 6.10] failed.
> 
> The following steps ended up in failing state:
> 
>   [content-switchover]
> ```
> 
> I kinda think we should make this step never skippable. Its asking for too many problems, but i'm fine with whatever support says

I like the idea however we need to be 100% sure that whatever failures reported by the content-switchover are always legit. This means there are no situation where we will use whitelist, which I think difficult to agree on?

jlsherrill:
The switchover step is so vital to be run successfully that i don't think it should be skippable.  If a user skips it, they are likely completely destroying all the content views on their satellite and will have to start over from scratch.  I'd rather have 100 users debug this failing and work through it than 1 user skipping it and losing everything.

ekohl:
> I think that there are situations/steps that should not be disable-able, at all.

Should his be called "mandatory" instead?

upadhyeammit:
As per the discussion with Ashish I have modified the message for downstream situation. As example,
~~~

  [pulp-remove]

the command.

In case the failures are false positives,
use --whitelist="pulp-remove"

~~~

test/lib/reporter_test.rb
```diff
@@ -151,6 +211,44 @@ def decision_question(description)
       MESSAGE
     end
 
+    it 'informs the user about warnings and failures of the last scenario for upstream' do
+      assume_feature_present(:satellite)
```
evgeni, line 215: `for upstream` and `assume_feature_present(:satellite)` seems odd to me? :)

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -317,19 +317,38 @@ def scenario_failure_message(scenario)
...
-          recommend << format(<<-MESSAGE.strip_heredoc, whitelist_labels)
-          Resolve the failed steps and rerun
-          the command. In case the failures are false positives,
-          use --whitelist="%s"
-          MESSAGE
+          unless whitelist_labels.empty?
+            recommend << if scenario.detector.feature(:instance).downstream
+                           format(<<-MESSAGE.strip_heredoc, whitelist_labels)
+              Resolve the failed steps and rerun
+              the command.
```
evgeni, line 336:
```suggestion
              Resolve the failed steps and rerun the command.
```

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -317,19 +317,38 @@ def scenario_failure_message(scenario)
...
+
+              If the situation persists and, you are unclear what to do next,
+              contact Red Hat Technical Support.
+
+              In case the failures are false positives,
+              use --whitelist="%s"
```
evgeni, line 342:
```suggestion
              In case the failures are false positives, use
              --whitelist="%s"
```

lib/foreman_maintain/reporter/cli_reporter.rb
```diff
@@ -317,19 +317,38 @@ def scenario_failure_message(scenario)
...
+              MESSAGE
+                         else
+                           format(<<-MESSAGE.strip_heredoc, whitelist_labels)
+              Resolve the failed steps and rerun
+              the command. In case the failures are false positives,
+              use --whitelist="%s"
```
evgeni, line 348:
```suggestion
              Resolve the failed steps and rerun the command.
              In case the failures are false positives, use
              --whitelist="%s"
```

test/lib/reporter_test.rb
```diff
@@ -94,9 +118,41 @@ def decision_question(description)
...
+      Scenario [Scenarios::Dummy::FailMultiple] failed.
+
+      The following steps ended up in failing state:
+
+        [dummy-check-fail]
+        [dummy-check-fail2]
+
+      Resolve the failed steps and rerun the command.
+      In case the failures are false positives, use
+      --whitelist=\"dummy-check-fail,dummy-check-fail2\"
```
evgeni, line 155:
```suggestion
      --whitelist="dummy-check-fail,dummy-check-fail2"
```

no need to escape those quotes -- makes it easier to read ;)

test/lib/reporter_test.rb
```diff
@@ -142,9 +201,46 @@ def decision_question(description)
 
           [dummy-check-warn]
 
-        Resolve the failed steps and rerun
-        the command. In case the failures are false positives,
-        use --whitelist=\"dummy-check-fail\"
+        Resolve the failed steps and rerun the command.
+        In case the failures are false positives, use
+        --whitelist=\"dummy-check-fail\"
```
evgeni, line 206:
```suggestion
        --whitelist="dummy-check-fail"
```

test/lib/reporter_test.rb
```diff
@@ -142,9 +201,46 @@ def decision_question(description)
...
+
+          [dummy-check-warn]
+
+        Resolve the failed steps and rerun the command.
+
+        If the situation persists and, you are unclear what to do next,
+        contact Red Hat Technical Support.
+
+        In case the failures are false positives, use
+        --whitelist=\"dummy-check-fail\"
```
evgeni, line 243:
```suggestion
        --whitelist="dummy-check-fail"
```

evgeni:
Can we add a failing step to the tests, that has `do_not_whitelist=true`, so we can assert that it's *not* listed in the "you can whitelist things" output?

I think that's the last missing piece before I say ACK ;-)

upadhyeammit:
> Can we add a failing step to the tests, that has `do_not_whitelist=true`, so we can assert that it's _not_ listed in the "you can whitelist things" output?
> 
> I think that's the last missing piece before I say ACK ;-)

done :)

evgeni (approved):
ACK!

PR #602, opened 2022-04-01
Refs #33620 - limit target version upto N+1

upadhyeammit:
This makes,
1. The target-version optional, actually I think if we agree on this then we don't require target-version at all.
2. Validates if target version is only N+1 where N is the current installed version.

### DIFF
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -13,7 +13,8 @@ module ForemanMaintain::Scenarios
     end
 
     def target_version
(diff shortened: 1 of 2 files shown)


ehelms:
Looks good to me, I'd like to see what @evgeni thinks to to vet it across another.

evgeni:
Yeah, I would drop the "target version" completely in this case. What's the point of a parameter that can only ever have one value.

evgeni:
LGTM

ehelms (approved):

PR #603, opened 2022-04-01
Refs #33620 - capitalize env var and introduce repository-label option

upadhyeammit:


ehelms (approved):
I would guess this also helps out disconnected use cases?

upadhyeammit:
> I would guess this also helps out disconnected use cases?

That was already covered by ENV var however there is transparent option in cmd itself now. I kept both of them.

PR #605, opened 2022-04-06
Refs #34717 - update Dnf PackageManager with respect to foreman-prote…

upadhyeammit:
The foreman-maintain use to copy config, source files of
foreman-protector yum plugin. This change handles it using
reinstall of foreman-maintain package which puts the files in
correct location if those are missing.


lib/foreman_maintain/package_manager/dnf.rb
```diff
@@ -1,10 +1,44 @@
...
   class Dnf < Yum
+    PROTECTOR_CONFIG_FILE = '/etc/yum/pluginconf.d/foreman-protector.conf'.freeze
+    PROTECTOR_WHITELIST_FILE = '/etc/yum/pluginconf.d/foreman-protector.whitelist'.freeze
+
     def clean_cache(assumeyes: false)
       dnf_action('clean', 'all', :assumeyes => assumeyes)
       super
     end
 
+    def dnf_plugin_dirs
```
upadhyeammit, line 11: I am expecting python version will be newer in EL9 hence trying to catch all dirs which have child `dnf-plugin` dir

lib/foreman_maintain/package_manager/dnf.rb
```diff
@@ -1,10 +1,44 @@
...
+      end
+      super
+    end
+
+    def install_extras(src, dest, override: false)
+      super
+    end
+
+    def version_locking_enabled?
+      File.exist?(PROTECTOR_CONFIG_FILE) && File.exist?(PROTECTOR_WHITELIST_FILE) &&
```
upadhyeammit, line 38: hmm should I delegate this to parent class? However in parent class I would need to have if el7? condition for `PROTECTOR_PLUGIN_FILE` similar to `install_version_locking` method

lib/foreman_maintain/package_manager/dnf.rb
```diff
@@ -1,10 +1,44 @@
...
+        d + '/' + 'foreman-protector.py'
+      end
+    end
+
+    def install_version_locking(*)
+      protector_plugin_files.each do |f|
+        install_extras('foreman_protector/dnf/foreman-protector.py', f)
+      end
+      super
+    end
```
evgeni, line 31: I'm still struggling to understand, why these steps are needed at all.

The files are installed there by packaging correctly already (excerpt from EL7, but otherwise correct):
```
# rpm -ql rubygem-foreman_maintain |grep prote
/etc/yum/pluginconf.d/foreman-protector.conf
/etc/yum/pluginconf.d/foreman-protector.whitelist
/usr/lib/yum-plugins/foreman-protector.py
/usr/lib/yum-plugins/foreman-protector.pyc
/usr/lib/yum-plugins/foreman-protector.pyo
```

Moreso, I don't think tools should poke around in `/usr/lib` if they are not `rpm` themself.

I this solely for the case where f-m is installed from source/gem, vs packaging?

upadhyeammit: That is one reason and another could be if these files are missing or somehow deleted then there should be way to restore it without reinstalling rpm? I do think reinstalling rpm should be ideal way

evgeni: With the same logic, they could be missing from `/usr/share/gems/gems/foreman_maintain-1.0.3/extras/foreman_protector` where they are copied from too ;)

And yes, reinstalling the package should be the way to get the files back.

upadhyeammit: I have removed the old way of installing locking tooling and now implemented reinstall of the package if any of the files are missing. Putting dnf-plugin specific python file at correct location will be handled by packaging pr.

Let me know what you think

definitions/procedures/packages/enable_version_locking.rb
```diff
@@ -6,7 +6,11 @@ class EnableVersionLocking < ForemanMaintain::Procedure
     end
 
     def run
-      package_manager.install_version_locking(:assumeyes => @assumeyes)
+      installed_fm_packages = []
+      ['satellite-maintain', 'rubygem-foreman_maintain'].each do |pkg|
+        installed_fm_packages << find_package(pkg)
+      end
+      package_manager.reinstall(installed_fm_packages, :assumeyes => @assumeyes)
```
evgeni, line 13: why don't we just assume the files are there?

lib/foreman_maintain/package_manager/dnf.rb
```diff
@@ -5,6 +5,26 @@ def clean_cache(assumeyes: false)
...
+    def protector_plugin_files
+      @protector_plugin_files ||= dnf_plugin_dirs.map do |d|
+        d + '/' + 'foreman-protector.py'
+      end
+    end
+
+    def version_locking_enabled?
+      super &&
+        protector_plugin_files.each { |f| File.exist?(f) }
+    end
```
evgeni, line 26: what does "enabled?" mean?
* active? then the check is wrong
* available? then this should just return true as we support locking for dnf.

lib/foreman_maintain/package_manager/yum.rb
```diff
@@ -23,14 +23,8 @@ def versions_locked?
...
-      File.exist?(PROTECTOR_PLUGIN_FILE) && File.exist?(PROTECTOR_CONFIG_FILE) &&
-        File.exist?(PROTECTOR_WHITELIST_FILE)
-    end
-
-    def install_version_locking(*)
-      install_extras('foreman_protector/foreman-protector.py', PROTECTOR_PLUGIN_FILE)
-      install_extras('foreman_protector/foreman-protector.conf', PROTECTOR_CONFIG_FILE)
-      install_extras('foreman_protector/foreman-protector.whitelist', PROTECTOR_WHITELIST_FILE)
+      File.exist?(PROTECTOR_PLUGIN_FILE) if el7? && File.exist?(PROTECTOR_CONFIG_FILE) &&
+                                            File.exist?(PROTECTOR_WHITELIST_FILE)
```
evgeni, line 27:
```suggestion
      File.exist?(PROTECTOR_CONFIG_FILE) && File.exist?(PROTECTOR_WHITELIST_FILE)
```

evgeni:
I still think this does too much with the system.

The availability of the locking feature can (and should) be detected by either "this is yum or dnf, so supported" or "this is yum or dnf and the config file is there". Any smartness around finding the actual plugin is just bound to produce errors at some point.

evgeni:
CI is unhappy, but this looks like what I had intended!

upadhyeammit:
> CI is unhappy, but this looks like what I had intended!

updating the tests, as I have added extra check for existence of whitelist file its complaining

ehelms (approved):
Code seems sane

upadhyeammit:
> Code seems sane

thanks! Merging now.

definitions/procedures/packages/enable_version_locking.rb
```diff
@@ -6,7 +6,11 @@ class EnableVersionLocking < ForemanMaintain::Procedure
...
     end
```
evgeni, line 14: I think this whole file can be removed now -- it's unreferenced after you last commit.

PR #608, opened 2022-04-26
Fixes #34820 - nft rule to allow traffic from localhost

upadhyeammit:


vsedmik:
With this PR I was able to successfully run the z-stream upgrade, which localhost traffic is part of.

PR #609, opened 2022-04-26
Fixes #34821 - allow restoring miss-matched candlepin versions

evgeni:
force DB migrations to run on restore


upadhyeammit (approved):
Haven't tested however I believe in @evgeni ;) Code looks sane

evgeni:
> Haven't tested however I believe in @evgeni ;) Code looks sane

♥ I did test it on the reproducer machine for that issue :)

PR #611, opened 2022-05-06
Fixes #34874: foreman_cache_data should be a config file for backups …

ehelms:
…always

The foreman_cache_data directory contains a number of important secrets
that need to be present in a backup. This was listed as part of the Puppet Server
feature rather than Foreman server which means when we disabled Puppet
by default this path was being omitted.

### DIFF
--- a/definitions/features/foreman_server.rb
+++ b/definitions/features/foreman_server.rb
@@ -34,7 +34,8 @@ module ForemanMaintain
           '/etc/selinux/targeted/contexts/files/file_contexts.subs',
           '/etc/sysconfig/foreman',
           '/usr/share/ruby/vendor_ruby/puppet/reports/foreman.rb',
-          '/var/lib/foreman'
+          '/var/lib/foreman',
+          '/opt/puppetlabs/puppet/cache/foreman_cache_data'
         ]
       end
 
--- a/definitions/features/puppet_server.rb
+++ b/definitions/features/puppet_server.rb
@@ -11,8 +11,6 @@ class Features::PuppetServer < ForemanMaintain::Feature
     [
       '/etc/puppet',
       '/etc/puppetlabs',
-      '/opt/puppetlabs/puppet/cache/foreman_cache_data',
-      '/var/lib/puppet/foreman_cache_data',
       '/opt/puppetlabs/puppet/ssl/',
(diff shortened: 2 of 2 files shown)


ehelms:
@evgeni Please also have a look

The consequence of this not being present in the backup is that, for example, a new CA key password will be generated when it is not found in the cache and then no certificates can be generated as the CA key from the backup and the password file are mis-matched.

evgeni:
I don't think the word cache means what you think it means? ;)

No, seriously, I'm fine with the change, as it essentially moves the folder to a different feature, but can this be better stored than in a cache, which by definition is ephemeral

ehelms:
> No, seriously, I'm fine with the change, as it essentially moves the folder to a different feature, but can this be better stored than in a cache, which by definition is ephemeral

Issues welcome :duck:

evgeni (approved):

upadhyeammit (approved):

ehelms:
For additional info, the "caching" feature we use for a variety of things is this -- https://github.com/voxpupuli/puppet-extlib/blob/master/lib/puppet/functions/extlib/cache_data.rb

Don't dig too deeply into the history of who wrote it :smile: 

Other things we store there:

```
cat /opt/puppetlabs/puppet/cache/foreman_cache_data/
admin_password             candlepin_db_password      db_password                oauth_consumer_key         qpid_router_sasl_password
ca_key_password            candlepin_oauth_secret     keystore_password-file     oauth_consumer_secret      truststore_password-file
```

And looking a little closer, seems we may have another gap, although this one may not have broken anything:

```
ls /opt/puppetlabs/puppet/cache/pulpcore_cache_data/
db_password  secret_key
```

evgeni:
So this also needs to be backed up on the proxies? Wouldn't it then better be in the installer feature?

evgeni:
> Don't dig too deeply into the history of who wrote it 😄

I think I've met every single one who ever touched that file (and its previous versions) in person ;)

ehelms:
> So this also needs to be backed up on the proxies? Wouldn't it then better be in the installer feature?

Oh there is one -- I included that in a commit here https://github.com/theforeman/foreman_maintain/pull/612

PR #613, opened 2022-05-09
Fixes #34878 - enable Red Hat repositories to resolve dependencies if…

upadhyeammit:
… any

### DIFF
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -79,6 +79,20 @@ module ForemanMaintain::Scenarios
 
       true
     end
+
+    def main_rh_repos
+      if el7?
+        ["rhel-#{el_major_version}-server-rpms",
+         "rhel-server-rhscl-#{el_major_version}-rpms"]
+      else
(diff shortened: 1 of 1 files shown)


upadhyeammit:
There could be systems not having psql/nftables packages installed and newer version of foreman-maintain requires these as dependencies.
I am yet to cherry pick https://projects.theforeman.org/issues/34855 for 6.11 and I think we need this change as well.

definitions/scenarios/self_upgrade.rb
```diff
@@ -79,6 +79,20 @@ def use_rhsm?
 
       true
     end
+
+    def main_rh_repos
```
evgeni, line 83: can't we use the same method from the `downstream` concern?

upadhyeammit: > can't we use the same method from the `downstream` concern?

I felt including whole concern for a method which does not depend anything from downstream concern is too much to include, but yes we can

evgeni: I can see that, but it also means that we have to places to maintain those repo lists, and you know how that will end ;-)

evgeni (approved):

PR #614, opened 2022-05-11
Fixes #34893 - correct yum-config-manager repo options

upadhyeammit:


evgeni (approved):

upadhyeammit:
both options seems valid, there is something else to fix maybe as I can see disable repos sometimes fail with yum-config-manager error. Accordingly not merging this, will revisit this week.

upadhyeammit:
closing this pull request as we are opting for better self-upgrade approach: https://github.com/theforeman/foreman_maintain/pull/619

PR #615, opened 2022-05-13
Fixes #34922 - don't disable all repositories in self-upgrade

upadhyeammit:


definitions/scenarios/self_upgrade.rb
```diff
@@ -102,10 +104,9 @@ def compose
         pkgs_to_update = %w[satellite-maintain rubygem-foreman_maintain]
         add_step(Procedures::Repositories::BackupEnabledRepos.new)
         disable_repos
-        add_step(Procedures::Repositories::Enable.new(repos: req_repos_to_update_pkgs,
-                                                      use_rhsm: use_rhsm?))
+        enable_repos(req_repos_to_update_pkgs)
         add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true))
-        disable_repos('*')
+        disable_repos([maintenance_repo_id(maintenance_repo_version)])
```
evgeni, line 109: can you elaborate a bit on the "why"? because I don't get it XD

the old procedure was:
* get a list of enabled repos (via `BackupEnabledRepos`)
* disable all repos from that list
* enable the repos we need for the upgrade
* perform upgrade
* disable all repos
* enable repos on the "reenable" list (which is fed from the backup above)

the new one replaces the "disable all repos" step with "disable maintenance we needed", but shouldn't the end result be the same (as the reenable step converges the system to what it should look like)?

upadhyeammit: Yes the end result is same however if we disable all repos then it takes long time to complete that step. Instead if we disable only the maintenance repo + rh repos then it completes faster.

upadhyeammit:
I am closing this pull request as we are opting for better approach to handle self-upgrade itself -> https://github.com/theforeman/foreman_maintain/pull/619

PR #617, opened 2022-05-17
Refs #34878 - don't enable RH repos when custom repo mentioned

upadhyeammit:


evgeni (approved):
ACK, even tho I start thinking `use_rhsm?` is the wrong name for that function ;)

upadhyeammit:
> ACK, even tho I start thinking `use_rhsm?` is the wrong name for that function ;)

Agree, I will refactor the code for sure

PR #618, opened 2022-05-17
Fixes #34928 - sanitize the repository label for EL7

upadhyeammit:


lib/foreman_maintain/repository_manager/el.rb
```diff
@@ -76,10 +76,21 @@ def enabled_repos
...
     def hash_of_repoids_urls(repos, regex)
-      Hash[*repos.split("\n").grep(regex).map do |entry|
-             entry.split(':', 2).last.strip
-           end
+      ids_urls = Hash[*repos.split("\n").grep(regex).map do |entry|
+                        entry.split(':', 2).last.strip
+                      end
       ]
+
+      if el7?
```
evgeni, line 84: given it doesn't hurt on EL8 either, why not always execute this, removing the branching complexity here?

(we can still have a comment a la `# remove anything after the first slash in the url, like EL7 returns it`

evgeni (commented):
can we have some tests for this, it's super hard to remember how `repos` looks like and what the result should look like ;)

lib/foreman_maintain/repository_manager/el.rb
```diff
@@ -76,10 +76,22 @@ def enabled_repos
     end
 
     def hash_of_repoids_urls(repos, regex)
-      Hash[*repos.split("\n").grep(regex).map do |entry|
-             entry.split(':', 2).last.strip
-           end
+      ids_urls = Hash[*repos.split("\n").grep(regex).map do |entry|
+                        entry.split(':', 2).last.strip
+                      end
```
ehelms, line 81: This looks like a tab snuck in here?

evgeni (approved):

PR #620, opened 2022-06-01
Backup and restore for Debian

upadhyeammit:


lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -4,11 +4,21 @@ module BaseDatabase
...
         if el7? && package_manager.installed?('rh-postgresql12-postgresql-server-syspaths')
           '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'
+        elsif debian?
+          deb_psql_data_dir
         else
           '/var/lib/pgsql/data/'
         end
       end
 
+      def deb_psql_data_dir
```
upadhyeammit, line 14: Is this good or should I query the package, grep the version, trim it to two digits and then use it in the path?

lib/foreman_maintain/package_manager/apt.rb
```diff
@@ -2,7 +2,15 @@ module ForemanMaintain::PackageManager
...
-      sys.execute?(%(dpkg-query -l #{packages_list}))
+      status, output = sys.execute_with_status(%(dpkg --status #{packages_list}))
+      return false if status != 0
+
+      status_of_pkg = output.split("\n").grep(/^Status:/).first
+      if status_of_pkg.include?('installed')
+        return true
+      end
+
+      false
```
ehelms, line 13:
```suggestion
      status_of_pkg.include?('installed')
```

This returns a boolean so you don't need to do anymore than call it here.

ehelms (approved):
Small suggestion, from a code perspective LGTM

definitions/procedures/selinux/set_file_security.rb
```diff
@@ -7,6 +7,9 @@ class SetFileSecurity < ForemanMaintain::Procedure
             'Is the backup incremental?',
             :required => true
       manual_detection
+      confine do
+        package_manager.installed?(['policycoreutils'])
```
evgeni, line 11: Should this instead check for existence if `/sys/fs/selinux`? IMHO cleaner than checking for a package.

lib/foreman_maintain/concerns/base_database.rb
evgeni: Given the config path in etc also contains the version, I'd go with "extract version from pkg and user"

Keep in mind, on Debian PostgreSQL versions are co-installable, so you can have 11 and 12 and 13 on one machine (and idk if there is a good way to learn from the installer answers which one is the "right" one).

lib/foreman_maintain/concerns/firewall/maintenance_mode.rb
```diff
@@ -0,0 +1,37 @@
...
+          question, pkg = question_and_pkg_name
+          answer = ask_decision(question, actions_msg: 'y(yes), q(quit)')
+          if answer == :yes
+            packages_action(:install, pkg)
+            feature(:instance).firewall.enable_maintenance_mode
+          end
+        end
+
+        def can_install_nft?
+          nft_kernel_version = Gem::Version.new('3.13')
```
evgeni, line 16: I think instead if checking "would this kernel support nft", I'd go with "did that OS release come with nft by default". That'd be el 8+ and for Debian/Ubuntu I can look it up for you if you want.

evgeni (commented):
Not a full review, but some food for thought :)

upadhyeammit:
@evgeni I can't find path or version of Postgresql stored in installer answers. When you say 'pkg and user' does it mean ask user for the version being used for Foreman installation?

evgeni:
"extract from pkg and USE it" damn typo 🫣

Essentially, find all packages of the form "postgresql-N" and backup all of the matching paths.

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -4,13 +4,27 @@ module BaseDatabase
...
+          '/var/lib/postgresql/11/main/'
+        end
+      end
+
       def postgresql_conf
-        "#{data_dir}/postgresql.conf"
+        if el?
+          "#{data_dir}/postgresql.conf"
+        else
+          '/etc/postgresql/11/main/postgresql.conf'
```
ehelms, line 26:
```suggestion
          "#{deb_psql_data_dir}/postgresql.conf"
```

evgeni: Contrary to (stupid) Red Hat packaging, Debian actually has it's PostgreSQL config not in data but in etc ;)

upadhyeammit:
> "extract from pkg and USE it" damn typo 🫣
> 
> Essentially, find all packages of the form "postgresql-N" and backup all of the matching paths.

After giving thought for stuff like, having separate Procedure to backup the Foreman offline db on Debian, handling it in base_database concern I zeroed down on using the append mode of tar and separating the debian methods completely in existing backup procedure. The https://projects.theforeman.org/issues/35021 is tracking the changes for that.

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -5,42 +5,38 @@ def data_dir
...
+        deb_postgresql_conf_dirs.map do |conf_dir|
           "#{conf_dir}postgresql.conf"
         end
       end
 
-      def deb_psql_conf_dirs
-        deb_psql_conf = []
-        deb_psql_versions.each do |ver|
-          deb_psql_conf << "/etc/postgresql/#{ver}/main/"
+      def deb_postgresql_conf_dirs
```
ehelms, line 36: Tiny nitpick for consistency -- `deb_postgresql_config_dirs` as in other places (https://github.com/theforeman/foreman_maintain/pull/620/files/1fcac25d189d442a7cddcef850d5dd622f8c3ef0..5212ee92e5c807b781a78661830988881db85bd6#diff-b991207d7ead365fb38e633cfb372602615689abf9563a097287c996fd94d687R80) `config` is used instead of an abbreviated wording.

definitions/procedures/backup/offline/foreman_db.rb
```diff
@@ -32,20 +32,60 @@ def run
...
+          elsif pg_data_dir.size == 1
+            do_backup(pg_data_dir.first)
+          else
+            deb_local_backup
+          end
         end
       end
 
+      def deb_local_backup
+        # We take backup of first dir
```
ehelms, line 46: You could omit this comment, your code is very readable and expresses this.

definitions/procedures/backup/offline/foreman_db.rb
```diff
@@ -32,20 +32,60 @@ def run
...
+        first_dir = pg_data_dir.first
+        pg_data_dir.delete(first_dir)
+        do_backup(first_dir)
+
+        # We append the rest of the dirs to pgsql_data.tar file
```
ehelms, line 51: You could omit this comment, your code is very readable and expresses this.

ehelms:
This is not related directly to backup/restore, but I did encounter the following trying to test this:

```
root@debian11-foreman-nightly:~/foreman_maintain# bundle exec ./bin/foreman-maintain
bundler: failed to load command: ./bin/foreman-maintain (./bin/foreman-maintain)
Traceback (most recent call last):
	48: from /usr/bin/bundle:23:in `<main>'
	47: from /usr/bin/bundle:23:in `load'
	46: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/exe/bundle:37:in `<top (required)>'
	45: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/friendly_errors.rb:130:in `with_friendly_errors'
	44: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/exe/bundle:49:in `block in <top (required)>'
	43: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/cli.rb:24:in `start'
	42: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
	41: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/cli.rb:30:in `dispatch'
	40: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
	39: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
	38: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
	37: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/cli.rb:494:in `exec'
	36: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/cli/exec.rb:28:in `run'
	35: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/cli/exec.rb:63:in `kernel_load'
	34: from /usr/share/rubygems-integration/all/gems/bundler-2.2.5/lib/bundler/cli/exec.rb:63:in `load'
	33: from /root/foreman_maintain/bin/foreman-maintain:14:in `<top (required)>'
	32: from /root/foreman_maintain/bin/foreman-maintain:14:in `require'
	31: from /root/foreman_maintain/lib/foreman_maintain/cli.rb:8:in `<top (required)>'
	30: from /root/foreman_maintain/lib/foreman_maintain/cli.rb:8:in `require'
	29: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced_command.rb:1:in `<top (required)>'
	28: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced_command.rb:1:in `require'
	27: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/procedure_command.rb:1:in `<top (required)>'
	26: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/procedure_command.rb:1:in `require'
	25: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/procedure/run_command.rb:3:in `<top (required)>'
	24: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/procedure/run_command.rb:4:in `<module:ForemanMaintain>'
	23: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/procedure/run_command.rb:5:in `<module:Cli>'
	22: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/procedure/run_command.rb:6:in `<module:Procedure>'
	21: from /root/foreman_maintain/lib/foreman_maintain/cli/advanced/procedure/run_command.rb:7:in `<class:RunCommand>'
	20: from /root/foreman_maintain/lib/foreman_maintain.rb:134:in `allowed_available_procedures'
	19: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:45:in `available_procedures'
	18: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:72:in `ensure_features_detected'
	17: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:72:in `each_key'
	16: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:73:in `block in ensure_features_detected'
	15: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:112:in `detect_feature'
	14: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:112:in `find'
	13: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:112:in `each'
	12: from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:204:in `present?'
	11: from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:229:in `evaluate_confines'
	10: from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:229:in `all?'
	 9: from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:230:in `block in evaluate_confines'
	 8: from /root/foreman_maintain/lib/foreman_maintain/concerns/metadata.rb:230:in `instance_exec'
	 7: from /root/foreman_maintain/definitions/features/foreman_1_11_x.rb:6:in `block (2 levels) in <class:Foreman_1_11_x>'
	 6: from /root/foreman_maintain/lib/foreman_maintain/concerns/system_helpers.rb:37:in `check_min_version'
	 5: from /root/foreman_maintain/lib/foreman_maintain/concerns/system_helpers.rb:234:in `check_version'
	 4: from /root/foreman_maintain/lib/foreman_maintain/concerns/system_helpers.rb:124:in `package_version'
	 3: from /root/foreman_maintain/lib/foreman_maintain/concerns/system_helpers.rb:146:in `version'
	 2: from /usr/lib/ruby/vendor_ruby/rubygems/version.rb:203:in `new'
	 1: from /usr/lib/ruby/vendor_ruby/rubygems/version.rb:203:in `new'
/usr/lib/ruby/vendor_ruby/rubygems/version.rb:214:in `initialize': Malformed version number string 9999-3.4.0-bullseye+scratchbuild+20220609152101 (ArgumentError)
```

definitions/procedures/backup/config_files.rb
```diff
@@ -60,17 +60,27 @@ def config_files
...
+    def foreman_proxy_configs
+      feature(:foreman_proxy).config_files(@proxy_features)
+    end
+
+    def foreman_proxy_exclude_configs
+      feature(:foreman_proxy).config_files_to_exclude(@proxy_features)
+    end
+
+    def deb_config_dirs
+      feature(:foreman_database).config_dirs if debian
```
ehelms, line 81:
```suggestion
      feature(:foreman_database).config_dirs if debian?
```

ehelms:
Other than what I noted, I was back to run offline / online backups and restore with success.

definitions/procedures/restore/extract_files.rb
```diff
@@ -71,8 +71,18 @@ def extract_pgsql_data(backup)
...
+
+    def del_data_dir_param
       # workaround for https://tickets.puppetlabs.com/browse/MODULES-11160
-      execute("sed -i '/data_directory/d' #{any_database.postgresql_conf}")
+      if el?
+        execute("sed -i '/data_directory/d' #{any_database.postgresql_conf}")
+      else
+        any_database.postgresql_conf.each do |conf_file|
+          execute("sed -i '/data_directory/d' #{conf_file}")
+        end
```
evgeni, line 84: I'd actually omit that else, the hack was only needed to make el7 and el8 backups cross-restorable, which we don't need on debian

lib/foreman_maintain/concerns/firewall/maintenance_mode.rb
```diff
@@ -0,0 +1,36 @@
...
+          answer = ask_decision(question, actions_msg: 'y(yes), q(quit)')
+          if answer == :yes
+            packages_action(:install, pkg)
+            feature(:instance).firewall.enable_maintenance_mode
+          end
+        end
+
+        def can_install_nft?
+          # The nftables is default from EL8 and Debian 10(Buster)
+          el_major_version >= 8 || deb_major_version >= 10
```
evgeni, line 17: What about Ubuntu? If my memory is correct, they do have nftables by default since 21.10 (or rather 22.04 LTS, as that's what we would support in the future)

evgeni:
The test failure looks related

definitions/features/foreman_database.rb
```diff
@@ -16,7 +16,15 @@ def configuration
   end
 
   def config_files
-    [postgresql_conf]
+    [postgresql_conf] if el?
```
ehelms, line 19: For a guard clause to work, this needs to be an explicit return I think.

```suggestion
    return [postgresql_conf] if el?
```

ehelms: You could also use a ternary operator:

```suggestion
    el? ? [postgresql_conf] : []
```

definitions/procedures/backup/offline/foreman_db.rb
```diff
@@ -31,21 +31,46 @@ def run
...
       def pg_backup_file
         File.join(@backup_dir, 'pgsql_data.tar')
       end
 
       def pg_data_dir
+        if el?
+          [pg_data_dir_el]
+        else
+          pg_data_dirs_deb
+        end
```
ehelms, line 61: Similar to above, you can also write this with a ternary:

```
  el? ? [pg_data_dir_el] : pg_data_dirs_deb
```

evgeni (changes requested):

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -55,7 +55,11 @@ def el?
       end
 
       def debian?
-        File.exist?('/etc/debian_version')
+        os_id == 'debian'
+      end
```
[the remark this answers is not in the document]
upadhyeammit: I just pushed [Fixes #35094 - identify debian and ubuntu and use it](https://github.com/theforeman/foreman_maintain/pull/620/commits/d410dfba0f2e374382868ef57d1b33739361e4bd) does that make sense?

definitions/procedures/backup/metadata.rb
```diff
@@ -37,12 +46,7 @@ def save_metadata(metadata, spinner)
 
     def release_info(spinner)
       spinner.update('Collecting system release info')
-      execute!('cat /etc/redhat-release').chomp
-    end
-
-    def rpms(spinner)
-      spinner.update('Collecting installed RPMs')
-      execute!('rpm -qa').split("\n")
+      "#{os_name} #{os_version_id}"
```
ekohl, line 49: Does it make sense to replace this with `cat /etc/os-release` now?

definitions/procedures/backup/offline/foreman_db.rb
```diff
@@ -31,21 +31,42 @@ def run
...
+          :data_dir => pg_dir,
+          :command => cmd
+        )
+      end
+
       def pg_backup_file
         File.join(@backup_dir, 'pgsql_data.tar')
       end
 
       def pg_data_dir
```
ekohl, line 56: If you're changing this from a single value to multiple I'd let the name reflect that:
```suggestion
      def pg_data_dirs
```

definitions/procedures/restore/extract_files.rb
```diff
@@ -71,6 +71,10 @@ def extract_pgsql_data(backup)
         :transform => any_database.restore_transform
       )
       feature(:tar).run(pgsql_data_tar)
+      del_data_dir_param if el?
```
ekohl, line 74: I'm not 100% sure this is limited to EL. It's the puppetlabs-postgresql module which applies this and if they used our installer it's probably present. However, I'm unsure if that's a problem since I'm too unfamiliar with foreman-maintain and where this code is used.

definitions/procedures/backup/metadata.rb
upadhyeammit: Do you mean to collect whole file? If not then `os_name` and `os_version_id` is getting pulled from `/etc/os-release` already?

ekohl: I do indeed mean to collect the entire file. My reasoning is that previously it did `cat /etc/redhat-release` but `/etc/os-release` feels like a more natural replacement. On Ubuntu it also shows the exact patch version (`VERSION="20.04.3 LTS (Focal Fossa)"`) which could be useful, while `VERSION_ID` doesn't contain that.

That said, I don't know for sure where this is exactly used so it may be inappropriate to collect the entire file.

upadhyeammit: Its not getting used as such in restore process, however I agree that instead of VERSION_ID we should collect the VERSION. Adding the method for same in os_facts concern.

definitions/procedures/backup/config_files.rb
```diff
@@ -45,10 +45,24 @@ def run
...
+      all_configs = available_features_config
+      configs = all_configs.first
+      exclude_configs = all_configs.last
+
+      if feature(:foreman_proxy)
+        configs += foreman_proxy_configs
+        exclude_configs += foreman_proxy_exclude_configs
+      end
+
+      configs += deb_config_dirs if debian_or_ubuntu?
```
ekohl, line 59: A bit of a nit, but there is a bit of duplication going on here. I think I'd name it `config_dirs` and have all the `debian_or_ubuntu?` logic there. If it isn't, return an empty array. Then this becomes:
```suggestion
      configs += config_dirs
```

definitions/procedures/backup/config_files.rb
```diff
@@ -45,10 +45,24 @@ def run
         end
       end
     end
-    # rubocop:enable Metrics/MethodLength
 
-    # rubocop:disable Metrics/AbcSize
     def config_files
+      all_configs = available_features_config
+      configs = all_configs.first
+      exclude_configs = all_configs.last
```
ekohl, line 52: You can write:
```suggestion
      configs, exclude_configs = available_features_config
```

definitions/procedures/backup/config_files.rb
```diff
@@ -59,17 +73,20 @@ def config_files
...
 
-      configs.compact.select { |path| Dir.glob(path).any? }
-      exclude_configs.compact.select { |path| Dir.glob(path).any? }
-      [configs, exclude_configs]
+    def foreman_proxy_exclude_configs
+      feature(:foreman_proxy).config_files_to_exclude(@proxy_features)
+    end
+
+    def deb_config_dirs
+      feature(:foreman_database).config_dirs if debian_or_ubuntu?
```
ekohl, line 88: And then this should be
```suggestion
    def config_dirs
      debian_or_ubuntu? ? feature(:foreman_database).config_dirs : []
```

definitions/procedures/backup/offline/foreman_db.rb
```diff
@@ -31,21 +31,42 @@ def run
...
-          feature(:foreman_database).backup_local(
-            pg_backup_file,
-            :listed_incremental => File.join(@backup_dir, '.postgres.snar'),
-            :volume_size => @tar_volume_size,
-            :data_dir => pg_data_dir
-          )
+        with_spinner("Collecting data from #{pg_data_dir.join(',')}") do
+          pg_data_dirs.each_with_index do |pg_dir, index|
+            do_backup(pg_dir, 'create') if index == 0
+            do_backup(pg_dir, 'append') if index != 0
```
ekohl, line 37: Personally I'm not a big fan of using guards this way and would expand it. My main reason is that it's much easier (at least to me) to follow the code flow when reading it.
```suggestion
            if index == 0
              do_backup(pg_dir, 'create') 
            else
              do_backup(pg_dir, 'append')
            end
```

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -63,7 +71,23 @@ def el8?
...
-        return os_version_id.to_i if el?
+        os_version_id.to_i if el?
+      end
+
+      def deb_major_version
+        os_version_id.to_i if debian?
+      end
+
+      def ubuntu_major_version
+        os_version_id.to_i if ubuntu?
```
ekohl, line 82: I don't think this is correct: in Ubuntu the full version numbers are strings and should be treated as such. Otherwise you can't distinguish the releases:
```
[1] pry(main)> '20.04'.to_i
=> 20
[2] pry(main)> '20.10'.to_i
=> 20
```

ekohl (commented):
Mostly some minor stylistic suggestions. Only the Ubuntu major version handling doesn't appear to be correct. Perhaps these methods should be in some helper?
https://github.com/theforeman/foreman_maintain/blob/35352533facf2288c5f431098306b3fa3b51739a/lib/foreman_maintain/concerns/downstream.rb#L8-L14

lib/foreman_maintain/concerns/os_facts.rb
evgeni: While technically true, we really only support Ubuntu LTS versions, so differentiating 18 vs 20 vs 22 is perfectly sufficient. Whether it's valid to call that the *major* number might be a different q ;)

upadhyeammit: I vote for having separate pr to move the Gem::Version methods separate and then use it for versions and comparisons all over the project.

lib/foreman_maintain/concerns/firewall/maintenance_mode.rb
```diff
@@ -0,0 +1,38 @@
...
+            packages_action(:install, pkg)
+            feature(:instance).firewall.enable_maintenance_mode
+          end
+        end
+
+        def can_install_nft?
+          # The nftables is default from EL8 and Debian 10(Buster)
+          el_major_version >= 8 ||
+            deb_major_version >= 10 ||
+            ubuntu_major_version.to_i >= 22
```
ekohl, line 19: How does this work? On Debian `el_major_version` will return `nil` and that will fail:
```
[1] pry(main)> nil >= 8
NoMethodError: undefined method `>=' for nil:NilClass
from (pry):1:in `__pry__'
```

ekohl (commented):
Sorry, one last item that came to mind.

lib/foreman_maintain/concerns/firewall/maintenance_mode.rb
```diff
@@ -0,0 +1,38 @@
...
+          pkg_to_install = []
+          if can_install_nft?
+            question << 'nftables?'
+            pkg_to_install << 'nftables'
+          else
+            question << 'iptables?'
+            pkg_to_install << 'iptables'
+          end
+          question << "\nand start maintenance mode?"
+          [question, pkg_to_install]
```
ekohl, line 33: You are modifying the string in place here (using <<) which conflicts with frozen strings (default starting with Ruby 3.0). I'd avoid it altogether and reduce it to:
```suggestion          
          pkg_to_install = can_install_nft? ? 'nftables' : 'iptables'
          question = "Do you want to install missing netfilter utility #{pkg_to_install}?\nand start maintenance mode?"
          [question, [pkg_to_install]]
```

definitions/procedures/backup/offline/foreman_db.rb
```diff
@@ -31,22 +31,47 @@ def run
...
-            :volume_size => @tar_volume_size,
-            :data_dir => pg_data_dir
-          )
+        with_spinner("Collecting data from #{pg_data_dirs.join(',')}") do
+          pg_data_dirs.each_with_index do |pg_dir, index|
+            if index == 0
+              do_backup(pg_dir, 'create')
+            else
+              do_backup(pg_dir, 'append')
+            end
```
ekohl, line 40: Nit: you could even reduce it down to:
```suggestion
            do_backup(pg_dir, index == 0 ? 'create' : 'append')
```

ekohl (approved):
I'd say the string modification is something that is a real difference, the other isn't really so feel free to ignore that. Otherwise :+1:

upadhyeammit:
Thanks a lot @ehelms @ekohl and @evgeni for reviews! Merging this now.

PR #621, opened 2022-06-01
Regenerate api docs after Puppet removal

upadhyeammit:


evgeni (approved):
Unrelated to the changes in this PR: we stopped generating the cache in Foreman 3.3+, so I think `Procedures::Foreman::ApipieCache` should become a NOOP in this case.

PR #623, opened 2022-06-07
Fixes #35029 - Require Candlepin CA cert prior running installer

upadhyeammit:


definitions/procedures/installer/run_for_6_11.rb
```diff
@@ -0,0 +1,58 @@
...
+    def ask_for_ca_path
+      msg = 'The system is using external database over SSL protocol!'\
+			      "\nThe Satellite 6.11 has introduced new installer flag"\
+			      "\n--katello-candlepin-db-ssl-ca which is required to proceed further."\
+			      "\nRequest to provide path of Candlepin DB SSL CA:"
+      ask(msg)
+    end
+
+    def run_installer_with_extra_option(spinner)
+      ssl_ca_path = ask_for_ca_path
```
upadhyeammit, line 43: Should we fail here if file does not exist?

definitions/procedures/installer/run_for_6_11.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+
+    def ext_db?
+      !feature(:foreman_database).local?
+    end
+
+    def server_db_with_ssl?
+      answers = feature(:installer).answers
+      db_sslmode = answers.fetch('foreman')['db_sslmode']
```
ehelms, line 24: I lean towards we should check `answers.fetch('katello')['candlepin_db_ssl']` as for this step we care about Candlepin database configuration the most and this is the leading indicator it's been configured this way.

definitions/procedures/installer/run_for_6_11.rb
```diff
@@ -0,0 +1,58 @@
...
+      db_sslmode == 'verify-full'
+    end
+
+    def extdb_and_ssl?
+      ext_db? && server_db_with_ssl?
+    end
+
+    def ask_for_ca_path
+      msg = 'The system is using external database over SSL protocol!'\
+			      "\nThe Satellite 6.11 has introduced new installer flag"\
```
ehelms, line 34:
```suggestion
			      "\nSatellite 6.11 has introduced new installer flag"\
```

ehelms:
One idea would be to not ask the user and just copy the value of `--foreman-db-root-cert`.

upadhyeammit:
> One idea would be to not ask the user and just copy the value of `--foreman-db-root-cert`.

done!

ehelms (approved):
I ran this version of the code through my test setup and it worked great!

PR #625, opened 2022-06-14
Fixes #35056 - add repository manager for Debian

upadhyeammit:


evgeni:
Where is the repository manager used anyways? So far, I see "self upgrade" and the CheckUpstreamRepository check, which both are downstream only and thus don't really affect Debian.

Read: do we actually *need* this functionality today?

evgeni:
Ah, okay. That way round, it makes sense :)

evgeni:
One thing that I realized while looking at this again:
Today, the EL-centric repomanager supports *enable* and *disable* actions, which means that the repo needs to be defined by some other means (in the downstream case it's the CDN which via some indirection populates the `/etc/yum.repos.d/redhat.repo` of a client). However, for upstream, there is no such thing (neither for RPM nor DEB). So while you will be able to *disable* the old repository, there is no repository definition for the new one to be enabled.

Now, if we think this further, we don't even *need* to disable the old repo. We can just install the new `foreman-release.rpm` and that will replace the old repo definitions with the new one.
(And then, for debian, this also becomes an edit of the `/etc/apt/sources.list.d/foreman.list` file?)

Makes me wonder how to handle "disconnected" scenarios, where the user has custom repos.

upadhyeammit:
> One thing that I realized while looking at this again: Today, the EL-centric repomanager supports _enable_ and _disable_ actions, which means that the repo needs to be defined by some other means (in the downstream case it's the CDN which via some indirection populates the `/etc/yum.repos.d/redhat.repo` of a client). However, for upstream, there is no such thing (neither for RPM nor DEB). So while you will be able to _disable_ the old repository, there is no repository definition for the new one to be enabled.
> 
> Now, if we think this further, we don't even _need_ to disable the old repo. We can just install the new `foreman-release.rpm` and that will replace the old repo definitions with the new one. (And then, for debian, this also becomes an edit of the `/etc/apt/sources.list.d/foreman.list` file?)
> 
> Makes me wonder how to handle "disconnected" scenarios, where the user has custom repos

For disconnected situations user needs to take care of the repo setup and then whitelist the repository validate and setup steps. I am revisiting my thoughts, and mostly this pull request wont be needed.

upadhyeammit:
This stuff is not required, closing this pull request!

PR #626, opened 2022-06-14
Fixes #35057 - the concern for upstream upgrades support

upadhyeammit:


definitions/features/upstream.rb
```diff
@@ -9,6 +9,58 @@ class Features::Upstream < ForemanMaintain::Feature
   end
 
   def setup_repositories(_version)
-    raise NotImplementedError
+    activation_key = ENV['EXTERNAL_KATELLO_ACTIVATION_KEY']
```
upadhyeammit, line 12: I am thinking to keep the activation key option?

definitions/features/upstream.rb
```diff
@@ -9,6 +9,58 @@ class Features::Upstream < ForemanMaintain::Feature
...
+                --activationkey #{shellescape(activation_key)} --force))
+  end
+
+  def update_release_pkg_el(pkg_url)
+    package_manager.install(pkg_url, assumeyes: true)
+  end
+
+  def update_release_pkg_deb(pkg_url)
+    # apt don't support pkg installation using URL
+    exit_status, = execute_with_status("wget -q -P /tmp/ #{pkg_url}")
```
upadhyeammit, line 50: I don't think this is the best way to handle file download and update? or its fine?

definitions/features/upstream.rb
```diff
@@ -9,6 +9,58 @@ class Features::Upstream < ForemanMaintain::Feature
...
+    if exit_status == 0
+      package_manager.install('/tmp/foreman-release.deb', assumeyes: true)
+    else
+      warn! "Couldn't install Foreman Release Package: #{pkg_url}"
+    end
+  end
+
+  def el_repositories(el_major_version)
+    if rhel?
```
upadhyeammit, line 59: I thought to add the repository validation, but feels like we can get cases where repository labels are not standard.. so stopped here, maybe we should not do that and let the yum/dnf/apt fail if it does not find correct repositories?

definitions/features/upstream_repositories.rb
```diff
@@ -14,6 +14,36 @@ def repoids_and_urls
     repoids_and_urls
   end
 
+  VERSION_MAPPING = {
```
upadhyeammit, line 17: Thought to have some version mappings, can be removed once we test the upgrades with scenarios

definitions/features/upstream_repositories.rb
```diff
@@ -14,6 +14,36 @@ def repoids_and_urls
...
+
+    'https://deb.theforeman.org/'
+  end
+
+  def foreman_release_pkg_url(version)
+    if el?
+      return "#{server_url}/releases/#{version}/#{el_short_name}/x86_64/foreman-release.rpm"
+    end
+
+    "#{server_url}/pool/#{os_version_codename}/#{version}/f/foreman-release/foreman-release.deb"
```
upadhyeammit, line 35: Do we have some symlink for latest deb package of foreman-release? if not then is that what needs to be changed in packaging?

definitions/features/upstream.rb
```diff
@@ -8,7 +8,59 @@ class Features::Upstream < ForemanMaintain::Feature
     end
   end
 
-  def setup_repositories(_version)
-    raise NotImplementedError
+  def setup_repositories(version)
+    activation_key = ENV['EXTERNAL_KATELLO_ACTIVATION_KEY']
+    org = ENV['EXTERNAL_KATELLO_ORG']
```
ehelms, line 13: Organization's are in Foreman, so I would proffer we either call it `_ORG` or `_FOREMAN_ORG`.

What does the `EXTERNAL_` imply in this context? external to what?

definitions/features/upstream.rb
ehelms: @evgeni to answer that

definitions/features/upstream_repositories.rb
ehelms: Shall we mark it with a TODO? If it has the potential to be longer lived, it would be nice to put this in a stand-alone file for easier locating and updating.

definitions/features/upstream_repositories.rb
```diff
@@ -14,6 +14,36 @@ def repoids_and_urls
...
+  end
```
ehelms, line 36: Similar point here as https://github.com/theforeman/foreman_maintain/pull/626/files#r923806221

definitions/features/upstream.rb
upadhyeammit: I thought keeping EXTERNAL should help to avoid self registrations?

definitions/features/upstream_repositories.rb
```diff
@@ -14,6 +14,41 @@ def repoids_and_urls
...
 
+  # TO DO - have a special file to keep Foreman and Katello mapping
+  VERSION_MAPPING = {
+    '3.2' => '4.4',
+    '3.3' => '4.5'
+  }.freeze
+
+  def server_url
+    if el?
+      return 'https://yum.theforeman.org/'
```
ehelms, line 25:
```suggestion
      'https://yum.theforeman.org/'
```

ehelms:
Are there unit tests that can be written for this new functionality?

upadhyeammit:
> Are there unit tests that can be written for this new functionality?

While writing the tests I am leaning more towards having concern similar to downstream for upstream and include it in foreman_install, katello_install scenarios. Changing it accordingly.

ehelms (approved):

ehelms (approved):
Updates look good, once you merge, if you could rebase https://github.com/theforeman/foreman_maintain/pull/633 as it does not rebase cleanly and I want to take it for a test drive.

lib/foreman_maintain/concerns/upstream.rb
```diff
@@ -0,0 +1,109 @@
...
+          raise 'Unknown operating system detected!'
+        end
+      end
+
+      def foreman_release_pkg_url(version)
+        if el?
+          "#{server_url}releases/#{version}/#{el_short_name}/x86_64/foreman-release.rpm"
+        elsif debian_or_ubuntu?
+          "#{server_url}pool/#{os_version_codename}/#{version}"\
+          '/f/foreman-release/foreman-release.deb'
```
evgeni, line 26: Today, this URL is 404 as we don't have a `latest` symlink. I'm fine merging this PR without it being present, but we need a redmine to track it?

lib/foreman_maintain/concerns/upstream.rb
```diff
@@ -0,0 +1,109 @@
...
+        end
+      end
+
+      def katello_pkgs_url(katello_version)
+        "#{server_url}katello/#{katello_version}/katello/#{el_short_name}/x86_64/"
+      end
+
+      def katello_release_pkg(version)
+        katello_version = VERSION_MAPPING[version]
+        "#{katello_pkgs_url(katello_version)}katello-repos-latest.rpm"
```
evgeni, line 38: I really wish we wouldn't need this mapping.
Can you please open a tracking redmine to "simplify" this? Probably by just placing the `katello-repos` RPM in both repos (foreman and katello)?

evgeni (approved):
two todo-items, but both are outside of foreman-maintain, so let's GOOOOOOO!

lib/foreman_maintain/concerns/upstream.rb
upadhyeammit: yes, here we go: https://projects.theforeman.org/issues/35348 and [RFC](https://community.theforeman.org/t/introduction-of-a-new-katello-release-rpm/29749) you might have already checked that though

lib/foreman_maintain/concerns/upstream.rb
upadhyeammit: here we go: https://projects.theforeman.org/issues/35349

lib/foreman_maintain/concerns/upstream.rb
evgeni: I did not, wasn't following discourse the last few weeks ;)

PR #629, opened 2022-06-30
run all tests via ruby/setup-ruby and enable bundler caching

evgeni:


upadhyeammit:
Thanks @evgeni for these changes!

PR #632, opened 2022-07-12
Fixes #35201 - Add Satellite and Capsule 6.12 upgrade scenarios

upadhyeammit:


upadhyeammit:
I think its time to remove 2.0 from test matrix as the new code won't run on this version.

evgeni:
> I think its time to remove 2.0 from test matrix as the new code won't run on this version.

Given that will be RHEL8 only, I am OK with that!

PR #633, opened 2022-07-18
Fixes #35058 - upgrade scenario for Foreman and Katello nightly

upadhyeammit:
Depends on  https://github.com/theforeman/foreman_maintain/pull/626

The flow for the Foreman and Katello upgrade should be as below,

1. Have the correct repositories as per the version of the scenario. If scenario is for .z then update the release packages, in this case if package is missing update should install it and make relevant repositories available.
2. After getting the repositories, run the pre upgrade checks, procedures.
3. Enable correct modules if those are not enabled already.
4. Update the packages.
5. Run the installer
6. Execute any post upgrade procedures and checks.

### DIFF
--- a/definitions/features/installer.rb
+++ b/definitions/features/installer.rb
@@ -92,9 +92,16 @@ class Features::Installer < ForemanMaintain::Feature
   end
 
   def run(arguments = '', exec_options = {})
-    out = execute!("LANG=en_US.utf-8 #{installer_command} #{arguments}".strip, exec_options)
+    out = execute!("#{installer_command} #{arguments}".strip, exec_options)
     @configuration = nil
     out
+  end
+
+  def run_with_status(arguments = '', exec_options = {})
+    cmd_with_arguments = "#{installer_command} #{arguments}".strip
+    cmd_status, out = execute_with_status(cmd_with_arguments, exec_options)
+    @configuration = nil
+    [cmd_status, out]
   end
 
   def upgrade(exec_options = {})
@@ -105,13 +112,15 @@ class Features::Installer < ForemanMaintain::Feature
     installer_args = ''
 
     if feature(:foreman_proxy) &&
-       feature(:foreman_proxy).with_content?
+       feature(:foreman_proxy).with_content? &&
+       check_max_version('foreman-installer', '3.4')
       installer_args += ' --disable-system-checks'
     end
 
-    unless check_min_version('foreman', '2.1') || check_min_version('foreman-proxy', '2.1')
+    unless check_min_version('foreman-installer', '2.1')
       installer_args += ' --upgrade' if can_upgrade?
     end
+
     installer_args
   end
 
--- a/definitions/features/instance.rb
+++ b/definitions/features/instance.rb
@@ -60,8 +60,13 @@ class Features::Instance < ForemanMaintain::Feature
     end
   end
 
+  def foreman_server_ssl_ca
+    @foreman_server_ssl_ca ||= feature(:installer).answers['foreman']['server_ssl_ca']
+  end
+
   def server_connection
     net = Net::HTTP.new(ForemanMaintain.config.foreman_url, ForemanMaintain.config.foreman_port)
+    net.ca_file = foreman_server_ssl_ca
     net.use_ssl = true
     net
   end
--- a/definitions/procedures/packages/enable_modules.rb
+++ b/definitions/procedures/packages/enable_modules.rb
@@ -1,0 +1,19 @@
+module Procedures::Packages
+  class EnableModules < ForemanMaintain::Procedure
+    metadata do
+      description 'Enable the given stream modules'
+      confine do
+        package_manager.class.name == 'ForemanMaintain::PackageManager::Dnf'
+      end
+      param :module_names, 'Module names', :array => true, :required => true
+      param :assumeyes, 'Do not ask for confirmation'
+      advanced_run false
+    end
+
+    def run
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+      dnf_option = assumeyes_val ? ' -y' : ''
+      execute!("dnf module enable #{@module_names.join(' ')} #{dnf_option}")
+    end
+  end
+end
--- a/definitions/procedures/packages/enable_modules.rb
+++ b/definitions/procedures/packages/enable_modules.rb
@@ -6,14 +6,11 @@ module Procedures::Packages
         package_manager.class.name == 'ForemanMaintain::PackageManager::Dnf'
       end
       param :module_names, 'Module names', :array => true, :required => true
-      param :assumeyes, 'Do not ask for confirmation'
       advanced_run false
     end
 
     def run
-      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
-      dnf_option = assumeyes_val ? ' -y' : ''
-      execute!("dnf module enable #{@module_names.join(' ')} #{dnf_option}")
+      execute!("dnf module enable #{@module_names.join(' ')} -y")
     end
   end
 end
--- a/definitions/scenarios/upgrade_to_foreman_nightly.rb
+++ b/definitions/scenarios/upgrade_to_foreman_nightly.rb
@@ -1,0 +1,90 @@
+module Scenarios::Foreman_Nightly
+  class Abstract < ForemanMaintain::Scenario
+    def self.upgrade_metadata(&block)
+      metadata do
+        tags :upgrade_scenario
+        confine do
+          feature(:foreman_install) || ForemanMaintain.upgrade_in_progress == 'nightly'
+        end
+        instance_eval(&block)
+      end
+    end
+
+    def target_version
+      'nightly'
+    end
+  end
+
+  class PreUpgradeCheck < Abstract
+    upgrade_metadata do
+      description 'Checks before upgrading to Foreman nightly'
+      tags :pre_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+    end
+  end
+
+  class PreMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures before upgrading to Foreman nightly'
+      tags :pre_migrations
+    end
+
+    def compose
+      add_steps(find_procedures(:pre_migrations))
+    end
+  end
+
+  class Migrations < Abstract
+    upgrade_metadata do
+      description 'Upgrade steps for Foreman nightly'
+      tags :migrations
+      run_strategy :fail_fast
+    end
+
+    def set_context_mapping
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      add_step(Procedures::Repositories::Setup.new(:version => 'nightly'))
+      if el?
+        modules_to_enable = ["foreman:#{el_short_name}"]
+        add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable,
+                                                         :assumeyes => true))
+      end
+      add_step(Procedures::Packages::Update.new(:assumeyes => true))
+      add_step_with_context(Procedures::Installer::Upgrade)
+    end
+  end
+
+  class PostMigrations < Abstract
+    upgrade_metadata do
+      description 'Post upgrade procedures for Foreman nightly'
+      tags :post_migrations
+    end
+
+    def compose
+      add_step(Procedures::RefreshFeatures)
+      add_step(Procedures::Service::Start.new)
+      add_steps(find_procedures(:post_migrations))
+    end
+  end
+
+  class PostUpgradeChecks < Abstract
+    upgrade_metadata do
+      description 'Checks after upgrading to Foreman nightly'
+      tags :post_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:post_upgrade))
+    end
+  end
+end
--- a/definitions/scenarios/upgrade_to_foreman_nightly.rb
+++ b/definitions/scenarios/upgrade_to_foreman_nightly.rb
@@ -54,8 +54,7 @@ module Scenarios::Foreman_Nightly
       add_step(Procedures::Repositories::Setup.new(:version => 'nightly'))
       if el?
         modules_to_enable = ["foreman:#{el_short_name}"]
-        add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable,
-                                                         :assumeyes => true))
+        add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
       end
       add_step(Procedures::Packages::Update.new(:assumeyes => true))
       add_step_with_context(Procedures::Installer::Upgrade)
--- a/definitions/scenarios/upgrade_to_katello_nightly.rb
+++ b/definitions/scenarios/upgrade_to_katello_nightly.rb
@@ -1,0 +1,88 @@
+module Scenarios::Katello_Nightly
+  class Abstract < ForemanMaintain::Scenario
+    def self.upgrade_metadata(&block)
+      metadata do
+        tags :upgrade_scenario
+        confine do
+          feature(:katello_install) || ForemanMaintain.upgrade_in_progress == 'nightly'
+        end
+        instance_eval(&block)
+      end
+    end
+
+    def target_version
+      'nightly'
+    end
+  end
+
+  class PreUpgradeCheck < Abstract
+    upgrade_metadata do
+      description 'Checks before upgrading to Katello nightly'
+      tags :pre_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+    end
+  end
+
+  class PreMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures before upgrading to Katello nightly'
+      tags :pre_migrations
+    end
+
+    def compose
+      add_steps(find_procedures(:pre_migrations))
+    end
+  end
+
+  class Migrations < Abstract
+    upgrade_metadata do
+      description 'Upgrade steps for Katello nightly'
+      tags :migrations
+      run_strategy :fail_fast
+    end
+
+    def set_context_mapping
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      add_step(Procedures::Repositories::Setup.new(:version => 'nightly'))
+      modules_to_enable = ["katello:#{el_short_name}", "pulpcore:#{el_short_name}"]
+      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable,
+                                                       :assumeyes => true))
+      add_step(Procedures::Packages::Update.new(:assumeyes => true))
+      add_step_with_context(Procedures::Installer::Upgrade)
+    end
+  end
+
+  class PostMigrations < Abstract
+    upgrade_metadata do
+      description 'Post upgrade procedures for Katello nightly'
+      tags :post_migrations
+    end
+
+    def compose
+      add_step(Procedures::RefreshFeatures)
+      add_step(Procedures::Service::Start.new)
+      add_steps(find_procedures(:post_migrations))
+    end
+  end
+
+  class PostUpgradeChecks < Abstract
+    upgrade_metadata do
+      description 'Checks after upgrading to Katello nightly'
+      tags :post_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:post_upgrade))
+    end
+  end
+end
--- a/definitions/scenarios/upgrade_to_katello_nightly.rb
+++ b/definitions/scenarios/upgrade_to_katello_nightly.rb
@@ -53,8 +53,7 @@ module Scenarios::Katello_Nightly
     def compose
       add_step(Procedures::Repositories::Setup.new(:version => 'nightly'))
       modules_to_enable = ["katello:#{el_short_name}", "pulpcore:#{el_short_name}"]
-      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable,
-                                                       :assumeyes => true))
+      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
       add_step(Procedures::Packages::Update.new(:assumeyes => true))
       add_step_with_context(Procedures::Installer::Upgrade)
     end
--- a/definitions/features/capsule.rb
+++ b/definitions/features/capsule.rb
@@ -1,5 +1,6 @@
 class Features::Capsule < ForemanMaintain::Feature
   include ForemanMaintain::Concerns::Downstream
+  include ForemanMaintain::Concerns::Versions
 
   metadata do
     label :capsule
--- a/definitions/features/foreman_1_11_x.rb
+++ b/definitions/features/foreman_1_11_x.rb
@@ -1,9 +1,0 @@
-require 'features/foreman_1_7_x'
-
-class Features::Foreman_1_11_x < Features::Foreman_1_7_x
-  metadata do
-    confine do
-      check_min_version('foreman', '1.11')
-    end
-  end
-end
--- a/definitions/features/foreman_1_7_x.rb
+++ b/definitions/features/foreman_1_7_x.rb
@@ -1,9 +1,0 @@
-class Features::Foreman_1_7_x < ForemanMaintain::Feature
-  metadata do
-    label :foreman
-
-    confine do
-      check_min_version('foreman', '1.7')
-    end
-  end
-end
--- a/definitions/features/foreman_install.rb
+++ b/definitions/features/foreman_install.rb
@@ -1,0 +1,12 @@
+class Features::ForemanInstall < ForemanMaintain::Feature
+  include ForemanMaintain::Concerns::Upstream
+  include ForemanMaintain::Concerns::Versions
+
+  metadata do
+    label :foreman_install
+
+    confine do
+      !feature(:instance).downstream && !feature(:katello) && feature(:foreman_server)
+    end
+  end
+end
--- a/definitions/features/foreman_server.rb
+++ b/definitions/features/foreman_server.rb
@@ -1,6 +1,7 @@
 module ForemanMaintain
   module Features
     class ForemanServer < ForemanMaintain::Feature
+      include ForemanMaintain::Concerns::Versions
       metadata do
         label :foreman_server
         confine do
@@ -46,6 +47,10 @@ module ForemanMaintain
 
       def services_running?
         services.all?(&:running?)
+      end
+
+      def current_version
+        @current_version ||= package_version('foreman')
       end
     end
   end
--- a/definitions/features/katello.rb
+++ b/definitions/features/katello.rb
@@ -1,4 +1,6 @@
 class Features::Katello < ForemanMaintain::Feature
+  include ForemanMaintain::Concerns::Versions
+
   metadata do
     label :katello
 
--- a/definitions/features/katello_install.rb
+++ b/definitions/features/katello_install.rb
@@ -1,0 +1,12 @@
+class Features::KatelloInstall < ForemanMaintain::Feature
+  include ForemanMaintain::Concerns::Upstream
+  include ForemanMaintain::Concerns::Versions
+
+  metadata do
+    label :katello_install
+
+    confine do
+      !feature(:instance).downstream && feature(:katello)
+    end
+  end
+end
--- a/definitions/features/satellite.rb
+++ b/definitions/features/satellite.rb
@@ -1,5 +1,6 @@
 class Features::Satellite < ForemanMaintain::Feature
   include ForemanMaintain::Concerns::Downstream
+  include ForemanMaintain::Concerns::Versions
 
   metadata do
     label :satellite
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -1,6 +1,8 @@
 module ForemanMaintain::Scenarios
   class SelfUpgradeBase < ForemanMaintain::Scenario
     include ForemanMaintain::Concerns::Downstream
+    include ForemanMaintain::Concerns::Versions
+
     def target_version
       @target_version ||= current_version.bump
     end
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -21,6 +21,7 @@ module ForemanMaintain
   require 'foreman_maintain/concerns/hammer'
   require 'foreman_maintain/concerns/base_database'
   require 'foreman_maintain/concerns/directory_marker'
+  require 'foreman_maintain/concerns/versions'
   require 'foreman_maintain/concerns/downstream'
   require 'foreman_maintain/concerns/upstream'
   require 'foreman_maintain/concerns/primary_checks'
--- a/lib/foreman_maintain/concerns/downstream.rb
+++ b/lib/foreman_maintain/concerns/downstream.rb
@@ -5,18 +5,6 @@ module ForemanMaintain
         raise NotImplementedError
       end
 
-      def less_than_version?(version)
-        Gem::Version.new(current_version) < Gem::Version.new(version)
-      end
-
-      def at_least_version?(version)
-        Gem::Version.new(current_version) >= Gem::Version.new(version)
-      end
-
-      def current_minor_version
-        current_version.to_s[/^\d+\.\d+/]
-      end
-
       def repository_manager
         ForemanMaintain.repository_manager
       end
--- a/lib/foreman_maintain/concerns/versions.rb
+++ b/lib/foreman_maintain/concerns/versions.rb
@@ -1,0 +1,17 @@
+module ForemanMaintain
+  module Concerns
+    module Versions
+      def less_than_version?(version)
+        Gem::Version.new(current_version) < Gem::Version.new(version)
+      end
+
+      def at_least_version?(version)
+        Gem::Version.new(current_version) >= Gem::Version.new(version)
+      end
+
+      def current_minor_version
+        current_version.to_s[/^\d+\.\d+/]
+      end
+    end
+  end
+end
--- a/lib/foreman_maintain/scenario.rb
+++ b/lib/foreman_maintain/scenario.rb
@@ -4,6 +4,7 @@ module ForemanMaintain
     include Concerns::SystemHelpers
     include Concerns::ScenarioMetadata
     include Concerns::Finders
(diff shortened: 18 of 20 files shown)


definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,91 @@
...
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+    end
+  end
+
+  class PreMigrations < Abstract
```
ehelms, line 33: Apologies if this is nomenclature that is already used throughout, and consider this more like educating me. There is 'Upgrade' lingo used in these procedures, e.g. `PreUpgradeCheck` and then there is Migration lingo used such as this step. 

What in this context does upgrade vs migration mean?

definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,91 @@
...
+      run_strategy :fail_fast
+    end
+
+    def set_context_mapping
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      add_step(Procedures::Repositories::Setup.new(:version => '3.3'))
+      modules_to_enable = ['ruby:2.7', 'foreman:el8']
```
ehelms, line 57: The foreman module by itself should be enough here, it has the dependency on the Ruby module.

definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,91 @@
...
+      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
+      add_step(Procedures::Packages::Update.new(:assumeyes => true))
+      add_step(Procedures::Service::Stop.new)
```
ehelms, line 60: This ought not be needed.

ehelms:
Ought the commit message only mention Foreman and not Katello?

definitions/scenarios/upgrade_to_foreman_3_3.rb
upadhyeammit: I agree, the different steps should be indeed like,
~~~
PreUpgradeChecks
PreUpgradeProcedures
Upgrade
PostUpgradeProcedures
PostUpgradeChecks
~~~
As per these the descriptions should be updated as well. These were introduced with upgrades feature in early stage and being used as is, maybe no one questioned about the confusions.

definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,91 @@
+module Scenarios::Foreman_3_3
+  class Abstract < ForemanMaintain::Scenario
+    def self.upgrade_metadata(&block)
+      metadata do
+        tags :upgrade_scenario
+        confine do
+          !feature(:satellite) && feature(:foreman)
```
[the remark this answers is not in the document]
upadhyeammit: yeah, we need to explicitly mark not for katello, updating it.

definitions/scenarios/upgrade_to_foreman_3_3.rb
upadhyeammit: removed the stop step

definitions/scenarios/upgrade_to_foreman_3_3.rb
upadhyeammit: removed

definitions/scenarios/upgrade_to_foreman_3_3.rb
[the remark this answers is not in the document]
upadhyeammit: Thought to have separate features for multiple conditions. Like only_foreman, only_katello. I thought to name those foreman_upstream and katello_upstream but felt that does not fit well. Let me know what you think of it.

definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
+    end
+  end
+
+  class PreMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures before migrating to Foreman 3.3'
```
ehelms, line 35:
```suggestion
      description 'Procedures before upgrading to Foreman 3.3'
```

Since we cannot change the internal step names, we could at least align our nomenclature that user's see.

definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    def compose
+      add_steps(find_procedures(:pre_migrations))
+    end
+  end
+
+  class Migrations < Abstract
+    upgrade_metadata do
+      description 'Migration scripts to Foreman 3.3'
```
ehelms, line 46:
```suggestion
      description 'Upgrade steps for Foreman 3.3'
```

definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+      add_step(Procedures::Packages::Update.new(:assumeyes => true))
+      add_step_with_context(Procedures::Installer::Upgrade)
+      add_step(Procedures::Installer::UpgradeRakeTask)
+    end
+  end
+
+  class PostMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures after migrating to Foreman 3.3'
```
ehelms, line 69:
```suggestion
      description 'Post upgrade procedures for Foreman 3.3'
```

definitions/features/only_foreman.rb
```diff
@@ -0,0 +1,9 @@
+class Features::OnlyForeman < ForemanMaintain::Feature
+  metadata do
+    label :only_foreman
```
ehelms, line 3: A couple of ideas for this and the Katello one as well:

 * foreman_scenario / katello_scenario (ties it to the installation type)
 * foreman_installer_scenario / katello_installer_scenario (same as above but more verbose)
 * foreman_install / katello_install

definitions/scenarios/upgrade_to_katello_4_5.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def target_version
+      '4.5'
+    end
+  end
+
+  class PreUpgradeCheck < Abstract
+    upgrade_metadata do
+      description 'Checks before upgrading to Katello 4.5'
```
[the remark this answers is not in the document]
upadhyeammit: Yes description is required as thats what we show while doing the step on terminal. I absolutely agree that its repetitive, actually the whole scenarios are having too much duplicate code. I feel we should address the whole scenarios refactoring all together.

https://projects.theforeman.org/issues/35252

definitions/scenarios/upgrade_to_foreman_3_3.rb
upadhyeammit: Yeah we can't maybe while refactoring the whole upgrade scenarios this can be address, forked the issue for same : https://projects.theforeman.org/issues/35252

ehelms:
Could you include on here a comment describing what the expected workflow for upgrades will be?

definitions/features/installer.rb
```diff
@@ -105,13 +105,19 @@ def installer_arguments
     installer_args = ''
 
     if feature(:foreman_proxy) &&
-       feature(:foreman_proxy).with_content?
+       feature(:foreman_proxy).with_content? &&
+       !feature(:katello_install)
       installer_args += ' --disable-system-checks'
```
upadhyeammit, line 110: I think this option with whole if block can be removed now?

definitions/features/installer.rb
```diff
@@ -105,13 +105,19 @@ def installer_arguments
...
     end
 
     unless check_min_version('foreman', '2.1') || check_min_version('foreman-proxy', '2.1')
       installer_args += ' --upgrade' if can_upgrade?
     end
+
+    if feature(:instance).upstream_install
+      installer_args += ' --tuning development'
```
upadhyeammit, line 118: I was able to install the Katello 4.4 with default tuning profile I think the default tuning requirement of 20GB changed in 4.5 or later,
~~~
2022-08-05 10:32:40 [NOTICE] [root] Running installer with log based terminal output at level NOTICE.
2022-08-05 10:32:40 [NOTICE] [root] Use -l to set the terminal output log level to ERROR, WARN, NOTICE, INFO, or DEBUG. See --full-help for definitions.
Insufficient memory for tuning size
Tuning profile 'default' requires at least 20 GB of memory and 4 CPU cores
                                      [FAIL]
~~~

How we should handle the situation where existing profile was default but we changed the defaults and now system fails to upgrade?

upadhyeammit:
Current status:
- Tried Foreman upgrades
  - Foreman 3.2 to 3.3 upgrade worked successfully, cmd,
  -  `# foreman-maintain upgrade run --target-version 3.3 --disable-self-upgrade `
  - Foreman 3.3 to nightly upgrade worked successfully, cmd,
  -  `# foreman-maintain upgrade run --target-version nightly --disable-self-upgrade `
- Tried Katello upgrades
  - Katello 4.4 to 4.5 worked successfully, cmd,
  -  `# foreman-maintain upgrade run --target-version 4.5 --disable-self-upgrade `
  - Katello 4.5 to nightly worked successfully, cmd,
  -  `# foreman-maintain upgrade run --target-version nightly --disable-self-upgrade `
 
- I tried EL upgrades only, on deb we are yet to solve foreman-release.deb situation.
- There is no nightly.z as users can just use nightly version to keep updating the system.
- Nightly upgrades are not bound to current version of Foreman or Katello, which means can be tried from older versions as well.
- The repositories we care for are only of Foreman, Katello and the current implementation expects other repositories are in place and configured correctly.

ehelms:
Is `--disable-self-upgrade` required in every case? What happens if its not included? If so, is there a way to drop the need to include the flag?

upadhyeammit:
> Is `--disable-self-upgrade` required in every case? What happens if its not included? If so, is there a way to drop the need to include the flag?

I used that to save the time, however now I am feeling the `foreman-maintain` package update should happen after setting up the new version or target version repositories which makes more sense. I will raise separate pr for same.

upadhyeammit:
> > Is `--disable-self-upgrade` required in every case? What happens if its not included? If so, is there a way to drop the need to include the flag?
> 
> I used that to save the time, however now I am feeling the `foreman-maintain` package update should happen after setting up the new version or target version repositories which makes more sense. I will raise separate pr for same.

I was tempted to create another pr for this one, however codewise I feel its not big change, so added it here. However happy to create separate pr if required.

ehelms (approved):
I did some release to nightly testing (of simpler scenarios) and everything ran smooth. I am good with squash commits, merge, get a release and start incorporating this into our nightly upgrade pipelines.

definitions/features/installer.rb
ehelms: That is a good question, we are faced with two choices that I see:

 1) Ensure the upgrade happens no matter what the case
 2) Throw an error if the system is out of compliance with the requirements so that the user can fix that before upgrading

@evgeni @ekohl Would appreciate your thoughts here

evgeni: I vote 2?

ehelms: A question is, if we want to ensure the upgrade continues should we run this with `--dont-save-answers` to only set the profile during runtime and not change it for the user permanently.

definitions/features/installer.rb
ehelms: This was dropped for Foreman 3.4+.

ekohl: And for vanilla Foreman it never existed in the first place.

definitions/features/installer.rb
ekohl: IMHO this is one of those things we can't answer so I'd also vote option 2. In the future if we do need to raise requirements the user should also be aware of this. They may be fine with increasing the server size rather than suffering reduced performance.

The best way to detect this would be to use a specific exit code. Then foreman-maintain can check for it. The good news is that we already do: codes 102 (insufficient CPU cores) and 103 (insufficient memory) are used.

definitions/features/instance.rb
```diff
@@ -62,6 +62,8 @@ def ping
 
   def server_connection
     net = Net::HTTP.new(ForemanMaintain.config.foreman_url, ForemanMaintain.config.foreman_port)
+    # the foreman_ping doesn't work without VERIFY_NONE
+    net.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
ekohl, line 66: In general I'm not too happy about `VERIFY_NONE` use. Perhaps you can get the foreman_ca_cert from the installer's answer files to verify the connection?

definitions/scenarios/upgrade_to_katello_nightly.rb
```diff
@@ -0,0 +1,91 @@
...
+      tags :migrations
+      run_strategy :fail_fast
+    end
+
+    def set_context_mapping
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      # The version 3.3 is of relevant Foreman version on which Katello 4.5 released
```
ekohl, line 54: What does 3.3 refer to here?

definitions/scenarios/upgrade_to_katello_nightly.rb
```diff
@@ -0,0 +1,91 @@
...
+      add_step(Procedures::Repositories::Setup.new(:version => 'nightly'))
+      modules_to_enable = ['katello:el8', 'pulpcore:el8']
```
ekohl, line 56: In other places you automated the `el8` bit, which will help in the future when we add EL9.

definitions/scenarios/upgrade_to_katello_4_5_z.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    def set_context_mapping
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      # The version 3.3 is of relevant Foreman version on which Katello 4.5 released
+      add_step(Procedures::Repositories::Setup.new(:version => '3.3'))
+      modules_to_enable = ["foreman:#{el_short_name}",
+                           "katello:#{el_short_name}", "pulpcore:#{el_short_name}"]
```
ekohl, line 59: katello depends on foreman
```suggestion
      modules_to_enable = ["katello:#{el_short_name}", "pulpcore:#{el_short_name}"]
```

definitions/features/installer.rb
```diff
@@ -105,13 +105,20 @@ def installer_arguments
     installer_args = ''
 
     if feature(:foreman_proxy) &&
-       feature(:foreman_proxy).with_content?
+       feature(:foreman_proxy).with_content? &&
+       !feature(:katello_install) &&
```
ekohl, line 109: This looks odd. Why not `katello_install`? Katello itself also has `--disable-system-checks`.

definitions/checks/foreman_installer/installer_noop_run.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      def run
+        check_error_codes = [102, 103]
+        cmd_status, output = feature(:installer).run_with_status(' -n')
+        msg = "\nThe system does not comply with checks in the installer!\n"\
+              "Fix the requirements:\n#{output}"
+        if check_error_codes.include?(cmd_status)
+          fail! msg
+        elsif cmd_status == 1
+          fail! output
```
ekohl, line 17: Why only on code 1? all non-zero exit codes should be treated as a problem.

Looking at https://github.com/theforeman/kafo#exit-code I wonder if we should have reused code 20 instead of 102 and 103. Besides kafo there are also exit codes in `foreman-installer` but various are not `1` (see `grep fail_and_exit hooks`).

definitions/features/installer.rb
```diff
@@ -97,6 +97,13 @@ def run(arguments = '', exec_options = {})
     out
   end
 
+  def run_with_status(arguments = '', exec_options = {})
+    cmd_with_arguments = "LANG=en_US.utf-8 #{installer_command} #{arguments}".strip
```
ekohl, line 101: This is an interesting one. Have we always forced the `en_US` locale? In the past it was needed to work around bugs, but we no longer need that. The only thing is that we can't run with `LOCALE=C`. However, with EL8 it defaults to `LOCALE=C.UTF-8` if the requested locale isn't found so that check probably doesn't hit anything anymore. I think EL8 doesn't even ship the `C` locale anymore.

definitions/scenarios/upgrade_to_foreman_3_3.rb
```diff
@@ -0,0 +1,93 @@
...
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      add_step(Procedures::Repositories::Setup.new(:version => '3.3'))
+      if el?
+        modules_to_enable = ["foreman:#{el_short_name}"]
+        add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable,
+                                                         :assumeyes => true))
+      end
```
ekohl, line 61: Foreman 3.3 still supports EL7. Will we not package it for 3.2/3.3 on EL7 or is there a mechanism to make this only apply to EL 8? If so, why is the `if el?` even there?

lib/foreman_maintain/cli/upgrade_command.rb
```diff
@@ -52,12 +55,22 @@ def print_versions(target_versions)
...
       end
 
+      def do_self_upgrade?
+        if ForemanMaintain.detector.feature(:foreman_install) ||
+           ForemanMaintain.detector.feature(:katello_install) ||
+           disable_self_upgrade?
+          false
+        else
+          true
+        end
```
ekohl, line 65:
```suggestion
        !(ForemanMaintain.detector.feature(:foreman_install) ||
          ForemanMaintain.detector.feature(:katello_install) ||
          disable_self_upgrade?)
```

definitions/scenarios/upgrade_to_foreman_3_3.rb
upadhyeammit: Apologies for confusion! The whole upstream upgrade story should be supporting from Foreman 3.4, I mean 3.4 to 3.5 upgrade. Considering we already have 3.4 near to GA then maybe from 3.5. However to test the upgrade scenarios I added the older upgrade paths.

ekohl: Ah, fair enough.

definitions/checks/foreman_installer/installer_noop_run.rb
upadhyeammit: oh yes, sorry for the miss

upadhyeammit: Added 20 error code and rest all are getting caught in elseif block.

definitions/features/installer.rb
upadhyeammit: I have added the Check which runs the installer in noop mode and throws error if any in pre upgrade stage. As I am at it and as per the IRC discussion changed the unless condition to depend on `foreman-installer` package instead of `foreman` and `foreman-proxy` to include `--upgrade` flag.

definitions/checks/foreman_installer/installer_noop_run.rb
```diff
@@ -0,0 +1,22 @@
...
+  module ForemanInstaller
+    class InstallerNoopRun < ForemanMaintain::Check
+      metadata do
+        description 'Run the installer in noop mode for system checks'
+        tags :pre_upgrade
+      end
+
+      def run
+        check_error_codes = [20, 102, 103]
+        cmd_status, output = feature(:installer).run_with_status(' -n')
```
evgeni, line 11:
```suggestion
        cmd_status, output = feature(:installer).run_with_status(' --noop')
```

nitpick, but long options are always easier to read ;)

definitions/checks/foreman_installer/installer_noop_run.rb
```diff
@@ -0,0 +1,22 @@
+module Checks
+  module ForemanInstaller
+    class InstallerNoopRun < ForemanMaintain::Check
+      metadata do
+        description 'Run the installer in noop mode for system checks'
```
evgeni, line 5: This will do a full `--noop` run of the installer, which can take some time.

I wonder if the installer should have a `--checks-only` option? cc @ehelms @ekohl

definitions/procedures/packages/enable_modules.rb
```diff
@@ -0,0 +1,18 @@
...
+        package_manager.class.name == 'Dnf'
+      end
+      param :module_names, 'Module names', :array => true, :required => true
+      param :assumeyes, 'Do not ask for confirmation'
+    end
+
+    def run
+      assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
+      dnf_option = assumeyes_val ? ' -y' : ''
+      execute!("dnf module enable #{@module_names.join(' ')} #{dnf_option}")
```
evgeni, line 15: if we don't pass `-y`, this becomes interactive - is that something that actually works when we spawn `dnf` from inside maintain? like, can the user press "Y" and it is passed to `dnf`?

lib/foreman_maintain/cli/upgrade_command.rb
```diff
@@ -52,12 +55,18 @@ def print_versions(target_versions)
         target_versions.sort.each { |version| puts version }
       end
 
+      def do_self_upgrade?
```
evgeni, line 58: `do_self_upgrade?` sounds (to me) like "this will tell you if we're in the middle of a self-upgrade", not "check whether self-upgrade is supported in this scenario".

`allow_self_upgrade?`, `self_upgrade_supported?`, idk what's better (feel free to ignore me, too!)

definitions/procedures/packages/enable_modules.rb
upadhyeammit: The answer to your question is no, the foreman-maintain does not have capability to prompt.
But yeah, from scenario I am only calling it with assumeyes `true` explicitly, then I also feel should I not keep assumeyes option because it can't be used to false.

definitions/checks/foreman_installer/installer_noop_run.rb
upadhyeammit: Yeah that would be nice!

definitions/procedures/packages/enable_modules.rb
evgeni: sounds good!

upadhyeammit:
I have moved changes with respect to self-upgrade in another pr: https://github.com/theforeman/foreman_maintain/pull/637

definitions/checks/foreman_installer/installer_noop_run.rb
ekohl: Just the checks implies running `run-parts /usr/share/foreman-installer/checks/*`, but we've implemented the system requirements via a hook. That's tricky since you'll also run all other hooks.

Looking at https://github.com/theforeman/kafo/blob/master/doc/kafo_run.png we can run everything up to (and including) pre_commit. I'm not sure how you'd name it though.

The easiest is some hook in the installer (like `hooks/pre/00-no-real-run.rb` or `hooks/pre_commit/99-no-real-run.rb`) that has a certain condition that makes it quit. Right now I'm thinking of some environment variable. That way you can run with `--noop` AND the env var. If it's supported, it'll be faster and if it isn't, no harm is done since it's still a noop run.

Thoughts?

definitions/scenarios/upgrade_to_katello_4_5.rb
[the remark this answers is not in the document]
ekohl: Can't you extract it from the class name, at least by default?
```ruby
module Scenarios
  module Katello_4_5_z
    class Base
      def self.target_version
        self.name.split('::')[-2].split('_', 2).last
      end

      def target_version
        self.class.target_version
      end
    end

    class Abstract < Base
      def self.upgrade_metadata
        self.target_version
      end
    end
  end
end

puts Scenarios::Katello_4_5_z::Abstract.upgrade_metadata
puts Scenarios::Katello_4_5_z::Abstract.new.target_version
```

Obviously you'll need to follow a pattern in module names and use `target_version` consistently. I'd probably also introduce a similar `minor_version` method.

definitions/scenarios/upgrade_to_katello_4_5_z.rb
```diff
@@ -0,0 +1,93 @@
...
+      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable,
+                                                       :assumeyes => true))
+      add_step(Procedures::Packages::Update.new(:assumeyes => true))
+      add_step_with_context(Procedures::Installer::Upgrade)
+      add_step(Procedures::Installer::UpgradeRakeTask)
```
ekohl, line 64: It was a bit confusing to me that this is here, but looks like there's an `if feature(:satellite)` in the UpgradeRakeTask itself.

upadhyeammit: Removing it, considering the Foreman and Katello scenarios already do that

ehelms:
A couple of thoughts to help get in an initial version of this work:

 1) Unless we plan to backport this work to 3.3 and 3.4 release streams I think this should for now only focus on nightly scenarios
 2) If we want to add 3.4 or 3.3 scenarios, I'd layer those in in a follow up
 3) The pre_upgrade issue for the tuning profile / system requirements is something we need to tackle, but as has been noted this requires doing an additional feature to the installer. That feature would correlate well with the nightly goal. I think the order could be: merge this PR with nightly scenario and no handling for failing installer checks, add installer feature for checks, update the scenario here with that check handling

upadhyeammit:
> A couple of thoughts to help get in an initial version of this work:
> 
>     1. Unless we plan to backport this work to 3.3 and 3.4 release streams I think this should for now only focus on nightly scenarios
> 
>     2. If we want to add 3.4 or 3.3 scenarios, I'd layer those in in a follow up
> 
>     3. The pre_upgrade issue for the tuning profile / system requirements is something we need to tackle, but as has been noted this requires doing an additional feature to the installer. That feature would correlate well with the nightly goal. I think the order could be: merge this PR with nightly scenario and no handling for failing installer checks, add installer feature for checks, update the scenario here with that check handling

I have removed the version based scenarios of Foreman and Katello and kept nightly. The tests were updated as per the change in installer feature.

definitions/checks/foreman_installer/installer_noop_run.rb
ehelms: I think implementing one of these strategies is the right way forward. For this PR, I would prefer:

 1) drop the use of noop here and merge this
 2) let us add a feature to the installer
 3) add a fix later that uses the installer's feature

ekohl: > drop the use of noop here and merge this

With this you mean drop the entire installer run and let it fail when it actually runs? I'd be OK with that, but I can't see all the implications in foreman-maintain.

ehelms (approved):

upadhyeammit:
@ekohl @evgeni like to know if you are good with this for initial start and later on we improve as per requirements?

definitions/procedures/packages/enable_modules.rb
evgeni: This one is still outstanding.

evgeni:
No comments besides the outstanding one around `dnf module enable`

ekohl (changes requested):
I played a bit with refactoring things so you don't repeat the target version all the time. In 3f99cc465c01cdbaa123bd67f919f892b55f4366 I restructured the directory layout a bit (it may break other things). Then in f0520bfca16b24b52e4eb46656087700b2dc16be I introduced a `target_version` method. In ab755a9cce249657cf0685a57763e8ae43a13118 I tried to reduce duplication by introducing a base class for various upgrade scenarios.

While writing this, I realized I could apply the same changes in a more minimal patch to this one: dab29a9f96a6e736564873040e19265d4443f083 does that. Then 10132217f88185519e0f371ab5ca8a4e7a77b837 makes the diff even smaller. Now it's:
```diff
--- definitions/scenarios/upgrade_to_foreman_nightly.rb	2022-09-08 13:05:33.742733736 +0200
+++ definitions/scenarios/upgrade_to_katello_nightly.rb	2022-09-08 13:08:45.405873682 +0200
@@ -1,4 +1,4 @@
-module Scenarios::Foreman_Nightly
+module Scenarios::Katello_Nightly
   class Abstract < ForemanMaintain::Scenario
     def self.target_version
       'nightly'
@@ -10,7 +10,7 @@
       metadata do
         tags :upgrade_scenario
         confine do
-          feature(:foreman_install) || ForemanMaintain.upgrade_in_progress == target_version
+          feature(:katello_install) || ForemanMaintain.upgrade_in_progress == target_version
         end
 
         @target_version = target_version
@@ -19,7 +19,7 @@
         end
 
         def target
-          "Foreman #{target_version}"
+          "Katello #{target_version}"
         end
 
         instance_eval(&block)
@@ -67,7 +67,7 @@
     end
 
     def modules_to_enable
-      el? ? ["foreman:#{el_short_name}"] : []
+      ["katello:#{el_short_name}", "pulpcore:#{el_short_name}"]
     end
 
     def compose
```

Now the next step could be to enhance `modules_to_enable` to look at the features to determine it. Once you have that, we can simplify it to just `upgrade_to_upstream_nightly` and we would simply have an upstream scenario that does the right thing for all installations.

Please include my patches (feel free to squash them in) and unify to a single upstream scenario.

definitions/scenarios/upgrade_to_foreman_nightly.rb
```diff
@@ -0,0 +1,90 @@
+module Scenarios::Foreman_Nightly
```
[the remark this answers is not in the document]
upadhyeammit: It seems not correct but I am searching for more legit source https://gist.github.com/alexpchin/f5d2be2ef3735889d315

definitions/scenarios/upgrade_to_foreman_nightly.rb
```diff
@@ -0,0 +1,90 @@
...
+  class PreUpgradeCheck < Abstract
+    upgrade_metadata do
+      description 'Checks before upgrading to Foreman nightly'
+      tags :pre_upgrade_checks
+      run_strategy :fail_slow
+    end
+
+    def compose
+      add_steps(find_checks(:default))
+      add_steps(find_checks(:pre_upgrade))
```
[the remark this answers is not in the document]
upadhyeammit: This should work if I see the series of methods starting from find_checks, however I would still like to delegate this work to https://projects.theforeman.org/issues/35508

definitions/scenarios/upgrade_to_foreman_nightly.rb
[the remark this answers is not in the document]
ehelms: I think you just go with `ForemanNightlyUpgrade` and that should be sufficient. Generally, the file name should match that convention `foreman_nightly_upgrade.rb`.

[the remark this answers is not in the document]
ehelms: @upadhyeammit For me this is the last little bit to tackle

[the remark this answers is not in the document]
ehelms: I do see that this is following the already established convention within that repository for upgrade files. Given that, I am OK with leaving this as is and doing a future re-factor to align these files to Ruby conventions.

definitions/checks/foreman_installer/installer_noop_run.rb
ehelms: Tracking this need here -- https://projects.theforeman.org/issues/35555

upadhyeammit:
Can we merge this one?

definitions/procedures/packages/enable_modules.rb
ehelms: @upadhyeammit Can you fix this last issue?

definitions/scenarios/upgrade_to_katello_nightly.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    def set_context_mapping
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      add_step(Procedures::Repositories::Setup.new(:version => 'nightly'))
+      modules_to_enable = ["katello:#{el_short_name}", "pulpcore:#{el_short_name}"]
+      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable,
+                                                       :assumeyes => true))
```
ehelms, line 57:
```suggestion
      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
```

I see this locally:

```
raceback (most recent call last):
	21: from ./bin/foreman-maintain:15:in `<main>'
	20: from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/command.rb:140:in `run'
	19: from /root/foreman_maintain/lib/foreman_maintain/cli.rb:46:in `run'
	18: from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/command.rb:66:in `run'
	17: from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/subcommand/execution.rb:18:in `execute'
	16: from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/command.rb:66:in `run'
	15: from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/subcommand/execution.rb:18:in `execute'
	14: from /usr/share/gems/gems/clamp-1.3.2/lib/clamp/command.rb:66:in `run'
	13: from /root/foreman_maintain/lib/foreman_maintain/cli/upgrade_command.rb:73:in `execute'
	12: from /root/foreman_maintain/lib/foreman_maintain/cli/upgrade_command.rb:44:in `upgrade_runner'
	11: from /root/foreman_maintain/lib/foreman_maintain/cli/upgrade_command.rb:27:in `validate_target_version!'
	10: from /root/foreman_maintain/lib/foreman_maintain/upgrade_runner.rb:19:in `available_targets'
	 9: from /root/foreman_maintain/lib/foreman_maintain/concerns/finders.rb:29:in `find_scenarios'
	 8: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:63:in `available_scenarios'
	 7: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:63:in `map'
	 6: from /root/foreman_maintain/lib/foreman_maintain/detector.rb:63:in `new'
	 5: from /root/foreman_maintain/lib/foreman_maintain/scenario.rb:87:in `initialize'
	 4: from /root/foreman_maintain/definitions/scenarios/upgrade_to_katello_nightly.rb:56:in `compose'
	 3: from /root/foreman_maintain/definitions/scenarios/upgrade_to_katello_nightly.rb:56:in `new'
	 2: from /root/foreman_maintain/lib/foreman_maintain/executable.rb:14:in `initialize'
	 1: from /root/foreman_maintain/lib/foreman_maintain/executable.rb:33:in `setup_params'
/root/foreman_maintain/lib/foreman_maintain/core_ext.rb:20:in `validate_options!': Unexpected options ["assumeyes"]. Valid keys are: "module_names" (ArgumentError)
```

ehelms (approved):

PR #634, opened 2022-07-22
Fixes #35261 - Run procedure FixPulpcoreArtifactOwnership during 6.11 upgrade

sjha4:
Correct permissions of artifacts directory with FixPulpcoreArtifactOwnership procedure automatically during 6.11 upgrade.

### DIFF
--- a/definitions/scenarios/upgrade_to_capsule_6_11.rb
+++ b/definitions/scenarios/upgrade_to_capsule_6_11.rb
@@ -40,6 +40,7 @@ module Scenarios::Capsule_6_11
     def compose
       add_steps(find_procedures(:pre_migrations))
       add_step(Procedures::Pulp::Remove.new(:assumeyes => true))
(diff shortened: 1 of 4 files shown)


upadhyeammit:
Can this be done on Capsule as well?

sjha4:
@upadhyeammit : Updated to also run on capsule upgrade. 👍🏼

upadhyeammit:
@sjha4 few more questions before we can merge,
1. This will only run when 6.10 to 6.11 upgrade will happen and not when 6.11.z upgrades. So in that case should we also do this for 6.11.z scenarios of Satellite and Capsule?
2. However I feel if the issue is there on 6.11 mostly users will correct it first before upgrade? because basic repo sync is failing?

sjha4:
Added the procedure to 6.11.z scenarios for cases where the 6.11 upgrade missed the step. 👍🏼

ianballou (approved):
Looks fine to me!

PR #635, opened 2022-08-19
Fixes #35252 - declare version map in separate file

upadhyeammit:


lib/foreman_maintain/foreman_and_katello_version_map.rb
```diff
@@ -0,0 +1,7 @@
+module ForemanMaintain
+  FOREMAN_AND_KATELLO_VERSION_MAP = {
```
ehelms, line 2: I think it would be fine to declare this inside of https://github.com/theforeman/foreman_maintain/pull/635/files#diff-b341f799f1a5ff007ab136ed4c9dad33fe4d87a564afd2994fc799abab02e1efR5

That is the only file using it and it prevents someone from having to hunt down and figure out which file is which given they have similar names.

ehelms (approved):

PR #637, opened 2022-08-22
Fixes #35401 - self upgrade for Foreman and Katello

upadhyeammit:
Depends on #633 and #635

### DIFF
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -85,14 +85,21 @@ module ForemanMaintain::Scenarios
       add_step(Procedures::Repositories::Setup.new(:version => upstream_target_version))
       add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true))
     ensure
+      rollback_repositories
+    end
+
+    def rollback_repositories
       installed_release_pkg = package_manager.find_installed_package('foreman-release',
                                                                      '%{VERSION}')
-      current_major_version = current_version[0..2]
-      installed_foreman_release_major_version = installed_release_pkg[0..2]
-      if installed_foreman_release_major_version != current_major_version
-        add_step(Procedures::Packages::Uninstall.new(packages: %w[foreman-release katello-repos],
-                                                     assumeyes: true))
-        add_step(Procedures::Repositories::Setup.new(:version => current_major_version))
+
+      unless current_version.nil? && installed_release_pkg.nil?
+        current_major_version = current_version[0..2]
+        installed_foreman_release_major_version = installed_release_pkg[0..2]
+        if installed_foreman_release_major_version != current_major_version
+          add_step(Procedures::Packages::Uninstall.new(packages: %w[foreman-release katello-repos],
+                                                       assumeyes: true))
+          add_step(Procedures::Repositories::Setup.new(:version => current_major_version))
+        end
       end
     end
 
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -48,6 +48,14 @@ module ForemanMaintain::Scenarios
         [maintenance_repo_id(target_version)]
       end
     end
+
+    def upstream_target_version
+      if feature(:katello_install)
+        return foreman_version_by_katello(target_version)
+      else
+        target_version
+      end
+    end
   end
 
   class SelfUpgrade < SelfUpgradeBase
@@ -58,14 +66,37 @@ module ForemanMaintain::Scenarios
       manual_detection
     end
 
-    def compose
+    def downstream_self_upgrade(pkgs_to_update)
       if check_min_version('foreman', '2.5') || check_min_version('foreman-proxy', '2.5')
-        pkgs_to_update = %w[satellite-maintain rubygem-foreman_maintain]
         yum_options = req_repos_to_update_pkgs.map do |id|
           "--enablerepo=#{id}"
         end
         add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true,
                                                   yum_options: yum_options))
+      end
+    end
+
+    def upstream_self_upgrade(pkgs_to_update)
+      add_step(Procedures::Repositories::Setup.new(:version => upstream_target_version))
+      add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true))
+      installed_release_pkg = package_manager.find_installed_package('foreman-release',
+                                                                     '%{VERSION}')
+      current_major_version = current_version[0..2]
+      installed_foreman_release_major_version = installed_release_pkg[0..2]
+      if installed_foreman_release_major_version != current_major_version
+        add_step(Procedures::Packages::Uninstall.new(packages: %w[foreman-release katello-repos],
+                                                     assumeyes: true))
+        add_step(Procedures::Repositories::Setup.new(:version => current_major_version))
+      end
+    end
+
+    def compose
+      pkgs_to_update = %w[rubygem-foreman_maintain]
+      if feature(:instance).downstream
+        pkgs_to_update << 'satellite-maintain'
+        downstream_self_upgrade(pkgs_to_update)
+      elsif feature(:instance).upstream_install
+        upstream_self_upgrade(pkgs_to_update)
       end
     end
   end
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -77,8 +77,14 @@ module ForemanMaintain::Scenarios
     end
 
     def upstream_self_upgrade(pkgs_to_update)
+      # This method is responsible for
+      # 1. Setup the repositories of next major version
+      # 2. Update the foreman-maintain packages from next major version repository
+      # 3. Rollback the repository to current major version
+
       add_step(Procedures::Repositories::Setup.new(:version => upstream_target_version))
       add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true))
+    ensure
       installed_release_pkg = package_manager.find_installed_package('foreman-release',
                                                                      '%{VERSION}')
       current_major_version = current_version[0..2]
--- a/lib/foreman_maintain/cli/upgrade_command.rb
+++ b/lib/foreman_maintain/cli/upgrade_command.rb
@@ -53,10 +53,14 @@ module ForemanMaintain
       end
 
       def allow_self_upgrade?
-        !disable_self_upgrade? ||
-          ((respond_to?(:target_version) && !target_version.nil?) &&
-          (!target_version.end_with?('z') ||
-           target_version == 'nightly'))
+        if disable_self_upgrade?
+          return !disable_self_upgrade?
+        elsif respond_to?(:target_version) && !target_version.nil?
+          target_version.end_with?('z') ||
+            target_version == 'nightly'
+        else
+          true
+        end
       end
 
       subcommand 'list-versions', 'List versions this system is upgradable to' do
--- a/lib/foreman_maintain/cli/upgrade_command.rb
+++ b/lib/foreman_maintain/cli/upgrade_command.rb
@@ -53,14 +53,7 @@ module ForemanMaintain
       end
 
       def allow_self_upgrade?
-        if disable_self_upgrade?
-          false
-        elsif respond_to?(:target_version) && !target_version.nil?
-          target_version.end_with?('z') ||
-            target_version == 'nightly'
-        else
-          true
-        end
+        !disable_self_upgrade?
       end
 
       subcommand 'list-versions', 'List versions this system is upgradable to' do
--- a/lib/foreman_maintain/cli/upgrade_command.rb
+++ b/lib/foreman_maintain/cli/upgrade_command.rb
@@ -52,12 +52,19 @@ module ForemanMaintain
         target_versions.sort.each { |version| puts version }
       end
 
+      def allow_self_upgrade?
+        !disable_self_upgrade? ||
+          ((respond_to?(:target_version) && !target_version.nil?) &&
+          (!target_version.end_with?('z') ||
+           target_version == 'nightly'))
+      end
+
       subcommand 'list-versions', 'List versions this system is upgradable to' do
         disable_self_upgrade_option
 
         def execute
           ForemanMaintain.validate_downstream_packages
-          ForemanMaintain.perform_self_upgrade unless disable_self_upgrade?
+          ForemanMaintain.perform_self_upgrade if allow_self_upgrade?
           print_versions(UpgradeRunner.available_targets)
         end
       end
@@ -69,7 +76,7 @@ module ForemanMaintain
 
         def execute
           ForemanMaintain.validate_downstream_packages
-          ForemanMaintain.perform_self_upgrade unless disable_self_upgrade?
+          ForemanMaintain.perform_self_upgrade if allow_self_upgrade?
           upgrade_runner.run_phase(:pre_upgrade_checks)
           exit upgrade_runner.exit_code
         end
@@ -89,7 +96,7 @@ module ForemanMaintain
 
         def execute
           ForemanMaintain.validate_downstream_packages
-          ForemanMaintain.perform_self_upgrade unless disable_self_upgrade?
+          ForemanMaintain.perform_self_upgrade if allow_self_upgrade?
           if phase
             upgrade_runner.run_phase(phase.to_sym)
           else
--- a/lib/foreman_maintain/cli/upgrade_command.rb
+++ b/lib/foreman_maintain/cli/upgrade_command.rb
@@ -54,7 +54,7 @@ module ForemanMaintain
 
       def allow_self_upgrade?
         if disable_self_upgrade?
-          return !disable_self_upgrade?
+          false
         elsif respond_to?(:target_version) && !target_version.nil?
           target_version.end_with?('z') ||
             target_version == 'nightly'
--- a/lib/foreman_maintain/package_manager/apt.rb
+++ b/lib/foreman_maintain/package_manager/apt.rb
@@ -1,3 +1,5 @@
+require 'open3'
+
 module ForemanMaintain::PackageManager
   class Apt < Base
     def installed?(packages)
@@ -39,6 +41,11 @@ module ForemanMaintain::PackageManager
 
     def check_update(packages: nil, with_status: false)
       apt_action('upgrade --dry-run', packages, :with_status => with_status)
+    end
+
+    def update_available?(package)
+      output, status = Open3.capture2("apt-get install #{package} --dry-run")
+      status.success? && output.include?("Inst #{package}")
     end
 
     def list_installed_packages(queryfm = '${binary:Package}-${VERSION}\n')
--- a/test/lib/cli/upgrade_command_test.rb
+++ b/test/lib/cli/upgrade_command_test.rb
@@ -1,5 +1,4 @@
 require 'test_helper'
-
 require 'foreman_maintain/cli'
 
 module ForemanMaintain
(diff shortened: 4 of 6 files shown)


ehelms:
Is this intended to go in after https://github.com/theforeman/foreman_maintain/pull/633 ?

upadhyeammit:
> Is this intended to go in after #633 ?

yeah as I am using some of the stuff from #633

upadhyeammit:
Okay, tests are green on my system Fedora, maybe something with respect to Ubuntu is causing the issue. Will check that, however this is ready for review!

lib/foreman_maintain/package_manager/apt.rb
```diff
@@ -41,6 +41,10 @@ def check_update(packages: nil, with_status: false)
       apt_action('upgrade --dry-run', packages, :with_status => with_status)
     end
 
+    def update_available?(package)
+      sys.execute!('apt-get upgrade --dry-run &> /dev/null').include?(package)
```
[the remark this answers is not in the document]
evgeni: `&> /dev/null` drops all output, so, uh, this will never be true? ;-)

That said, no, there is no *good* alternative for that.

You could check the output of `apt-cache policy <package>` and check if `Installed` and `Candidate` are the same.
Or you could check the output of `apt-get install <package> --dry-run` to contain a line like `Inst <package> …`

The problem with your approach is, if there are two packages `fun` and `well-this-isnt-funny`, your `update_available?('fun')` would also trigger if the available update is actually for `well-this-isnt-funny`.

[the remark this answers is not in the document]
upadhyeammit: > `&> /dev/null` drops all output, so, uh, this will never be true? ;-)
> 
> That said, no, there is no _good_ alternative for that.
> 
> You could check the output of `apt-cache policy <package>` and check if `Installed` and `Candidate` are the same. Or you could check the output of `apt-get install <package> --dry-run` to contain a line like `Inst <package> …`
> 
> The problem with your approach is, if there are two packages `fun` and `well-this-isnt-funny`, your `update_available?('fun')` would also trigger if the available update is actually for `well-this-isnt-funny`.

Yeah I also thought there command won't return anything, but that was not the case,
~~~
2] pry(#<ForemanMaintain::PackageManager::Apt>)> sys.execute!("apt-get upgrade --dry-run &> /dev/null")
=> "Reading package lists...\nBuilding dependency tree...\nReading state information...\nCalculating upgrade...\nThe following packages have been kept back:\n  linux-cloud-tools-generic linux-cloud-tools-virtual linux-generic linux-headers-generic linux-image-generic\nThe following packages will be upgraded:\n  base-files language-pack-en language-
~~~

I am checking on the `apt-cache` and `dry-run` and will update the code accordingly.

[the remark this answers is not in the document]
evgeni: I'd still omit the `&>/dev/null` then, it's confusing. (I recall it can be a bashism, and Debian doesn't use Bash by default)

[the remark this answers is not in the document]
upadhyeammit: > I'd still omit the `&>/dev/null` then, it's confusing. (I recall it can be a bashism, and Debian doesn't use Bash by default)

Agree, i just wanted to hide foreman-maintain printing whole command output on terminal and I can't see anything like `-q` flag which we have with yum. So lets just print the output on terminal while checking if updates for package is available? or maybe I should be using shell commands to trim the output and check it instead of ruby which will help a bit.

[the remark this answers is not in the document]
evgeni: Why not using
```ruby
output, status = Open3.capture2('apt-get …')
if status.success? && output.includes?('search')
  …
end
```

lib/foreman_maintain/package_manager/apt.rb
```diff
@@ -41,6 +43,13 @@ def check_update(packages: nil, with_status: false)
       apt_action('upgrade --dry-run', packages, :with_status => with_status)
     end
 
+    def update_available?(package)
+      output, status = Open3.capture2("apt-get install #{package} --dry-run")
+      return true if status.success? && output.include?("Inst #{package}")
+
+      false
```
ekohl, line 50:
```suggestion
      status.success? && output.include?("Inst #{package}")
```

definitions/scenarios/self_upgrade.rb
```diff
@@ -58,15 +66,37 @@ class SelfUpgrade < SelfUpgradeBase
...
     end
 
-    def compose
+    def downstream_self_upgrade(_pkgs)
       if check_min_version('foreman', '2.5') || check_min_version('foreman-proxy', '2.5')
-        pkgs_to_update = %w[satellite-maintain rubygem-foreman_maintain]
         yum_options = req_repos_to_update_pkgs.map do |id|
           "--enablerepo=#{id}"
         end
         add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true,
```
evgeni, line 74: shouldn't this use `_pkgs` now? or is it using `pkgs_to_update` from `compose`? seems odd.

definitions/scenarios/self_upgrade.rb
```diff
@@ -58,15 +66,37 @@ class SelfUpgrade < SelfUpgradeBase
...
                                                   yum_options: yum_options))
       end
     end
+
+    def upstream_self_upgrade(_pkgs)
+      add_step(Procedures::Repositories::Setup.new(:version => upstream_target_version))
+      add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true))
```
evgeni, line 81: same, `pkgs_to_update` seem undefined in this scope?

upadhyeammit: seems like i literally pushed the change when I was half awake!

definitions/scenarios/self_upgrade.rb
```diff
@@ -58,15 +66,38 @@ class SelfUpgrade < SelfUpgradeBase
...
+      installed_release_pkg = package_manager.find_installed_package('foreman-release',
+                                                                     '%{VERSION}')
+      current_major_version = current_version[0..2]
+      installed_foreman_release_major_version = installed_release_pkg[0..2]
+      if installed_foreman_release_major_version != current_major_version
+        add_step(Procedures::Packages::Uninstall.new(packages: %w[foreman-release katello-repos],
+                                                     assumeyes: true))
+        add_step(Procedures::Repositories::Setup.new(:version => current_major_version))
+      end
```
evgeni, line 90: I had to look multiple times at this to understand what it's doing.
1. setup `target_version` repositories
2. update `foreman_maintain`
3. roll back the repositories to `current_version`

Given there is no good way to use a temporary repo definition (there is with DNF, but it would require us to get the `.repo` file from somewhere, which is also hard), I think this is as nice as it gets.

No idea if this deserves a comment or anything, but thought I mention that I stumbled on reading it.

lib/foreman_maintain/cli/upgrade_command.rb
```diff
@@ -52,12 +52,19 @@ def print_versions(target_versions)
         target_versions.sort.each { |version| puts version }
       end
 
+      def allow_self_upgrade?
+        !disable_self_upgrade? ||
+          ((respond_to?(:target_version) && !target_version.nil?) &&
+          (!target_version.end_with?('z') ||
+           target_version == 'nightly'))
+      end
```
evgeni, line 60: I think this logic is wrong. The old code would read `perform_self_upgrade unless disable_self_upgrade?` which means that unless the user passed `--disable-self-upgrade` the upgrade would happen.

Now, if they pass `--disable-self-upgrade --target-version nightly` the upgrade will happen, as the logic is now `if not disable_self_upgrade? OR target_version` (shortened)

definitions/scenarios/self_upgrade.rb
upadhyeammit: Added the comments and added `ensure` as well to rollback repositories.

lib/foreman_maintain/cli/upgrade_command.rb
upadhyeammit: Corrected this, now the method `allow_self_upgrade?` respects the `disable_self_upgrade?` and behaves as the name says. And yes we always do the self-upgrade for the `list-versions` subcommand.

upadhyeammit: and I am checking the tests now!

lib/foreman_maintain/cli/upgrade_command.rb
```diff
@@ -52,12 +52,23 @@ def print_versions(target_versions)
         target_versions.sort.each { |version| puts version }
       end
 
+      def allow_self_upgrade?
+        if disable_self_upgrade?
+          return !disable_self_upgrade?
```
evgeni, line 57:
```suggestion
          return false
```

Easier to read ;-) (also, I think the `return` is *technically* not necessary, and you don't `return` in the other branches)

definitions/scenarios/self_upgrade.rb
```diff
@@ -58,15 +66,44 @@ class SelfUpgrade < SelfUpgradeBase
...
+      add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true))
+    ensure
+      installed_release_pkg = package_manager.find_installed_package('foreman-release',
+                                                                     '%{VERSION}')
+      current_major_version = current_version[0..2]
+      installed_foreman_release_major_version = installed_release_pkg[0..2]
+      if installed_foreman_release_major_version != current_major_version
+        add_step(Procedures::Packages::Uninstall.new(packages: %w[foreman-release katello-repos],
+                                                     assumeyes: true))
+        add_step(Procedures::Repositories::Setup.new(:version => current_major_version))
```
ekohl, line 95: I wonder if `Setup` should be able to deal with `foreman-release` and `katello-repos` being present, possibly with a newer version. That would remove the need to uninstall and I mostly say that because if `Uninstall` successed but `Setup` fails you end up with no repositories present. If it's a single call to the package manager then it becomes atomic.

definitions/scenarios/self_upgrade.rb
```diff
@@ -58,15 +66,44 @@ class SelfUpgrade < SelfUpgradeBase
...
+
+    def upstream_self_upgrade(pkgs_to_update)
+      # This method is responsible for
+      # 1. Setup the repositories of next major version
+      # 2. Update the foreman-maintain packages from next major version repository
+      # 3. Rollback the repository to current major version
+
+      add_step(Procedures::Repositories::Setup.new(:version => upstream_target_version))
+      add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true))
+    ensure
```
ekohl, line 87: I'm debating if this should be `rescue` instead of `ensure`. Now it happens also after success, but that feels redundant to me.

definitions/scenarios/self_upgrade.rb
```diff
@@ -58,15 +66,44 @@ class SelfUpgrade < SelfUpgradeBase
...
+      installed_release_pkg = package_manager.find_installed_package('foreman-release',
+                                                                     '%{VERSION}')
+      current_major_version = current_version[0..2]
+      installed_foreman_release_major_version = installed_release_pkg[0..2]
```
ekohl, line 91: If `foreman-release` is not installed for some reason, won't this be `nil[0..2]` which raises an exception?

definitions/scenarios/self_upgrade.rb
upadhyeammit: We need to revert the repositories to installed major version in successful run because doing the self-upgrade does not mean users will do the Foreman or Katello major upgrade just after the self-upgrade?

lib/foreman_maintain/cli/upgrade_command.rb
```diff
@@ -52,12 +52,23 @@ def print_versions(target_versions)
...
+        elsif respond_to?(:target_version) && !target_version.nil?
+          target_version.end_with?('z') ||
+            target_version == 'nightly'
```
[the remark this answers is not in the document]
upadhyeammit: Hardcoding the `nightly` will help to upgrade foreman-maintain itself to latest version when doing `nightly`=>`nightly` upgrade?
As per the scenarios `upgrade_to_foreman_nightly` and `upgrade_to_katello_nightly` the nightly upgrades are possible from any version https://github.com/theforeman/foreman_maintain/pull/633 . Is it not the case today that using installer we can upgrade to nightly from any version?

definitions/scenarios/self_upgrade.rb
upadhyeammit: Hmm.. I can install the next version or updated release packages without uninstalling the existing one and let the `dnf/yum/apt` do the job. So removing the uninstall step should be a go

definitions/scenarios/self_upgrade.rb
upadhyeammit: yepp :| Guarding it now

definitions/scenarios/self_upgrade.rb
upadhyeammit: Yeah so for downgrade I think Uninstall and Setup is correct way. Yeah Setup step takes care of installing the specified version(new/any).

test/lib/cli/upgrade_command_test.rb
```diff
@@ -46,10 +84,34 @@ module ForemanMaintain
 
     describe 'check' do
       let :command do
-        %w[upgrade check --disable-self-upgrade]
+        %w[upgrade check]
+      end
+
+      it 'run self upgrade if upgrade available for foreman-maintain' do
```
ehelms, line 90: This is an example of one of the failing tests that is running the `:pre_upgrade_checks` and then failing cause it tries to start a fake service. So I think you either need to do some additional mocking here or include a line like you have below:

```
UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
```

lib/foreman_maintain/cli/upgrade_command.rb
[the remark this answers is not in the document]
ehelms: This check is breaking a number of tests because the tests use versions like `1.15`. However, I think this whole `elseif` could be dropped because the version of Foreman that self upgrades intend to support are the only versions that this ode change will be released for. In that sense, shouldn't we be able to rely strictly on `disable_self_upgrade?`

ehelms:
The following got tests passing for me, with a few assumptions removed:

```
diff --git a/lib/foreman_maintain/cli/upgrade_command.rb b/lib/foreman_maintain/cli/upgrade_command.rb
index ca641e1..273078f 100644
--- a/lib/foreman_maintain/cli/upgrade_command.rb
+++ b/lib/foreman_maintain/cli/upgrade_command.rb
@@ -53,14 +53,7 @@ module ForemanMaintain
       end
 
       def allow_self_upgrade?
-        if disable_self_upgrade?
-          false
-        elsif respond_to?(:target_version) && !target_version.nil?
-          target_version.end_with?('z') ||
-            target_version == 'nightly'
-        else
-          true
-        end
+        !disable_self_upgrade?
       end
 
       subcommand 'list-versions', 'List versions this system is upgradable to' do
diff --git a/test/lib/cli/upgrade_command_test.rb b/test/lib/cli/upgrade_command_test.rb
index 71b8a58..cbef56a 100644
--- a/test/lib/cli/upgrade_command_test.rb
+++ b/test/lib/cli/upgrade_command_test.rb
@@ -55,7 +55,6 @@ module ForemanMaintain
       end
       it 'run self upgrade if upgrade available for foreman-maintain' do
         foreman_maintain_update_available
-        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
         assert_cmd <<-OUTPUT.strip_heredoc
         Checking for new version of rubygem-foreman_maintain...
 
@@ -68,7 +67,6 @@ module ForemanMaintain
 
       it 'inform if no updates available for foreman-maintain' do
         foreman_maintain_update_unavailable
-        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
         assert_cmd <<-OUTPUT.strip_heredoc
         Checking for new version of rubygem-foreman_maintain...
         Nothing to update, can't find new version of rubygem-foreman_maintain.
@@ -92,7 +90,6 @@ module ForemanMaintain
       it 'run self upgrade if upgrade available for foreman-maintain' do
         foreman_maintain_update_available
         command << '--target-version=1.15'
-        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
         assert_cmd <<-OUTPUT.strip_heredoc
         Checking for new version of rubygem-foreman_maintain...
 
@@ -120,7 +117,7 @@ module ForemanMaintain
       end
 
       it 'should raise UsageError and exit with code 1' do
-        Cli::MainCommand.any_instance.expects(:exit!)
+        Cli::MainCommand.any_instance.stubs(:exit!)
 
         run_cmd([])
       end
@@ -134,7 +131,7 @@ module ForemanMaintain
       it 'run self upgrade if upgrade available for foreman-maintain' do
         foreman_maintain_update_available
         command << '--target-version=1.15'
-        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
+
         assert_cmd <<-OUTPUT.strip_heredoc
         Checking for new version of rubygem-foreman_maintain...
 
@@ -188,7 +185,7 @@ module ForemanMaintain
 
       it 'remembers the current target version and informs no update available' do
         foreman_maintain_update_unavailable
-        Cli::MainCommand.any_instance.expects(:exit!)
+        Cli::MainCommand.any_instance.expects(:exit!).twice
         assert_cmd <<-OUTPUT.strip_heredoc
         Checking for new version of rubygem-foreman_maintain...
         Nothing to update, can't find new version of rubygem-foreman_maintain.
@@ -199,7 +196,6 @@ module ForemanMaintain
 
         UpgradeRunner.current_target_version = '1.15'
         UpgradeRunner.any_instance.expects(:run)
-        Cli::MainCommand.any_instance.expects(:exit!)
 
         run_cmd
```

upadhyeammit:
> The following got tests passing for me, with a few assumptions removed:
> 
> ```
> diff --git a/lib/foreman_maintain/cli/upgrade_command.rb b/lib/foreman_maintain/cli/upgrade_command.rb
> index ca641e1..273078f 100644
> --- a/lib/foreman_maintain/cli/upgrade_command.rb
> +++ b/lib/foreman_maintain/cli/upgrade_command.rb
> @@ -53,14 +53,7 @@ module ForemanMaintain
>        end
>  
>        def allow_self_upgrade?
> -        if disable_self_upgrade?
> -          false
> -        elsif respond_to?(:target_version) && !target_version.nil?
> -          target_version.end_with?('z') ||
> -            target_version == 'nightly'
> -        else
> -          true
> -        end
> +        !disable_self_upgrade?
>        end
>  
>        subcommand 'list-versions', 'List versions this system is upgradable to' do
> diff --git a/test/lib/cli/upgrade_command_test.rb b/test/lib/cli/upgrade_command_test.rb
> index 71b8a58..cbef56a 100644
> --- a/test/lib/cli/upgrade_command_test.rb
> +++ b/test/lib/cli/upgrade_command_test.rb
> @@ -55,7 +55,6 @@ module ForemanMaintain
>        end
>        it 'run self upgrade if upgrade available for foreman-maintain' do
>          foreman_maintain_update_available
> -        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
>          assert_cmd <<-OUTPUT.strip_heredoc
>          Checking for new version of rubygem-foreman_maintain...
>  
> @@ -68,7 +67,6 @@ module ForemanMaintain
>  
>        it 'inform if no updates available for foreman-maintain' do
>          foreman_maintain_update_unavailable
> -        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
>          assert_cmd <<-OUTPUT.strip_heredoc
>          Checking for new version of rubygem-foreman_maintain...
>          Nothing to update, can't find new version of rubygem-foreman_maintain.
> @@ -92,7 +90,6 @@ module ForemanMaintain
>        it 'run self upgrade if upgrade available for foreman-maintain' do
>          foreman_maintain_update_available
>          command << '--target-version=1.15'
> -        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
>          assert_cmd <<-OUTPUT.strip_heredoc
>          Checking for new version of rubygem-foreman_maintain...
>  
> @@ -120,7 +117,7 @@ module ForemanMaintain
>        end
>  
>        it 'should raise UsageError and exit with code 1' do
> -        Cli::MainCommand.any_instance.expects(:exit!)
> +        Cli::MainCommand.any_instance.stubs(:exit!)
>  
>          run_cmd([])
>        end
> @@ -134,7 +131,7 @@ module ForemanMaintain
>        it 'run self upgrade if upgrade available for foreman-maintain' do
>          foreman_maintain_update_available
>          command << '--target-version=1.15'
> -        UpgradeRunner.any_instance.expects(:run_phase).with(:pre_upgrade_checks)
> +
>          assert_cmd <<-OUTPUT.strip_heredoc
>          Checking for new version of rubygem-foreman_maintain...
>  
> @@ -188,7 +185,7 @@ module ForemanMaintain
>  
>        it 'remembers the current target version and informs no update available' do
>          foreman_maintain_update_unavailable
> -        Cli::MainCommand.any_instance.expects(:exit!)
> +        Cli::MainCommand.any_instance.expects(:exit!).twice
>          assert_cmd <<-OUTPUT.strip_heredoc
>          Checking for new version of rubygem-foreman_maintain...
>          Nothing to update, can't find new version of rubygem-foreman_maintain.
> @@ -199,7 +196,6 @@ module ForemanMaintain
>  
>          UpgradeRunner.current_target_version = '1.15'
>          UpgradeRunner.any_instance.expects(:run)
> -        Cli::MainCommand.any_instance.expects(:exit!)
>  
>          run_cmd
> ```

This does mean we only depend on the `disable_self_upgrade?` ?

ehelms (approved):

PR #639, opened 2022-09-05
Fixes #35487 - add missing psql and nftables dependencies

upadhyeammit:


extras/foreman_protector/foreman-protector.whitelist
```diff
@@ -17,8 +17,28 @@ boost-random
 boost-iostreams
 boost-thread
 yum-utils
+# el7 psql client dependencies
+rh-postgresql12-postgresql
+rh-postgresql12-postgresql-libs
+rh-postgresql12-runtime
```
ehelms, line 23: These are only needed if cherry picking this change back to an EL7 based release -- is that the intent?

upadhyeammit: As per the bz the change should be available upto 6.10, so I added it.

ehelms (approved):

PR #641, opened 2022-09-06
Run the tests on CentOS Stream 8

upadhyeammit:


ehelms:
What was your thinking behind the change in OS? Does the underlying test OS influence the tests themselves?

upadhyeammit:
> What was your thinking behind the change in OS? Does the underlying test OS influence the tests themselves?

Yeah, there were few instances where failures were specifically on Debian and not on Fedora. I don't remember it explicitly, or maybe need to dig in prs but this was the intention to catch all.

evgeni:
IMHO the tests should never depend on the OS they are being executed on.

PR #643, opened 2022-09-16
Fixes #35533 - remove duplicate ssl params of cpdb cmd

upadhyeammit:


upadhyeammit:
The url already has the ssl params from candlepin configuration file, so its not needed to again pass those via cpdb cmd.

ehelms:
Looks good to me -- do you think as part of this it'd be feasible to add a `candlepin.conf` fixture and test the parsing of it to reduce having to guess that this does the right thing?

ehelms (approved):
Thanks for adding the test and fixture!

PR #644, opened 2022-09-19
Fixes #35544 - Remove ansible-runner from non-RH package exception list

ehelms:
The ansible-runner package is now shipped by both Foreman repositories and continues to be built and shipped by Red Hat repositories.

For additional references, see: https://github.com/theforeman/puppet-foreman_proxy/pull/780


definitions/checks/non_rh_packages.rb
```diff
@@ -27,8 +26,4 @@ def rh_regexp_list
      /-qpid-router-server/, /java-client/, /pulp-client/, /katello-default-ca/, /katello-server-ca/,
      /katello-ca-consumer/, /gpg-pubkey/, /-tomcat/]
```
ekohl, line 27: A lot of these are the certificate packages. I'd love to get rid of those as well. Is there anything tracking the complete removal of those? I'm mostly wondering when we can drop it due to the whole n-1 support.

ekohl (approved):
Code wise :+1: but a Redmine issue is needed.

PR #651, opened 2022-10-06
Fixes #35612: Add check that system meets tuning profile requirements

ehelms:


lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -105,6 +105,15 @@ def centos?
       def el_short_name
         "el#{el_major_version}"
       end
+
+      def memory
+        meminfo = execute('cat /proc/meminfo')
```
evgeni, line 110: there has to be a better way to read files in Ruby ;)

definitions/checks/foreman/check_tuning_requirements.rb
```diff
@@ -0,0 +1,78 @@
...
+      def check_cpu_cores(tuning_profile)
+        tuning_cpu_cores = tuning_sizes[tuning_profile][:cpu_cores]
+        system_cpu_cores = cpu_cores
+
+        unless system_cpu_cores.to_i >= tuning_cpu_cores
+          "The number of CPU cores for the system is #{system_cpu_cores} but the currently configured tuning profile requires #{tuning_cpu_cores}." # rubocop:disable Metrics/LineLength
+        end
+      end
+
+      def tuning_sizes
```
evgeni, line 60: I wonder if we somehow could read this from `hooks/pre_commit/13-tuning.rb`, but for sure is not a blocker here

evgeni:
besides the `cat` part, I like!

definitions/checks/foreman/check_tuning_requirements.rb
ehelms: I thought about that and the current structure of the file and such seemed not a good interface in the present design.

evgeni: Yeah, I didn't see a good way either, but hoped you'd do!

evgeni (approved):

PR #653, opened 2022-10-12
Include socket services while stopping services during restart

Gauravtalreja1:
**Fixes** https://bugzilla.redhat.com/show_bug.cgi?id=2067120

**Results**:
```
# foreman-maintain service restart --only foreman
Running Restart Services
Restart applicable services:

foreman, foreman.socket
\ All services stopped

foreman
/ All services started                                                [OK]
```

### DIFF
--- a/definitions/procedures/service/restart.rb
+++ b/definitions/procedures/service/restart.rb
@@ -13,7 +13,7 @@ module Procedures::Service
     PING_RETRY_INTERVAL = 30
 
     def run
-      run_service_action('restart', common_options)
+      run_service_action('restart', common_options.merge(:include_sockets => true))
       server_ping_retry if @wait_for_server_response
     end
 


ehelms:
@evgeni @ekohl do y'all remember where we landed on this?

definitions/procedures/service/restart.rb
```diff
@@ -13,7 +13,7 @@ class Restart < Base
     PING_RETRY_INTERVAL = 30
 
     def run
-      run_service_action('stop', common_options)
+      run_service_action('stop', common_options.merge(:include_sockets => true))
       run_service_action('start', common_options)
```
ekohl, line 17: I really wonder why we don't call the restart action. I mean, `systemctl` is perfectly capable of calling stop and start itself if there's no restart defined. That also gets rid of the warning.

Gauravtalreja1: @ehelms @ekohl Could you please take a look at this?

evgeni:
Two things:
* In the meantime https://bugzilla.redhat.com/show_bug.cgi?id=2067120 was closed as a dupe of https://bugzilla.redhat.com/show_bug.cgi?id=2154917, which I'd say is wrong (fixing the later fixes the former, but it's not a dupe per se)
* Doesn't this change mean that `foreman-maintain service restart` will *stop* sockets, but not start them again?

/packit build

Gauravtalreja1:
@ehelms Thanks for addressing the comment above in https://github.com/theforeman/foreman_maintain/pull/733, I've updated this PR to use `restart` and include sockets for restart action to avoid this socket warnings, please can you re-visit this PR, if you think this PR is still valid, else we can close this

ehelms:
From my quick test, when we switched to using systemd restart, this was inherently fixed, here is output without this fix:

```
# foreman-maintain service restart --only foreman
Running Restart Services
Restart applicable services: 

Restarting the following service(s):
foreman
/ All services restarted                                              [OK]      
```

Gauravtalreja1:
@ehelms Yes, I've also tried, this warning disappears when using restart, but sockets aren't included in it, so do you think if it makes sense to include sockets while restart? if not, then we're good to close this PR I think, Thanks!

ehelms:
I do not think we need to include sockets as systemd handles this for us.

PR #654, opened 2022-10-13
Refs #35612 - Tuning requirements check when successful should return…

ehelms:
… nothing

I goobered this and the test originally confirmed my goobering. This actually results in bad output and incorrect result.

```
Check if system requirements match current tuning profile:            [FAIL]
true
```


evgeni (approved):
Glad we talked about this ;)

PR #656, opened 2022-10-26
Fixes #35678 - Trim RPM Changelogs when upgrading to pulp_rpm 3.17.5+

evgeni:


definitions/scenarios/upgrade_to_satellite_6_12.rb
```diff
@@ -73,6 +73,7 @@ class PostMigrations < Abstract
     def compose
       add_step(Procedures::RefreshFeatures)
       add_step(Procedures::Service::Start.new)
+      add_step(Procedures::Pulpcore::TrimRpmChangelogs.new)
```
ehelms, line 76: Will this also do it for z-stream of 6.12? If not, I think we should.

evgeni: No it will not, can add!

evgeni: Done

ehelms (approved):

PR #657, opened 2022-10-31
Bump version to 1.2.0

ehelms:


ehelms:
@evgeni Mind taking a quick look? I'd like to get a version into nightly that has the upgrade features

PR #661, opened 2022-11-17
Fixes #35765 - don't pass --disable-system-checks on 3.4+

ehelms:
(cherry picked from commit 2cc876301c72530c15e546c5ebf556258e33c2ea)

### DIFF
--- a/definitions/procedures/restore/installer_reset.rb
+++ b/definitions/procedures/restore/installer_reset.rb
@@ -21,7 +21,8 @@ module Procedures::Restore
       # already ran since this is to be run on an existing system, which means installer checks
       # has already been skipped
       if feature(:foreman_proxy) &&
-         feature(:foreman_proxy).with_content?
+         feature(:foreman_proxy).with_content? &&
+         check_max_version('foreman-installer', '3.4')
         installer << '--disable-system-checks '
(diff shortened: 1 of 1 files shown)


evgeni:
Oooh, I didn't mean to *pick* the change to 1.1.z, but pick 1.2.z packaging to 3.4.

But I am also open to this.

evgeni (approved):

ehelms:
> Oooh, I didn't mean to _pick_ the change to 1.1.z, but pick 1.2.z packaging to 3.4.
> 
> But I am also open to this.

Ohhh, I did not look at the diff to see if that made sense or not.

evgeni:
So looking at the diff between `v1.1.x` and `master`:
* `definitions/checks/foreman/check_tuning_requirements.rb` was added, which IMHO makes sense for 3.4 as that has the `--tuning development` change
* The `foreman_install` and `katello_install` features were added for the upstream upgrade feature -- doesn't hurt in 3.4
* The `packages` procedure learned how to enable modules and uninstall packages -- doesn't hurt in 3.4
* Upstream Self Upgrade feature was added -- doesn't hurt (and we don't need to advertise it as working)

PR #663, opened 2022-11-21
Fixes #35775 - manage salt services

sbernhard:


definitions/features/salt_server.rb
```diff
@@ -0,0 +1,22 @@
...
+      find_package('salt-master') &&
+        ForemanMaintain::Utils::Service::Systemd.new('salt-master', 0).exist? &&
+        ForemanMaintain::Utils::Service::Systemd.new('salt-master', 0).enabled?
+    end
+  end
+
+  def services
+    salt_services = []
+    salt_services += [system_service('salt-master', 30)]
+    salt_services += if ForemanMaintain::Utils::Service::Systemd.new('salt-api', 0).exist?
```
evgeni, line 15: does it only need to exist, or also be enabled like the master?

definitions/features/salt_server.rb
```diff
@@ -0,0 +1,22 @@
...
+                       [system_service('salt-api', 30)]
+                     else
+                       []
+                     end
```
evgeni, line 19: I would have written this as
```suggestion
    salt_services = [system_service('salt-master', 30)]
    if ForemanMaintain::Utils::Service::Systemd.new('salt-api', 0).exist?
      salt_services += [system_service('salt-api', 30)]
    end
```

But wouldn't block on syntax either way

evgeni:
Are there things that are worth backing up when Salt is used? Like config files and stuff?

If so, you can provide them like this, and the backup scenario will automatically pick them up.
https://github.com/theforeman/foreman_maintain/blob/474d54b1147270f1aecb6f8558c0c50b79df1e04/definitions/features/puppet_server.rb#L10-L20

sbernhard:
Thanks @evgeni

evgeni (approved):

PR #668, opened 2023-01-02
Fixes #35892 - use server_ssl_chain to check certs

evgeni:
Fixes: b0337ee3

### DIFF
--- a/definitions/features/instance.rb
+++ b/definitions/features/instance.rb
@@ -60,13 +60,13 @@ class Features::Instance < ForemanMaintain::Feature
     end
   end
 
-  def foreman_server_ssl_ca
-    @foreman_server_ssl_ca ||= feature(:installer).answers['foreman']['server_ssl_ca']
+  def server_connection_ca_file
+    @server_connection_ca_file ||= feature(:installer).answers['foreman']['server_ssl_chain']
   end
 
   def server_connection
     net = Net::HTTP.new(ForemanMaintain.config.foreman_url, ForemanMaintain.config.foreman_port)
-    net.ca_file = foreman_server_ssl_ca
+    net.ca_file = server_connection_ca_file
     net.use_ssl = true
     net
   end


definitions/features/instance.rb
```diff
@@ -60,13 +60,13 @@ def ping
     end
   end
 
-  def foreman_server_ssl_ca
-    @foreman_server_ssl_ca ||= feature(:installer).answers['foreman']['server_ssl_ca']
+  def foreman_server_ssl_chain
+    @foreman_server_ssl_chain ||= feature(:installer).answers['foreman']['server_ssl_chain']
```
evgeni, line 64: @ekohl @ehelms would you please double check me that this is *really* the right cert?

ekohl: I always end up reading what's what. https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcacertificatefile states the CA file is the one we allow client auth with and the chain is where we put the CA that signed the certificate.

Perhaps the method name shouldn't use the Apache values but rather something that describes the purpose. Like `foreman_connection_ca_file`?

evgeni: I can do that rename, yeah. I think the confusing part (and also the source of this issue) is the use of `server_ssl` prefix for all those variables in our installer:
```
  server_ssl_ca: "/etc/pki/katello/certs/katello-default-ca.crt"
  server_ssl_chain: "/etc/pki/katello/certs/katello-server-ca.crt"
  server_ssl_cert: "/etc/pki/katello/certs/katello-apache.crt"
  server_ssl_key: "/etc/pki/katello/private/katello-apache.key"
```

While they are all really different ;)

But to get back to my original question, is using the `server_ssl_chain` answer value correct here?

ekohl: It certainly is confusing, but in this case you are right. If anything, we should have used `internal` vs `external` instead of `default` and `server`.

ekohl (approved):

PR #670, opened 2023-01-04
Fixes #35897 - add Satellite 6.13 scenarios

evgeni:


evgeni:
I made two commits to illustrate how little the difference between 6.12 and 6.13 is right now ;)

definitions/scenarios/upgrade_to_capsule_6_13.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    def set_context_mapping
+      context.map(:assumeyes, Procedures::Installer::Upgrade => :assumeyes)
+    end
+
+    def compose
+      add_step(Procedures::Repositories::Setup.new(:version => '6.13'))
+      modules_to_enable = ["satellite-capsule:#{el_short_name}"]
+      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
```
evgeni, line 60: this is *technically* not required, but also absolutely doesn't hurt, so I left it in

(we could also probably move it to `Procedures::Repositories::Setup`, as it feels like it's part of it anyway?

definitions/scenarios/upgrade_to_satellite_6_13.rb
```diff
@@ -0,0 +1,95 @@
...
+  class PostMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures after migrating to Satellite 6.13'
+      tags :post_migrations
+    end
+
+    def compose
+      add_step(Procedures::RefreshFeatures)
+      add_step(Procedures::Service::Start.new)
+      add_step(Procedures::Pulpcore::TrimRpmChangelogs.new)
```
evgeni, line 78: We originally said we would not do it on the "to 6.12" upgrade, but as we didn't ship it in 6.12 yet… I left it in.

ekohl:
> I made two commits to illustrate how little the difference between 6.12 and 6.13 is right now ;)

I had a hard time reviewing this, so to verify:
```console
$ for file in definitions/scenarios/*6_12* ; do sed 's/\(6.\)12/\113/g ; s/6.11/6.12/g' $file > ${file/12/13} ; done
$ git diff
$
```

So it's really just replacing those values. I should really revisit extracting the target version to a separate method, which I suggested a while back. See https://github.com/theforeman/foreman_maintain/pull/671 for that.

ehelms (approved):

evgeni:
> So it's really just replacing those values. I should really revisit extracting the target version to a separate method, which I suggested a while back. See #671 for that.

Yeah, in *many* cases it's identical (or just a diff of one step or so). I agree that we can do better and will have a look at your PR.

Merging this as-is for now, to unblock users.

PR #671, opened 2023-01-04
Factor out the target version to a method

ekohl:
This allows copying the scenario without having to modify the target version everywhere. Ideally more steps would be taken out and reused.

This should probably be applied to many files, but first I'd like to get approval on the idea before I put in too much effort only to have to redo it.


ekohl:
I think Rubocop gets confused by the complex ways to define classes. There's some heavy meta-programming going on.

ehelms:
I think this is a good change, https://github.com/theforeman/foreman_maintain/pull/705 means we won't have to make it in but a few extra places.

ehelms:
This is being addressed through various changes, in large, culminating through https://github.com/theforeman/foreman_maintain/pull/879

PR #674, opened 2023-01-13
Fixes #35939 - Procedure for fixing NIC type from Nic::Base to Nic::Managed

stejskalleos:
See https://bugzilla.redhat.com/show_bug.cgi?id=1979976#c4

### DIFF
--- a/definitions/procedures/foreman/fix_nic_type.rb
+++ b/definitions/procedures/foreman/fix_nic_type.rb
@@ -1,0 +1,35 @@
+module Procedures::Foreman
+  class FixNicType < ForemanMaintain::Procedure
+    metadata do
+      advanced_run true
+      description 'Fix NIC type from Nic::Base to Nic::Managed'
+    end
+
+    def run
+      nics = feature(:foreman_database).query(sql_select)
+
+      return unless nics.any?
+
+      answer = ask_decision("Found #{nics.size} interfaces with 'Nic::Base' type.\n\n" \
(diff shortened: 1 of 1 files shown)


adamruzicka:
It works well, however the user has to know that if they have this problem, they can run this thing. Would it make sense to split this into a check and a procedure so that f-m checks for the wrong records when it runs and offers to fix it if it finds anything?

definitions/procedures/foreman/fix_nic_type.rb
```diff
@@ -0,0 +1,35 @@
...
+      abort! unless answer == :yes
+
+      update_base_nics
+    end
+
+    private
+
+    def sql_select
+      <<-SQL
+        SELECT id FROM nics WHERE type = 'Nic::Base'
```
ofedoren, line 25: Whilst reproducing this with @adamruzicka we've noticed that it's possible to create (in not supported way, but Rails/Foreman is okay with that) `Nic` with empty `type`, which can be considered in similar way as invalid. I guess we could catch here more false-valid types: `nil`, `""`, `Nic::Base`, `Nic::Interface`.

ofedoren (changes requested):

definitions/procedures/foreman/fix_nic_type.rb
adamruzicka: Just got my hands on a reproducer machine, the offending records had `NULL` type

PR #679, opened 2023-01-25
Fixes #35998: Setting token_duration fetching via database is broken

ehelms:
As part of the tftp_storage check before an upgrade, the setting token_duration was being fetched using a direct database query. As the table changed,
 this query also changed. This method of direct database access is unreliable
as it is not beholden to an API that is stable. This switches to using hammer to extract the value.

### DIFF
--- a/definitions/checks/foreman_proxy/check_tftp_storage.rb
+++ b/definitions/checks/foreman_proxy/check_tftp_storage.rb
@@ -42,12 +42,8 @@ module Checks::ForemanProxy
     end
 
     def lookup_token_duration
-      data = feature(:foreman_database). \
-             query("select s.value, s.default from settings s \
-                    where category IN ('Setting::Provisioning','Setting') \
-                    and name = 'token_duration'")
-
-      YAML.load(data[0]['value'] || data[0]['default'])
+      data = feature(:hammer).run('--output yaml settings info --id token_duration')
+      YAML.load(data)['Value']
     end
   end
 end


definitions/checks/foreman_proxy/check_tftp_storage.rb
```diff
@@ -42,12 +42,8 @@ def token_duration
...
 
     def lookup_token_duration
-      data = feature(:foreman_database). \
-             query("select s.value, s.default from settings s \
-                    where category IN ('Setting::Provisioning','Setting') \
-                    and name = 'token_duration'")
-
-      YAML.load(data[0]['value'] || data[0]['default'])
+      data = feature(:hammer).run('--output yaml settings info --id token_duration')
+      YAML.load(data['Value'])
```
evgeni, line 46: Shouldn't this be (untested)
```suggestion
      YAML.load(data)['Value']
```
As `data` will be the YAML hammer returns, which you need to parse and then get the Value out of it?

evgeni: And if you wonder why this seems to work:

```
irb(main):001:0> a = "dgdfg Value sgdfg"
=> "dgdfg Value sgdfg"
irb(main):002:0> a['Value']
=> "Value"
```

So you get `Value` as the return of `lookup_token_duration`

evgeni (changes requested):
see inline comment

definitions/checks/foreman_proxy/check_tftp_storage.rb
ehelms: Nutty

evgeni (approved):

definitions/checks/foreman_proxy/check_tftp_storage.rb
evgeni: Ruby, but I'll take yours too.

PR #680, opened 2023-01-26
Fixes #36012 - restore properly backups created with -t option

pmoravec:
Use --new-volume-script option for every tar (not gz) command to ensure also tar extract can "swap tapes" automatically.

### DIFF
--- a/definitions/features/tar.rb
+++ b/definitions/features/tar.rb
@@ -51,13 +51,11 @@ class Features::Tar < ForemanMaintain::Feature
     end
 
     if volume_size
-      split_tar_script = default_split_tar_script
       tar_command << "--tape-length=#{volume_size}"
-      tar_command << "--new-volume-script=#{split_tar_script}"
     end
 
     tar_command << '--overwrite' if options[:overwrite]
-    tar_command << '--gzip' if options[:gzip]
+    tar_command << (options[:gzip] ? '--gzip' : "--new-volume-script=#{default_split_tar_script}")
 
     exclude = options.fetch(:exclude, [])
     exclude.each do |ex|


pmoravec:
It is not an ideal fix to use the `--new-volume-script` every time we create `*.tar` backup (even without `-t` / `--tape-length` option on backup), but a check "we do extract of an archive with tape length" would be redundantly complex, I think.

This option works well also for creating or extracting tarballs without `--tape-length`.

pmoravec:
Failure

```
definitions/features/tar.rb:59:5: C: Use the return of the conditional for variable assignment and comparison.
    if options[:gzip] ...
    ^^^^^^^^^^^^^^^^^
```

should be resolved by

```
tar_command << '--gzip' if options[:gzip] else "--new-volume-script=#{default_split_tar_script}"
```

? (I dont have means to verify it atm)

definitions/features/tar.rb
```diff
@@ -51,13 +51,11 @@ def tar_command(options)
...
 
     if volume_size
-      split_tar_script = default_split_tar_script
       tar_command << "--tape-length=#{volume_size}"
-      tar_command << "--new-volume-script=#{split_tar_script}"
     end
 
     tar_command << '--overwrite' if options[:overwrite]
-    tar_command << '--gzip' if options[:gzip]
+    tar_command << (options[:gzip] ? '--gzip' : "--new-volume-script=#{default_split_tar_script}")
```
ehelms, line 58: Previously this was gated by `volume_size` being true, is that no longer necessary?

pmoravec: Yes and no: https://github.com/theforeman/foreman_maintain/pull/680#issuecomment-1405684413 . Technically not necessary, `tar` does cope well (I successfully tested it, anyway an independent verification is welcomed). "Logically" or methodologically it is required (why to treat archive as "split to tapes" every time, even though it is usually not the case?). But I havent found out a simple test to distinguish that ("was the tar archive created with -t option, or not?").

ehelms (approved):

PR #681, opened 2023-01-31
Fixes #36024 - ignore already deleted sync plans

sbernhard:


definitions/features/sync_plans.rb
```diff
@@ -33,12 +33,24 @@ def sync_plan_ids_by_status(enabled = true, filter_ids = nil)
...
+    feature(:foreman_database).query(query).map { |r| r['id'].to_i }
+  end
+
   def make_disable(ids)
     update_records(ids, false)
   end
 
   def make_enable
-    update_records(data[:disabled], true)
+    # remove ids of sync plans which no longer exist in DB
```
ekohl, line 49: Shouldn't we tell users to remove IDs from their config?

ekohl (commented):
I had a look and this generally looks ok, but I wonder if we should silently discard invalid options or rather error out and tell users to fix their config.

definitions/features/sync_plans.rb
sbernhard: Is it a "config" or just the internal "brain" of the foreman-maintain which shouldn't be adapted by a user itself?

evgeni: data.yml is purely an internal thing that is used for storing things between runs.

definitions/features/sync_plans.rb
```diff
@@ -33,12 +33,24 @@ def sync_plan_ids_by_status(enabled = true, filter_ids = nil)
     feature(:foreman_database).query(query).map { |r| r['id'].to_i }
   end
 
+  def validate_sync_plan_ids(ids)
+    ids_condition = ids.map { |id| "'#{id}'" }.join(',')
+    query = <<-SQL
+      SELECT id FROM katello_sync_plans WHERE id IN (#{ids_condition})
+    SQL
+    feature(:foreman_database).query(query).map { |r| r['id'].to_i }
+  end
```
ehelms, line 42: Can we do this without querying the database directly? Either through hammer or the API?

sbernhard: Would be possible but in the same file, there are other SQL querries. Therefore I thought, this is the way to implement this.

ehelms: Yea... I was hoping we could start to not make that problem worse.

sbernhard: Well, currently there is a bug which should be solved. Receiving the still existing sync plan ids via hammer wouldn't be pretty performant. 

In another PR, I'm pretty sure the required_new_implementation can bei removed.  Maybe then its a good point in time to make some parts easier.

definitions/features/sync_plans.rb
```diff
@@ -33,12 +33,24 @@ def sync_plan_ids_by_status(enabled = true, filter_ids = nil)
...
+
   def make_disable(ids)
     update_records(ids, false)
   end
 
   def make_enable
-    update_records(data[:disabled], true)
+    # remove ids of sync plans which no longer exist in DB
+    ids = validate_sync_plan_ids(data[:disabled])
+    @data[:disabled] &= ids
```
ehelms, line 51: Is this step necessary? I would have thought the line above returned only valid ids already.

definitions/features/sync_plans.rb
ehelms: Welp, you can't blame me for trying to stem the tide of bad design in the project. Thanks for at least considering it.

definitions/features/sync_plans.rb
sbernhard: @data[:disabled] has all IDs from the local storage, which were disabled before (1)
validate_sync_plan_ids returns all IDs of still existing sync plans from the DB (2)

Only sync plans which where disabled before (1) and still exist in the DB (2) can be enabled using hammer in https://github.com/theforeman/foreman_maintain/pull/681/files#diff-32fb2f52b059a8d16272774a6261b7cf27c1ccc7d5ad4ce057a66a2fcbed3719L78

definitions/features/sync_plans.rb
sbernhard: This was surely not my intention. The question is, should we first improve the design of sync-plan feature and afterwards fix the issue (a), fix the bug first and then improve the design (b) or do it in one PR (c)?

definitions/features/sync_plans.rb
ehelms: Should this only be passing `@data[:disabled]` to `update_records`? If that is the subset of identified disabled intersection with what was found in the database?

definitions/features/sync_plans.rb
ehelms: I am mostly worried doing it systematically will be hard, so where we can, try not to introduce new instances of database access. That does not mean you have to solve it here to solve the real bug, not as a follow up (unless you really really want to). I just have the one logic question -- https://github.com/theforeman/foreman_maintain/pull/681/files#r1112268541 and then we can proceed with merging

definitions/features/sync_plans.rb
sbernhard: You were right. Thank you. `@data[:disabled]` is of course the right variable to pass to update_records()

evgeni: >  validate_sync_plan_ids returns all IDs of still existing sync plans from the DB (2)

But you're passing the *current* `data[:disabled]` to `validate_sync_plan_ids` so the return value should be "all sync plan ids from the old data disabled list that still exist in the DB"?

So, wouldn't this be correct too?
```ruby
data[:disabled] = validate_sync_plan_ids(data[:disabled])
update_records(data[:disabled], true)
```

evgeni: also the `data` vs `@data` confuses me :)

(but overall I think this is doing what it's supposed to be doing and we're just talking cosmetics)

sbernhard: data vs @data => https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/sync_plans.rb#L90

true :) thanks again.

evgeni: OMGWTFBBQ ;-)

Thanks!

ehelms (approved):

PR #683, opened 2023-02-02
Fixes #36034 - add mosquitto feature definition

evgeni:


definitions/features/mosquitto.rb
```diff
@@ -0,0 +1,28 @@
...
+    label :mosquitto
+
+    confine do
+      # Luckily, the service name is the same as the package providing it
+      find_package(service_name)
+    end
+  end
+
+  def services
+    [system_service(self.class.service_name, 10)]
```
evgeni, line 12: no idea about the real priority, so I took the same as `qpidd` had

ehelms:
Per our previous conversations, mind adding a basic set of tests for this in https://github.com/theforeman/foreman_maintain/tree/master/test/definitions/features ?

ehelms (approved):
Does this automatically add the service into the service list? Does it only do so if it's available on the system? If it's enabled?

evgeni:
> Does this automatically add the service into the service list? Does it only do so if it's available on the system? If it's enabled?

If the package is installed (see `confine` block) the service is added to the list.

PR #687, opened 2023-02-23
Refs #36024 - don't fail on empty sync plan list

evgeni:


definitions/features/sync_plans.rb
```diff
@@ -34,10 +34,10 @@ def sync_plan_ids_by_status(enabled = true, filter_ids = nil)
   end
 
   def validate_sync_plan_ids(ids)
+    return [] if ids.empty?
+
     ids_condition = ids.map { |id| "'#{id}'" }.join(',')
-    query = <<-SQL
-      SELECT id FROM katello_sync_plans WHERE id IN (#{ids_condition})
-    SQL
+    query = "SELECT id FROM katello_sync_plans WHERE id IN (#{ids_condition})"
```
evgeni, line 40: this is only cosmetic so that the stub below doesn't have to care for indentation and newlines

sbernhard (approved):

PR #691, opened 2023-03-01
Fixes #36152 - don't try to backup or restore pg_globals

evgeni:
We used to backup pg_globals as those contain roles (users) and should allow us to restore the same on the new system.

However, the restore never has worked, as the code as introduced in 1ae71c3ad827503eb370310a6b27232f80289351 tried to restore the globals using the "foreman_database" connection, so the "foreman" user which means this user had to *exist* for the restore and more crucially would need superuser privileges, which it never had.

In the logs, you can see that during a restore it actually throws errors:

    DEBUG -- : Running command PGPASSWORD='[FILTERED]' psql -h localhost  -p 5432 -U foreman -f /backup/pg_globals.dump postgres 2>/dev/null with stdin nil
    DEBUG -- : output of the command:
    SET
    SET
    SET
    psql:/backup/pg_globals.dump:14: ERROR:  permission denied to create role
    psql:/backup/pg_globals.dump:15: ERROR:  must be superuser to alter superusers
    psql:/backup/pg_globals.dump:16: ERROR:  permission denied to create role
    psql:/backup/pg_globals.dump:17: ERROR:  must be superuser to alter superusers
    psql:/backup/pg_globals.dump:18: ERROR:  permission denied to create role
    psql:/backup/pg_globals.dump:19: ERROR:  must be superuser to alter superusers
    psql:/backup/pg_globals.dump:20: ERROR:  permission denied to create role
    psql:/backup/pg_globals.dump:21: ERROR:  must be superuser to alter superusers
    INFO -- : --- Execution step 'Restore any existing postgresql global objects from backup' finished ---

The code has been ported over from katello-restore [1] where it ran as the "postgres" user, which *does* have superuser permissions and thus worked.

The good thing is, we run the installer *before* we try to restore things, and that already sets up all the users correctly for us, and there is no need to try to restore pg_globals.

[1] https://github.com/theforeman/foreman-packaging/blob/ed161b7ac3fe3e6a31e7d3b329f9a0ce4aa50b41/packages/katello/katello/restore.rb#L93

Fixes: 1ae71c3ad827503eb370310a6b27232f80289351


sayan3296:
> @sayan3296 promised to test this one out, as I couldn't reproduce the original issue he had that started the chase of this functionality.

I confirm that , On the reproducer where I was always able to reproduce the issue, After installing rubygem-foreman_maintain-1.2.6-1.20230301111458534583.pr691.1.g53dac01.el8.noarch.rpm and trying to restore the online backup, Everything worked great.  My restored instance looks perfectly fine.

ehelms:
> > @sayan3296 promised to test this one out, as I couldn't reproduce the original issue he had that started the chase of this functionality.
> 
> I confirm that , On the reproducer where I was always able to reproduce the issue, After installing rubygem-foreman_maintain-1.2.6-1.20230301111458534583.pr691.1.g53dac01.el8.noarch.rpm and trying to restore the online backup, Everything worked great. My restored instance looks perfectly fine.

Awesome! Using the Packit build++

ehelms (approved):
0 lines added
84 lines removed

My kind of change.

evgeni:
I think the only remaining question is: are we saying this fixes the issue you reported (even tho I wasn't able to reproduce it) or should we convert it to "fm needlessly tries to restore pg globals and silently fails doing so" which is something one can *always* observe?

evgeni:
> 0 lines added
> 84 lines removed
> 
> My kind of change.

I'm sure it's also something @tbrisker would enjoy ;)

tbrisker (approved):
I approve of that message.

sayan3296:
> I think the only remaining question is: are we saying this fixes the issue you reported (even tho I wasn't able to reproduce it) or should we convert it to "fm needlessly tries to restore pg globals and silently fails doing so" which is something one can _always_ observe?

Well, As you had seen in my reproducer, It was evidently reproducible for me always. But the fact that, pg_globals restoration was failing whereas it was not even needed from the restoration point of view, Let's say `"fm needlessly tries to restore pg globals and silently fails doing so"`  is the right statement here. 

I am good with anything here as the fix works and works great ..

PR #693, opened 2023-03-02
Add 6.14 upgrade scenario

ehelms:


evgeni (changes requested):
this is missing the 6.14.z and capsule scenarios, which we usually add in one go

definitions/scenarios/upgrade_to_satellite_6_14_z.rb
```diff
@@ -0,0 +1,94 @@
...
+  class PostMigrations < Abstract
+    upgrade_metadata do
+      description 'Procedures after migrating to Satellite 6.14.z'
+      tags :post_migrations
+    end
+
+    def compose
+      add_step(Procedures::RefreshFeatures)
+      add_step(Procedures::Service::Start.new)
+      add_step(Procedures::Pulpcore::TrimRpmChangelogs.new)
```
evgeni, line 77: same as above

PR #701, opened 2023-03-23
add check whether postgres user can read dump files

evgeni:


definitions/checks/restore/validate_postgresql_dump_permissions.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def run
+      backup = ForemanMaintain::Utils::Backup.new(@backup_dir)
+      if feature(:instance).postgresql_local? && backup.online_backup?
+        errors = []
+        [:candlepin_dump, :foreman_dump, :pulpcore_dump].each do |dump|
+          if backup.file_map[dump][:present]
+            success = system("runuser - postgres -c 'test -r #{backup.file_map[dump][:path]}'")
+            errors << backup.file_map[dump][:path] unless success
+          end
```
ehelms, line 20: Just another way to write parts of this (not required, food for thought):

```suggestion
    next unless backup.file_map[dump][:present]
    unless system("runuser - postgres -c 'test -r #{backup.file_map[dump][:path]}'")
      errors << backup.file_map[dump][:path]
    end
```

ehelms (approved):

definitions/checks/restore/validate_postgresql_dump_permissions.rb
evgeni: Like!

evgeni: (besides the bad syntax and all of that, fixed)

PR #702, opened 2023-03-27
Fixes #36230 - Puppet removal parameter to installer is no longer used

ehelms:


ehelms:
I've a proposed a test to ensure this over in bats: https://github.com/theforeman/forklift/pull/1664

ehelms:
> Is this something which should take the version into account? If this is cherry picked to an older release then it'll break things. If not, I think the min version should be raised to 3.6.

I would assume if you cherry pick this back to an older release you are doing so with the intent to provide this functionality to that older release. 

But since those checks exist, I can update them. They are a hold over from the previous release model foreman_maintain was used in with Satellite.

ekohl (approved):
Thanks

PR #703, opened 2023-03-27
Fixes #36233 - Run container-repair-media-type script on upgrade

wbclark:
I think we could run this on other upgrades as well. Does foreman_maintain have a check to ensure that the procedure doesn't run again next time if we already ran it, though?

### DIFF
--- a/definitions/procedures/pulpcore/repair_container_media_type.rb
+++ b/definitions/procedures/pulpcore/repair_container_media_type.rb
@@ -1,0 +1,23 @@
+module Procedures::Pulpcore
+  class RepairContainerMediaType < ForemanMaintain::Procedure
+    include ForemanMaintain::Concerns::SystemService
+
+    metadata do
+      description 'Repair container media type in the pulpcore db'
+      for_feature :pulpcore
+    end
+
+    def run
+      with_spinner('Repairing container media type in the pulpcore db') do |spinner|
+        necessary_services = feature(:pulpcore_database).services
+
+        feature(:service).handle_services(spinner, 'start', :only => necessary_services)
+
+        spinner.update('Repairing container media type')
+        execute!('PULP_SETTINGS=/etc/pulp/settings.py '\
+                 'DJANGO_SETTINGS_MODULE=pulpcore.app.settings '\
+                 'pulpcore-manager container-repair-media-type')
+      end
+    end
+  end
+end
--- a/definitions/scenarios/upgrade_to_satellite_6_11.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_11.rb
@@ -77,6 +77,7 @@ module Scenarios::Satellite_6_11
       add_step(Procedures::RefreshFeatures)
       add_step(Procedures::Service::Start.new)
       add_step(Procedures::Pulpcore::TrimRpmChangelogs.new)
+      add_step(Procedures::Pulpcore::ContainerRepairMediaType.new)
       add_steps(find_procedures(:post_migrations))
     end
   end
--- a/definitions/scenarios/upgrade_to_satellite_6_11_z.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_11_z.rb
@@ -72,6 +72,7 @@ module Scenarios::Satellite_6_11_z
     def compose
       add_step(Procedures::RefreshFeatures)
       add_step(Procedures::Service::Start.new)
+      add_step(Procedures::Pulpcore::ContainerRepairMediaType.new)
       add_steps(find_procedures(:post_migrations))
     end
   end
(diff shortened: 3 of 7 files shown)


ehelms:
> I think we could run this on other upgrades as well. Does foreman_maintain have a check to ensure that the procedure doesn't run again next time if we already ran it, though?

It will run every time, so need to make sure its idempotent if possible and that its performant.

wbclark:
Thanks! I've added the procedure to additional upgrades -- it now targets 6.11, 6.11.z, 6.12, 612.z, and 6.13.

@ipanova @ianballou would you mind giving this approach a quick sanity check?

ehelms:
> Thanks! I've added the procedure to additional upgrades -- it now targets 6.11, 6.11.z, 6.12, 612.z, and 6.13.

That will somewhat help, but it'll still have to be cherry picked back and conflicts addressed for any version you want to bring this back to.

wbclark:
> > Thanks! I've added the procedure to additional upgrades -- it now targets 6.11, 6.11.z, 6.12, 612.z, and 6.13.
> 
> That will somewhat help, but it'll still have to be cherry picked back and conflicts addressed for any version you want to bring this back to.

The maintenance repository is unversioned, and FM always updates itself to the latest version before doing anything else, correct?

If CPs are needed, I should be able to open those after this is finalized and merged. Are you anticipating merge conflicts?

ianballou (commented):
This is looking good to me! I haven't tested it, but the planning and the running of the cleanup script appears to be correct.

wbclark:
I tracked down the PR which introduced the repair command, at https://github.com/pulp/pulp_container/pull/1003/files

If I'm reading this correctly it won't be the most performant implementation for repeated runs, since it forces an update for all manifests in manifests_schema_v1 if the media type as determined by the associated json metadata is not MEDIA_TYPE.MANIFEST_V1

This means that for manifests with other media types, an update will be attempted each time the repair command is run, even if the media type is already set to the correct value. The result would still be the same after running the script twice, but the second run would not be a no-op.

I also need to refresh my memory regarding whether it's possible (or even default) to run Katello with the Pulpcore Cache enabled, in which case it looks like a potential issue would occur if we tried to run the repair script while Pulpcore is configured to use the Cache but Redis is not running.

ipanova:
@wbclark besides mentioned PR there is one more https://github.com/pulp/pulp_container/pull/1051 So this commanded fixes 2 things.
You're right this is not really meant to be repeatedly run. It is not performant because unfortunately we need to inspect each manifest json file stored on the FS. It is enough to be run it once, for content that was in DB prior the bugfix, with the bugfix backport it will ensure *new* incoming content is correct.

ehelms:
Is there a clear way to know if it *needs* to be run based on some data? Or if it has run before based on some existing data?

ipanova:
It needs to be ran on the upgrade ( just once) as long as the DB is not empty and there will be manifests with `MEDIA_TYPE.MANIFEST_V1` , which can point to incorrectly stored data.
There is no way to determine whether the script ran before.

wbclark:
Looking into it more, there is https://github.com/theforeman/foreman_maintain/commit/a625d952a5697520da95a65f7fbbff97472fde77

The comment at https://github.com/theforeman/foreman_maintain/blob/72adb771d2c99a4ca0e998711d472a04152d4ed6/lib/foreman_maintain/concerns/metadata.rb#L93-L97 implies it may not work out of the box across different upgrade scenarios, however

ehelms:
Note I am planning to remove the irrelevant upgrade scenarios (https://github.com/theforeman/foreman_maintain/pull/705) which will force a rebase on this PR. The changes will then need to be cherry-picked back to any version relevant to a Satellite version where this change is desired (which would have to happen regardless).

ehelms:
I'd like to check again based on re-reading and checking the code. Is there a way to query Pulp and ask if there are any containers with `MEDIA_TYPE.MANIFEST_V1_SIGNED` ? If I understand correctly, that is the logic that would indicate whether the repair needs to run?

Alternatively, looking at the logic, this appears to do a fetch for any containers that have `MEDIA_TYPE.MANIFEST_V1_SIGNED`, is that correct? Which means if it returns 0 then the update will do nothing and it will be a noop basically.

wbclark:
The latest (internal) 6.12 build has python39-pulp-container-2.10.12-1.el8pc.noarch.rpm and the latest 6.13 build has python39-pulp-container-2.14.3-1.el8pc.noarch.rpm, while 6.11 has python38-pulp-container-2.9.9-1.el8pc.noarch.rpm

The original bug https://bugzilla.redhat.com/show_bug.cgi?id=2102078 is in VERIFIED state for 6.13 but it was never cloned into 6.11 or 6.12

Satellite 6.12 and some later versions of Satellite 6.11 do contain pulp-container versions with the script; however this is a consequence of using later pulp releases to address other bugs -- the script was never QE'd for those releases and there was never any erratum.

Therefore, we can execute this script just on the 6.13 upgrades for Satellite and Capsule, since that is the first version where the fix is "officially" present.

Users who are aware of the script on earlier releases could still try to run it if they are not yet ready for the full upgrade to 6.13, but since we won't target multiple upgrades for the fix, there won't be any mechanism to skip executing the repair script if it has already run.

As https://github.com/theforeman/foreman_maintain/pull/705 is now merged, I'll close this PR and open a new PR for the 6.13 branch.

PR #705, opened 2023-03-28
Remove upgrade scenarios that no are no longer applicable

ehelms:
I think it's time and safe to remove all previous upgrade scenarios based upon the way foreman_maintain is being released. There are effectively versions that map to each Satellite release for which most upgrade scenarios exist and therefore the main branch only needs the latest and greatest.


ekohl (approved):

evgeni:
I think I would keep the more recent ones (say 6.12+), as they are mostly identical to the latest ones (so it's not like you can remove more supporting code) and that would allow us to keep (unofficial) compatibility to quicker try out fixes via packit and friends.

(But I wouldn't block on that, as technically you're right and those *can* go)

One side quests: if someone starts an upgrade to the next version (6.14 here, as everything older is removed) by installing the new maintain, but then does not do it, do they have to revert to the older maintain to get 6.13.z upgrades back?

ehelms:
> One side quests: if someone starts an upgrade to the next version (6.14 here, as everything older is removed) by installing the new maintain, but then does not do it, do they have to revert to the older maintain to get 6.13.z upgrades back?

Yes. And that sounds like the correct answer regardless.


> I think I would keep the more recent ones (say 6.12+), as they are mostly identical to the latest ones (so it's not like you can remove more supporting code) and that would allow us to keep (unofficial) compatibility to quicker try out fixes via packit and friends.

You can by opening a PR to the related branch of foreman-maintain that matches the Satellite version right? Which would in the end better match the environment to test anyway?

PR #706, opened 2023-03-28
Remove all references to EL7

ehelms:


definitions/features/redis.rb
```diff
@@ -17,24 +17,12 @@ def config_files
   end
 
   class << self
-    SCL_NAME = 'rh-redis5'.freeze
-
     def etc_prefix
```
ekohl, line 20: I debated if this is still needed. In EL9 the config is `/etc/redis/redis.conf` and on Debian it's always been in `/etc/redis/redis.conf` too. Looking at `config_files` it already always probes `/etc/redis` now so perhaps it is redundant.

lib/foreman_maintain/concerns/os_facts.rb
```diff
@@ -62,10 +62,6 @@ def ubuntu?
         os_id == 'ubuntu'
       end
 
-      def el7?
-        el_major_version == 7
-      end
-
       def el8?
```
ekohl, line 65: This is now unused so I wonder if we should drop it.

ekohl (approved):

PR #707, opened 2023-03-28
Remove all mongo and Pulp 2 references

ehelms:
There is a lot here, as there was a lot to remove. And this will fail tests currently. The trickiest bit has been getting the backup code squared away as there is some funk happening in there. The most confusing thing I am trying to understand is the "hybrid" backup scenario.

  1. Is hybrid (some local, some remote) something we actually want to support?
  2. Is hybrid a holdover from 6.9 -> 6.10 where we missed the Pulpcore database having options to be remote for external database users initially?


ehelms:
Some further data about hybrid.

In Satellite 6.9, we did not provide a clean upgrade workflow for external database users. This resulted in users with an internal Pulpcore database and external Foreman + Candlepin databases. The hybrid represents users in this situation and needing backup.

I would propose we do the following:

 * Remove hybrid backups
 * Add an upgrade check to determine if the user has a local Pulpcore database and external Candlepin / Foreman
 * Get users to fix this situation before upgrade

definitions/scenarios/restore.rb
```diff
@@ -31,10 +31,6 @@ def compose
         Procedures::Selinux::SetFileSecurity,
         Procedures::Restore::Configs)
       add_step_with_context(Procedures::Crond::Stop) if feature(:cron)
-      unless backup.incremental?
-        add_steps_with_context(Procedures::Restore::EnsureMongoEngineMatches,
-          Procedures::Restore::InstallerReset)
```
evgeni, line 36: this step (InstallerReset) should remain, otherwise restore will be broken.

definitions/scenarios/restore.rb
```diff
@@ -45,17 +41,16 @@ def compose
...
       end
 
-      add_steps_with_context(Procedures::Pulp::Migrate,
+      add_steps_with_context(
         Procedures::Pulpcore::Migrate,
-        Procedures::Restore::CandlepinResetMigrations)
+        Procedures::Restore::CandlepinResetMigrations
+      )
 
-      add_steps_with_context(Procedures::Restore::RegenerateQueues) if backup.online_backup?
```
evgeni, line 58: why is this removed? it doesn't seem pulp2 related?

lib/foreman_maintain/utils/backup.rb
```diff
@@ -19,26 +19,16 @@ def initialize(backup_dir)
         @foreman_offline_files = ['pgsql_data.tar.gz']
         @katello_online_files = ['candlepin.dump', 'foreman.dump']
         @katello_offline_files = ['pgsql_data.tar.gz']
-        if feature(:pulp2)
-          @katello_online_files  << 'mongo_dump'
-          @katello_offline_files << 'mongo_data.tar.gz'
-          @fpc_online_files = ['mongo_dump']
-          @fpc_offline_files = ['mongo_data.tar.gz']
-        elsif feature(:pulpcore_database)
-          @katello_online_files << 'foreman.dump'
```
evgeni, line 28: while this line was wrong, I think `katello_online_files` should include `pulpcore.dump` ;)

test/lib/cli/upgrade_command_test.rb
```diff
@@ -230,6 +230,7 @@ def foreman_maintain_update_unavailable
       end
 
       it 'with --phase it runs only a specific phase of the upgrade' do
+        foreman_maintain_update_unavailable
```
evgeni, line 233: why was that needed?

evgeni (changes requested):

definitions/scenarios/restore.rb
ehelms: Solid question, according to https://github.com/theforeman/foreman_maintain/pull/707/files#diff-f9d25557678350ba1c6282e05ea917d8eb58185b49203038a43aba238c149b56L6-L8 this was confined to Pulp 2 only, so it has not been used for quite a few releases now.

test/lib/cli/upgrade_command_test.rb
ehelms: I'll remove it, as I have come to realize there is a intermittent test failure that happens in some runs that I will try to fix stand-alone.

definitions/scenarios/restore.rb
evgeni: amazing(ly confusing)

evgeni (approved):

evgeni:
Welp, let's see what we break

PR #710, opened 2023-03-29
always add Rubocop tasks

evgeni:
We don't support anything older than Ruby 2.0 for a looong time.


ehelms (approved):
Tested locally since GH actions was having a bad day.

PR #711, opened 2023-03-30
Add JUnit reporter for Jenkins

ehelms:


lib/tasks/jenkins.rake
```diff
@@ -0,0 +1,18 @@
...
+      task :pre_ci do
+        ENV["CI_REPORTS"] = 'jenkins/reports/unit/'
+        gem 'ci_reporter'
+      end
+      task :minitest  => [:pre_ci, "ci:setup:minitest"]
+    end
+  end
+rescue LoadError
+  # ci/reporter/rake/rspec not present, skipping this definition
+end
```
ekohl, line 18: Perhaps this comes from Python, but I've grown fond of limiting the part that can raise an exception and then putting the real code in `else`.
```suggestion
begin
  require "ci/reporter/rake/minitest"
rescue LoadError
  # ci/reporter/rake/rspec not present, skipping this definition
else
  namespace :jenkins do
    task :unit => ["jenkins:setup:minitest", 'rake:test']

    namespace :setup do
      task :pre_ci do
        ENV["CI_REPORTS"] = 'jenkins/reports/unit/'
        gem 'ci_reporter'
      end
      task :minitest  => [:pre_ci, "ci:setup:minitest"]
    end
  end
end
```

foreman_maintain.gemspec
```diff
@@ -30,4 +30,5 @@ the Foreman/Satellite up and running."
   s.add_development_dependency 'rake'
   s.add_development_dependency 'rubocop', '0.50.0' # rubocop >= 0.51.0 drops support for ruby 2.0
   s.add_development_dependency 'minitest-stub-const'
+  s.add_development_dependency 'ci_reporter_minitest'
```
ekohl, line 33: In Foreman I've migrated to [minitest-reporters](https://github.com/minitest-reporters/minitest-reporters) because it's maintained. The last commit to `ci_reporter_minitest` was in 2014. I've yet to migrate the Smart Proxy repositories, but that's mostly for a lack of time. See https://github.com/theforeman/foreman/commit/b1f03b375e18c8da498ba0944b2565f53441833c for the conversion, which is rather trivial.

ekohl (changes requested):

lib/tasks/jenkins.rake
ehelms: I've not seen the pattern before using begin/rescue/else. Structurally though I think it confuses me a bit. As I am not clear reading it if the `else` is tied to the begin or the rescue? And I thought the `namespace :jenkins` block should only happen if the require actually passes?

BTW I borrowed this task from -- https://github.com/theforeman/smart-proxy/blob/develop/tasks/jenkins.rake

ekohl: > I've not seen the pattern before using begin/rescue/else. Structurally though I think it confuses me a bit. As I am not clear reading it if the else is tied to the begin or the rescue? And I thought the namespace :jenkins block should only happen if the require actually passes?

I couldn't find a more recent version of the doc, but https://docs.ruby-lang.org/en/2.4.0/syntax/exceptions_rdoc.html has:
```ruby
begin
  # ...
rescue
  # ...
else
  # this runs only when no exception was raised
ensure
  # ...
end
```

Or from https://stackify.com/rescue-exceptions-ruby/ an even clearer example:

```ruby
begin
    #... process, may raise an exception
rescue =>
    #... error handler
else
    #... executes when no error
ensure
    #... always executed
end
```

It is the block that is executed if no exception was raised.

> BTW I borrowed this task from -- https://github.com/theforeman/smart-proxy/blob/develop/tasks/jenkins.rake

It was a syntax that Ruby introduced somewhere after it was added to Smart Proxy. In my memory it was Ruby 2.5, but that it was documented in 2.4 tells me I'm wrong.

The idea is that any code in could raise an exception, and you want it to be as specific as possible.

ehelms:
Here's another swing at it using reporters, but I find I get an odd result when I run it locally.

lib/tasks/jenkins.rake
```diff
@@ -0,0 +1,23 @@
...
+
+    namespace :setup do
+      task :minitest do
+        ENV['MINITEST_REPORTER'] = 'JUnitReporter'
+        ENV['MINITEST_REPORTERS_REPORTS_DIR'] = 'jenkins/reports/unit'
+      end
+    end
+
+    task :rubocop do
+      system("bundle exec rubocop \
```
ekohl, line 16: I think this isn't really needed anymore and we should migrate away from it, in favor of `recordIssues` with the `ruboCop` parser. See https://github.com/theforeman/jenkins-jobs/blob/1041458b9bb8e68e53b8d863e0ff251540632304/theforeman.org/pipelines/test/smart-proxy.groovy#L45-L54

Again, a lot of legacy we don't really need anymore, but haven't gotten around to cleaning up.

lib/tasks/jenkins.rake
```diff
@@ -0,0 +1,23 @@
+begin
+  require 'English'
```
ekohl, line 2: `English` is part of Ruby itself, so I don't think you need to rescue it. If that's broken, something is seriously wrong.

Having said that: if you change the rubocop task (or drop it altogether) to not use `system()` you don't need `$CHILD_STATUS` and can safely drop it.

lib/tasks/jenkins.rake
```diff
@@ -0,0 +1,23 @@
...
+rescue LoadError
+else
+  namespace :jenkins do
+    task :unit => ['jenkins:setup:minitest', 'rake:test']
+
+    namespace :setup do
+      task :minitest do
+        ENV['MINITEST_REPORTER'] = 'JUnitReporter'
+        ENV['MINITEST_REPORTERS_REPORTS_DIR'] = 'jenkins/reports/unit'
```
ekohl, line 11: I'd prefer to set this in Jenkins via env vars:
https://github.com/theforeman/jenkins-jobs/blob/1041458b9bb8e68e53b8d863e0ff251540632304/theforeman.org/pipelines/test/smart-proxy.groovy#L98-L100

ehelms: That would make this task irrelevant, which is fine, but it does mean that it's harder to just replicate this locally. As in, a user has to jump through more hoops rather than just running a nicely encapsulated rake task.

Thoughts?

ekohl (approved):

PR #715, opened 2023-04-03
Drop old satellite

ehelms:
Requires other PRs getting in first.


ehelms:
This is now ready for review rebased on the big drop Pulp 2 and EL7 work. After this change goes in, I will do a new release to push this through nightly testing.

definitions/checks/pulpcore/group_ownership_check.rb
```diff
@@ -8,10 +8,7 @@ class GroupOwnershipCheck < ForemanMaintain::Check
       end
 
       def run
-        group_id = File.stat('/var/lib/pulp/content/').gid
-        if Etc.getgrgid(group_id).name != 'pulp'
-          fail! "Please run 'foreman-maintain prep-6.10-upgrade' prior to upgrading."
-        end
+        File.stat('/var/lib/pulp/content/').gid
```
evgeni, line 11: This check now does nothing (besides fetching the `gid` of `/var/lib/pulp/content` and never using it for anything). I'd argue the whole check can go away now?

definitions/checks/foreman/check_https_proxies.rb
```diff
@@ -1,34 +0,0 @@
...
-      description 'Check for HTTPS proxies from the database'
-      manual_detection
-    end
-
-    def run
-      https_proxies = find_https_proxies
-      unless https_proxies.empty?
-        https_proxy_names = https_proxies.map { |proxy| proxy['name'] }
-        question = "Syncing repositories through an 'HTTP Proxy' that uses the HTTPS\n"\
-                    "protocol is not supported directly with Satellite 6.10.\n"\
```
evgeni, line 15: technically, that BZ is still open and unsolved, so I would keep that check?

evgeni: (but it's `manual_detection` and the last use was in `definitions/scenarios/upgrade_to_satellite_6_10.rb` :woman_shrugging: )

definitions/checks/pulpcore/group_ownership_check.rb
evgeni: (also, it's `manual_detection` and the last use was in the already removed `definitions/scenarios/upgrade_to_satellite_6_10.rb`)

ehelms: You aint got to tell me twice to delete code, even though you tried.

definitions/checks/foreman/check_https_proxies.rb
ehelms: Linkie to the BZ?

ehelms: Hrmm, yea so when upgrading to 6.10 it was an issue, and there is nothing about the "upgrade" itself that would cause this to be better or worse. I think for upgrades this check does not make sense anymore. The question then is is this worth a health check? I would expect users to be broken if they tried this at any given time and then they would revert the change.

evgeni: probably, yeah

nuke it

evgeni (approved):

PR #690, opened 2023-02-28
Add enabling satellite-maintenance if present

ehelms:
When performing a self-upgrade for Satellite, it should first check if satellite-maintenance:el8 is present and enabled. If present, it should enable it before upgrading.

PR #716, opened 2023-04-04
Add enabling satellite-maintenance if present (#690)

ehelms:
* Add enabling satellite-maintenance if present

When performing a self-upgrade for Satellite, it should first check if satellite-maintenance:el8 is present and enabled. If present, it should enable it before upgrading.

---------

Co-authored-by: Evgeni Golov <evgeni@golov.de>
(cherry picked from commit f5002fc640574c7337b61820cca44a5004f76403)

### DIFF
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -58,7 +58,9 @@ module ForemanMaintain::Scenarios
       manual_detection
     end
 
-    def compose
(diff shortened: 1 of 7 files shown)


.bundle/config
Odilhao, line 1: Do we need this bundle_path?

ehelms: Thanks for noticing, cleaned up my mess.

PR #719, opened 2023-04-05
Fixes #36282: Pre-download packages before stopping services

ehelms:
In upgrade scenarios, update packages using --downloadonly option with dnf to stage the packages without having to stop services. Then stop services and actually perform the update. This will cut down on the downtime during the upgrade process.

### DIFF
--- a/definitions/procedures/packages/update.rb
+++ b/definitions/procedures/packages/update.rb
@@ -27,7 +27,11 @@ module Procedures::Packages
     end
 
     def description
-      "Update package(s) #{@packages.join(', ')}"
+      if @yum_options.include?('--downloadonly')
+        "Download package(s) #{@packages.join(', ')}"
+      else
+        "Update package(s) #{@packages.join(', ')}"
+      end
     end
   end
 end
--- a/definitions/scenarios/upgrade_to_capsule_6_14.rb
+++ b/definitions/scenarios/upgrade_to_capsule_6_14.rb
@@ -40,7 +40,6 @@ module Scenarios::Capsule_6_14
 
     def compose
       add_steps(find_procedures(:pre_migrations))
-      add_step(Procedures::Service::Stop.new)
     end
   end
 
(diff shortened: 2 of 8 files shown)


definitions/procedures/packages/update.rb
```diff
@@ -27,7 +27,11 @@ def necessary?
     end
 
     def description
-      "Update package(s) #{@packages.join(', ')}"
+      if @yum_options.include?('--downloadonly')
```
ekohl, line 30: It surprises me that there's a `yum_options` parameter to the generic `Update` procedure. My immediate question is whether it also applies those on DNF.

test/test_helper.rb
```diff
@@ -10,7 +10,7 @@
 Mocha.configure do |c|
   c.strict_keyword_argument_matching = true
 end
-Minitest::Reporters.use!
+Minitest::Reporters.use! [Minitest::Reporters::DefaultReporter.new(:color => true)]
```
ekohl, line 13: This looks unrelated. Can this be separated, at least in a different commit?

definitions/procedures/packages/update.rb
ehelms: > It surprises me that there's a `yum_options` parameter to the generic `Update` procedure. My immediate question is whether it also applies those on DNF.

It does, as the DNF package manager inherits the yum one. I tested this on a real system.

test/test_helper.rb
ehelms: Done.

definitions/scenarios/upgrade_to_capsule_6_14.rb
```diff
@@ -58,7 +57,9 @@ def compose
       add_step(Procedures::Repositories::Setup.new(:version => '6.14'))
       modules_to_enable = ["satellite-capsule:#{el_short_name}"]
       add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
-      add_step(Procedures::Packages::UnlockVersions.new)
```
ekohl, line 61: Is unlocking no longer needed?

ehelms: Unlocking is designed to happen whenever an action requires it, explicitly unlocking here serves no purpose and further wastes time.

evgeni (approved):

PR #721, opened 2023-04-12
Fixes #33947 - only check directory permissions for backups with dumps

evgeni:
offline backups are performed as root and don't need write permissions for the postgres user


ekohl (commented):
Can't we avoid the whole permission problem by using `cat` which runs as root and then pipe to `runuser`?

ehelms:
@evgeni can you address the question left in the comment?

evgeni:
Technically, I guess yes. But I have no idea what performance implications this might have as now we'd be writing to "memory" and then cat writes to disk

ehelms (approved):
If this solves the problem, I'm good with it.

PR #722, opened 2023-04-12
Fix intermittent test failure

ehelms:
On some test runs this test fail like so:

```
 FAIL ForemanMaintain::Cli::UpgradeCommand::run#test_0008_with --phase it runs only a specific phase of the upgrade (0.07s)
        --- expected
        +++ actual
        @@ -1,3 +1,3 @@
         "Checking for new version of rubygem-foreman_maintain...
        -Nothing to update, can't find new version of rubygem-foreman_maintain.
        +undefined method `module_exists?' for #<ForemanMaintain::PackageManager::Apt:0xXXXXXX>
         "
        /home/ehelms/workspace/upstream/installer/foreman_maintain/test/test_helper.rb:22:in `assert_cmd'
        /home/ehelms/workspace/upstream/installer/foreman_maintain/test/lib/cli/upgrade_command_test.rb:234:in `block (3 levels) in <module:ForemanMaintain>'
```

### DIFF
--- a/lib/foreman_maintain.rb
+++ b/lib/foreman_maintain.rb
@@ -180,10 +180,11 @@ module ForemanMaintain
       puts "Checking for new version of #{package_name}..."
 
       enable_maintenance_module
+      package_manager = ForemanMaintain.package_manager
 
-      if ForemanMaintain.package_manager.update_available?(main_package_name)
+      if package_manager.update_available?(main_package_name)
         puts "\nUpdating #{package_name} package."
-        ForemanMaintain.package_manager.update(main_package_name, :assumeyes => true)
+        package_manager.update(main_package_name, :assumeyes => true)
         puts "\nThe #{package_name} package successfully updated."\
              "\nRe-run #{command} with required options!"
         exit 75
--- a/lib/foreman_maintain/package_manager.rb
+++ b/lib/foreman_maintain/package_manager.rb
@@ -5,12 +5,12 @@ require 'foreman_maintain/package_manager/apt'
 
 module ForemanMaintain
   def self.package_manager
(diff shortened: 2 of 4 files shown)


evgeni:
I don't think this is the correct "fix". The failure hints that we (*in some cases*) still try to call `module_exists?` on the `Apt` manager, which obviously doesn't support this.

ehelms:
> I don't think this is the correct "fix". The failure hints that we (_in some cases_) still try to call `module_exists?` on the `Apt` manager, which obviously doesn't support this.

Yes, I was trying to just fix that test because this is a rare, test environment only issue that causes things to fail. Here is a more direct to the problem fix. The general problem is that because we cache the package_manager instance, which in production makes sense given we are not jumping around OSes, that it can get cached in our test suite and then the logic for el? may not match the cached package manager.

We could opt to stop caching the package_manager during a run. I wasn't clear the performance impact for multiple operations in something like upgrades.


Here is the failure seed:
```
bundle exec rake test TESTOPTS="--seed=52677"
```

ehelms:
> I think my biggest worry is that this means that our tests don't _actually_ test how things are executed on a real system.
> 
> The clear cache trick in tests seems OK to me as a bandaid (more than the `foreman_maintain_update_unavailable`), but I'd really love us to have a way to say `on supported_oses` like we do in puppet…

I could probably derive something here. This should have been done when we added Debian support, but is something we can try and back-port on for features we are enabling on Debian.

ehelms:
@evgeni This is not exactly what you asked for, but I think it's a step in the direction to be able to achieve that while solving the intermittent test failure.

PR #723, opened 2023-04-12
Change --whitelist to --skip

ehelms:
The whitelist option doesn't accurately capture what will happen when steps are supplied to the option. The change to --skip can help to better understand that steps included with it will be skipped.


lib/foreman_maintain/cli/upgrade_command.rb
```diff
@@ -44,7 +44,7 @@ def upgrade_runner
         validate_target_version!
         @upgrade_runner = ForemanMaintain::UpgradeRunner.new(target_version, reporter,
           :assumeyes => assumeyes?,
-          :whitelist => whitelist || [],
+          :whitelist => skip || [],
```
ekohl, line 47: Should it also internally refactor?

ekohl (commented):
:+1: for renaming because it really is a `blacklist`, rather than a `whitelist`. Still, `skip` is an even better description.

ekohl (commented):
Would it make sense to keep `--whitelist` but issue a deprecation warning?

PR #728, opened 2023-05-08
Refactor pulpcore-manager call to a function

ehelms:


ekohl (approved):
I had been thinking about including the `execute!` method, but either way is fine by me.

ehelms:
> > @ekohl @sayan3296 small re-factor to help in the future if new commands get added
> 
> Looks neat. I was scrolling up the old PR and now see your suggestion about doing something like that.
> 
> Two questions for my own clarity :
> 
> * Should this be a Refs to #36365  ?

I do not think so. The re-factor was inspired by but not required by the issue.

> * Would this [test](https://github.com/theforeman/foreman_maintain/blob/master/test/definitions/procedures/pulpcore/trim_rpm_changelogs_test.rb#L15-L19) require an update ?

Nope. As this is a re-factor of internals, we should see no test breakages as that tells us that we changed the internals but the expectations remained the same.

PR #730, opened 2023-05-17
Fix foreman service being enabled after disable

ehelms:
The conditional is no longer needed since having Puma as our app server for foreman. This code path introduced a bug where once foreman was disabled it would never be enabled again.

### DIFF
--- a/definitions/features/foreman_server.rb
+++ b/definitions/features/foreman_server.rb
@@ -10,11 +10,7 @@ module ForemanMaintain
       end
 
       def services
-        if execute?('systemctl is-enabled foreman')
-          [system_service('foreman', 30, :socket => 'foreman')]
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (approved):
Followed steps in the BZ [here](https://bugzilla.redhat.com/show_bug.cgi?id=1995783) with this build.

Result after reboot:
```
● foreman.service - Foreman
   Loaded: loaded (/usr/lib/systemd/system/foreman.service; enabled; vendor pre>
  Drop-In: /etc/systemd/system/foreman.service.d
           └─installer.conf
   Active: active (running) since Wed 2023-05-17 14:14:48 EDT; 2min 36s ago
     Docs: https://theforeman.org
```

PR #732, opened 2023-05-20
Convert all code to use DNF

ehelms:


ehelms:
@evgeni When you get a moment, I'd like to get this in at the start of the dev cycle for the next Foreman release.

evgeni (changes requested):
makes sense overall

evgeni (approved):

PR #735, opened 2023-05-26
BZ#2192841 - only protect a select list of dnf commands

evgeni:
before:

```console
# dnf search vim
Reading Foreman protector configuration
*** Excluded total: 42168

WARNING: Excluding 42168 packages due to foreman-protector. 

=== Name & Summary Matched: vim ===
vim-minimal.x86_64 : A minimal version of the VIM editor
```

after:

```console
# dnf search vim
=== Name & Summary Matched: vim ===
vim-X11.x86_64 : The VIM version of the vi editor for the X Window System - GVim
vim-common.x86_64 : The common files needed by any version of the VIM editor
vim-enhanced.x86_64 : A version of the VIM editor which includes recent enhancements
vim-filesystem.noarch : VIM filesystem layout
vim-minimal.x86_64 : A minimal version of the VIM editor

# dnf install vim
Reading Foreman protector configuration
*** Excluded total: 42168

WARNING: Excluding 42168 packages due to foreman-protector. 

All matches were filtered out by exclude filtering for argument: vim
Error: Unable to find a match: vim
```


extras/foreman_protector/dnf/foreman-protector.py
```diff
@@ -53,6 +55,8 @@ def _add_obsoletes(self):
         return final_query
 
     def sack(self):
+        if self.cli is not None and self.cli.command._basecmd not in PROTECT_COMMANDS:
```
evgeni, line 58: the `self.cli is not None` part is required so that e.g. Ansible (which uses the library, not the cli) still honors the excludes.

extras/foreman_protector/dnf/foreman-protector.py
```diff
@@ -4,6 +4,8 @@
 
 import configparser
 
+PROTECT_COMMANDS = ('install', 'downgrade', 'reinstall', 'distro-sync', 'swap', 'upgrade', 'upgrade-minimal')
```
evgeni, line 7: tbh not sure about the commands here. given `f-m` only has wrappers for install and upgrade, the others *seem* wrong here, but then again they too can introduce issues…

extras/foreman_protector/dnf/foreman-protector.py
lpramuk: isn't it due to safe access to `self.cli.command._basecmd` ? (to avoid `None. _basecmd` )

evgeni: kinda… dnf is smart enough to not choke on a failing plugin here, so instead of erroring out with an exception, the filtering just doesn't happen and ansible can install things w/o our supervision.

ehelms:
Not sure if its possible, or another way to attack the problem - I feel like it would be nice if it told you when running a command if it was an excluded package. So, when you say, searched vim it indicated this is in the excluded package list so that there was less surprise when trying to install it.

evgeni:
A brief look at the search command didn't yield anything obvious how we could alter the output.

Also, keep in mind, the `search` above was just an example to show that it works. The original BZ was filed because `reposync` would be also affected by the exclude, and we don't want this.

PR #736, opened 2023-06-01
Fixes #36464 - always manage pulpcore services

ehelms:
Old, old code confined Pulp services to if they were enabled or not. However, if the 'service disable' command is used those services are naturally disabled and now no longer manageable by the service command. This also cleans up some old bits that no longer apply to the Pulp architecture.

### DIFF
--- a/definitions/features/pulpcore.rb
+++ b/definitions/features/pulpcore.rb
@@ -5,17 +5,13 @@ class Features::Pulpcore < ForemanMaintain::Feature
 
   metadata do
     label :pulpcore
-
-    confine do
-      ForemanMaintain::Utils::Service::Systemd.new('pulpcore-api', 0).enabled?
-    end
   end
 
   def services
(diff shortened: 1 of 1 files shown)


ehelms:
@Griffin-Sullivan here's another one that should look familiar

definitions/features/pulpcore.rb
```diff
@@ -50,16 +37,4 @@ def self.pulpcore_common_services
     ]
     common_services + pulpcore_resource_manager_service
```
evgeni, line 38: you can't delete `pulpcore_resource_manager_service` below, but still reference it here ;-)

evgeni (approved):
is there a BZ/redmine for that?

ehelms:
> is there a BZ/redmine for that?

There is a BZ -- https://bugzilla.redhat.com/show_bug.cgi?id=2055790

PR #738, opened 2023-06-06
Fixes #36480 - add restore_dir option to backup_local

martin-schlossarek:
Without a restore_dir option, snapshot backups cannot be restored with "foreman-maintain restore".

### DIFF
--- a/definitions/procedures/backup/offline/candlepin_db.rb
+++ b/definitions/procedures/backup/offline/candlepin_db.rb
@@ -36,7 +36,8 @@ module Procedures::Backup
             pg_backup_file,
             :listed_incremental => File.join(@backup_dir, '.postgres.snar'),
             :volume_size => @tar_volume_size,
-            :data_dir => pg_data_dir
+            :data_dir => pg_data_dir,
+            :restore_dir => feature(:candlepin_database).data_dir
           )
         end
       end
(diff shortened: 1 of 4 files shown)


ehelms:
@evgeni Could you take a look at this?

evgeni:
I think the patch is correct for EL-based installs, but might break things on Debian where we might have multiple DB paths -- need to test this out.

(Also, I am not sure I like the name of the parameter, `restore_dir`, as at the time of its use we do not perform a restore but a backup. Maybe something like `archive_base_dir` or so is more appropriate, but I won't insist on the naming either way, just wanted to point it out)

evgeni:
Okay, on CentOS 8 this worked fine, on Debian, this produces a bad backup.

working one:
```
root@debian11-foreman-nightly:~# tar tvf /var/backup/foreman-backup-2023-08-15-12-12-28/pgsql_data.tar.gz | grep main/PG_VERSION
-rw------- postgres/postgres 3 2023-08-15 12:08 var/lib/postgresql/13/main/PG_VERSION
```

after this patch:
```
root@debian11-foreman-nightly:~# tar tvf /var/backup/foreman-backup-2023-08-15-12-14-28/pgsql_data.tar.gz | grep PG_VERSION
-rw------- postgres/postgres 3 2023-08-15 12:08 []PG_VERSION
```

Going to investigate how we can fix that nicely.

evgeni (approved):

PR #718, opened 2023-04-05
Fixes #36233 - Run container-repair-media-type script on upgrade

wbclark:
Follow up from #703

PR #740, opened 2023-06-15
Fixes #36233 - Run container-repair-media-type script on upgrade

Odilhao:
Fix one typo for server and capsule upgrade to 6.13 Change  ContainerRepairMediaType to RepairContainerMediaType


Odilhao:
Fix the typo mentioned by @lpramuk  at #718

wbclark:
actually, the actual pulpcore-manager command is called `pulpcore-manager container-repair-media-type` so I think we should keep it as `ContainerRepairMediaType`

PR #744, opened 2023-07-11
Fixes #36578 - ensure puppet/qpidd are present before restore

evgeni:


lib/foreman_maintain/utils/backup.rb
```diff
@@ -253,6 +253,18 @@ def incremental?
       def online_backup?
         !!metadata.fetch('online', false)
       end
+
+      def installed_rpms
+        metadata.fetch('rpms', metadata.fetch(:rpms, []))
```
evgeni, line 258: depending on the Ruby version used to generate the metadata, it will be either a string or a symbol :roll_eyes:

evgeni: actually, no, it's 0018ba9397130893413135b646f3cee2ed88e43d that broke the API and changed from string to symbol…

still :exploding_head:

ekohl (approved):

PR #751, opened 2023-08-07
Allow to define product support entity in features file

sbernhard:


definitions/features/instance.rb
```diff
@@ -28,6 +28,14 @@ def product_name
     end
   end
 
+  def product_support_entity
```
ehelms, line 31:
```suggestion
  def project_support_entity
```

ehelms: I like the change here! I forget things like that got hard-coded.

ehelms (approved):

PR #752, opened 2023-08-10
Fixes #36668 - only reindex DB when the OS changes

evgeni:


ehelms (approved):
@Griffin-Sullivan @jameerpathan111 would one of y'all take a look?

Griffin-Sullivan (approved):
LGTM

PR #757, opened 2023-09-04
Fix #36716 - Add check for missing product-content

m-bucher:
source https://github.com/ATIX-AG/orcharhino-scripts/tree/main/find_missing_candlepin_product_contents


ekohl (commented):
I'm generally not a fan of SQL in a tool. Could the same information be retrieved from Candlepin using the official API? If so, could Katello do so and implement it as a rake command?

I'm mostly worried about Candlepin version updates which changes the DB somehow and verifying this part can easily be forgotten.

We don't have a good group to ping, but @ianballou may know or may know who can make it happen.

ianballou:
I agree with @ekohl that this is a little strange to have in foreman-maintain. It would be really perfect to have it exist as a rake script in Katello.

However, this script does look like it would be incredibly helpful to have around for when users have issues with repositories not appearing in their `redhat.repo` file.

It looks like you can get product content from products at least via the Candlepin API:

```
{
  "created": "2020-04-23 03:23:50.927000",
  "updated": "2020-04-23 03:23:50.927000",
  "id": "5051",
  "uuid": "ff808081554a3e4101554a3e9033005d",
  "name": "Admin OS Developer Bits",
  "multiplier": 1,
  "attributes": [
    {
      "name": "string",
      "value": "string"
    }
  ],
  "productContent": [
    {
      "content": {
        "created": "2020-04-23 03:23:50.927000",
        "updated": "2020-04-23 03:23:50.927000",
        "uuid": "ff808081554a3e4101554a3e9033005d",
        "id": "5001",
        "type": "yum",
        "label": "content_label",
        "name": "content_name",
        "vendor": "example-vendor",
        "contentUrl": "/admin/foo/example/path",
        "requiredTags": "TAG1,TAG2",
        "releaseVer": "1.2.3",
        "gpgUrl": "/admin/foo/example/gpg/path",
        "modifiedProductIds": [
          5051,
          5052,
          5053
        ],
        "arches": "x86_64",
        "metadataExpire": 1467124079
      },
      "enabled": true
    }
  ],
...
```

If rewriting with the Candlepin API ends up being too much to prioritize, I'd be for keeping it as-is

m-bucher:
@ianballou @ekohl , I created a rake-task, which does similar things: https://github.com/Katello/katello/pull/10740

At the moment I do not have a system that really has the issue this is intended to fix.

PR #762, opened 2023-09-14
Check if CONFIG_FILE is defined

ehelms:


evgeni:
In which case will that be undefined?

(I was looking at the same part of code the other day, as I wanted to be able to override the config from `ENV` or something, but didn't see a case where it can be undefined).

ehelms:
> In which case will that be undefined?
> 
> (I was looking at the same part of code the other day, as I wanted to be able to override the config from `ENV` or something, but didn't see a case where it can be undefined).

I ran into this running code locally, trying to call parts of the stack to generate lists of items and realized this had a strong expectation that this exists instead of either defining a default CONFIG_FILE or checking for existence.

evgeni (approved):

PR #766, opened 2023-09-16
drop old and unused DB backup code

evgeni:
the last use of it was removed in 5c2a705e18a9ec26d5461cdd7e1b211cd06d9971

this means we can also finally drop the "can postgres user access" check during backup


ekohl (approved):
Hard to review, but it looks ok. I also noticed `delete_records_by_ids` is unused since 5c2a705e18a9ec26d5461cdd7e1b211cd06d9971.

evgeni:
> Hard to review, but it looks ok. I also noticed `delete_records_by_ids` is unused since [5c2a705](https://github.com/theforeman/foreman_maintain/commit/5c2a705e18a9ec26d5461cdd7e1b211cd06d9971).

and I found another, so here is https://github.com/theforeman/foreman_maintain/pull/772

PR #771, opened 2023-09-20
only collect installer metadata in backup if installer present

evgeni:
there are setups which do not use the installer, and we should be able to backup those too


definitions/procedures/backup/metadata.rb
```diff
@@ -65,8 +65,10 @@ def proxy_feature_list(spinner)
 
     def proxy_config(spinner)
       spinner.update('Collecting proxy configuration')
```
ekohl, line 67: Should this also be inside the `if` statement, just like all the other methods?

evgeni: I guess yeah

ekohl (approved):

PR #773, opened 2023-09-25
Add final upgrade step to check if reboot needed

ehelms:
This is intended to automate into the procedure and be able to drop from our documentation running the command explicitly. I have included example outputs below. My only concern is whether this is too "hidden" of output for the user and if instead we need more of special post upgrade output (similar to the installer) that clearly articulates any upgrade considerations.

Output if reboot needed:

```
Running Checks after upgrading to Satellite 6.15.z
Clean old Kernel and initramfs files from tftp-boot:                  [OK]
Check if only installed assets are present on the system: 
| Checking for presence of non-original assets...                     [OK]      
Check if system needs reboot: 
Updating Subscription Management repositories.
Core libraries or services have been updated since boot-up:
  * dbus
  * dbus-daemon
  * kernel
  * linux-firmware
  * systemd

Reboot is required to fully utilize these updates.
More information: https://access.redhat.com/solutions/27943           [OK]


Upgrade finished.
```

Example if no reboot:

```
Running Checks after upgrading to Satellite 6.15.z
Clean old Kernel and initramfs files from tftp-boot:                  [OK]
Check if only installed assets are present on the system: 
| Checking for presence of non-original assets...                     [OK]      
Check if system needs reboot: 
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
More information: https://access.redhat.com/solutions/27943           [OK]


Upgrade finished.
```

### DIFF
--- a/definitions/procedures/packages/check_for_reboot.rb
+++ b/definitions/procedures/packages/check_for_reboot.rb
@@ -1,0 +1,16 @@
+module Procedures::Packages
+  class CheckForReboot < ForemanMaintain::Procedure
+    metadata do
+      description 'Check if system needs reboot'
+    end
+
+    def run
+      status, output = execute_with_status('dnf needs-restarting --reboothint')
+      if status == 1
+        set_status(:warning, output)
+      else
+        set_status(:success, output)
+      end
+    end
+  end
+end
--- a/definitions/scenarios/upgrade_to_capsule_6_15_z.rb
+++ b/definitions/scenarios/upgrade_to_capsule_6_15_z.rb
@@ -87,6 +87,7 @@ module Scenarios::Capsule_6_15_z
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:post_upgrade))
+      add_step(Procedures::Packages::CheckForReboot)
     end
   end
 end
--- a/definitions/scenarios/upgrade_to_capsule_6_15.rb
+++ b/definitions/scenarios/upgrade_to_capsule_6_15.rb
@@ -87,6 +87,7 @@ module Scenarios::Capsule_6_15
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:post_upgrade))
+      add_step(Procedures::Packages::CheckForReboot)
     end
   end
 end
--- a/definitions/scenarios/upgrade_to_satellite_6_15.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_15.rb
@@ -89,6 +89,7 @@ module Scenarios::Satellite_6_15
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:post_upgrade))
+      add_step(Procedures::Packages::CheckForReboot)
     end
   end
 end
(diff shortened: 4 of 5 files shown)


definitions/scenarios/upgrade_to_capsule_6_15_z.rb
```diff
@@ -87,6 +87,7 @@ class PostUpgradeChecks < Abstract
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:post_upgrade))
+      add_step(Procedures::Packages::CheckForReboot)
```
evgeni, line 90: why isn't the step just tagged with `post_upgrade`?

ehelms: I want it to be the final step so the user sees it.

definitions/procedures/packages/check_for_reboot.rb
```diff
@@ -0,0 +1,18 @@
...
+  class CheckForReboot < ForemanMaintain::Procedure
+    metadata do
+      description 'Check if system needs reboot'
+    end
+
+    def run
+      status, output = execute_with_status('dnf needs-restarting --reboothint')
+      puts output
+    rescue ForemanMaintain::Error::ExecutionError => e
+      if @warn_on_errors
```
evgeni, line 11: I don't see `@warn_on_errors` being defined anywhere here. so it will always be `nil`

evgeni (approved):

definitions/procedures/packages/check_for_reboot.rb
```diff
@@ -0,0 +1,14 @@
+module Procedures::Packages
+  class CheckForReboot < ForemanMaintain::Procedure
+    metadata do
+      description 'Check if system needs reboot'
+    end
+
+    def run
+      _status, output = execute_with_status('dnf needs-restarting --reboothint')
+      puts output
```
evgeni, line 9: if a reboot is required, _status will be 1 (instead of 0), can we mark the step as "warn" instead of "OK" then?

```suggestion
      status, output = execute_with_status('dnf needs-restarting --reboothint')
      if status != 0:
        set_warn(output)
      else
        puts output
      end
```

ehelms:
Here is the updated output:

```
Check if system needs reboot:                                         [WARNING]
Updating Subscription Management repositories.
Core libraries or services have been updated since boot-up:
  * dbus
  * dbus-daemon
  * glibc
  * kernel
  * linux-firmware
  * systemd

Reboot is required to fully utilize these updates.
More information: https://access.redhat.com/solutions/27943
Scenario [Checks after upgrading to Satellite 6.15.z] failed.

The following steps ended up in warning state:

  [packages-check-for-reboot]

The steps in warning state itself might not mean there is an error,
but it should be reviewed to ensure the behavior is expected
```

I am not the biggest fan of that "what does warning mean" output, but we can live with it.

evgeni (approved):

ehelms:
I learned that this results in an exit code `78` as that is what we issue for a warning state, I wonder if that will result in this being interpreted as failed by users / scripts?

evgeni:
If they take notice, this is what we wanted, no? :)

PR #782, opened 2023-10-23
Fixes #36877 - Add a pre-upgrade check for katello-agent

wbclark:


definitions/checks/katello_agent.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  INSTALLER_ANSWERS_FILE = '/etc/foreman-installer/scenarios.d/satellite-answers.yaml'.freeze
+
+  def run
+    assert(
+      !katello_agent_enabled?,
+      "The katello-agent feature is enabled on this system. As of Satellite 6.15.0, katello-agent"\
+      " is removed and will no longer function. Before proceeding with the upgrade, you should ensure"\
+      " that you have deployed and configured an alternative tool for remote package management and"\
+      " patching for content hosts, such as the Remote Execution (REX) Pull Provider. See the"\
```
jeremylenz, line 22: Since technically it's not a "transport" any more, they standardized on different terms in the [documentation](https://docs.theforeman.org/nightly/Installing_Proxy/index-katello.html#enabling-remote-execution_smart-proxy):

```suggestion
      " patching for content hosts, such as remote execution (REX) with pull-based transport. See the"\
```

jeremylenz (commented):
Thanks @wbclark!

definitions/checks/katello_agent.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  private
+
+  def katello_agent_enabled?
+    return false unless File.exist?(INSTALLER_ANSWERS_FILE)
+
+    answers = YAML.load_file(INSTALLER_ANSWERS_FILE)
+
+    answers['foreman_proxy_content']['enable_katello_agent'] == true
```
evgeni, line 37:
```suggestion
    feature(:installer).answers['foreman_proxy_content']['enable_katello_agent']
```

we have a feature for the installer, and that knows how to load answers, and comparing to `true` is unnecessary

definitions/checks/katello_agent.rb
```diff
@@ -0,0 +1,39 @@
...
+    label :check_katello_agent_enabled
+    description 'Check whether the katello-agent feature is enabled before upgrading'
+    tags :pre_upgrade
+
+    confine do
+      feature(:instance).downstream
+    end
+  end
+
+  INSTALLER_ANSWERS_FILE = '/etc/foreman-installer/scenarios.d/satellite-answers.yaml'.freeze
```
evgeni, line 14:
```suggestion
```

see below.

definitions/checks/katello_agent.rb
```diff
@@ -0,0 +1,39 @@
+require 'yaml'
+
+class Checks::CheckKatelloAgentEnabled < ForemanMaintain::Check
+  metadata do
+    label :check_katello_agent_enabled
+    description 'Check whether the katello-agent feature is enabled before upgrading'
+    tags :pre_upgrade
+
+    confine do
+      feature(:instance).downstream
```
evgeni, line 10: why is that confined to downstream?

also, should that run on proxies, or just on the server?

definitions/checks/katello_agent.rb
```diff
@@ -0,0 +1,39 @@
...
+    assert(
+      !katello_agent_enabled?,
+      "The katello-agent feature is enabled on this system. As of Satellite 6.15.0, katello-agent"\
+      " is removed and will no longer function. Before proceeding with the upgrade, you should"\
+      " ensure that you have deployed and configured an alternative tool for remote package"\
+      " management and patching for content hosts, such as Remote Execution (REX) with pull-based"\
+      " transport. See the Managing Hosts guide in the Satellite documentation for more info."\
+      " Disable katello-agent with the command"\
+      " `satellite-installer --foreman-proxy-content-enable-katello-agent false`"\
+      " before proceeding with the upgrade."
```
evgeni, line 26: This should probably be written in a way that also applies to upstream?

jeremylenz: I thought upstream users don't use foreman-maintain? Does foreman-maintain have a downstream repo that can be branded?

evgeni: no, this repo is capable of branding itself. like `feature(:installer).installer_command` will give you foreman-installer/satellite-installer as needed.

and while upstream users don't use f-m much for upgrades, the code is capable of doing it (mostly) and we should strive not to break that.

jeremylenz: oh, and the `satellite-installer` command :yum:

evgeni: It can't do the version translation, but I also think it doesn't have to?
This check makes only sense at two points:
Katello 4.9 → 4.10 upgrade, Satellite 6.14 → 6.15 upgrade, right?
So you could just write "The katello-agent feature is enabled on this system, but it will be removed during the upgrade".

definitions/checks/katello_agent.rb
```diff
@@ -0,0 +1,39 @@
...
+  def run
+    assert(
+      !katello_agent_enabled?,
+      "The katello-agent feature is enabled on this system. As of Satellite 6.15.0, katello-agent"\
+      " is removed and will no longer function. Before proceeding with the upgrade, you should"\
+      " ensure that you have deployed and configured an alternative tool for remote package"\
+      " management and patching for content hosts, such as Remote Execution (REX) with pull-based"\
+      " transport. See the Managing Hosts guide in the Satellite documentation for more info."\
+      " Disable katello-agent with the command"\
+      " `satellite-installer --foreman-proxy-content-enable-katello-agent false`"\
```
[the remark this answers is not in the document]
jeremylenz: I think our idea was to stop the upgrade _on purpose_, even though technically it may not be strictly required. This way the user would wake up and say, wait a minute, I need to make sure all my ducks are in a row for host patching before I upgrade!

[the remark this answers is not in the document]
jeremylenz: after all, our user story does say

"As a user with katello-agent enabled, I am prevented from upgrading to Satellite 6.15 / Katello 4.10 until I disable katello-agent" :)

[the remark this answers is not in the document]
evgeni: Mhh. Installer runs are rather expensive and I'd love to avoid them.

Would your story be also happy if we warn the user that they will lose the feature and they have to accept that by pressing "yes"?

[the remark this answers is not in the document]
myee111: > Mhh. Installer runs are rather expensive and I'd love to avoid them.
> 
> Would your story be also happy if we warn the user that they will lose the feature and they have to accept that by pressing "yes"?

I think this is fine.

definitions/checks/katello_agent.rb
wbclark: Regarding whether it should be run on proxies, I guess it's not necessary, because the installer migration already disables the feature and the puppet modules already perform the cleanup. The purpose of the check is to be an informative courtesy reminder to users, so running the check on each proxy seems excessive.

My understanding of foreman_maintain is that it's not used much by upstream, so I originally restricted this check to run downstream only in case we want to directly link any downstream documentation. I could lift this restriction if it's desired.

definitions/checks/katello_agent.rb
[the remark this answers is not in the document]
wbclark: I am curious to know what @jeremylenz thinks about this as well.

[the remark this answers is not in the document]
wbclark: An interactive choice seems fine as well. Let me research the best way to implement it.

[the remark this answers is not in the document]
evgeni: I am quite sure we had similar in the past, if you don't find something, I can have a look tomorrow morning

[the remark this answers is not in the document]
jeremylenz: Closing the loop: let's keep this the way it is, as discussed offline. Expensive installer run that adds a few minutes is better than being able to proceed thru an installation that breaks host patching unbeknownst to you.

definitions/checks/katello_agent.rb
ehelms: I agree, by the time the server is upgraded, katello-agent support is gone already and the proxies don't matter anymore.

definitions/checks/katello_agent.rb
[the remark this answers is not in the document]
ehelms: Do you want this to be skippable? Users can whitelist this step and get the automatic removal. That seems fine but I want to make sure that matches your desired behaviour.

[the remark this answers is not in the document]
wbclark: I added some language explaining what will happen if the user uses `--whitelist` to skip the check, which indeed should be the fastest way to proceed if they elect to do so.

definitions/checks/katello_agent.rb
```diff
@@ -0,0 +1,41 @@
...
+      " #{instance_version}, katello-agent is removed and will no longer function."\
+      " Before proceeding with the upgrade, you should ensure that you have deployed"\
+      " and configured an alternative tool for remote package management and patching"\
+      " for content hosts, such as Remote Execution (REX) with pull-based transport."\
+      " See the Managing Hosts guide in the #{instance_name} documentation for more info."\
+      " Disable katello-agent with the command"\
+      " `#{installer_command} --foreman-proxy-content-enable-katello-agent false`"\
+      " before proceeding with the upgrade. Alternatively, you may skip this check and proceed by"\
+      " running #{maintain_command} again with the `--whitelist` option, which will automatically"\
+      " uninstall katello-agent."
```
jeremylenz, line 32: As someone new to foreman-maintain, I'm still not sure what command it's telling me to run here. I assume it's not just `foreman-maintain --whitelist`? Can it give me the exact command I would run?

wbclark: I need to figure out how the name to pass to `--whitelist` is constructed from the label `:check_katello_agent_enabled`

wbclark: OK, so it would be `--whitelist="check-katello-agent-enabled" but it turns out the exact language shouldn't be necessary here.

The reason why is that if ANY checks fail, there is general language explaining how to whitelist the failed step.

jeremylenz (approved):
In that case, LGTM :+1:

chris1984 (approved):

ehelms (approved):

PR #792, opened 2023-12-14
Update to next version of upgrade scenarios, bump to 1.5.0

ehelms:


Griffin-Sullivan (approved):
Looks good to me! Noob question: Is there no way to set this as a constant so you only have to update in one place? Could also be done when branching.

ehelms:
> Looks good to me! Noob question: Is there no way to set this as a constant so you only have to update in one place? Could also be done when branching.

Not a noob question -- see https://github.com/theforeman/foreman_maintain/pull/671

PR #793, opened 2024-01-19
Fixes #37114: Download packages during upgrade before maintenance mode 

ehelms:


ehelms:
When testing a Satellite update with no package updates, the commit I added to only refresh cache on download step shaves 25 seconds off the downtime of the update run.

ehelms:
On another note, because we stop services, an update with no packages having changed (so bare minimum downtime) results (on my system) in the installer taking ~3 minutes of the 5 minutes of downtime. The primary culprits:

```
[28.25 seconds] /Service[foreman]: 
[27.37 seconds] /Service[dynflow-sidekiq@orchestrator]: 
[26.42 seconds] /Service[dynflow-sidekiq@worker-hosts-queue-1]: 
[26.32 seconds] /Service[dynflow-sidekiq@worker-1]: 
[18.67 seconds] /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]: 
```

I'll be curious if the change to Ruby 3+ has any direct impact on this.

definitions/procedures/packages/update.rb
```diff
@@ -11,7 +11,7 @@ class Update < ForemanMaintain::Procedure
 
     def run
       assumeyes_val = @assumeyes.nil? ? assumeyes? : @assumeyes
-      package_manager.clean_cache(:assumeyes => assumeyes_val)
+      package_manager.clean_cache(:assumeyes => assumeyes_val) if download_only?
```
evgeni, line 14: IMHO we should rather fix the definition of `clean_cache` in https://github.com/theforeman/foreman_maintain/blob/fa2e3cb52fc6d224124bd9d3e9d1c888578625a8/lib/foreman_maintain/package_manager/dnf.rb#L94-L96

`clean metadata` should be usually sufficient as it cleans everything but packages: https://github.com/rpm-software-management/dnf/blob/b5d25c075ba12d7543a4fb6ac39c6b20e5e5088a/dnf/cli/commands/clean.py#L40-L46

evgeni: Or maybe both. Re-reading your explanation above makes sense not to clean the cache "inbetween", but even for the non-inbetween case we should not `clean all` IMHO

evgeni:
> On another note, because we stop services, an update with no packages having changed (so bare minimum downtime) results (on my system) in the installer taking ~3 minutes of the 5 minutes of downtime. The primary culprits:
> 
> ```
> [28.25 seconds] /Service[foreman]: 
> [27.37 seconds] /Service[dynflow-sidekiq@orchestrator]: 
> [26.42 seconds] /Service[dynflow-sidekiq@worker-hosts-queue-1]: 
> [26.32 seconds] /Service[dynflow-sidekiq@worker-1]: 
> [18.67 seconds] /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]: 
> ```
> 
> I'll be curious if the change to Ruby 3+ has any direct impact on this.

Just to spell out what we talkted about: I think we can omit restarting `dynflow-sidekiq` by Puppet as it is part of `foreman.service` and already got restarted, but Puppet probably doesn't know that.

ehelms:
> > On another note, because we stop services, an update with no packages having changed (so bare minimum downtime) results (on my system) in the installer taking ~3 minutes of the 5 minutes of downtime. The primary culprits:
> > ```
> > [28.25 seconds] /Service[foreman]: 
> > [27.37 seconds] /Service[dynflow-sidekiq@orchestrator]: 
> > [26.42 seconds] /Service[dynflow-sidekiq@worker-hosts-queue-1]: 
> > [26.32 seconds] /Service[dynflow-sidekiq@worker-1]: 
> > [18.67 seconds] /Stage[main]/Foreman::Database/Foreman::Rake[db:migrate]/Exec[foreman-rake-db:migrate]: 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'll be curious if the change to Ruby 3+ has any direct impact on this.
> 
> Just to spell out what we talkted about: I think we can omit restarting `dynflow-sidekiq` by Puppet as it is part of `foreman.service` and already got restarted, but Puppet probably doesn't know that.

@ekohl Do you know if there is a way at the Puppet level to allow foreman to start before the dynflow service definition is evaluated so that a start is not triggered of the dynflow service? Given we connect things at the systemd level (https://github.com/theforeman/foreman/blob/develop/extras/systemd/dynflow-sidekiq%40.service#L5).  

Perhaps we just need to manage the file but not the state as we do today (https://github.com/theforeman/puppet-foreman/blob/master/manifests/dynflow/worker.pp#L49-L53) ?

ekohl:
> The primary culprits:

Just plain `foreman-rake -T` (list tasks) is already slow. We execute that at least once (to see if there are any migrations). If we can speed that up, that would help.

It may help to go through the various rake tasks that we define and perhaps make sure they aren't defined in production.

Note that https://guides.rubyonrails.org/active_record_migrations.html#running-migrations suggests to run `bin/rails db:migrate`. That ends up just being a wrapper around the rake task though, so it doesn't help at all with being faster.

ehelms:
Here is an idea I tested to solve the added time due to dynflow restarts:

https://github.com/theforeman/puppet-foreman/pull/1148

ehelms:
@evgeni Your re-review would be appreciated.

definitions/procedures/packages/update.rb
evgeni, line 14: I think gating this on `download_only` being true is wrong.

We use `Procedures::Packages::Update` in other places, that don't use `download_only` and thus would not get cache refreshes anymore:
https://github.com/theforeman/foreman_maintain/blob/fa2e3cb52fc6d224124bd9d3e9d1c888578625a8/definitions/scenarios/self_upgrade.rb#L67-L87
https://github.com/theforeman/foreman_maintain/blob/fa2e3cb52fc6d224124bd9d3e9d1c888578625a8/definitions/scenarios/upgrade_to_foreman_nightly.rb#L59

I think we need an own param that controls the cache cleanup, which defaults to `true` and then the "second" runs (after download_only) set it explicitly to `false`

definitions/scenarios/upgrade_to_capsule_6_16.rb
```diff
@@ -38,6 +38,15 @@ class PreMigrations < Abstract
     end
 
     def compose
+      add_step(Procedures::Repositories::Setup.new(:version => '6.16'))
+      modules_to_enable = ["satellite-capsule:#{el_short_name}"]
+      add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
+      add_step(Procedures::Packages::Update.new(
+        :assumeyes => true,
+        :clean_cache => false,
```
evgeni, line 46: shouldn't it be `true` (default) here, and then `false` in Line 66?

ehelms: Errr, yes.

definitions/scenarios/upgrade_to_foreman_nightly.rb
```diff
@@ -56,7 +56,7 @@ def compose
         modules_to_enable = ["foreman:#{el_short_name}"]
         add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
       end
-      add_step(Procedures::Packages::Update.new(:assumeyes => true))
+      add_step(Procedures::Packages::Update.new(:assumeyes => true, :clean_cache => false))
```
evgeni, line 59: I think this one should clean cache, as there is no dedicated `download_only` step (and we'd need to implement that for debian if we wanted it here)

evgeni:
LGTM besides the "foreman nightly" file.

evgeni (approved):

PR #794, opened 2024-01-25
Fixes #37096 - Add rubocop to github actions

dosas:


dosas:
I do not understand why ht erubocop target did not run foir this PR? 
Does it require configuration in the settings or does the file have to be on main?

sbernhard (approved):
LGTM

ehelms:
The current workflow runs `bundle exec rake` which captures rubocop and tests based on the default. If we want to break out the rubocop to it's own entry, I think we should update the workflow so that it only runs the unit tests alongside the change.

https://github.com/theforeman/foreman_maintain/blob/master/Rakefile#L23

ehelms:
> I do not understand why the rubocop target did not run foir this PR?
> Does it require configuration in the settings or does the file have to be on main?

The tests do not automatically run unless you are a public member of theforeman organization. I approved and ran it but see my comment.

ehelms (approved):

PR #795, opened 2024-02-06
Use rhel-8 chroot for packit

ehelms:


.packit.yaml
```diff
@@ -30,8 +30,8 @@ jobs:
   - job: copr_build
     trigger: pull_request
     targets:
-      centos-stream-8:
-        additional_modules: "foreman:el8,nodejs:12"
+      rhel-8:
+        additional_modules: "foreman:el8,nodejs:14"
```
evgeni, line 34:
```suggestion
        additional_modules: "foreman:el8"
```

We don't use nodejs (and if we'd did, we should use `foreman-devel:el8` module instead)

evgeni (changes requested):

evgeni (approved):

PR #796, opened 2024-02-07
Fixes #37149: Include certs tar in backup

ehelms:
I think this has not worked for a while :/


ehelms:
@evgeni Could I bother you for a review?

PR #798, opened 2024-02-14
Add procedure to remove corrupted host statuses

sbernhard:
From time to time we have some hosts, which can not be deleted because of errorrs like:

```
PG::ForeignKeyViolation: ERROR: update or delete on table "hosts" violates foreign key constraint "host_status_hosts_host_id_fk" on table "host_status" DETAIL: Key (id)=(HOST_ID) is still referenced from table "host_status".
```

This procedure is able to find and fix these issues.


definitions/procedures/foreman/remove_corrupted_host_statuses.rb
```diff
@@ -0,0 +1,43 @@
+module Procedures::Foreman
```
ehelms, line 1: This is not included in any scenarios - where are you wanting it to run?

sbernhard: Yes. Actually I was unsure if we want to have this to run on every upgrade or backup or other places. 
The procedures can be called as a single routine and I think it make sense to have a collection of these "maintenance" routines which can solve issues. What are your thougths?

PR #799, opened 2024-02-15
Combine Foreman and Katello upgrade scenarios into one

ehelms:


definitions/procedures/repositories/setup.rb
```diff
@@ -3,7 +3,7 @@ class Setup < ForemanMaintain::Procedure
     metadata do
       description 'Setup repositories'
       confine do
-        feature(:instance).downstream || feature(:upstream)
```
evgeni, line 6: this… was undefined? :)

definitions/scenarios/foreman_upgrade.rb
```diff
@@ -0,0 +1,145 @@
+module Scenarios::ForemanUpgrade
+  class Abstract < ForemanMaintain::Scenario
+    def self.upgrade_metadata(&block)
+      metadata do
+        tags :upgrade_scenario
+        confine do
+          feature(:foreman_install)
```
evgeni, line 7: you dropped the `|| ForemanMaintain.upgrade_in_progress == 'nightly'` is that because the "foreman_install" feature will always be there?

definitions/procedures/repositories/setup.rb
ehelms: Exactly.

definitions/scenarios/foreman_upgrade.rb
ehelms: Yea

evgeni (approved):

PR #801, opened 2024-02-16
Fixes #37172 - Use Run class for installer procedures

ehelms:
The Upgrade class was a leftover from when the installer had `--upgrade` option and never got cleaned up.


Griffin-Sullivan (approved):
Ran the upgrade on stream with packit and everything passed normally.

PR #802, opened 2024-02-16
Fixes #37175: Add dedicated update command

ehelms:
This changes adds a dedicated command for z-stream style updates and is intended to be run like:

```
foreman-maintain update run
```

This introduces a dedicated update scenario intended to work for all installation scenarios letting the various checks and processes handle the state of the system (e.g. what installer scenario is present). This does not remove using the upgrade command to do a z-stream "upgrade" as my goal is to introduce a new update command alongside these first before doing major demolition.

I am putting this in draft status for now in order to add testing.

### DIFF
--- a/definitions/scenarios/update.rb
+++ b/definitions/scenarios/update.rb
@@ -1,0 +1,126 @@
+module Scenarios::Update
+  class Abstract < ForemanMaintain::Scenario
+    def self.update_metadata(&block)
+      metadata do
+        tags :update_scenario
+        instance_eval(&block)
+      end
+    end
+  end
+
+  class PreUpdateCheck < Abstract
+    update_metadata do
+      description 'Checks before updating'
+      tags :pre_update_checks
+      run_strategy :fail_slow
+    end
+
+    # rubocop:disable Metrics/MethodLength
+    def compose
+      add_steps(
+        Checks::Foreman::FactsNames, # if Foreman database present
+        Checks::ForemanProxy::CheckTftpStorage, # if Satellite with foreman-proxy+tftp
+        Checks::ForemanProxy::VerifyDhcpConfigSyntax, # if foreman-proxy+dhcp-isc
+        Checks::ForemanTasks::NotPaused, # if foreman-tasks present
+        Checks::Puppet::VerifyNoEmptyCacertRequests, # if puppetserver
+        Checks::ServerPing,
+        Checks::ServicesUp,
+        Checks::SystemRegistration,
+        Checks::CheckHotfixInstalled,
+        Checks::CheckTmout,
+        Checks::CheckUpstreamRepository,
+        Checks::Disk::AvailableSpace,
+        Checks::Disk::AvailableSpaceCandlepin, # if candlepin
+        Checks::Foreman::ValidateExternalDbVersion, # if external database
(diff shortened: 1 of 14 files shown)


Griffin-Sullivan (approved):
Approving from the point of view that I tested this and it works, but can't speak to the code here.

lib/foreman_maintain/package_manager/dnf.rb
```diff
@@ -121,9 +121,13 @@ def info(name)
       dnf_action('module info', name, with_status: true, assumeyes: true)
     end
 
+    def package_version(package_name)
+      sys.execute!("rpm -q #{package_name} --qf '%{version}'").strip
+    end
```
evgeni, line 126: this already exists in `lib/foreman_maintain/concerns/system_helpers.rb`, no need to re-invent

definitions/scenarios/update.rb
```diff
@@ -0,0 +1,126 @@
...
+        Checks::Foreman::CheckCorruptedRoles,
+        Checks::Foreman::CheckDuplicatePermissions,
+        Checks::Foreman::TuningRequirements, # if katello present
+        Checks::ForemanOpenscap::InvalidReportAssociations, # if foreman-openscap
+        Checks::ForemanTasks::Invalid::CheckOld, # if foreman-tasks
+        Checks::ForemanTasks::Invalid::CheckPendingState, # if foreman-tasks
+        Checks::ForemanTasks::Invalid::CheckPlanningState, # if foreman-tasks
+        Checks::ForemanTasks::NotRunning, # if foreman-tasks
+        Checks::NonRhPackages,
+        Checks::PackageManager::Dnf::ValidateDnfConfig,
```
evgeni, line 44: picking this particular, but we should double check this either way: not all those checks have sufficient "confines" -- like this one does not confine to the package manager actually being DNF and it might break on Debian-like systems.
(not that this would be different with the current upgrade commands, but I still wanted to point it out)

evgeni:
Overall I think this makes sense

definitions/scenarios/update.rb
ehelms: Funny, this one doesn't throw a problem:

```
Check to validate dnf configuration before upgrade:                   [OK]
```

Not that we can't confine it better. I actually found other problems with running this on Debian that I can work on addressing separately.

ehelms: Funny, this one doesn't throw a problem:

```
Check to validate dnf configuration before upgrade:                   [OK]
```

Not that we can't confine it better. I actually found other problems with running this on Debian that I can work on addressing separately.

evgeni: Maybe it doesn't find any offending lines as the configuration is not there at all, I didn't look too closely, just that it's not confined.

Thanks!

ehelms: First item found to fix -- https://github.com/theforeman/foreman_maintain/pull/832

ehelms:
Rebased on to the fixes for the Debian

Griffin-Sullivan:
https://github.com/SatelliteQE/robottelo/pull/14883 passed PRT

evgeni (approved):

PR #803, opened 2024-02-16
Restrict katello-agent check to only 6.15 upgrade

ehelms:


definitions/checks/katello_agent.rb
```diff
@@ -5,7 +5,7 @@ class Checks::CheckKatelloAgentEnabled < ForemanMaintain::Check
     tags :pre_upgrade
 
     confine do
-      !feature(:capsule)
+      feature(:satellite) && feature(:satellite).current_minor_version == '6.14'
```
evgeni, line 8: !capsule would also trigger on upstream katello, now it doesn't anymore. I think that's (technically) a regression :)

ehelms: This check applied with the 4.9 to 4.10 release, which was Foreman 3.8 which has formean-maintain 1.3. So I think we are safe here?

evgeni: Ah, fair point

evgeni (approved):

PR #804, opened 2024-02-16
Cleaning up unused code paths

ehelms:


definitions/features/foreman_tasks.rb
```diff
@@ -93,10 +93,8 @@ def delete(state)
       DELETE FROM foreman_tasks_locks as ftl where ftl.task_id NOT IN (SELECT id FROM foreman_tasks_tasks);
     SQL
 
-    if check_min_version(foreman_plugin_name('foreman-tasks'), '4.0.0')
-      sql += 'DELETE FROM foreman_tasks_links as ftl ' \
-             'where ftl.task_id NOT IN (SELECT id FROM foreman_tasks_tasks);'
-    end
+    sql += 'DELETE FROM foreman_tasks_links as ftl ' \
+           'where ftl.task_id NOT IN (SELECT id FROM foreman_tasks_tasks);'
```
evgeni, line 97: should we just roll this into the `sql` defined in L87-94? no need to `+=` IMHO.

definitions/procedures/restore/installer_reset.rb
```diff
@@ -12,28 +12,12 @@ def run
 
     def installer_cmd
       installer = "yes | #{feature(:installer).installer_command} "
```
evgeni, line 14: someone needs to remind me, why this `yes |` is required, but out of scope for this PR probably.

evgeni (approved):
some nitpicks, but nothing that I'd not merge as-is.

PR #805, opened 2024-02-22
remove snapshot backup feature

evgeni:


ehelms:
Code looks good, we should great a 1.5-stable branch first. And this raises the question of -- do we got to 1.6 next or 2.0?

evgeni:
> Code looks good, we should great a 1.5-stable branch first. And this raises the question of -- do we got to 1.6 next or 2.0?

We never did semver for maintain, and snapshot backups are a rather niche feature, so I don't think it warrants a 2.0.

ehelms (approved):

PR #809, opened 2024-02-26
add PostgreSQL 13 storage check

evgeni:


definitions/checks/disk/available_space_postgresql13.rb
```diff
@@ -0,0 +1,43 @@
+module Checks
+  module Disk
+    class AvailableSpacePostgresql13 < ForemanMaintain::Check
+      metadata do
+        label :available_space_for_postgresql13
+        description 'Check to make sure PostgreSQL 13 work directory has enough space for upgrade'
+        confine do
+          (feature(:foreman_database) || feature(:candlepin_database) || feature(:pulpcore_database)) && \
+            file_exists?('/var/lib/pgsql/data')
```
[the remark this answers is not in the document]
evgeni: No idea 😁
Copied from old check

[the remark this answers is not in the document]
evgeni: Ah, it was to check that the DB is local. Now better.

definitions/checks/disk/available_space_postgresql13.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      def psql_13_available_space
+        io_obj = ForemanMaintain::Utils::Disk::IODevice.new(pgsql_dir(13))
+        io_obj.available_space
+      end
+
+      def warning_message
+        "PostgreSQL will be upgraded from 12 to 13. \n" \
+        "During the upgrade a backup is created in /var/lib/pgsql/data-old and requires " \
```
ehelms, line 37: Is the backup removed on successful upgrade or does it persist?

evgeni: It persists and the user has to remove it

ehelms: This is the same tactic we took before?

definitions/checks/disk/available_space_postgresql13.rb
```diff
@@ -0,0 +1,42 @@
...
+      metadata do
+        label :available_space_for_postgresql13
+        description 'Check to make sure PostgreSQL 13 work directory has enough space for upgrade'
+        confine do
+          feature(:instance).postgresql_local?
+        end
+      end
+
+      def run
+        assert(psql_13_available_space >= psql_12_consumed_space, warning_message, :warn => true)
```
ehelms, line 13: I am worried about warning being strong enough here. If there is not enough space, the upgrade will out right fail won't it?

definitions/checks/disk/available_space_postgresql13.rb
evgeni: That's what my memory says, yeah

(At least I recall having to special case el7 systems with /var/lib/pgsql still containing psql9 data in leapp)

definitions/checks/disk/available_space_postgresql13.rb
evgeni: Yeah, will change.

ehelms (approved):

PR #811, opened 2024-03-18
switch-to PostgreSQL 13 module during the upgrade

evgeni:


evgeni:
The refactoring slightly escalated, sorry :D

definitions/scenarios/upgrade_to_capsule_6_16.rb
```diff
@@ -41,6 +41,8 @@ class PreMigrations < Abstract
     def compose
       add_step(Procedures::Repositories::Setup.new(:version => '6.16'))
       if el8?
+        modules_to_switch = ['postgresql:13']
+        add_step(Procedures::Packages::SwitchModules.new(:module_names => modules_to_switch))
```
evgeni, line 45: I pondered whether this should also be added to the upstream nightly upgrade definitions, and couldn't decide.

ehelms: That seems correct -- is there an issue if it is added?

ehelms (approved):

definitions/scenarios/upgrade_to_capsule_6_16.rb
evgeni: Nah, switch-to to a module that is already enabled is a noop.
So you say "add"?

ehelms: Yes

evgeni: done

PR #813, opened 2024-03-25
Fixes #37294 - SCA-Only: pre-upgrade check for entitlement content access mode

wbclark:


definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,42 @@
...
+  private
+
+  def execute_rake_task
+    execute!('foreman-rake katello:upgrades:4.12:update_content_access_modes[dry_run]')
+  end
+
+  def parse_orgs_to_migrate(output)
+    orgs = []
+    output.each_line do |line|
+      match = line.match(/Setting content access mode for (.+)/)
```
jeremylenz, line 37: We say "Checking" instead of "Setting" if it's a dry run.

definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def run
+    with_spinner('Checking organization content access modes') do
+      output = execute_rake_task
+      orgs_to_migrate = parse_orgs_to_migrate(output)
+
+      assert(
+        orgs_to_migrate.empty?,
+        "The following organizations are using EntitlementContentAccess mode:\n"\
```
jeremylenz, line 19: "EntitlementContentAccess" is a term I hadn't heard before. I prefer just "entitlement mode."

```suggestion
        "The following organizations are using entitlement mode:\n"\
```

definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,42 @@
...
+        "#{orgs_to_migrate.join(', ')}\n"\
+        "As of Katello 4.12, EntitlementContentAccess mode is removed, and these organizations"\
+        " will be automatically migrated to SimpleContentAccess mode during the upgrade."\
```
jeremylenz, line 22: I think we use this or the non-capitalized version ("simple content access"), in most user-facing places I've seen. cc @bangelic

```suggestion
        " will be automatically migrated to Simple Content Access during the upgrade."\
```

definitions/checks/sca_only.rb
wbclark: conveniently, this should get a few lines under rubocop's length limit as well

definitions/checks/sca_only.rb
wbclark: I made the change, but I'll leave this review item open for now in case @bangelic wants to give input

jeremylenz (approved):
Okay just two more comments. Code LGTM; I'll leave any functional testing to colleagues.

Thanks @wbclark !

definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,42 @@
...
+        " will be automatically migrated to Simple Content Access during the upgrade."\
+        " Please ensure that you have reviewed the documentation and are prepared for this change."\
+      )
+    end
+  end
+
+  private
+
+  def execute_rake_task
+    execute!('foreman-rake katello:upgrades:4.12:update_content_access_modes[dry_run]')
```
jeremylenz, line 31: If we're upgrading from Katello 4.11, and this rake task is only included in 4.12, will we have access to it here?

definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,42 @@
...
+  def run
+    with_spinner('Checking organization content access modes') do
+      output = execute_rake_task
+      orgs_to_migrate = parse_orgs_to_migrate(output)
+
+      assert(
+        orgs_to_migrate.empty?,
+        "The following organizations are using entitlement mode:\n"\
+        "#{orgs_to_migrate.join(', ')}\n"\
+        "As of Katello 4.12, entitlement mode is removed, and these organizations"\
```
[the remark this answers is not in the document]
ehelms: We do not perform any branding.

definitions/checks/sca_only.rb
ehelms: If this check is intended to run before upgrading then no, as the Katello code will only be updated during the upgrade steps after pre upgrade checks.

jeremylenz: Hm, we'll need a different solution here then.

Can the pre-upgrade check run a `rails runner` command?

```
foreman-rake runner "Organization.all.map(&:simple_content_access?).count(false)"
```

should give you the number of orgs you need to compose the message here, without needing to backport the rake task.

definitions/checks/sca_only.rb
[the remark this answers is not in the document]
wbclark: We should be able to use `feature(:instance).downstream` for this, I think

I'd like to add separate upstream/downstream documentation links using this as well

definitions/checks/sca_only.rb
wbclark: I'm spinning up a katello 4.11 to test this

wbclark: trying to figure out why this is happening:

```
[root@centos8-stream-katello-4-11 ~]# foreman-maintain upgrade list-versions --disable-self-upgrade
[root@centos8-stream-katello-4-11 ~]# 
```

wbclark:
@jeremylenz @ehelms I got this working on Satellite 6.14... for ease of review, I left the changes as a separate commit now and will fixup & force push once ACKed

here's the output I got in the failure case:

```
[root@sat-6-14-qa-rhel8 checks]# foreman-maintain upgrade run --target-version=6.14.z
Check whether system has any non Red Hat repositories (e.g.: EPEL) enabled: 
/ Checking repositories enabled on the system                         [OK]      


Running Checks before upgrading to Satellite 6.14.z
Check to make sure root(/) partition has enough space:                [OK]
Check to make sure /var/lib/candlepin has enough space:               [OK]
Check for running tasks:                                              [OK]
Check for old tasks in paused/stopped state:                          [OK]
Check for pending tasks which are safe to delete:                     [OK]
Check for tasks in planning state:                                    [OK]
Check to verify if any hotfix installed on system: 
\ Checking for presence of hotfix(es). It may take some time to verify.         
                                                                      [OK]
Check whether system has any non Red Hat repositories (e.g.: EPEL) enabled: 
| Checking repositories enabled on the system                         [OK]      
Check if any organizations are using entitlement mode: 
| Checking organization content access modes                          [FAIL]    
The following organizations are using entitlement mode:
Default Organization
As of Satellite 6.16, entitlement mode is removed, and these organizations will be automatically migrated to Simple Content Access during the upgrade. Please ensure that you have reviewed the documentation and are prepared for this change. Documentation for SCA mode is available at https://access.redhat.com/articles/simple-content-access
Check if TMOUT environment variable is set:                           [OK]
Check if system requirements match current tuning profile:            [OK]
Check if system requirements match current tuning profile:            [OK]
Check if any upstream repositories are enabled on system: 
\ Checking for presence of upstream repositories                      [OK]      
Check for roles that have filters with multiple resources attached:   [OK]
Check for duplicate permissions from database:                        [OK]
Check if system has any non Red Hat RPMs installed (e.g.: Fedora):    [OK]
Check whether reports have correct associations:                      [OK]
Check to validate yum configuration before upgrade:                   [OK]
Check if checkpoint_segments configuration exists on the system:      [OK]
\ Validating availability of repositories for 6.14                    [OK]      
Scenario [Checks before upgrading to Satellite 6.14.z] failed.

  [check-organization-content-access-mode]

--whitelist="check-organization-content-access-mode"


```

definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,55 @@
...
+        " will be automatically migrated to Simple Content Access during the upgrade."\
+        " Please ensure that you have reviewed the documentation and are prepared for this change."\
+        " Documentation for SCA mode is available at #{doc_link}"
+      )
+    end
+  end
+
+  private
+
+  def execute_rake_task
```
jeremylenz, line 39: nit: this is technically no longer a rake task. Perhaps `def execute_content_access_mode_check` or `def execute_org_check`?

jeremylenz (commented):
Awesome! Just one comment below

definitions/checks/sca_only.rb
wbclark: `execute_org_check` sounds good to me, thanks

jeremylenz (approved):
LGTM :+1: 

thanks @wbclark !

definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+  end
+
+  def run
+    if feature(:instance).downstream
+      versioned_instance_name = "Satellite 6.16"
+      doc_link = "https://access.redhat.com/articles/simple-content-access"
+    else
+      versioned_instance_name = "Katello 4.12"
+      doc_link = "https://theforeman.org/plugins/katello/nightly/user_guide/simple_content_access/index.html"
```
wbclark, line 18: Katello 4.12 docs weren't available yet when I originally wrote this, so I pointed it at nightly as a temporary measure

But maybe it should point at 4.11 docs instead? I'm not sure if the docs will continue to document a distinction that no longer exists.

jeremylenz: Indeed, 4.12 docs will have all SCA-related info removed. 4.11 docs is a better bet I'd say.

jeremylenz: Actually docs.theforeman.org would be the more correct upstream documentation in general (the link you had is for the old doc site.) But docs.theforeman.org doesn't really have any good overview of SCA. I'm actually thinking maybe just use https://access.redhat.com/articles/transition_of_subscription_services_to_the_hybrid_cloud_console

for both upstream and downstream. This is what we also ended up using in https://bugzilla.redhat.com/show_bug.cgi?id=2255222. Thoughts?

definitions/checks/sca_only.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  private
+
+  # rubocop:disable Lint/InterpolationCheck
+  def execute_org_check
+    execute!(
+      'echo "Organization.all.each ' \
+      '{ |org| puts \"NONSCA: #{org.name}\" unless org.simple_content_access? }" ' \
+      '| foreman-rake console'
+    )
```
ehelms, line 45: Are we able to use our API here? I suppose this is a one off check so it doesn't matter as much for a longer-term support.

definitions/checks/sca_only.rb
wbclark: ...and SCA is only relevant for folks using Red Hat content to begin with, so they'd already have an account

cool, I'll just link that one in either case.

definitions/checks/sca_only.rb
wbclark: Well, I actually broke the damn thing again while trying to appease rubocop:

```
The following organizations are using entitlement mode:
#{org.name}" unless org.simple_content_access? }, Default Organization
```

I'd still rather fix that up and also test that it works properly with multiple SCA org and no SCA orgs, but could pivot to API if it ends up being more trouble than it's worth.

definitions/checks/sca_only.rb
jeremylenz: Oh, and it looks like this was recently updated, too! So this is the link to use. https://access.redhat.com/articles/transition_of_subscription_services_to_the_hybrid_cloud_console#satellite-13

ehelms:
Could you create a bit more spacing around the output, e.g.

```
| Checking organization content access modes                          [FAIL]    

The following organizations are using entitlement mode:
  Default Organization

As of Satellite 6.16, entitlement mode is removed, and these organizations will be 
automatically migrated to Simple Content Access during the upgrade. Please ensure that
you have reviewed the documentation and are prepared for this change. Documentation 
for SCA mode is available at https://access.redhat.com/articles/simple-content-access
```

ehelms:
I also think this check would benefit from not being in the middle and being the last check of the pre_upgrade steps. To do this, add this step explicitly and drop the `tags :pre_upgrade`, for example like this -- https://github.com/theforeman/foreman_maintain/blob/master/definitions/scenarios/upgrade_to_satellite_6_16.rb#L31

definitions/checks/sca_only.rb
wbclark: The absence of an answer for disconnected users here is concerning to me:

```
Satellite

Question: What if I do not want to use the Simple Content Access?
Answer: Satellite versions <= 6.15 are the only versions that support entitlement mode. Eventually, as per the [Red Hat Satellite Lifecycle](https://access.redhat.com/support/policy/updates/satellite) document, those versions will be out of support eventually.

Question: As a satellite user, how do I track my usage of subscriptions?
Answer: Install and/or enable the [Satellite inventory upload plugin](https://access.redhat.com/documentation/en-us/subscription_central/1-latest/html-single/getting_started_with_the_subscriptions_service/index#proc-installing-satellite-inventory-upload-plugin_assembly-setting-up-subscriptionwatch-ctxt)

Question: I am a disconnected user of Satellite and I cannot send usage data to console.redhat.com?
Answer:

Question: The manifest UI in console.redhat.com doesn't have any way to attach subscriptions. How do I attach subscriptions to a manifest?
Answer: The preferred pattern for a connected Satellite customer is to download the manfiest from console.redhat.com, import it, and add subscriptions via the Satellite user interface. [Learn more about creating manifests for a connected Satellite Server](https://access.redhat.com/documentation/en-us/subscription_central/2023/html/creating_and_managing_manifests_for_a_connected_satellite_server/index). If you are a disconnected Satellite or Ansible Automation Platform user, please continue to use the [Subscription Allocations](https://access.redhat.com/management/subscription_allocations) page on access.redhat.com.
```

definitions/scenarios/upgrade_to_satellite_6_16.rb
```diff
@@ -29,6 +29,7 @@ def compose
       add_steps(find_checks(:pre_upgrade))
       add_step(Checks::Disk::AvailableSpacePostgresql13)
       add_step(Checks::Repositories::Validate.new(:version => '6.16'))
+      add_step(Checks::CheckOrganizationContentAccessMode)
```
wbclark, line 32: @ehelms I don't see a corresponding upgrade scenario for Katello 4.12 -- the only upstream upgrade scenario is Katello Nightly.

Is this what you were referring to when you previously said we don't do any branding here?

Should I add this check to the Nightly upgrade scenario as well, or should I leave it Satellite only and remove the branding of instance_version string?

definitions/checks/sca_only.rb
jeremylenz: We're working on that in rh_cloud; users are already able to see the data that would otherwise have been sent to console, and I believe there's work going on to make that easier as well. Rich and @chris1984 may know more

definitions/scenarios/upgrade_to_satellite_6_16.rb
ehelms: Fair question - as the Katello scenario is nightly only and I do not think our docs mention using foreman-maintain upgrade then being Satellite specific about this is OK. Please verify the docs part though.

wbclark:
Current output with the updates I've pushed (using 6.14.z update to test):

```
- Validating availability of repositories for 6.14                    [OK]      
Check if any organizations are using entitlement mode: 
\ Checking organization content access modes                          [FAIL]    

The following organizations are using entitlement mode:
Default Organization

As of Satellite 6.16, entitlement mode is removed, and these organizations will be automatically migrated to Simple Content Access during the upgrade. Please ensure that you have reviewed the documentation and are prepared for this change. Documentation for SCA mode is available at https://access.redhat.com/articles/simple-content-access
Scenario [Checks before upgrading to Satellite 6.14.z] failed.
```

definitions/scenarios/upgrade_to_satellite_6_16.rb
jeremylenz: I was just looking at the upgrade docs, and indeed they don't mention foreman-maintain. The upstream upgrade procedure is just dnf update followed by an installer run.

wbclark: Thanks. So upstream docs at https://docs.theforeman.org/nightly/Upgrading_Project/index-katello.html use foreman-maintain for starting/stopping services only; the upgrade itself is still done manually

I've pushed an update that removes references to Katello and always uses downstream names

wbclark:
Output formatting is broken with multiple non-SCA orgs:

```
Check if any organizations are using entitlement mode: 
| Checking organization content access modes                          [FAIL]    

The following organizations are using entitlement mode:
Default Organization\nMyNewOrg

As of Satellite 6.16, entitlement mode is removed, and these organizations will be automatically migrated to Simple Content Access during the upgrade. Please ensure that you have reviewed the documentation and are prepared for this change. Documentation for SCA mode is available at https://access.redhat.com/articles/simple-content-access
```

Fix incoming....

chris1984:
See how much more fun this is than Python :)

definitions/checks/sca_only.rb
wbclark: I know answers/options exist, I just don't want to point customers to a document that creates the appearance that we either don't know or don't care about what this means for disconnected users. Even for connected users who have no reason to care, it just _looks_ sloppy.

Should we keep the current link for now given that we can always change the string later before releasing? Or do we have a guarantee that someone is actively working on the other document and it will be completed in time?

wbclark:
Output works correctly for multiple non-SCA orgs:

```
Check if any organizations are using entitlement mode: 
\ Checking organization content access modes                          [FAIL]    

The following organizations are using entitlement mode:
Default Organization
MyNewOrg

As of Satellite 6.16, entitlement mode is removed, and these organizations will be automatically migrated to Simple Content Access during the upgrade. Please ensure that you have reviewed the documentation and are prepared for this change. Documentation for SCA mode is available at https://access.redhat.com/articles/simple-content-access
```

And after migrating all orgs to SCA mode:

```
Check if any organizations are using entitlement mode: 
\ Checking organization content access modes                          [OK]      
```

definitions/checks/sca_only.rb
jeremylenz: Given that we already link to this from the UI, I think it's better to link to it here as well. I'm confident that question won't remain blank for long.

sideangleside: Yeah, I totally forget to actually answer that one. I'll do the needful and update

sideangleside: Also, the [SCA KCS](https://access.redhat.com/articles/simple-content-access) should be accessible without login. I purpose keep that doc outside the paywall as SCA does have value for non-Satellite users (if for no other reason than they don't have to attach custom subs either). 

That said, you can also use [the official product doc](https://access.redhat.com/documentation/en-us/subscription_central/1-latest/html-single/getting_started_with_simple_content_access/index)

wbclark: I think ideally, we link something here that provides a quick summary, and that quick summary itself contains a link to the full documentation. Does that make sense on your end @sideangleside and is there a preferred link to use?

sideangleside: @wbclark Lets just use the KCS. It's easier to update in place.

wbclark: Thanks for the feedback. I'll keep the link to the KCS as-is, and mark this conversation resolved.

ehelms (approved):

PR #814, opened 2024-03-27
Refs #37275: Enable modules and download updates after enabling postgresql module

ehelms:
The dnf module switch-to functionality not only swaps module streams but forces the installation of the new packages which can cause the system to be in a bad state. For example, sync plan disable will break in some cases.

I should also mention that it breaks the paradigm we designed where package updates are downloaded but not applied before entering maintenance mode.

### DIFF
--- a/definitions/scenarios/upgrade_to_satellite_6_16.rb
+++ b/definitions/scenarios/upgrade_to_satellite_6_16.rb
@@ -39,17 +39,6 @@ module Scenarios::Satellite_6_16
     end
 
     def compose
-      add_step(Procedures::Repositories::Setup.new(:version => '6.16'))
-      if el8?
-        modules_to_switch = ['postgresql:13']
-        add_step(Procedures::Packages::SwitchModules.new(:module_names => modules_to_switch))
-        modules_to_enable = ["satellite:#{el_short_name}"]
-        add_step(Procedures::Packages::EnableModules.new(:module_names => modules_to_enable))
-      end
-      add_step(Procedures::Packages::Update.new(
-        :assumeyes => true,
-        :dnf_options => ['--downloadonly']
-      ))
       add_steps(find_procedures(:pre_migrations))
     end
   end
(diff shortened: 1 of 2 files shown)


evgeni (changes requested):
You can't enable the `satellite:el8` module as it now has broken deps. That has to happen *after* the switch.
And because if that the download of packages needs to also happen after the switch, as otherwise no packages are downloaded (dnf discards modules with broken deps, and thus all RPMs provided by them)

ehelms:
> You can't enable the `satellite:el8` module as it now has broken deps. That has to happen _after_ the switch. And because if that the download of packages needs to also happen after the switch, as otherwise no packages are downloaded (dnf discards modules with broken deps, and thus all RPMs provided by them)

I was afraid of that, I hadn't tested this yet. I have another idea but need to test it.

evgeni:
Yeah, I think you need to move this whole block:
https://github.com/theforeman/foreman_maintain/blob/fba91a5b0548f42a9b10ef70d1e30911692e26c8/definitions/scenarios/upgrade_to_satellite_6_16.rb#L42-L52
down to `Migrations`.

Also, frankly, it makes more sense there. PreMigrations should be non-destructive, while this code (now) is.

definitions/scenarios/upgrade_to_satellite_6_16.rb
```diff
@@ -39,6 +39,7 @@ class PreMigrations < Abstract
     end
 
     def compose
+      add_step(Procedures::SyncPlans::Disable.new)
```
evgeni, line 42: Should the same change happen in capsule? OTOH, capsules don't have syncplans.

evgeni:
I think this should refer to https://projects.theforeman.org/issues/37275 in some way?

ehelms:
> Yeah, I think you need to move this whole block:
> 
> https://github.com/theforeman/foreman_maintain/blob/fba91a5b0548f42a9b10ef70d1e30911692e26c8/definitions/scenarios/upgrade_to_satellite_6_16.rb#L42-L52
> 
> 
> down to `Migrations`.
> Also, frankly, it makes more sense there. PreMigrations should be non-destructive, while this code is.

I was hoping I could find a way to avoid that :/ since it defeats the whole point of reducing the downtime by caching downloads. Why must modularity break everything.

ehelms:
Updated to move all of that logic (sadly) into the Migration block after maintenance mode

evgeni:
you have a typo in the redmine, it's 37275 not 37375 ;)

lpramuk:
`Upgrade finished.`  

I am able to upgrade after applying this PR

evgeni (approved):

PR #826, opened 2024-04-09
Allow restarting all pulpcore or dynflow workers

ehelms:
This is now possible:

```
foreman-maintain service restart --only=pulpcore-worker@*.service
```


lib/foreman_maintain/utils/service/abstract.rb
```diff
@@ -31,7 +31,11 @@ def inspect
 
       def matches?(service)
         if service.is_a? String
-          service == @name
+          if service.include?('@*')
+            @name.include?(service.split('@*')[0])
```
evgeni, line 35: this is a systemd-specific feature, so shouldn't this be in `lib/foreman_maintain/utils/service/systemd.rb` instead?

(and then with a nice test added? :D)

lib/foreman_maintain/utils/service/abstract.rb
```diff
@@ -31,7 +31,11 @@ def inspect
 
       def matches?(service)
         if service.is_a? String
-          service == @name
+          if service.include?('@*')
```
ekohl, line 34: What if they only want to restart part of it? Shouldn't it just split on `@`?

ehelms:
On Wed, Apr 10, 2024, 6:27 AM Ewoud Kohl van Wijngaarden <
***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In lib/foreman_maintain/utils/service/abstract.rb
> <https://github.com/theforeman/foreman_maintain/pull/826#discussion_r1559208083>
> :
>
> > @@ -31,7 +31,11 @@ def inspect
>
>        def matches?(service)
>          if service.is_a? String
> -          service == @name
> +          if service.include?('@*')
>
> What if they only want to restart part of it? Shouldn't it just split on @
> ?
>
Define part of it? Or an example?

lib/foreman_maintain/utils/service/abstract.rb
ehelms: Define part of it? Or provide me an example of the case you are thinking of @ekohl ?

evgeni: I think something like `restart pulpcore-worker@1` to only restart the first one. (Probably makes more sense with dynflow).

ehelms: Well, the original code already supported restarting a specific worker, I went on a wild goose chase.

evgeni: Ah, interesting.
And I should not review stuff on my phone ;)

ehelms: However, I think with systemd you can pass the wildcard at any point after the `@` such as `dynflow-sidekiq@worker*` and should we support that?

I am questioning all this indirection and "service checking" against a known list of services rather than just passing things directly to systemd. I am asking myself, what is the benefit of checking the service passed to `service restart` against a known list of services?

ehelms:
This change at least solves the original request to be able to do:

```
foreman-maintain service restart --only=pulpcore-worker@*.service
```

And while it doesn't cover every possible scenario, I think we would need to address this change more directly to systemd.

ehelms:
@evgeni Thoughts on taking this narrow approach based on my [comment](https://github.com/theforeman/foreman_maintain/pull/826#issuecomment-2051876877)

lib/foreman_maintain/utils/service/abstract.rb
evgeni: For restart, not much (but I personally also would never use FM for that).
For status, it's useful to only see a subset of services :woman_shrugging: 

But then I started reading https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#Parameter%20Syntax and think your solution only goes half as far as it could (to be correct).

How about

```ruby
      def matches?(service)
        if service.is_a? String
          service == @name || File.fnmatch(service, @name)
        elsif service.instance_of?(self.class)
          service.name == @name
        else
          false
        end
      end
```

This now also allows `foreman-maintain service status --only dynflow-sidekiq@worker*` and friends.

evgeni (approved):

test/lib/utils/service/systemd_test.rb
```diff
@@ -109,6 +109,10 @@ module ForemanMaintain
       it 'can compare with other obejcts' do
         _(httpd_service.matches?(nil)).must_equal false
       end
+
+      it 'can compare by wildcard service' do
+        _(httpd_service.matches?('httpd@*.service')).must_equal true
```
evgeni, line 114:
```suggestion
        _(httpd_service.matches?('http*.service')).must_equal true
```
there is no `httpd@` service defined, only `httpd`

evgeni (approved):

PR #828, opened 2024-04-12
disable foreman-protector when dnf is executed inside leapp

evgeni:
otherwise it will prevent dnf from installing anything, breaking the upgrade


evgeni:
discussed with @pirat89, this is the way to go for us, so un-drafting and ready to review!

PR #829, opened 2024-04-16
Fixes #37357 - Add container migration to upgrade

sjha4:
Adding container-handle-image-data migration to post-upgrade tasks with upgraded pulp.

### DIFF
--- a/definitions/procedures/pulpcore/container_handle_image_metadata.rb
+++ b/definitions/procedures/pulpcore/container_handle_image_metadata.rb
@@ -1,0 +1,23 @@
+module Procedures::Pulpcore
+  class ContainerHandleImageMetadata < ForemanMaintain::Procedure
+    include ForemanMaintain::Concerns::SystemService
+    include ForemanMaintain::Concerns::PulpCommon
+
+    metadata do
+      description 'Initialize and expose container image metadata in the pulpcore db'
+      for_feature :pulpcore
+    end
+
+    def run
+      with_spinner('Initialize and expose container image metadata in the pulpcore db') do |spinner|
+        necessary_services = feature(:pulpcore_database).services
+
+        feature(:service).handle_services(spinner, 'start', :only => necessary_services)
+
+        spinner.update('Adding image metadata to pulp. You can continue using the ' \
+               'system normally while the task runs in the background.')
+        execute!(pulpcore_manager('container-handle-image-data'))
+      end
+    end
+  end
+end
--- a/definitions/scenarios/upgrade_to_capsule_6_16.rb
+++ b/definitions/scenarios/upgrade_to_capsule_6_16.rb
@@ -95,6 +95,7 @@ module Scenarios::Capsule_6_16
       add_steps(find_checks(:default))
       add_steps(find_checks(:post_upgrade))
       add_step(Procedures::Packages::CheckForReboot)
+      add_step(Procedures::Pulpcore::ContainerHandleImageMetadata)
     end
   end
 end
--- a/definitions/scenarios/foreman_upgrade.rb
+++ b/definitions/scenarios/foreman_upgrade.rb
@@ -134,7 +134,8 @@ module Scenarios::ForemanUpgrade
         Checks::ServerPing,
         Checks::ServicesUp,
         Checks::SystemRegistration,
-        Procedures::Packages::CheckForReboot
+        Procedures::Packages::CheckForReboot,
+        Procedures::Pulpcore::ContainerHandleImageMetadata
       )
     end
   end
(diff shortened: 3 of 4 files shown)


evgeni (changes requested):
can you address the rubocop issues please?

evgeni:
This is not needed on content proxies? Just the main server?

sjha4:
> This is not needed on content proxies? Just the main server?

It is..Added to proxy upgrade.. Do the same scenarios run for upstream upgrades as well?

definitions/scenarios/upgrade_to_capsule_6_16.rb
```diff
@@ -80,6 +80,7 @@ class PostMigrations < Abstract
     def compose
       add_step(Procedures::RefreshFeatures)
       add_step(Procedures::Service::Start.new)
+      add_step(Procedures::Pulpcore::ContainerHandleImageData)
```
ehelms, line 83: I am thinking this should be the last step? One thing I talked to @ianballou about is making this step happen in a way that we communicate to users that the system is completely operable and usable while this is happening in case for some user it's a longer running task. The text output should also indicate this I think to help guide users.

ianballou: Indeed indeed, the migration should be triggered once the Satellite is back up and running after the upgrade. If there's an error, it shouldn't block the upgrade as well.
+1 to the text output too giving the user some info.

ianballou (commented):
Just wanted to note that we'll also need to add the indexing task to this afterwards.

definitions/scenarios/upgrade_to_capsule_6_16.rb
sjha4: @ehelms : Ack.. 👍🏼 
Having the step after `add_step(Procedures::Service::Start.new)` in the PostMigrations would get the services up but not return from the upgrade I guess..

Would adding it to PostMigrationCheck block the upgrade too?

ehelms: Services would be started where you have it now, but the firewall from maintenance mode would still be up (I know, not obvious from this). Take a look at https://github.com/theforeman/foreman_maintain/pull/763/files#diff-574e3a009c6f79e4ce0c1bbab07fc007071ea541bc8985ca4ffc4a6c7505d212L78-R113 where this lays out each individual item very clearly.

sjha4: Looking at that, feels like the right place would be in PostUpgradeChecks after post_upgrade checks step once the services are checked to be running etc.

sjha4:
Will get to the merge conflict. Does the change look good @ehelms ?

ianballou (approved):
Just gave this a test with a mock Satellite upgrade, the command is indeed called right before the last step, and I saw the correct message, so I think it's all good.

ehelms:
@sjha4 I will get to checking this early this week.

definitions/scenarios/upgrade_to_capsule_6_16.rb
```diff
@@ -94,6 +94,7 @@ class PostUpgradeChecks < Abstract
     def compose
       add_steps(find_checks(:default))
       add_steps(find_checks(:post_upgrade))
+      add_step(Procedures::Pulpcore::ContainerHandleImageMetadata)
```
ehelms, line 97: For Satellite and Capsule you opted to put this before check for reboot and in the foreman upgrade it's after that check. I think we should be consistent.

definitions/procedures/pulpcore/container_handle_image_metadata.rb
```diff
@@ -0,0 +1,23 @@
...
+    def run
+      with_spinner('Initialize and expose container image metadata in the pulpcore db') do |spinner|
+        necessary_services = feature(:pulpcore_database).services
+
+        feature(:service).handle_services(spinner, 'start', :only => necessary_services)
+
+        spinner.update('Adding image metadata to pulp. You can continue using the ' \
+               'system normally while the task runs in the background.')
+        execute!(pulpcore_manager('container-handle-image-data'))
+      end
```
sjha4, line 20: Wondering if we can add a step to run the [rake task](https://github.com/Katello/katello/pull/10988/files#diff-055c9a3444ca692a4bf7d89c9e46f3495096c13c3b0b2355a731086e1128d116R75) to reindex katello data right after this in the file like 
`execute!('foreman-rake katello:import_docker_manifest_labels')`

ianballou: I think indexing the content right after running this would be ideal since it has to happen anyway. It's just a matter of us having one or two commands in the documentation.

sjha4: I'll update it here..

sjha4: Ok..I will let this PR move independently of the other PR for now and make any updates to this file when the rake task is merged.

ehelms (approved):

PR #834, opened 2024-04-22
Support needs reboot check on Debian

ehelms:


evgeni (approved):

lib/foreman_maintain/package_manager/apt.rb
```diff
@@ -60,6 +60,12 @@ def version_locking_supported?
       false
     end
 
+    def reboot_required?
+      status = File.exist?('/var/run/reboot-required') ? 1 : 0
+      output = ''
```
ekohl, line 65: How is the user supposed to know that it's needed? This output is printed directly and it's always empty.

evgeni: You say this should be:
```suggestion
      output = 'A reboot is required.'
```

ekohl: Only if it is. Otherwise probably not because I think the message is printed regardless.

PR #837, opened 2024-04-29
Drop z-stream upgrade scenarios

ehelms:


ehelms:
@Griffin-Sullivan @evgeni ok with this change or do you feel we should wait?

evgeni:
I'm cool with the change, but I think we need to first (?) migrate at least robotello to use the new command? tests/foreman/maintain/test_upgrade.py and tests/upgrades/test_satellite_maintain.py refer to that (and https://github.com/SatelliteQE/robottelo/pull/14883 is fixing it)

evgeni (approved):

Griffin-Sullivan (approved):
I think we can continue with merging now. I have https://github.com/SatelliteQE/robottelo/pull/14970 open.

PR #840, opened 2024-05-02
Check service status in preparation steps before hammer setup

ehelms:
Before this change a user would just see the following if a service was down:

```
Setup hammer: 
Configuring Hammer CLI...
Hammer admin password: 
                                                                      [FAIL]
Hammer configuration failed: Incorrect credential for admin user.
```

Which is rather confusing as it's not the password that's the problem.

### DIFF
--- a/definitions/procedures/hammer_setup.rb
+++ b/definitions/procedures/hammer_setup.rb
@@ -2,6 +2,9 @@ class Procedures::HammerSetup < ForemanMaintain::Procedure
   metadata do
     description 'Setup hammer'
     for_feature :hammer
+    preparation_steps do
+      Checks::ServicesUp.new
+    end
   end
 
   def run


definitions/procedures/hammer_setup.rb
```diff
@@ -2,6 +2,9 @@ class Procedures::HammerSetup < ForemanMaintain::Procedure
   metadata do
     description 'Setup hammer'
     for_feature :hammer
+    preparation_steps do
+      Procedures::Service::Status.new(:brief => true, :failing => true)
```
evgeni, line 6: Shouldn't this be rather `Checks::ServicesUp.new`?

ehelms: Done

evgeni (approved):

PR #841, opened 2024-05-02
Drop list-versions command

ehelms:


Griffin-Sullivan (commented):
If we remove list-versions shouldn't we simultaneously remove `--target-version`?

Griffin-Sullivan (approved):
Automation is running successfully.

ehelms:
> If we remove list-versions shouldn't we simultaneously remove `--target-version`?

I wasn't going to quite yet as it ties into a lot of code and makes it more explicit what's happening. However, as there is only ever one version to upgrade to, I didn't see a reason to keep this command around.

PR #845, opened 2024-05-06
drop dnf_options from P::P::Update

evgeni:


ehelms:
Is this largely because it's unused now?

evgeni:
> Is this largely because it's unused now?

And because something leaks and this fixes it.

ehelms:
> > Is this largely because it's unused now?
> 
> And because something leaks and this fixes it.

Were you able to figure out how or why? Or are we chalking this up to "foreman-maintain does some crazy meta programming crap?"

evgeni:
I *think* because you're passing in a global variable `@dnf_options` it ends up modifying that, instead of a local one

(that's why in python you never define the signature of such functions as `def do(something=[])` but def `do(something=None)`  and then do `if something is None: something = []`)

evgeni:
```
% cat /tmp/test.rb 
def update(options: [])
  options << 'nice'
  puts(options)
end

@my_fancy_options = ['lol']

update(options: @my_fancy_options)
puts(@my_fancy_options)

% ruby /tmp/test.rb
lol
nice
lol
nice
```

but

```
% cat /tmp/test.rb 
def update(options: nil)
  options |= []
  options << 'nice'
  puts(options)
end

@my_fancy_options = ['lol']

update(options: @my_fancy_options)
puts(@my_fancy_options)

% ruby /tmp/test.rb
lol
nice
lol
```

ehelms (approved):

PR #847, opened 2024-05-13
Add ipv6 disable check

ehelms:


definitions/checks/check_ipv6_disable.rb
```diff
@@ -0,0 +1,23 @@
+class Checks::CheckIpv6Disable < ForemanMaintain::Check
+  metadata do
+    label :check_ipv6_disable
+    description 'Check if ipv6.disable=1 is set at kernel level'
+  end
+
+  def run
+    grub_file = File.read('/etc/default/grub')
+
+    assert(!grub_file.include?("ipv6.disable=1"), error_message)
```
evgeni, line 10: I'd rather check `/proc/cmdline` as that shows what's currently booted vs "what could be booted if the user didn't fiddle with things"

test/definitions/checks/check_ipv6_disable_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  subject { Checks::CheckIpv6Disable.new }
+
+  context 'throw an error message when ipv6.disable=1 is set' do
+    before do
+      File.expects(:read).with('/proc/cmdline').returns('ipv6.disable=1')
+    end
+
+    it 'system is self registered' do
+      result = run_step(subject)
+      msg = 'test'
```
ekohl, line 15: This doesn't look very useful to add

test/definitions/checks/check_ipv6_disable_test.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  context 'throw an error message when ipv6.disable=1 is set' do
+    before do
+      File.expects(:read).with('/proc/cmdline').returns('ipv6.disable=1')
+    end
+
+    it 'system is self registered' do
+      result = run_step(subject)
+
+      assert result.fail?, msg
```
evgeni, line 16: `msg` is undefined :)

ehelms:
@evgeni Fixed up if you can review again

test/definitions/checks/check_ipv6_disable_test.rb
```diff
@@ -0,0 +1,31 @@
...
+  include DefinitionsTestHelper
+
+  subject { Checks::CheckIpv6Disable.new }
+
+  context 'throw an error message when ipv6.disable=1 is set' do
+    before do
+      File.expects(:read).with('/proc/cmdline').returns('ipv6.disable=1')
+    end
+
+    it 'system is self registered' do
```
evgeni, line 13: I bet you can come up with a better description ;)

test/definitions/checks/check_ipv6_disable_test.rb
```diff
@@ -0,0 +1,31 @@
...
+      assert result.fail?
+    end
+  end
+
+  context 'success when ipv6.disable=1 is not set' do
+    before do
+      File.expects(:read).with('/proc/cmdline').returns('test.net=0')
+    end
+
+    it 'system is self registered' do
```
evgeni, line 25: Same

evgeni (approved):

test/definitions/checks/check_ipv6_disable_test.rb
ehelms: How much you willing to bet?

evgeni: A dime?

Griffin-Sullivan (approved):
Passed testing in https://github.com/SatelliteQE/robottelo/pull/15062

ekohl:
I just saw https://github.com/voxpupuli/facterdb/pull/340 and noticed you can also read the sysctl. So technically you can enable it even if it's in `/proc/cmdline`, but I still think it's a good check because it means that after a reboot it'll also work. Or at least, more likely.

evgeni:
No, you really can't:

```console
[root@centos9-stream ~]# cat /proc/cmdline 
BOOT_IMAGE=(hd0,msdos1)/boot/vmlinuz-5.14.0-354.el9.x86_64 root=UUID=a91e762e-6199-45c4-8989-1bba94306968 ro no_timer_check console=tty0 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0 elevator=noop ipv6.disable=1
[root@centos9-stream ~]# sysctl -a |grep ipv6
[root@centos9-stream ~]# sysctl -w net.ipv6.conf.all.disable_ipv6=0
sysctl: cannot stat /proc/sys/net/ipv6/conf/all/disable_ipv6: No such file or directory
```

ekohl:
Ah, so you can enable/disable it at runtime if the kernel was booted with IPv6 support, but if it's in `/proc/cmdline` then it surely isn't there.

That makes me wonder if we need both checks.

evgeni:
> That makes me wonder if we need both checks.

I don't think so. `ipv6.disable` breaks things like `bind` to `[::]`. But disabling IPv6 on all interfaces (that's what the `all` in `net.ipv6.conf.all` is) means you have no `::1` etc, but I *think* a bind of `[::]` should still work, you'd just not get any v6 binds by that, so it's the same as `0.0.0.0`.

PR #851, opened 2024-05-23
Add usage report command

ares:
This is the PoC of generating yaml representing the usage of the instance. Test wit

```
foreman-maintain report generate
```

Currently, the command searches for all ReportChecks, executes them and collects the data from them. Then it prints out the resulting YAML to stdout.


lib/foreman_maintain/cli.rb
```diff
@@ -33,6 +34,7 @@ class MainCommand < Base
       subcommand 'self-upgrade', 'Perform major version self upgrade', SelfUpgradeCommand
       subcommand 'maintenance-mode', 'Control maintenance-mode for application',
         MaintenanceModeCommand
+      subcommand 'report', 'Generate the usage reports', ReportCommand
```
evgeni, line 37:
```suggestion
      subcommand 'report', 'Generate usage report', ReportCommand
```

I guess it's *one* report generated per run? :)

lib/foreman_maintain/cli/report_command.rb
```diff
@@ -0,0 +1,25 @@
+module ForemanMaintain
+  module Cli
+    class ReportCommand < Base
+      extend Concerns::Finders
+
+      option '--output', 'FILE', 'Output the generate report into FILE'
+      subcommand 'generate', 'Generates the usage reports' do
```
evgeni, line 7: having just one subcommand feels unnatural, but I guess this is preparation to also have `upload` or something?

lib/foreman_maintain/report.rb
```diff
@@ -0,0 +1,59 @@
...
+    include Concerns::Metadata
+    include Concerns::Finders
+
+    attr_accessor :data
+
+    def sql_count(sql, column: '*', cte: '')
+      sql_as_count("COUNT(#{column})", sql, cte: cte)
+    end
+
+    def sql_as_count(selection, sql, cte: '')
```
evgeni, line 14: what does `cte` stand for?

test/lib/concerns/os_facts_test.rb
```diff
@@ -8,13 +8,13 @@ module ForemanMaintain
   describe Concerns::OsFacts do
     let(:system) { FakeSystem.new }
 
-    it 'should report memory from /proc/meminfo' do
+    it 'should reports memory from /proc/meminfo' do
```
evgeni, line 11: nope

lib/foreman_maintain/cli/report_command.rb
```diff
@@ -0,0 +1,25 @@
...
+        def execute
+          scenario = run_scenario(Scenarios::Report::Generate.new({}, [:reports])).first
+
+          # description can be used too
+          report_data = scenario.steps.map(&:data).compact.reduce(&:merge).transform_keys(&:to_s)
```
evgeni, line 12: does that take all reports, their `.data` and merge those hashes? what if two reports accidentally define the same hash key?

definitions/reports/external_auth_source.rb
```diff
@@ -0,0 +1,28 @@
...
+        data["last_login_on_through_external_auth_source_in_days"] = nil
+
+        sql = <<~SQL
+          SELECT users.* FROM users
+          INNER JOIN auth_sources ON (auth_sources.id = users.auth_source_id)
+          WHERE auth_sources.type = 'AuthSourceExternal' AND users.last_login_on IS NOT NULL
+          ORDER BY users.last_login_on DESC
+        SQL
+        users = feature(:foreman_database).query(sql)
+        if (user = users.first)
```
evgeni, line 21:
```suggestion
        user = feature(:foreman_database).query(sql).first
        if (user)
```
isn't that the same?

definitions/reports/inventory.rb
```diff
@@ -0,0 +1,92 @@
...
+      metadata do
+        description 'Facts about hosts and the rest of the inventory'
+      end
+
+      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      def run
+        # Hosts
+        hosts_by_type_count =
+          feature(:foreman_database).
+          query("select type, count(*) from hosts group by type").
```
evgeni, line 14: I think `Report` should have a `query` itself, no need to call `feature()` all the time

definitions/reports/ldap_auth_source.rb
```diff
@@ -0,0 +1,75 @@
...
+          record_flavor_usage(flavor)
+        end
+      end
+
+      private
+
+      # rubocop:disable Metrics/AbcSize
+      def record_flavor_usage(flavor)
+        query_base = query_base(flavor)
+        data["ldap_auth_source_#{flavor}_count"] = sql_count(query_base)
```
evgeni, line 28: is the `query_base` inside `sql_count` the variable? the method? ;)

lib/foreman_maintain/cli/report_command.rb
ares: yes, this is open ended atm but we would like to have some direct upload in near future

lib/foreman_maintain/cli/report_command.rb
ares: Yes. I think that's very unlikely as the keys are very descriptive. Is it worth printing a warning to STDERR? Or perhaps better, having a some sort of a test that check for all keys in all checks? That may be a bit harder though as some of them are generated dynamically. Perhaps each check could have some explicit prefix that could that be verified, but that's likely schema v2 thing.

definitions/reports/ldap_auth_source.rb
ares: you mean it's better to rename the variable to something else so it's not confusing?

definitions/reports/external_auth_source.rb
ares: yeah I guess `users` was used below in some earlier version, good catch, fixing

definitions/reports/external_auth_source.rb
```diff
@@ -0,0 +1,28 @@
...
+      def run
+        self.data = {}
+        # nil means no user linked to external auth source ever logged in
+        data["last_login_on_through_external_auth_source_in_days"] = nil
+
+        sql = <<~SQL
+          SELECT users.* FROM users
+          INNER JOIN auth_sources ON (auth_sources.id = users.auth_source_id)
+          WHERE auth_sources.type = 'AuthSourceExternal' AND users.last_login_on IS NOT NULL
+          ORDER BY users.last_login_on DESC
```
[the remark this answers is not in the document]
ares: :+1:

definitions/reports/ldap_auth_source.rb
evgeni: yes

lib/foreman_maintain/cli.rb
ares: right

lib/foreman_maintain/report.rb
ares: introduced in fd7c0c7428ddf4d40e9945d19c47e1d2e480becf , I guess @adamruzicka would know the best :-)

lib/foreman_maintain/cli/report_command.rb
evgeni: What if we add an `add_data(key, value)` method to `Report`, and it does something like
```ruby
data[self.class.name+'_'+key] = value
```

I wouldn't block on this, but I think it's a potential source for hard to debug issues once a "not so descriptive" name leaks into the set.

definitions/reports/inventory.rb
ares: nope, undefined method #query

lib/foreman_maintain/cli/report_command.rb
ares: @adamruzicka @ShimShtein thoughts?

definitions/reports/inventory.rb
evgeni: I said "should", not "does" :)
As in: we should implement a `Report.query` and use that, instead of calling `feature(:db).query` all the  time ;)

lib/foreman_maintain/cli/report_command.rb
evgeni: It would also save the `transform_keys(to_s)` here, which makes the code more readable?

ares:
I address most of the comments, rebased the PR and added one more report definition around RBAC

test/lib/cli_test.rb
```diff
@@ -31,6 +31,7 @@ module ForemanMaintain
             plugin                        Manage optional plugins on your server
             self-upgrade                  Perform major version self upgrade
             maintenance-mode              Control maintenance-mode for application
+            report                        Generate the usage reports
```
evgeni, line 34:
```suggestion
            report                        Generate usage report
```

you updated the string on my suggestion, but not the test

lib/foreman_maintain/report.rb
adamruzicka: [common table expression(s)](https://www.postgresql.org/docs/current/queries-with.html)

evgeni:
Can we get some basic tests for the new `Report` class you're introducing and the new scenario you're adding?
Tests for the individual reports are probably overkill (and would result in a lot of mocking, so doubtfully add any value)

lib/foreman_maintain/report.rb
evgeni: TIL, thanks

test/lib/cli_test.rb
adamruzicka: https://github.com/ares/foreman_maintain/pull/15

definitions/reports/inventory.rb
adamruzicka: https://github.com/ares/foreman_maintain/pull/15

lib/foreman_maintain/cli/report_command.rb
adamruzicka: I'd leave this for v2. While we're at it, we should maybe shuffle things around a bit, the report classes are currently named somewhat arbitrarily

definitions/reports/compliance.rb
```diff
@@ -0,0 +1,30 @@
+module Checks
```
ehelms, line 1: These are all in the module Checks, but they live in `definitions/reports`. I think you should drop this line. If you want to differentiate checks from some other type for the report feature, then invert these, e.g.

```
module Report
  module Checks
```

definitions/reports/inventory.rb
```diff
@@ -0,0 +1,92 @@
...
+            <<-SQL
+              select fact_names.type,
+                     min(fact_values.updated_at) as min_update_time,
+                     max(fact_values.updated_at) as max_update_time,
+                     count(fact_values.id) as values_count
+              from fact_values inner join fact_names on fact_name_id = fact_names.id
+              group by fact_names.type
+            SQL
+          ).
+          to_h { |row| [row['type'].sub('FactName', ''), to_fact_hash(row)] }
```
ehelms, line 42: Be nice if each of these chunks of data were their own method instead of one giant method.

definitions/reports/oidc_usage.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      # Do you use OIDC/keycloak?
+      def run
+        oidc_issuer = sql_setting('oidc_issuer')
+        result = if oidc_issuer
+                   loaded = YAML.load(oidc_issuer)
+                   loaded.is_a?(String) && loaded != ''
+                 end
+
+        self.data = { oidc_use: !!result }
```
ehelms, line 16: Why is double bang used here?

definitions/reports/rbac.rb
```diff
@@ -0,0 +1,61 @@
...
+      metadata do
+        description 'Checks the RBAC use'
+      end
+
+      # How many users do you have in the system?
+      # How many non-admin users do you have?
+      # How many custom roles did you create and assigned to users?
+      # rubocop:disable Layout/LineLength
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/AbcSize
```
ehelms, line 13: Break this method up instead of disabling rubocop. This could also help with the comments listed in these files. You can use ruby to help, e.g. 

```
def number_of_users
def number_of_non_admin_users
def number_of_custom_roles
```

lib/foreman_maintain/cli/report_command.rb
ehelms: If you keep something big like this I would find it more readable to split it across lines:

```
          report_data = scenario.
            steps.
            map(&:data).
            compact.
            reduce(&:merge).
            transform_keys(&:to_s)
```

lib/foreman_maintain/report.rb
ehelms: And will forget. I hate to be pedantic.. actually I don;t hate it, maybe hate it for you :) Can we spell that out somehow please? The meaning of `cte` will be lost when someone has to read this later in life.

lib/foreman_maintain/report.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+      end
+    end
+
+    def table_exists(table)
+      sql_count("information_schema.tables WHERE table_name = '#{table}'").positive?
+    end
+
+    def run
+      raise NotImplementedError
```
ehelms, line 43: Fun fact I learned and stuck with me -- https://kirillshevch.medium.com/misusing-of-notimplementederror-in-ruby-531dd866f461

Yes this code base already does it, but sharing is caring.

ares:
All comments should be addressed atm, could you please re-review/merge?

definitions/reports/virt_who.rb
```diff
@@ -0,0 +1,21 @@
+module Reports
+  class Virtwho < ForemanMaintain::Report
+    metadata do
+      description 'Check if virt-who is being used and what hypervisor types are present'
+    end
+
+    HYPERVISOR_TYPES = %w[ahv esx fakevirt hyperv kubevirt libvirtd].freeze
+
+    def run
+      self.data = {}
```
evgeni, line 10:
```suggestion
```

`data_field` will do that for us

lib/foreman_maintain/report.rb
```diff
@@ -0,0 +1,75 @@
...
+          result[new_key] = value
+        end
+      end
+    end
+
+    def table_exists(table)
+      sql_count("information_schema.tables WHERE table_name = '#{table}'").positive?
+    end
+
+    def data_field(field_name)
```
evgeni, line 42: this was introduced (thanks!) but is not consistently used across the codebase

definitions/reports/template
```diff
@@ -0,0 +1,20 @@
...
+      end
+
+      def run
+        # make sure to populate data with the hash, that will be merged into the result set
+        self.data = { template: 5 }
+
+        # SQL example
+        # count = feature(:foreman_database).query("SELECT COUNT(*) FROM hosts WHERE managed = true AND created_at >= current_date - interval '3 months'")
+        # result = count.first['count'].to_i
+        # self.data = result
```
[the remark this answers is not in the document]
ShimShtein: It's a template. I didn't change it.

definitions/reports/virt_who.rb
ShimShtein: The `each` statement still uses it.
I didn't have the time to think about how to pass context parameters to the data_field/merge_data methods.

lib/foreman_maintain/report.rb
ShimShtein: There are a couple of places where I need to pass context between queries. I couldn't use the simple methods there.

definitions/reports/template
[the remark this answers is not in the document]
evgeni: shouldn't the template promote the best practices? :)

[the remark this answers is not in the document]
ShimShtein: Done, ready for another round

evgeni:
Are the remaining open points (method length in rbac.rb, NotImplementedError in report.rb, Report#query method and tests) planned to be addressed?

adamruzicka:
> Are the remaining open points (method length in rbac.rb, NotImplementedError in report.rb, Report#query method and tests) planned to be addressed?

I filed https://github.com/theforeman/foreman_maintain/issues/976 to track all of those. Could we accept this PR as is and resolve those in additional PRs so that we can save the additional round-trip through @ares's fork?

definitions/reports/platform.rb
```diff
@@ -0,0 +1,104 @@
...
+      data_field('custom_roles_count') { sql_count('roles where origin = null') }
+
+      merge_data('taxonomies_counts') do
+        feature(:foreman_database).
+          query("select type, count(*) from taxonomies group by type").
+          to_h { |row| [row['type'], row['count'].to_i] }
+      end
+    end
+
+    def settings_fields
```
evgeni, line 61: some settings (like `db_pending_seed`, `instance_id` and `bcrypt_cost`) are always modified -- is it worth logging them here or should they be filtered?

lib/foreman_maintain/cli/report_command.rb
evgeni, line 12: This produces quite a lot of output, which you probably don't want in the case where you run this via cron (as otherwise you'll get a mail from cron on every invocation. You could add `--cron` or `--silent` to suppress these?

evgeni (approved):
I'm fine with this a a first round as suggested by Adam

definitions/reports/rbac.rb
```diff
@@ -0,0 +1,59 @@
...
+        " INNER JOIN auth_sources ON auth_sources.id = users.auth_source_id" +
+        " WHERE ((users.admin = FALSE OR users.admin IS NULL) AND (usergroups.admin = FALSE OR usergroups.admin IS NULL))" +
+        " AND auth_sources.name != 'Hidden'")
+      result["non_admin_users_count"] = count
+
+      role_ids = feature(:foreman_database).query("SELECT id FROM roles WHERE roles.builtin != 2 AND roles.origin IS NULL")
+      result["custom_roles_count"] = role_ids.size
+      role_ids = role_ids.flat_map(&:values)
+      count = sql_count("cached_user_roles WHERE cached_user_roles.role_id IN (#{role_ids.join(',')})")
+      result["assigned_custom_roles_count"] = count
```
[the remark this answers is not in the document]
adamruzicka: Apparently, the query we generate isn't valid if there are no custom roles. Good catch
```
> query
=> " SELECT COUNT(*) AS COUNT FROM cached_user_roles WHERE cached_user_roles.role_id IN ()"

> feature(:foreman_database).query(query)
CSV::MalformedCSVError: Illegal quoting in line 1.
```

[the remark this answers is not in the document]
evgeni: Yeah, `IN ()` is not valid:

```
foreman=# SELECT COUNT(*) AS COUNT FROM cached_user_roles WHERE cached_user_roles.role_id IN ();
ERROR:  syntax error at or near ")"
LINE 1: ...ROM cached_user_roles WHERE cached_user_roles.role_id IN ();
                                                                     ^
```

PR #855, opened 2024-05-28
drop dynflowd support

evgeni:
we do not deploy dynflowd, only dynflow_sidekiq since
https://projects.theforeman.org/issues/28421 (RPM)
https://projects.theforeman.org/issues/28068 (DEB)


adamruzicka (approved):
lgtm

PR #857, opened 2024-05-29
make backup scenario more readable

evgeni:


definitions/scenarios/backup.rb
```diff
@@ -149,8 +140,8 @@ class BackupRescueCleanup < ForemanMaintain::Scenario
     end
 
     def compose
-      add_step_with_context(Procedures::Service::Start) if strategy != :online
-      add_steps_with_context(find_procedures(:maintenance_mode_off)) if strategy != :online
+      add_step_with_context(Procedures::Service::Start) if strategy == :offline
+      add_steps_with_context(find_procedures(:maintenance_mode_off)) if strategy == :offline
```
ekohl, line 144: Often I'm not a fan of guard clauses. Here they repeat so while you're at it, is this better?
```suggestion
      if strategy == :offline
        add_step_with_context(Procedures::Service::Start)
        add_steps_with_context(find_procedures(:maintenance_mode_off))
      end
```

ekohl (commented):
I'd be tempted to remove the `online_backup?` helper as well. The other code implies there could be more strategies (since it's a case statement), which could imply that `!online_backup?` isn't always equal to `strategy == :offline`.

evgeni:
> I'd be tempted to remove the `online_backup?` helper as well. The other code implies there could be more strategies (since it's a case statement), which could imply that `!online_backup?` isn't always equal to `strategy == :offline`.

Well, there used to be, but I really would like to only have as few as possible.

Done

ehelms (approved):

PR #859, opened 2024-05-31
Fix self-upgrade by adding enabled-repos support to dnf

ehelms:


evgeni:
While I like the refactoring, I do not understand what this is fixing.

Previously, self upgrade would self (lol) calculate the needed DNF options for `--enablerepo`, now the handling has been moved to the DNF helper itself, which is nice but should result in the very same dnf invocation?

definitions/scenarios/self_upgrade.rb
```diff
@@ -67,11 +67,8 @@ class SelfUpgrade < SelfUpgradeBase
     def downstream_self_upgrade(pkgs_to_update)
       ForemanMaintain.enable_maintenance_module
 
-      dnf_options = req_repos_to_update_pkgs.map do |id|
-        "--enablerepo=#{id}"
-      end
       add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true,
-        dnf_options: dnf_options))
```
ehelms, line 74: This should answer your question @evgeni -- we removed this option and so this is broken entirely :)

evgeni: Oooh, so this should have a

Fixes: 9ac8a20cda789c0d733396f0c92bf9b45d25901c

evgeni (approved):
It'd be cool if the commit message would refer to the commit that broke it ;)

PR #860, opened 2024-05-31
Fail self-upgrade if current and target version are different

ehelms:


definitions/features/satellite.rb
```diff
@@ -10,6 +10,10 @@ class Features::Satellite < ForemanMaintain::Feature
     end
   end
 
+  def target_version
```
ehelms, line 13: I intend in future updates not to replicate this in multiple places but rather to use this as the canonical location and make other places reference this (such as upgrade scenario).

definitions/scenarios/self_upgrade.rb
```diff
@@ -65,13 +75,26 @@ class SelfUpgrade < SelfUpgradeBase
     end
 
     def downstream_self_upgrade(pkgs_to_update)
+      unless upgrade_allowed?
+        if version_difference > 1
+          raise(
+            ForemanMaintain::Error::Warn,
+            "foreman-maintain is too many versions ahead, given the" \
+            "target version is #{target_version}. Please rollback " \
+            "foreman-maintain to the proper version."
```
ehelms, line 84: This message could use some word smithing to help get the right point across. I debated including a note about distrosync? That may only apply to EL systems?

definitions/scenarios/self_upgrade.rb
```diff
@@ -65,13 +75,26 @@ class SelfUpgrade < SelfUpgradeBase
...
+          )
+        else
+          raise(
+            ForemanMaintain::Error::Warn,
+            "self-upgrade has already been executed, and is ready to upgrade to #{target_version}"
```
ehelms, line 89: We could opt to instead run `dnf update` to update to the latest foreman-maintain from the maintenance repository if for some reason the user self-upgraded and then a new maintain was released? That seems rather corner case but perhaps more in-line?

definitions/scenarios/self_upgrade.rb
Griffin-Sullivan: Maybe adding a note about what the current version is so they can see the difference? Like make it more obvious that the version is < X.Y-1

definitions/scenarios/self_upgrade.rb
Griffin-Sullivan: Seems like a safe suggestion in case someone ran self upgrade but then left it alone for a while

Griffin-Sullivan (approved):

definitions/scenarios/self_upgrade.rb
```diff
@@ -4,11 +4,21 @@ class SelfUpgradeBase < ForemanMaintain::Scenario
     include ForemanMaintain::Concerns::Versions
 
     def target_version
-      @target_version ||= Gem::Version.new(current_version).bump.to_s
+      if feature(:instance).downstream
+        feature(:satellite).target_version
+      else
+        feature(:foreman_install).target_version
+      end
```
evgeni, line 11: this will break upgrades on capsules, as those do not have these features.

definitions/features/foreman_install.rb
```diff
@@ -9,4 +9,16 @@ class Features::ForemanInstall < ForemanMaintain::Feature
...
+    '3.11'
+  end
+
+  def current_version
+    @current_version ||= package_version(package_name)
+  end
+
+  def package_name
+    'foreman'
+  end
```
evgeni, line 23: this is already present in `foreman_server` feature, seems like duplication?
(and what about katello?)

definitions/scenarios/self_upgrade.rb
```diff
@@ -47,12 +57,12 @@ def req_repos_to_update_pkgs
...
 
-    def upstream_target_version
-      if feature(:katello_install)
-        return foreman_version_by_katello(target_version)
-      else
-        target_version
-      end
+    def upgrade_allowed?
+      Gem::Version.new(current_version) == Gem::Version.new(target_version)
+    end
```
evgeni, line 62: don't we need `current_version == target_version - 1` to allow  upgrades?

evgeni: oooh, wait. you want to say "if you have the foreman maintain that supports the upgrade to `current_version`, your foreman maintain is too old to upgrade to the next version and thus this is the only time you're allowed to install "next" foreman maintain"?

definitions/features/foreman_install.rb
ehelms: I don't know what the difference between the two features are and I didn't want to find out for this change :P

definitions/scenarios/self_upgrade.rb
ehelms: Yessir.

evgeni: I think I would have been less confused if the method would be `self_upgrade_required?`.

ehelms: Is it required though? Or simply possible? I think that naming ends up confusing me :)

evgeni: Well, I read is as "required" as otherwise it would have the wrong "target_release" and thus handle things wrongly?

ehelms: What other code does with this information is different than what is this method telling us. At least, that was my design desire - to have a method that could tell me if self-upgrade is possible.

definitions/scenarios/self_upgrade.rb
ehelms: @evgeni do you concur?

evgeni: You mean something like `dnf --disablerepo=* --enablerepo=NEW_MAINT rubygem-foreman_maintain satellite-maintain`? Yeah… I could see this being useful.

evgeni: So, uh, actually exactly what we already do for "self upgrade" a few lines later.

definitions/scenarios/self_upgrade.rb
```diff
@@ -65,6 +65,23 @@ class SelfUpgrade < SelfUpgradeBase
     end
 
     def downstream_self_upgrade(pkgs_to_update)
+      unless self_upgrade_allowed?
+        if version_difference > 1
```
evgeni, line 69: Do we need the complicated difference computation, or would the following suffice?
```suggestion
        if Gem::Version.new(current_version).bump != Gem::Version.new(target_version)
```

definitions/scenarios/self_upgrade.rb
```diff
@@ -47,12 +47,12 @@ def req_repos_to_update_pkgs
...
     end
 
-    def upstream_target_version
-      if feature(:katello_install)
-        return foreman_version_by_katello(target_version)
-      else
-        target_version
-      end
+    def self_upgrade_allowed?
+      Gem::Version.new(current_version) == Gem::Version.new(target_version)
```
evgeni, line 51: `current_version` is `X.Y.Z` while `target_version` is `X.Y`. So while this will work for `Z=0`, it won't for others.

definitions/features/instance.rb
```diff
@@ -91,6 +91,24 @@ def upstream_install
...
+              else
+                feature(:foreman_install).current_version
+              end
+
+    version.to_s[/^\d+\.\d+\.\d+/]
+  end
+
+  def target_version
+    if feature(:instance).downstream
+      Features::Satellite.new.target_version
```
evgeni, line 106:
```suggestion
      feature(:satellite).target_version
```

evgeni: what about capsules?

ehelms: `Features::Satellite.new.target_version` is intentional - I want the target_version from this class and I do not want it to try to autodetect and fail this feature being present. I am trying to avoid defining target_version for Satellite in two places. 

The `target_version` being mocked was a good call out, that did not need to happen anymore and I updated the tests so that we can see it happen in Capsule and Satellite scenarios properly.

evgeni: Just to have it in writing what we talked about: this is not obvious from reading the code, especially as the `else` branch uses `feature()` again. I'll be OK if both branches use the same notation.

evgeni (approved):

PR #863, opened 2024-06-03
refresh collation version when pulp_ansible is installed

evgeni:


definitions/procedures/restore/reindex_databases.rb
```diff
@@ -16,6 +17,11 @@ def run
 
         spinner.update('Reindexing the databases')
         execute!('runuser - postgres -c "reindexdb -a"')
+        if check_min_version('python3.11-pulp-ansible', '0.20.0')
+          execute!('runuser -c '\
+                   '\'echo "ALTER COLLATION pulp_ansible_semver REFRESH VERSION;"'\
+                   '| psql pulpcore\' postgres')
```
ekohl, line 23: This isn't needed for remote DBs?

ekohl (commented):
Do you have a reference on why this is needed?

definitions/procedures/restore/reindex_databases.rb
evgeni: Not when they don't switch the OS :)

evgeni:
> Do you have a reference on why this is needed?

pulp_ansible ships an "own" collation since https://github.com/pulp/pulp_ansible/commit/d442dded1e4831e4788d4f79007ebc7b0c3df0ea, which requires a "refresh" of the version (after the reindex) if the underlying collation library (icu) changes, which happens during 8to9. otherwise you get "WARNING:  collation "pulp_ansible_semver" has version mismatch" in the postgres logs.

we do the same in leapp.

ehelms (approved):

ekohl (approved):
I don't see why a custom collation is needed, but given it's used I think this is correct.

PR #871, opened 2024-06-04
refactor restoring of online dumps

evgeni:


evgeni:
contains #870

test/definitions/scenarios/restore_test.rb
```diff
@@ -51,6 +51,41 @@ module Scenarios
         ForemanMaintain::Utils::Backup.any_instance.stubs(:different_source_os?).returns(true)
         refute_scenario_has_step(scenario, Procedures::Restore::ReindexDatabases)
       end
+
+      it 'drops and restores DB dumps if present' do
```
evgeni, line 55: these tests pass both on master and here in the PR :)

PR #873, opened 2024-06-05
don't run restorecon before restoring a backup

evgeni:


ekohl (commented):
This was added in 1ae71c3ad827503eb370310a6b27232f80289351, but I have a hard time thinking of why this would be needed.

Perhaps this has to do with parent directories where children can inherit incorrect labels?

evgeni:
my basic test (`ansible-playbook pipelines/clone_pipeline.yml -e pipeline_os=centos9-stream -e pipeline_type=clone -e pipeline_version=nightly -e '{"packit_prs":["theforeman/foreman_maintain/873"]}') passed, so I call this OK :)

ekohl (approved):
If it was after restoring, I'd be a lot more understanding. Let's try this.

PR #877, opened 2024-06-11
Only reindex container metadata on Foreman server

sjha4:
We need to confine katello db indexing to run only on Satellite and not capsules. Refactoring a little to do that.

### DIFF
--- a/definitions/procedures/repositories/index_katello_repositories_container_metadata.rb
+++ b/definitions/procedures/repositories/index_katello_repositories_container_metadata.rb
@@ -1,0 +1,17 @@
+module Procedures::Repositories
+  class IndexKatelloRepositoriesContainerMetatdata < ForemanMaintain::Procedure
+    metadata do
+      description 'Import container manifest metadata'
+      confine do
+        feature(:katello)
+      end
+    end
+
+    def run
+      with_spinner(('Adding image metadata. You can continue using the ' \
+                    'system normally while the task runs in the background.')) do
+        execute!('foreman-rake katello:import_container_manifest_labels')
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


definitions/procedures/repositories/index_katello_repositories_container_metadata.rb
```diff
@@ -0,0 +1,17 @@
+module Procedures::Repositories
+  class IndexKatelloRepositoriesContainerMetatdata < ForemanMaintain::Procedure
+    metadata do
+      description 'Import container manifest metadata into Katello'
+      confine do
+        feature(:satellite)
```
ehelms, line 6: Is this procedure only applicable to Satellite? The description and text below both use 'Katello'. But if this is something a user of Katello can/should also be having run on a Foreman then we need to update the confines.

definitions/procedures/repositories/index_katello_repositories_container_metadata.rb
```diff
@@ -0,0 +1,17 @@
+module Procedures::Repositories
+  class IndexKatelloRepositoriesContainerMetatdata < ForemanMaintain::Procedure
+    metadata do
+      description 'Import container manifest metadata into Katello'
```
ehelms, line 4:
```suggestion
      description 'Import container manifest metadata'
```

definitions/procedures/repositories/index_katello_repositories_container_metadata.rb
```diff
@@ -0,0 +1,17 @@
...
+      confine do
+        feature(:katello)
+      end
+    end
+
+    def run
+      with_spinner(('Adding image metadata to katello. You can continue using the ' \
```
ehelms, line 11:
```suggestion
      with_spinner(('Adding image metadata. You can continue using the ' \
```

ehelms:
You might consider commit message like 'Only reindex container metadata on Foreman server'

sjha4:
@ehelms : Thanks..Updated PR and commit message.. ❤️

ehelms (approved):

PR #879, opened 2024-06-12
Single target version

ehelms:


ehelms:
This includes https://github.com/theforeman/foreman_maintain/pull/763 and should go in after it.

PR #888, opened 2024-06-25
don't fail when chown to postgres fails

evgeni:
- **add prepare directory tests**
- **move chown postgres inside unless preserve**
- **only chown to postgres group for online backups with local DB**
- **don't fail when chown to postgres fails**

### DIFF
--- a/definitions/procedures/backup/prepare_directory.rb
+++ b/definitions/procedures/backup/prepare_directory.rb
@@ -9,6 +9,7 @@ module Procedures::Backup
       param :online_backup, 'Select for online backup', :flag => true, :default => false
     end
 
+    # rubocop:disable Metrics/MethodLength
     def run
       puts "Creating backup folder #{@backup_dir}"
 
@@ -17,7 +18,16 @@ module Procedures::Backup
         FileUtils.chmod_R 0o770, @backup_dir
 
         if feature(:instance).postgresql_local? && @online_backup
-          FileUtils.chown_R(nil, 'postgres', @backup_dir)
+          begin
+            FileUtils.chown_R(nil, 'postgres', @backup_dir)
+          rescue Errno::EPERM
+            warn_msg = <<~MSG
+              #{@backup_dir} could not be made readable by the 'postgres' user.
+              This won't affect the backup procedure, but you have to ensure that
+              the 'postgres' user can read the data during restore.
+            MSG
+            set_status(:warning, warn_msg)
+          end
         end
       end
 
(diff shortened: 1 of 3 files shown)


definitions/procedures/backup/prepare_directory.rb
```diff
@@ -6,18 +6,29 @@ class PrepareDirectory < ForemanMaintain::Procedure
...
+              #{@backup_dir} could not be made readable by the 'postgres' user.
+              This won't affect the backup procedure, but you have to ensure that
+              the 'postgres' user can read the data during restore.
+            MSG
+            set_status(:warning, warn_msg)
+          end
+        end
       end
 
       FileUtils.rm(Dir.glob(File.join(@backup_dir, '.*.snar'))) if @preserve_dir
```
ekohl, line 34: Should this be in the `else` part of `unless @preserve_dir`?

ekohl (approved):

definitions/procedures/backup/prepare_directory.rb
evgeni: Maybe. But honestly I didn't want to touch too much of the code, just the parts that I actually wanted to change.

Also, I am not really sure why this is guarded by `if @preserve_dir` anyway.
As when it's `false`, the `@backup_dir` will be fresh and empty and the `.glob` will just return `[]`.

definitions/procedures/backup/prepare_directory.rb
```diff
@@ -6,18 +6,29 @@ class PrepareDirectory < ForemanMaintain::Procedure
...
       param :preserve_dir, 'Directory where to backup to', :flag => true
       param :incremental_dir, 'Changes since specified backup only'
+      param :online_backup, 'Select for online backup', :flag => true, :default => false
     end
 
+    # rubocop:disable Metrics/MethodLength
     def run
       puts "Creating backup folder #{@backup_dir}"
 
       unless @preserve_dir
```
ekohl, line 16: I think this would actually make a lot more sense, but I can understand why you don't want to touch those too much
```suggestion
      if @preserve_dir
        FileUtils.rm(Dir.glob(File.join(@backup_dir, '.*.snar')))
      else
        puts "Creating backup folder #{@backup_dir}"
```

evgeni: (moved the puts, but not the `rm`)

I think what my brain tries to fight is the fact that "not preserve_dir" is the *primary* mode of operation of this code, so it should come *first* vs the else branch that is, well, "else" ;-)

PR #891, opened 2024-06-26
ensure that the backup types match when doing incremental backups

evgeni:


evgeni:
```
# foreman-maintain backup offline --incremental /var/off-size/katello-backup-2024-06-26-10-03-42/ /var/lol
Starting backup: 2024-06-26 10:32:22 +0000
Make sure Foreman DB is up: 
Make sure Candlepin DB is up: 
- Checking connection to the Candlepin DB                             [OK]      
Make sure Pulpcore DB is up: 
\ Checking connection to the Pulpcore DB                              [OK]      


Running Backup
Check if the incremental backup has the right type:                   [FAIL]
The existing backup has PostgreSQL as a tarball, but the new one will have a dump.
Scenario [Backup] failed.

  [backup-incremental-parent-type]

--whitelist="backup-incremental-parent-type"



Running Failed backup cleanup
Start applicable services: 

redis, postgresql, pulpcore-api, pulpcore-content, pulpcore-worker@1.service, pulpcore-worker@2.service, tomcat, dynflow-sidekiq@orchestrator, foreman, httpd, dynflow-sidekiq@worker-1, dynflow-sidekiq@worker-hosts-queue-1, foreman-proxy
- All services started                                                [OK]      
Clean up backup directory:                                            [OK]

Done with backup: 2024-06-26 10:32:23 +0000
Backup didn't finish. Incomplete backup was removed.
```

Gotta find out how to move that check to "prep"

definitions/checks/backup/incremental_parent_type.rb
```diff
@@ -0,0 +1,31 @@
...
+      param :sql_tar, 'Will backup include PostgreSQL tarball', :flag => true, :default => false
+      manual_detection
+    end
+
+    def run
+      return unless @incremental_dir
+
+      backup = ForemanMaintain::Utils::Backup.new(@incremental_dir)
+
+      existing_type = backup.online_backup? ? 'online' : 'offline'
```
ehelms, line 17: How much are the words online and offline used in the code? This feels like a spot that could be prone to typos and failures - consider defining a constant somewhere that can be used.

definitions/checks/backup/incremental_parent_type.rb
```diff
@@ -0,0 +1,31 @@
...
+      new_type = @online_backup ? 'online' : 'offline'
+      msg = "The existing backup is an #{existing_type} one, but an #{new_type} one was requested."
```
ehelms, line 19:
```suggestion
      msg = "The existing backup is an #{existing_type} backup, but an #{new_type} backup was requested."
```

definitions/scenarios/backup.rb
```diff
@@ -14,6 +14,14 @@ class Backup < ForemanMaintain::Scenario
       param :proxy_features, 'List of proxy features to backup (default: all)', :array => true
       param :skip_pulp_content, 'Skip Pulp content during backup'
       param :tar_volume_size, 'Size of tar volume (indicates splitting)'
+
+      preparation_steps do
+        Checks::Backup::IncrementalParentType.new(
```
ehelms, line 19: Could this just be the first step in the compose method?

definitions/checks/backup/incremental_parent_type.rb
evgeni: three times "backup" in one sentence?

definitions/scenarios/backup.rb
evgeni: that's what I had before (and the current one doesn't work anyway), but it feels odd that it shows "preparation steps", those succeed, then it foes "now doing backup" and the first step fails the sanity check

definitions/checks/backup/incremental_parent_type.rb
evgeni: In which code? this check? or all of FM?

ehelms: I was thinking about the latter.

definitions/checks/backup/incremental_parent_type.rb
ehelms: Just don't say it in front of a mirror. Better to be explicit?

evgeni:
@ehelms ready!

definitions/checks/backup/incremental_parent_type.rb
```diff
@@ -0,0 +1,36 @@
+module Checks::Backup
+  class IncrementalParentType < ForemanMaintain::Check
+    metadata do
+      description 'Check if the incremental backup has the right type'
+      tags :backup
+      param :incremental_dir, 'Path to existing backup directory'
+      param :online_backup, 'Select for online backup', :flag => true, :default => false
```
ehelms, line 7: Is the supposition that if this is not set, it defaults to offline backup?

definitions/checks/backup/incremental_parent_type.rb
```diff
@@ -0,0 +1,36 @@
...
+      param :sql_tar, 'Will backup include PostgreSQL tarball', :flag => true, :default => false
+      manual_detection
+    end
+
+    def run
+      return unless @incremental_dir
```
ehelms, line 13: This parameter is required right? So can this ever be true?

ehelms (approved):

definitions/checks/backup/incremental_parent_type.rb
evgeni: Yes

definitions/checks/backup/incremental_parent_type.rb
evgeni: This parameter can be `nil` (as in: set to `nil` by the caller) when we're creating a non-incremental (=base) backup

PR #895, opened 2024-06-27
Drop RPM changelog procedure

ehelms:


ianballou (approved):
Confirmed with the Pulp team that this is no longer needed. It's no longer used anywhere else in foreman-maintain. ACK.

PR #898, opened 2024-07-04
Auto-detect if colors are supported

ekohl:
Not all output supports colors. For example, cron doesn't. This copies the autodetection from Kafo to disable colors if needed.

Currently untested. I'd also like to ask others for input: do we want (like the installer) `--[no-]color` flags? Should it have a config option?

### DIFF
--- a/lib/foreman_maintain/config.rb
+++ b/lib/foreman_maintain/config.rb
@@ -25,6 +25,11 @@ module ForemanMaintain
       @foreman_port = @options.fetch(:foreman_port, 443)
     end
 
+    def use_color?
+      ENV['TERM'] && ENV.fetch('NO_COLOR', '') == '' && \
+        system('command -v tput', out: File.open('/dev/null')) && `tput colors`.to_i > 0
+    end
+
     private
 
     def load_log_configs
--- a/test/lib/cli/health_command_test.rb
+++ b/test/lib/cli/health_command_test.rb
@@ -40,10 +40,10 @@ module ForemanMaintain
           [dummy-check-fail-skipwhitelist] Check that ends up with fail
           [dummy-check-success] Check that ends up with success
           [dummy-check-warn] Check that ends up with warning
-          [external-service-is-accessible] External_service_is_accessible         [pre-upgrade-check]
-          [present-service-is-running] Present service run check                  [default]
-          [service-is-stopped] Service not running check                          [default]
-          [upgrade-post-upgrade-check] Procedures::Upgrade::PostUpgradeCheck      [post-upgrade-checks]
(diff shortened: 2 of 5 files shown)


lib/foreman_maintain/config.rb
```diff
@@ -25,6 +25,11 @@ def initialize(options)
       @foreman_port = @options.fetch(:foreman_port, 443)
     end
 
+    def use_color?
+      # TODO: read config option?
+      ::ENV['TERM'] && !`which tput 2> /dev/null`.empty? && `tput colors`.to_i > 0
```
evgeni, line 30: `command -v` instead of `which` please

evgeni:
I think having a cli option is nice (but not a must-have), and I don't think a config option is necessary.

lib/foreman_maintain/config.rb
ekohl: Then we should update kafo too because I copied it from there

evgeni: Quite possibly, yeah.

ekohl: Now that I'm not on mobile I can easily find it: https://github.com/theforeman/kafo/blob/1e0258a8cc643b2bbef06eb0935d6ddf7681b87b/lib/kafo/color_scheme.rb#L6-L8

evgeni:
What do you think about adding support for https://no-color.org/ too?

ekohl:
I think that's good and slightly prefer that over implementing a CLI option (probably because I'm lazy).

evgeni:
untested, but LGTM :)

test/lib/cli/health_command_test.rb
```diff
@@ -40,10 +40,10 @@ module ForemanMaintain
...
           [dummy-check-success] Check that ends up with success
           [dummy-check-warn] Check that ends up with warning
-          [external-service-is-accessible] External_service_is_accessible         [pre-upgrade-check]
-          [present-service-is-running] Present service run check                  [default]
-          [service-is-stopped] Service not running check                          [default]
-          [upgrade-post-upgrade-check] Procedures::Upgrade::PostUpgradeCheck      [post-upgrade-checks]
+          [external-service-is-accessible] External_service_is_accessible                  [pre-upgrade-check]
+          [present-service-is-running] Present service run check                           [default]
+          [service-is-stopped] Service not running check                                   [default]
+          [upgrade-post-upgrade-check] Procedures::Upgrade::PostUpgradeCheck               [post-upgrade-checks]
```
ekohl, line 46: This was needed. Probably because it now no longer generates color codes and the alignment has changed. That tells me the `NO_COLOR` bit works.

PR #900, opened 2024-07-10
Update test to check full package update message

ehelms:


Griffin-Sullivan (approved):

test/definitions/procedures/packages/update_all_confirmation_test.rb
```diff
@@ -24,11 +24,11 @@ def skip_mock_package_manager
       To upgrade to the next Foreman version use 'foreman-maintain upgrade'.
       Do you want to proceed with update of everything regardless of
       the recommendations?
+      , [y(yes), q(quit)]
```
ekohl, line 27: What's the weird comma here?

ekohl (commented):
Wouldn't it make sense to keep a test where the question is answered? Effectively simulating `--assumeyes`?

test/definitions/procedures/packages/update_all_confirmation_test.rb
ehelms: That was the original point of me re-working the tests, the output is weird. And I first wanted to create a test that can show that and then follow up with a commit that fixes it.

ekohl: That's always useful information to include in a PR description.

ehelms:
> Wouldn't it make sense to keep a test where the question is answered? Effectively simulating `--assumeyes`?

I can look at adding a test for this, I do not believe the existing tests were previously testing this.

definitions/procedures/packages/update_all_confirmation.rb
```diff
@@ -17,10 +17,10 @@ def run
           version and/or operating system using '#{command} update'.
           To upgrade to the next #{feature(:instance).product_name} version use '#{command} upgrade'.
           Do you want to proceed with update of everything regardless of
-          the recommendations?
+          the recommendations
```
evgeni, line 20: While I get why you're dropping the `?` here, every other occasion of `ask_decision` uses a message that ends with a question-mark, and then adds the question.
Conceptually, I'd then expect the possibly answers on a new line, but that seems not to happen right now.

ehelms: > Conceptually, I'd then expect the possibly answers on a new line, but that seems not to happen right now.

Can you maybe provide example of what you think the output should look like?

evgeni:
```
This action will potentially eat a small kitten.
You should be using foreman-mrawntain to avoid this!
Do you want to proceed with update of everything regardless of the recommendations?
[y(yes), n(no), q(let me see if I have enough spare kittens)]
```

evgeni: You can (and I wouldn't be opposed to that), but I also didn't really want to stop you from doing a release ;)

Maybe, keep the `?`, assert the output is `the recommendations?, [y(yes), q(quit)]` (thus matching the other places we use decisions), refactor in a follow up?

(But again: this has an ack, and it's me being pedantic. feel free to merge as is and do more cleanups later)

evgeni (approved):

PR #893, opened 2024-06-26
always use pg_dump for DB backups

evgeni:
- **always use pg_dump for DB backups**
- **drop --include-db-dumps parameter, we always include them now**
- **unify DB backup steps**

includes https://github.com/theforeman/foreman_maintain/pull/891 for the checks

PR #905, opened 2024-07-23
Backup and restore container gateway postgres DB

ianballou:
Adds backup and restore support for the container gateway DB. Only adds support for the postgres version of the database for now.

One notable change was that the DB commands need to be run as `foreman-proxy` since no password is saved for the database. The container gateway uses unix socket auth for the connection, so only the owning user can access the DB (as far as I know).

As such, I've added support for running generic commands as a different user.

ToDos:
- [ ] Figure out why stderr is ending up in the container gateway database dump
- [ ] Write tests


ianballou:
I looked into the issue where stderr makes it into the dump file. It seems that this was always the case since the final command looks like `sudo -u foreman-proxy -- pg_dump -Fc database > dump_file 2>&1`. That would redirect stderr into stdout, which is being redirected into `dump_file`.

I'm guessing this wasn't an issue before because `pg_dump` was always getting run as root. In my case, foreman-proxy had no permissions to change directories into the place where I started the command, which caused an error which then made it literally into the dump file.

ianballou:
I also ended up adding lines to change the backup ownership to `foreman-proxy:foreman-proxy` in order for the foreman-proxy user to have access. I noticed we deal with this issue slightly differently for restore -- there's a simple error telling the user to change the backup permissions. This is a bit different for backup though, because the user isn't the one creating the inner backup directory. I'm curious if anyone can see any issues coming out of this.

definitions/checks/container_gateway/db_up.rb
```diff
@@ -0,0 +1,29 @@
+module Checks
+  module ContainerGateway
+    class DBUp < ForemanMaintain::Check
```
wbclark, line 3: [idea] It looks like we'll have nearly identical code at `definitions/checks/{foreman,candlepin,pulpcore,container_gateway}/db_up.rb`... what do you think about abstracting these into a common subclass of ForemanMaintain::Check?

wbclark (commented):
Hey @ianballou !

I've shared some thoughts from my initial reading of this.

Regarding the proposal for some more abstraction here now that we've added a fourth(!) DB, it looks like it could also make sense for the definitions for online DB backup procedures and DB restore procedures?

I am curious to know what you think about it, and perhaps @evgeni would have some insight as well about whether this has been discussed before, and whether it's worth doing now

definitions/checks/container_gateway/db_up.rb
ianballou: It sounds like a good idea to me, the files appear to be very similar.

ianballou:
I originally thought this would be required for Katello 4.14 since I assume the change to postgres would introduce a regression in backup & restore. It looks like it does not, in fact it appears that offline backup is backing up the container gateway DB for free somehow.  Once the installer is run, the container_gateway DB is correctly created. So, users restoring at least can still just do a smart proxy sync to restore their content.

As such, I don't think we need to rush this for the upcoming Foreman release in case there are other concerns with the implementation that pop up. I'm personally a bit busy with getting other things in shape for the release, but please let me know if anyone thinks there's a new regression that I'm missing.

ianballou:
@wbclark I've implemented your ideas

evgeni:
> It looks like it does not, in fact it appears that offline backup is backing up the container gateway DB for free somehow.

It will not anymore (since we merged https://github.com/theforeman/foreman_maintain/pull/893), so caution! :)

definitions/features/container_gateway_database.rb
```diff
@@ -0,0 +1,40 @@
...
+  private
+
+  def load_configuration
+    config = YAML.load(File.read(CONTAINER_GATEWAY_DB_CONFIG))
+    @configuration = {}
+    connection_string = config[:db_connection_string]
+    if connection_string
+      uri = URI.parse(connection_string)
+      @configuration['connection_string'] = connection_string
+      @configuration['user'] = 'foreman-proxy'
```
evgeni, line 35: can't the user configure a different db user, and this one is just the fallback?

evgeni: oh wait, no, reading further below, that's the "shell user" not "database user"? I'm confused :D

ianballou: It is indeed a shell user.  The container gateway database (and maybe more in the future) does not configure any password by default since it uses unix-domain socket authentication.  As such, `root` can't (as far as I can tell) log into the database in any way, so I've created a generic way to set the shell user for doing database commands.

ianballou:
Since this has slipped a little on our team priorities, I'm going to call this a PoC and get things updated once the strategy is determined to be sound.

evgeni:
* This needs a rebase, as we re-did quite some parts of `base_database.rb`, sorry.
* Is my assumption correct, we do not support external DB for proxies at all right now?

ianballou:
>     * Is my assumption correct, we do not support external DB for proxies at all right now?

I'll preface this by saying I'm not sure if we support external DBs for proxies, but, if we do ...

We did not test external container gateway external DBs. If I remember correctly, we have all of the connection information ready to be modified to an external host, so it should work to be remote.

If my PR here does not respect external databases but should, I'll fix it.

definitions/features/container_gateway_database.rb
```diff
@@ -0,0 +1,40 @@
...
+      @configuration['database'] = uri.path[1..]
+    end
```
evgeni, line 37: So looking at https://github.com/theforeman/puppet-foreman_proxy/pull/835, the `connection_string` is something like `postgres:///container_gateway` (for a local setup) or `postgres://configurableuser:changeme@test.example.com:5432/container_gateway` (for a remote DB, supported or not).

Now, the other code in this PR depends on the `user` definition whether it switches to that (shell) user or not.
That switching is unnecessary for the remote-DB case!

```suggestion
    connection_string = config[:db_connection_string]
    if connection_string
      uri = URI.parse(connection_string)
      @configuration['connection_string'] = connection_string
      @configuration['user'] = 'foreman-proxy' unless uri.host
      @configuration['database'] = uri.path[1..]
    end
```

definitions/procedures/repositories/index_katello_repositories_container_metadata.rb
```diff
@@ -2,6 +2,7 @@ module Procedures::Repositories
   class IndexKatelloRepositoriesContainerMetatdata < ForemanMaintain::Procedure
     metadata do
       description 'Import container manifest metadata'
+      for_feature :repositories
```
evgeni, line 5: There is no such feature in Foreman Maintain!

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -57,20 +58,25 @@ def psql(query, config = configuration)
...
       end
 
       def dump_db(file, config = configuration)
-        execute!(dump_command(config) + " > #{file}", :hidden_patterns => [config['password']])
+        if config['user']
+          execute!("chown -R #{config['user']}:#{config['user']} #{File.dirname(file)}")
+        end
+        execute!(dump_command(config) + " -f #{file}",
+          :hidden_patterns => [config['password']], :user => config['user'])
       end
```
evgeni, line 80: Right now, this relies on the DB credentials we parse from the config and in the super special case of gateway, the shell user of the proxy.

For restores, we have this "if local, use `postgres` user, else the parsed config as `root`", if we use the same logic for dumping, we reduce the special-casing for gateway (as we either use `postgres`, which has access to all databases anyway, or a remote connection that doesn't require `foreman-proxy` user).

evgeni (changes requested):
I think this can be simplified, see inline comments :)

definitions/procedures/repositories/index_katello_repositories_container_metadata.rb
ianballou: Huh, not sure how this slipped in there :)

lib/foreman_maintain/concerns/base_database.rb
ianballou: When I wrote this I didn't think I could use the postgres user for some reason. Using `postgres` would make the code here a whole lot simpler :+1:

ianballou:
@evgeni I've rebased this and changed the PR to use the `postgres` user for "local" database commands. 

I need to get the tests passing again.

I've tested this on a container gateway enabled smart proxy with online backup & restore.

ianballou:
The latest code ensures all local DB connections use unix sockets with a simple connection string (like `postgres:///foreman`). It's been tested on a capsule and I'm currently testing on a satellite system (so far so good).

definitions/features/container_gateway_database.rb
```diff
@@ -0,0 +1,39 @@
...
+  private
+
+  def load_configuration
+    config = YAML.load(File.read(CONTAINER_GATEWAY_DB_CONFIG))
+    @configuration = {}
+    connection_string = config[:db_connection_string]
+    if connection_string
+      uri = URI.parse(connection_string)
+      @configuration['connection_string'] = connection_string
+      @configuration['database'] = uri.path[1..]
```
adamruzicka, line 35: The `path[1..]` is there "just" to strip the leading forward slash?

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -57,20 +58,36 @@ def query_csv(sql)
...
+            :env => base_env)
+        end
       end
 
       def dump_db(file)
-        dump_command = "pg_dump -Fc -f #{file}"
-        execute!(dump_command, :env => base_env)
+        if local?
+          dump_command = "pg_dump --format=c #{configuration['connection_string']}"
+          execute!("chown -R postgres:postgres #{File.dirname(file)}")
```
adamruzicka, line 85: Why wasn't this needed previously?

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -57,20 +58,36 @@ def query_csv(sql)
...
       end
 
       def ping
-        execute?('psql',
-          :stdin => 'SELECT 1 as ping',
-          :env => base_env)
+        if local?
+          execute?(
+            "runuser - postgres -c 'psql -d #{configuration['database']} -c \"SELECT 1 as ping\"'"
+          )
```
adamruzicka, line 74: Could the query be passed on stdin same as on L76-78?

adamruzicka (commented):
Haven't tested, but looks reasonable at a glance

ianballou:
@adamruzicka I think I've addressed your comments.

PR #909, opened 2024-08-13
fix tests for update scenario after adding pulpcore tasks check

evgeni:


ehelms:
merged as part of https://github.com/theforeman/foreman_maintain/pull/908

PR #912, opened 2024-08-15
Always include the installer answer file in backups

ekohl:
If the user has changed the answer file location in their scenario to something outside of `/etc/foreman-installer` then it isn't backed up.

This changes the installer feature's config_files to include it if needed. It also stops globbing the files and simply trust that you can list a directory, which we already do for the cached data.

I'm unsure about the implications for restores.

Note that this also prepares for the case where we will move the answer files to `/var/lib/foreman-installer` (provided we keep the scenario files in `/etc`).

### DIFF
--- a/definitions/features/installer.rb
+++ b/definitions/features/installer.rb
@@ -8,7 +8,7 @@ class Features::Installer < ForemanMaintain::Feature
   end
 
   def answers
-    load_answers(configuration)
+    YAML.load_file(answer_file)
   end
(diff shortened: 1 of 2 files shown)


evgeni:
Rebased and fixed tests.

I think this should have no implications on restores.

ekohl:
Thanks for fixing up the tests.

evgeni (approved):

PR #917, opened 2024-08-21
use correct phase to check if an upgrade scenario is available

evgeni:
Fixes: c5890ab9445fa5f99dce2f511839638d3e9b6775


ehelms (approved):

lpramuk (commented):
Tested successfully.

PR #920, opened 2024-08-23
Print message if update is unavailable

ehelms:


evgeni (changes requested):

lib/foreman_maintain/cli/update_command.rb
```diff
@@ -19,16 +19,38 @@ def update_runner
...
+      def try_update
+        if update_runner.available?
+          yield
+        else
+          instance = ForemanMaintain.detector.feature(:instance)
+          msg = <<~BANNER
+
+            Update is not available.
+            The system appears to be in the middle of an upgrade.
+            The current version of #{instance.product_name} is #{instance.current_major_version}.
```
evgeni, line 31: The code works as designed. I am just wondering about the messaging here.

The current "trigger" for this message to be printed is when the currently installed version does NOT match the "target" version of the foreman_maintain code:
https://github.com/theforeman/foreman_maintain/blob/d274464b64fd7dcf3ded8813dc65fa8698b6d8e6/definitions/scenarios/update.rb#L8

This can happen when the user has a FM installed from a different release stream than the real system, and in reality is the situation where you install the N+1 FM to perform the upgrade. However, it does not imply the upgrade has already begun.

How about (untested, but you get the idea):
```suggestion
            This version of #{ForemanMaintain.command_name} only supports #{instance.target_version},
            but the installed version of #{instance.product_name} is #{instance.current_major_version}.
            
            Therefor the update command is not available right now.
            
            Please install a version of #{ForemanMaintain.command_name} that supports #{instance.current_major_version}
            or perform an upgrade to #{instance.target_version} using the upgrade command.
```

:eyes: @lpramuk as I think this was part of what you were complaining about recently about miss-matched maintain/product versions?

evgeni (approved):

lpramuk (approved):

PR #924, opened 2024-08-30
Fixes #37776 Improve available_space Check in foreman_maintain

pafernanr:
**Description of problem:**

The / free-space check was included in foreman_maintain but it was from the perspective of yum not able to complete the package updates due to lack of disk space in /var/cache , under the assumption that it was residing under / itself.

Mostly people would have a seperate FS for /var or have a large enough /  to work with. And those with separate /var partition may not even need 4 GB free space in /.

There are some scenarios where Check fails even when there is plenty of space to handle the yum transaction

**How reproducible:**

Always when / partition available space < 4096 and /var/ has enough partition

**Is this issue a regression from an earlier version:**

No

**Actual behavior:**
The space_available Check fails

### DIFF
--- a/definitions/checks/disk/available_space.rb
+++ b/definitions/checks/disk/available_space.rb
@@ -3,7 +3,8 @@ module Checks
     class AvailableSpace < ForemanMaintain::Check
       metadata do
         label :available_space
-        description 'Check if /var/cache partition has enough space for dnf transaction:'
+        # /var/cache usually contains the cache for the most used package-managers (rpm, apt...)
+        description 'Check if /var/cache partition has enough space for transaction:'
         tags :pre_upgrade
       end
 
--- a/definitions/checks/disk/available_space.rb
+++ b/definitions/checks/disk/available_space.rb
@@ -3,7 +3,7 @@ module Checks
     class AvailableSpace < ForemanMaintain::Check
       metadata do
         label :available_space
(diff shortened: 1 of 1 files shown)


pafernanr:
Hey @sayan3296, this is the PR we were talking about

definitions/checks/disk/available_space.rb
```diff
@@ -3,19 +3,21 @@ module Disk
     class AvailableSpace < ForemanMaintain::Check
       metadata do
         label :available_space
-        description 'Check to make sure root(/) partition has enough space'
+        # /var/cache usually contains the cache for the most used package-managers (rpm, apt...)
+        description 'Check if /var/cache partition has enough space for transaction:'
```
ehelms, line 7:
```suggestion
        description 'Check if /var/cache partition has enough space for transaction'
```

definitions/checks/disk/available_space.rb
```diff
@@ -3,19 +3,21 @@ module Disk
     class AvailableSpace < ForemanMaintain::Check
       metadata do
         label :available_space
-        description 'Check to make sure root(/) partition has enough space'
+        # /var/cache usually contains the cache for the most used package-managers (rpm, apt...)
```
ehelms, line 6:
```suggestion

```

The description says it all and this is superfluous.

PR #925, opened 2024-09-04
Implement "restart" for RemoteDB services

evgeni:
It's not really a restart, as we can't control it, but it checks the DB
and returns the status properly.

Fixes: 7cdffe78727b48ca148ba50e9eb069fc522429f8

### DIFF
--- a/lib/foreman_maintain/utils/service/remote_db.rb
+++ b/lib/foreman_maintain/utils/service/remote_db.rb
@@ -38,6 +38,14 @@ module ForemanMaintain::Utils
         [0, db_status.last]
       end
 
+      def restart
+        command_name = ForemanMaintain.command_name
+        db_status(<<~MSG
+          Remote databases are not managed by #{command_name} and therefore was not restarted.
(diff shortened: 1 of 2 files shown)


lib/foreman_maintain/utils/service/remote_db.rb
```diff
@@ -38,6 +38,10 @@ def stop
         [0, db_status.last]
       end
 
+      def restart
+        db_status("It can't be restarted")
```
ehelms, line 42: This is ambiguous, and I am trying to think through better wording:

- Must be restarted manually
- Cannot be restarted by #{tool-name}
- Remote databases are not managed by #{tool-name}, they must be restarted by you.

evgeni: Maybe. But it's in line with `enable`/`disable` in this file, and @lpramuk pointed out that having some kind of MSG for `start`/`stop` would also be cool.

Further improvement?

PR #930, opened 2024-09-24
Drop rotate-script from executables-section in gemspec

### LINKED ISSUE #929: gemspec defines useless wrapper-script
The gemspec lists `foreman-maintain-rotate-tar` in the `executables`-section. This creates a ruby-wrapper-script that, when called, tries to load the bash-script in `<gempath>/bin/foreman-maintain-rotate-tar` as ruby-code.

I propose to drop that script from the gemspec, because it is only used internally.

m-bucher:
Fixes #929

### ToDo:

- [ ] Drop from [foreman-packaging](https://github.com/theforeman/foreman-packaging/blob/36daef0ed33329be3106f14a870c9264c15ee69c/packages/foreman/rubygem-foreman_maintain/rubygem-foreman_maintain.spec#L92)

### DIFF
--- a/foreman_maintain.gemspec
+++ b/foreman_maintain.gemspec
@@ -18,8 +18,7 @@ the Foreman/Satellite up and running."
   s.files += `git ls-files extras`.split("\n")
   s.extra_rdoc_files = ['LICENSE', 'README.md']
   s.require_paths = ['lib']
-  s.executables = ['foreman-maintain',
-                   'foreman-maintain-complete', 'foreman-maintain-rotate-tar']
+  s.executables = ['foreman-maintain', 'foreman-maintain-complete']
   s.required_ruby_version = ">= 2.7", "< 4"
 
   s.add_dependency 'clamp'


foreman_maintain.gemspec
```diff
@@ -18,8 +18,7 @@ the Foreman/Satellite up and running."
   s.files += `git ls-files extras`.split("\n")
   s.extra_rdoc_files = ['LICENSE', 'README.md']
   s.require_paths = ['lib']
-  s.executables = ['foreman-maintain',
-                   'foreman-maintain-complete', 'foreman-maintain-rotate-tar']
```
ehelms, line 22: Can we just drop the whole script at this point?

m-bucher: Removing it from `s.executables` ensures that the gem-installer does not create a ruby-wrapper script in `/usr/bin/foreman-maintain-rotate-tar`.

I am not fully sure, why tar needs to use the shell-script when changing volume.
Also it is only set, if the tar-command does not use `--gzip`.
The history says we need it for the extract to work smoothly: 2beadc61aefc5caff5813198520d48027eb13baa

evgeni (approved):

PR #936, opened 2024-10-10
satellite maintain metadata

evgeni:
- **add test for Satellite Upgrade scenario confinement**
- **assert target_version**
- **move the logic which versions are supported into the downstream concern**

### DIFF
--- a/lib/foreman_maintain/concerns/downstream.rb
+++ b/lib/foreman_maintain/concerns/downstream.rb
@@ -1,6 +1,8 @@
 module ForemanMaintain
   module Concerns
     module Downstream
+      SATELLITE_MAINTAIN_CONFIG = '/usr/share/satellite-maintain/config.yml'.freeze
+
       def current_version
         raise NotImplementedError
       end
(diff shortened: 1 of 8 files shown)


lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -46,8 +48,25 @@ def fm_pkg_and_cmd_name
...
+          ForemanMaintain.upgrade_in_progress == satellite_maintain_config['current_sat_version']
+      end
+
       private
 
+      def satellite_maintain_config
+        if File.exist?(SATELLITE_MAINTAIN_CONFIG)
+          YAML.load_file(SATELLITE_MAINTAIN_CONFIG)
+        else
+          {}
```
ehelms, line 66: This is a good sane default, but I wondering if we shouldn't throw a more in-yo-face error if the config file is missing as that indicates the system is likely in a bad state, something is missing and you can't proceed.

test/definitions/test_helper.rb
```diff
@@ -209,6 +209,15 @@ def assume_service_missing(name, priority = 30)
       yield service if block_given?
     end
   end
+
+  def mock_satellite_maintain_config
+    config = {
+      'current_sat_version' => '6.16',
+      'previous_sat_version' => '6.15',
+    }
```
ehelms, line 217: My 2-cents, either spell out `satellite`, drop the `sat` or nest the data:

```
satellite: {
  current_version
  previous_version
}
```

evgeni:
@ehelms updated! (and offended the :policeman:)

lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -46,8 +48,25 @@ def fm_pkg_and_cmd_name
...
+          ForemanMaintain.upgrade_in_progress == satellite_maintain_target_version
+      end
+
       private
 
+      def satellite_maintain_config
+        if File.exist?(SATELLITE_MAINTAIN_CONFIG)
+          YAML.load_file(SATELLITE_MAINTAIN_CONFIG)
+        else
+          raise "Couldn't load satellite-maintain configuration file #{SATELLITE_MAINTAIN_CONFIG}."
```
ehelms, line 66:
```suggestion
          raise "Could not load satellite-maintain configuration file #{SATELLITE_MAINTAIN_CONFIG}."
```

Contractions get messy

ehelms (approved):
Small nit

PR #941, opened 2024-10-16
load pulpcore db configuration by asking django, not by parsing python

evgeni:


definitions/features/pulpcore_database.rb
```diff
@@ -26,18 +26,21 @@ def services
 
   private
 
-  # rubocop:disable Metrics/AbcSize
   def load_configuration
-    full_config = File.read(PULPCORE_DB_CONFIG).split(/[\s,'":]/).reject(&:empty?)
+    python_command = <<~PYTHON.strip
+      from django.conf import settings; import json; print(json.dumps(settings.DATABASES["default"]))
```
ehelms, line 31: This looks like fun code.

evgeni: how so?

ehelms (approved):

PR #953, opened 2024-11-15
Halt upgrades if evr is not owned by foreman

ianballou:
Halts upgrades if the `evr` extension in Katello external DBs are not owned by `foreman`. Skips the checks if the `evr` extension does not exist.

Related installer PR: https://github.com/theforeman/foreman-installer/pull/984


definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,55 @@
...
+        tags :pre_upgrade
+        confine do
+          feature(:foreman_database) && !feature(:foreman_database).local? && feature(:katello)
+        end
+      end
+
+      def run
+        # No check is needed if the evr extension does not exist
+        evr_exists = find_evr_exists
+        return if !evr_exists
```
ehelms, line 17: I think you could just turn this into more of a `evr_exists?` method, so that you can reduce these two lines to:

```
return unless evr_exists?
```

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        error_msg = 'The evr extension is not owned by the foreman DB owner. Please run the following command to fix it: ' \
+                    "UPDATE pg_extension SET extowner = (SELECT oid FROM pg_authid WHERE rolname='foreman') WHERE extname='evr';"
+        foreman_owns_evr = find_foreman_owns_evr
+        fail!(error_msg) if !foreman_owns_evr
```
ehelms, line 22: Same idea here, switch to `foreman_owns_evr?` method name and then do:

```
fail!(error_msg) unless foreman_own_evr?
```

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,55 @@
...
+        fail!('Could not determine if the evr extension is owned by the foreman DB owner')
+      end
+
+      def self.query_for_evr_existence
+        <<-SQL
+          SELECT 1 AS evr_exists FROM pg_extension WHERE extname = 'evr'
+        SQL
+      end
+
+      def self.query_to_find_evr_owner
```
ehelms, line 47: This does not need to be a class method

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,55 @@
...
+        for_feature :foreman_database
+        description 'Check that external DBs have proper EVR extension permissions'
+        tags :pre_upgrade
+        confine do
+          feature(:foreman_database) && !feature(:foreman_database).local? && feature(:katello)
+        end
+      end
+
+      def run
+        # No check is needed if the evr extension does not exist
```
ehelms, line 15: You can drop this, the code reads well enough to tell me what's happening.

definitions/checks/foreman/check_external_db_evr_permissions.rb
ianballou: The other checks have the queries as class methods so I assumed it was some foreman maintain convention. I can change it.

ianballou:
All the comments should be addressed now.

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      private
+
+      def evr_exists?
+        evr_exists = feature(:foreman_database).query(query_for_evr_existence)
+        if !evr_exists.empty? && evr_exists.first['evr_exists'] == '1'
+          return evr_exists.first['evr_exists'] == '1'
+        end
+        return false
```
ehelms, line 31:
```suggestion
        if evr_exists.empty?
          return false
        else
          return evr_exists.first['evr_exists'] == '1'
        end
```

Performing the check that it `== 1` twice is redundant as you can just return the value of that check.

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      def foreman_owns_evr?
+        evr_owned_by_postgres = feature(:foreman_database).query(query_if_postgres_owns_evr)
+        unless evr_owned_by_postgres.empty?
+          return evr_owned_by_postgres.first['evr_owned_by_postgres'] == '0'
+        end
+        fail!('Could not determine if the evr extension is owned by the foreman DB owner')
```
ehelms, line 39: What do we expect a user that hits this fail statement to do next? Can we give them some direction?

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,56 @@
...
+        tags :pre_upgrade
+        confine do
+          feature(:foreman_database) && !feature(:foreman_database).local? && feature(:katello)
+        end
+      end
+
+      def run
+        return unless evr_exists?
+
+        error_msg = 'The evr extension is not owned by the foreman DB owner. Please run the ' \
```
ehelms, line 17: I think we should include mention of running it on the external database server just to be as clear as possible about where they would be executing the statement.

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,56 @@
...
+        for_feature :foreman_database
+        description 'Check that external DBs have proper EVR extension permissions'
+        tags :pre_upgrade
+        confine do
+          feature(:foreman_database) && !feature(:foreman_database).local? && feature(:katello)
+        end
+      end
+
+      def run
+        return unless evr_exists?
```
ehelms, line 15: Should we explicitly return true since the check would pass? I am unsure of how it handles `nil` returns from checks but `nil` is often a negative case.

ianballou: In this case foreman maintain is accepting nil as a pass, but I'll return true.

definitions/checks/foreman/check_external_db_evr_permissions.rb
ianballou: The only failures reasons I can think of would be if the DB connection failed or if the config doesn't have the correct username information. I've updated the error message.

ianballou:
I stopped hardcoding 'foreman' as the DB name just in case. We are hardcoding it in the installer check, but I figured why not here since it seems to be supported.

definitions/checks/foreman/check_external_db_evr_permissions.rb
ianballou: The only failures reasons I can think of would be if the DB connection failed or if the config doesn't have the correct username information. I've updated the error message.

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,63 @@
...
+      def foreman_db_user
+        feature(:foreman_database).configuration['username'] || 'foreman'
+      end
+
+      def evr_exists?
+        evr_exists = feature(:foreman_database).query(query_for_evr_existence)
+        unless evr_exists.empty?
+          return evr_exists.first['evr_exists'] == '1'
+        end
+        return false
```
ehelms, line 35: This is not a required change, more food for thought. You could reduce this down to two lines:

```
return false if evr_exists.empty?
return evr_exists.first['evr_exists'] == '1'
```

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      def foreman_owns_evr?
+        evr_owned_by_postgres = feature(:foreman_database).query(query_if_postgres_owns_evr)
+        unless evr_owned_by_postgres.empty?
+          return evr_owned_by_postgres.first['evr_owned_by_postgres'] == '0'
+        end
+        failure_msg = 'Could not determine if the evr extension is owned by the ' \
+                'foreman DB owner. Check that the foreman database is accessible ' \
```
ehelms, line 44: Nitpick - spell out database instead of using DB here since we spell out database everywhere else and that is more explicit. The rest of the message looks good now, thanks for the update.

ehelms:
I had just a minor nit, @evgeni please also do a final review

definitions/checks/foreman/check_external_db_evr_permissions.rb
ianballou: works for me!

ehelms (approved):

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,61 @@
...
+        fail!(failure_msg)
+      end
+
+      def query_for_evr_existence
+        <<-SQL
+          SELECT 1 AS evr_exists FROM pg_extension WHERE extname = 'evr'
+        SQL
+      end
+
+      def query_if_postgres_owns_evr
```
evgeni, line 53: Technically, this doesn't check whether `postgres` owns it, but whether `foreman_db_user` (not) owns it.

definitions/checks/foreman/check_external_db_evr_permissions.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      def evr_exists?
+        evr_exists = feature(:foreman_database).query(query_for_evr_existence)
+        return false if evr_exists.empty?
+        return evr_exists.first['evr_exists'] == '1'
+      end
+
+      def foreman_owns_evr?
+        evr_owned_by_postgres = feature(:foreman_database).query(query_if_postgres_owns_evr)
+        unless evr_owned_by_postgres.empty?
```
evgeni, line 38: I guess it won't ever be empty, given at this point we do know the extension exists, so it must have *some* owner?
But it also doesn't hurt to play safe here :)

evgeni (approved):

evgeni:
> I stopped hardcoding 'foreman' as the DB name just in case. We are hardcoding it in the installer check, but I figured why not here since it seems to be supported.

We (you :smile: ) don't hard code it in the installer check ;)

PR #955, opened 2024-11-26
Add check for sha1 server certificate before upgrading

ehelms:


definitions/checks/check_sha1_certificate_authority.rb
```diff
@@ -0,0 +1,24 @@
+class Checks::CheckSha1CertificateAuthority < ForemanMaintain::Check
+  metadata do
+    label :check_sha1_certificate_authority
+    description 'Check if server certificate authority is sha1 signed'
+
+    confine do
+      feature(:katello)
```
evgeni, line 7: Wouldn't this also apply to proxies?

definitions/checks/check_sha1_certificate_authority.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def run
+    installer_answers = feature(:installer).answers
+    server_ca = installer_answers[:certs][:server_ca_cert]
+
+    certificate = OpenSSL::X509::Certificate.new(File.read(server_ca))
+
+    assert(
+      certificate.signature_algorithm != 'sha1WithRSAEncryption',
+      "Server CA certificate signed with sha1 which will break on upgrade."
```
evgeni, line 21: Ok, and what shall I (as a user) do?

(read: can we make the printed error somehow actionable?)

evgeni (approved):
you need to make the :policeman: happy, but otherwise ACK

PR #956, opened 2024-11-29
don't fail when pulp cli didn't return valid json

evgeni:


evgeni:
I'm still puzzled why https://github.com/theforeman/foreman_maintain/commit/95e9d5a84d6c40ecb0a02b13389b5e6a7d5d0dfd was not sufficient, aka: in which case `pulp cli` doesn't return valid JSON yet does not exit non-zero

(Waiting on feedback from the person who reported the issue)

evgeni:
```
#  pulp --format json task list --state-in running --state-in canceling
/usr/lib/python3.11/site-packages/urllib3/connection.py:463: SubjectAltNameWarning: Certificate for foreman.example.com has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/urllib3/urllib3/issues/497 for details.)
  warnings.warn(
[]
```

the warning is on stderr, so we mix those two up somewhere.

evgeni:
https://github.com/theforeman/foreman_maintain/blob/e31bbdb5fe5ba269643fb33226d8eed7bea7b3b8/lib/foreman_maintain/utils/command_runner.rb#L96

whyyyyyyyyyyyy

ehelms (approved):

PR #963, opened 2024-12-10
add Ruby 3.3 to the test matrix

evgeni:
3.3.5 is what's currently in CentOS Stream 10

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -11,7 +11,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: ['2.7.4', '3.0.2', '3.1.3', '3.2.1']
+        ruby-version:
+          - '2.7.4'
+          - '3.0.2'
+          - '3.1.3'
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby_tests.yml
```diff
@@ -11,7 +11,12 @@ jobs:
...
       matrix:
-        ruby-version: ['2.7.4', '3.0.2', '3.1.3', '3.2.1']
+        ruby-version:
+          - '2.7.4'
+          - '3.0.2'
+          - '3.1.3'
+          - '3.2.1'
+          - '3.3.5'
     steps:
       - uses: actions/checkout@v2
```
ekohl, line 21: Unrelated, but we should update this
```suggestion
      - uses: actions/checkout@v4
```

ekohl (approved):

.github/workflows/ruby_tests.yml
evgeni: https://github.com/theforeman/foreman_maintain/pull/964

PR #969, opened 2024-12-31
Fixes #38119: When the "server_ca_cert" is set and has an empty line, it will fail

phess:
When bundle_pem contains non-certificate contents after the final `---END CERTIFICATE---` line, the `OpenSSL::X509::Certificate.new(pem.join)` action will fail because `pem.join` will be an empty or non-certificate string and the `load_fullchain(bundle_pem)` function will fail.

This change adds a check for the presence of `-+BEGIN` followed (however many lines later) by `-+END` before trying to load it as an x509 certificate object, thus preventing blank contents from being loaded as x509 certificates.


ehelms:
I am first trying to update the tests to show this -- https://github.com/theforeman/foreman_maintain/pull/971

ehelms:
Closing in favor of solution and updates in https://github.com/theforeman/foreman_maintain/pull/971, thanks @phess for the pointer!

PR #981, opened 2025-02-11
don't merge stderr into the output when calling pulp commands

evgeni:


ekohl (approved):
Before you know it you'll even pass in commands as arrays to avoid spawning a shell.

lib/foreman_maintain/utils/command_runner.rb
```diff
@@ -82,18 +88,16 @@ def run_interactively
       end
 
       def run_non_interactively
-        IO.popen(@env, full_command, 'r+') do |f|
-          if @stdin
-            f.puts(@stdin)
-            f.close_write
-          end
-          @output = f.read.strip
```
evgeni, line 90: I should not have killed this `.strip`, it breaks some executions that do compare the output and don't expect a trailing newline :see_no_evil:

PR #986, opened 2025-02-28
Add report for collecting Katello content facts

ianballou:
Begins Katello content fact collection for the analytics report. Starts with repository type information.

Example output:

```
[root@centos9-stream-katello-nightly foreman_maintain]# bin/foreman-maintain report generate | grep repositories
custom_library_yum_repositories_count: 1
redhat_library_yum_repositories_count: 1
library_debian_repositories_count: 1
library_container_repositories_count: 1
library_file_repositories_count: 1
library_python_repositories_count: 1
library_ansible_collection_repositories_count: 1
library_ostree_repositories_count: 1
```

Also adds some alternate content source fact collection:

```
[root@centos9-stream-katello-nightly foreman_maintain]# bin/foreman-maintain report generate | grep alternate
custom_alternate_content_sources_count: 2
simplified_alternate_content_sources_count: 1
rhui_alternate_content_sources_count: 1
yum_alternate_content_sources_count: 3
file_alternate_content_sources_count: 1
```

### DIFF
--- a/definitions/reports/alternate_content_sources.rb
+++ b/definitions/reports/alternate_content_sources.rb
@@ -1,0 +1,46 @@
+module Checks
+  module Report
+    class AlternateContentSources < ForemanMaintain::Report
+      metadata do
+        description 'Facts about ACSs'
+        confine do
+          feature(:katello)
+        end
+      end
+
+      def run
+        data_field('custom_alternate_content_sources_count') { custom_alternate_content_sources }
+        data_field('simplified_alternate_content_sources_count') do
+          simplified_alternate_content_sources
+        end
+        data_field('rhui_alternate_content_sources_count') { rhui_alternate_content_sources }
+        data_field('yum_alternate_content_sources_count') { yum_alternate_content_sources }
+        data_field('file_alternate_content_sources_count') { file_alternate_content_sources }
+      end
+
+      def custom_alternate_content_sources
+        sql_count(
+          "katello_alternate_content_sources WHERE alternate_content_source_type = 'custom'"
+        )
+      end
+
+      def simplified_alternate_content_sources
+        sql_count(
+          "katello_alternate_content_sources WHERE alternate_content_source_type = 'simplified'"
+        )
+      end
+
+      def rhui_alternate_content_sources
+        sql_count("katello_alternate_content_sources WHERE alternate_content_source_type = 'rhui'")
+      end
+
+      def yum_alternate_content_sources
+        sql_count("katello_alternate_content_sources WHERE content_type = 'yum'")
+      end
+
+      def file_alternate_content_sources
+        sql_count("katello_alternate_content_sources WHERE content_type = 'file'")
+      end
+    end
+  end
+end
--- a/definitions/reports/content.rb
+++ b/definitions/reports/content.rb
@@ -11,67 +11,43 @@ module Checks
       def run
         data_field('custom_library_yum_repositories_count') { custom_library_yum_repositories }
         data_field('redhat_library_yum_repositories_count') { redhat_library_yum_repositories }
-        data_field('library_debian_repositories_count') { library_debian_repositories }
-        data_field('library_container_repositories_count') { library_container_repositories }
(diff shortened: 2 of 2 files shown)


ianballou:
Edit: updated to have "count" at the ends of the data fields.

definitions/reports/alternate_content_sources.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      def custom_alternate_content_sources
+        query =
+          query(
+            <<-SQL
+              SELECT count(*) as custom_acs_count FROM "katello_alternate_content_sources"
+                WHERE "katello_alternate_content_sources"."alternate_content_source_type" = 'custom'
+            SQL
+          )
+        query.first['custom_acs_count'].to_i
```
adamruzicka, line 29: You could simplify this by using the `sql_count` helper[1]. Something along the lines of this should work

```suggestion
        sql_count("katello_alternate_content_sources WHERE alternate_content_source_type = 'custom'")
```

[1] - https://github.com/theforeman/foreman_maintain/blob/02ce1e30bda950701b70e1aa4d65812ad5cbec46/lib/foreman_maintain/report.rb#L14

definitions/reports/content.rb
```diff
@@ -0,0 +1,78 @@
...
+          query(
+            <<-SQL
+              SELECT count(*) as yum_rh_count FROM "katello_root_repositories"
+                INNER JOIN "katello_products" ON "katello_products"."id" = "katello_root_repositories"."product_id"
+                INNER JOIN "katello_providers" ON "katello_providers"."id" = "katello_products"."provider_id"
+                  WHERE "katello_providers"."provider_type" = 'Red Hat'
+                AND "katello_root_repositories"."content_type" = 'yum'
+            SQL
+          )
+        query.first['yum_rh_count'].to_i
```
adamruzicka, line 50: nitpick: This could still be using the `sql_count` helper

definitions/reports/content.rb
```diff
@@ -0,0 +1,78 @@
...
+        sql_count("katello_root_repositories WHERE content_type = 'file'")
+      end
+
+      def library_python_repositories
+        sql_count("katello_root_repositories WHERE content_type = 'python'")
+      end
+
+      def library_debian_repositories
+        sql_count("katello_root_repositories WHERE content_type = 'deb'")
+      end
```
adamruzicka, line 75: nitpick: While this works, it feels a bit repetitive

adamruzicka (commented):
Seems to work well, although it could be made a bit DRYer, but that might be just personal preference.

ianballou:
I can clean these ^ up a bit.

ianballou:
I've updated the content.rb file. It isn't tested yet.

definitions/reports/content.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def custom_library_yum_repositories
+        query_snippet =
+            <<-SQL
+              "katello_root_repositories"
+                WHERE "katello_root_repositories"."id" NOT IN
+                (SELECT "katello_root_repositories"."id" FROM "katello_root_repositories" INNER JOIN "katello_products"
+                  ON "katello_products"."id" = "katello_root_repositories"."product_id" INNER JOIN "katello_providers"
+                  ON "katello_providers"."id" = "katello_products"."provider_id" WHERE "katello_providers"."provider_type" = 'Red Hat')
+                AND "katello_root_repositories"."content_type" = 'yum'
```
adamruzicka, line 32: Out of curiosity, is there a reason for going with a subquery directly joining all the necessary tables and then filtering out the rows we don't want to count?

adamruzicka (commented):
Seems to work well, rubocop is unhappy though

definitions/reports/content.rb
ianballou: I grabbed these queries directly from Katello & `to_sql`. Could be simplified perhaps.

adamruzicka (approved):

PR #990, opened 2025-03-07
Add networking report

adamruzicka:
Rather than trying to answer the loosely defined "are we running ipv4-only, ipv6-only or dualstack?" question directly, the aim here is to collect enough data to be able to answer the question later, depending on how exactly we define it.

```shell
# foreman-maintain report generate | grep -e ipv -e dualstack
subnet_ipv4_count: 0
subnet_ipv6_count: 0
hosts_with_ipv4only_interface_count: 5
hosts_with_ipv6only_interface_count: 0
hosts_with_dualstack_interface_count: 0
foreman_interfaces_ipv4only_count: 1
foreman_interfaces_ipv6only_count: 0
foreman_interfaces_dualstack_count: 0
setting_remote_execution_connect_by_ip_prefer_ipv6: false
setting_discovery_prefer_ipv6: true
```

I was also thinking about further scoping the `hosts_with_{ipv4-only,ipv6-only,dualstack}_interace_count` by the flag of the interface (`primary`, `provision` and so on), but that would probably be too granular.

I also though about including the ipv4 vs ipv6 preference settings (`remote_execution_connect_by_ip_prefer_ipv6`, `discovery_prefer_ipv6`), but we can already infer that from the list of modified settings.

### DIFF
--- a/definitions/reports/networking.rb
+++ b/definitions/reports/networking.rb
@@ -17,8 +17,10 @@ module Reports
     # How many ipv4 subnets are defined in Foreman?
     # How many ipv6 subnets are defined in Foreman?
     def subnet_counts_by_type
-      %w(ipv4 ipv6).each do |type|
-        data_field("subnet_#{type}_count") { sql_count("subnets where type = 'Subnet::Ipv#{type}'") }
+      %w[ipv4 ipv6].each do |type|
+        data_field("subnet_#{type}_count") do
+          sql_count("subnets where type = 'Subnet::Ipv#{type}'")
+        end
       end
     end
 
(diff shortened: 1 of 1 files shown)


adamruzicka:
@ShimShtein @ares any thoughts on this?

ShimShtein:
> but we can already infer that from the list of modified settings

It's quite painful to look back for changed settings, so I think it would be OK to duplicate the current setting value here.

> hosts_with_ipv6-only_interface_count

Having mixed `-` and `_` feels weird, even though I understand where it comes from.

> by the flag of the interface (primary, provision and so on)

Agreed, at least theoretically it should add cardinality, which is not desired.

ShimShtein (approved):

ShimShtein (changes requested):

definitions/reports/networking.rb
```diff
@@ -0,0 +1,79 @@
...
+    def hosts_by_address_family
+      { 'ipv4only': 'nics.ip IS NOT NULL AND nics.ip6 IS NULL',
+        'ipv6only': 'nics.ip IS NULL AND nics.ip6 IS NOT NULL',
+        'dualstack': 'nics.ip IS NOT NULL AND nics.ip6 IS NOT NULL' }.each do |kind, condition|
+        query = <<~SQL
+          hosts
+          INNER JOIN nics on nics.host_id = hosts.id
+          WHERE #{condition}
+        SQL
+        data_field("hosts_with_#{kind}_interface_count") { sql_count(query) }
```
ShimShtein, line 40: If I am not mistaken, for host with two nics, you will get 2 records in the joined table.

adamruzicka: Yeah, you're right, fixed

ShimShtein (approved):

PR #992, opened 2025-03-11
Collect count of hosts running RHEL AI

adamruzicka:


definitions/reports/inventory.rb
```diff
@@ -80,5 +81,16 @@ def to_audits_record(row)
...
       }
     end
+
+    def hosts_running_rhel_ai
+      query = <<~SQL
+        hosts
+        INNER JOIN katello_subscription_facets AS ksf ON hosts.id = ksf.host_id
+        INNER JOIN katello_subscription_facet_installed_products AS ksfip ON ksfip.subscription_facet_id = ksf.id
+        INNER JOIN katello_installed_products AS kip ON ksfip.installed_product_id = kip.id
+        WHERE kip.name = 'Red Hat Enterprise Linux AI'
```
adamruzicka, line 91: @ianballou is this the way to detect hosts running rhel ai or is there an alternative approach that would be preferred?

ianballou: I'm installing a RHEL AI machine to test this out. I think this should work well. My only other idea would be to use the bootc info Katello collects from sub-man. In that case we could check the bootc booted image, but only if it's a consistent value that can be relied upon.

lfu: Does foreman facts have any information about RHEL AI or Openshift AI?

ianballou: @lfu we could use the bootc_booted_image, which will have the `rhelai1` prefix in it.
Example: `registry.redhat.io/rhelai1/bootc-intel-rhel9:1.4`

@adamruzicka have you been able to get this to work in practice? I've registered a RHEL AI machine to my Satellite but the 'Red Hat Enterprise Linux AI' product is not getting associated to it.

ianballou: This will only work with RHEL AI boxes that have the version of bootc that knows how to send facts via sub-man, however.

ianballou: The RHEL 9.4 AI box I tried does not have the right version.

lfu: Given the AI popularity, maybe this information should be added at Foreman level so all other plugins may access it.
Does it have to be a bootc_booted_image to be AI enabled?

ianballou: Interesting, if that worked for you then it might be the best hope for now.
What setup steps did do first in Katello? Import a manifest with the right subs, sync the rhel AI repo, and attach the repo to your host?

adamruzicka: It was some months ago, but I don't think I did anything special. I had rocky linux synced into that instance, no rh manifest or anything, created an activation key giving access to library and just used hosts > register to get the host in.

ianballou: Gotcha, as far as I'm aware, this is the only way to get the information for hosts that don't send bootc facts via sub-man. So, +1 from me.

ianballou:
@adamruzicka do you need someone to test this? It looks right to me.

ianballou (approved):
Looks good :+1:

PR #993, opened 2025-03-28
Add report for collecting image mode host counts

adamlazik1:
Host counts or sorted by OS same as the existing `hosts_count_by_os`.

Open question: Should image mode hosts should still be counted in existing `hosts_count_by_os`? Or should the total host count be split between `image_mode_hosts_count_by_os` and `hosts_count_by_os`?

### DIFF
--- a/definitions/reports/image_mode_hosts.rb
+++ b/definitions/reports/image_mode_hosts.rb
@@ -1,0 +1,27 @@
+module Reports
+  class ImageModeHosts < ForemanMaintain::Report
+    metadata do
+      description 'Report number of image mode hosts registered by operating system'
+      confine do
+        feature(:katello)
+      end
+    end
+
+    def run
+      merge_data('image_mode_hosts_by_os_count') { image_mode_hosts_by_os_count }
+    end
+
+    # OS usage on image mode hosts
+    def image_mode_hosts_by_os_count
+      query(
+        <<-SQL
+          select max(operatingsystems.name) as os_name, count(*) as hosts_count
+          from hosts inner join operatingsystems on operatingsystem_id = operatingsystems.id inner join katello_content_facets on hosts.id = katello_content_facets.host_id
+          where bootc_booted_digest is not null
+          group by operatingsystems.name
+        SQL
+      ).
+        to_h { |row| [row['os_name'], row['hosts_count'].to_i] }
+    end
+  end
+end


adamruzicka:
> Should image mode hosts should still be counted in existing hosts_count_by_os? 

Yes, they're still hosts running that os, just in a different way.

> Or should the total host count be split between image_mode_hosts_count_by_os and hosts_count_by_os?

No, if we become interested in how many hosts with a certain os are running in "standard" mode, we can just do the math.

adamruzicka (approved):
Lgtm

definitions/reports/image_mode_hosts.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def run
+      merge_data('image_mode_hosts_by_os_count') { image_mode_hosts_by_os_count }
+    end
+
+    # OS usage on image mode hosts
+    def image_mode_hosts_by_os_count
+      query(
+        <<-SQL
+          select max(operatingsystems.name) as os_name, count(*) as hosts_count
```
ShimShtein, line 18: A couple of questions:
Are those name constant across Satellites?
How many names we expect here?
Can we have two os instances with the same name (with a different version perhaps)?

I am worried about the amount of records this can generate in the report. For now we agreed that we do not create nodes based on fields with unbound value counts.

adamlazik1: > Are those name constant across Satellites?
I am not sure I understand the question, but the name of OS should hopefully not depend on Satellite.

> How many names we expect here?
I'd say that depends on individual deployments, but I wouldn't expect too many. For example, I think that all RHEL systems should be grouped under RedHat. It is less of a name and more family name (I think? I copied the behavior for the existing host_count_by_os). 

> Can we have two os instances with the same name (with a different version perhaps)?
I am fairly certain that those are grouped under the same name.

ShimShtein: > I am not sure I understand the question, but the name of OS should hopefully not depend on Satellite.
If a user can change the name - we will have different names for different instances, making aggregation impossible. Since you're following the same scheme as `host_count_by_os`, I suppose we already decided that it was good enough.

@adamruzicka  It actually makes me wonder if we don't have an error in `host_count_by_os` - we're grouping by os_id there, but labeling by name. If we have two os's with the same name, but different ids, we will be in trouble.

adamruzicka: > It actually makes me wonder if we don't have an error in host_count_by_os


It seems we do https://github.com/theforeman/foreman_maintain/pull/997

adamlazik1: Thanks, I applied the same fix here.

adamruzicka (approved):

PR #994, opened 2025-04-02
Add report for remediations from advisor on prem

adamlazik1:


adamlazik1:
The current implementation relies on counting distinct insights hit entries which are associated with a successfully completed subtask of the `rh_cloud_remediate_hosts` job_invocation. There might be some hole in this logic, it is yet to be confirmed.

definitions/reports/advisor_on_prem_remediations.rb
```diff
@@ -0,0 +1,31 @@
+module Reports
+  class AdvisorOnPremRemediations < ForemanMaintain::Report
+    metadata do
+      description 'Report number of remediations based on advisor rules from advisor on premise'
+      confine do
+        config_file = '/etc/foreman/plugins/foreman_rh_cloud.yaml'
+        setting_regex = /:use_local_advisor_engine:\s*true/
```
ehelms, line 7: Can you rely on the installer answers instead?

`foreman::plugin::rh_cloud::enable_iop_advisor_engine`

definitions/reports/advisor_on_prem_remediations.rb
```diff
@@ -0,0 +1,27 @@
...
+      sql_count('(
+        SELECT DISTINCT hits.id
+        FROM job_invocations AS jobs
+        INNER JOIN remote_execution_features AS rexf ON jobs.remote_execution_feature_id = rexf.id
+        INNER JOIN template_invocations AS tinv ON jobs.id = tinv.job_invocation_id
+        INNER JOIN insights_hits AS hits ON tinv.host_id = hits.host_id
+        INNER JOIN foreman_tasks_tasks AS tasks ON tinv.run_host_job_task_id = tasks.id
+        WHERE rexf.label = \'rh_cloud_remediate_hosts\'
+        AND tasks.result = \'success\'
+      ) as subquery')
```
adamruzicka, line 24: The parens and `as subquery` is unnecessary

definitions/reports/advisor_on_prem_remediations.rb
```diff
@@ -0,0 +1,27 @@
...
+  class AdvisorOnPremRemediations < ForemanMaintain::Report
+    metadata do
+      description 'Report number of remediations based on advisor rules from advisor on premise'
+      confine do
+        feature(:installer)&.answers&.dig('foreman::plugin::rh_cloud', 'enable_iop_advisor_engine')
+      end
+    end
+
+    def run
+      data_field('advisor_on_prem_remediations') { advisor_on_prem_remediations }
```
adamruzicka, line 11: There is an unwritten convention to end counts with a `_count` suffix.
```suggestion
      data_field('advisor_on_prem_remediation_count') { advisor_on_prem_remediations }
```

adamruzicka (commented):
After talking with @ares we landed on counting the hosts - find all the relevant remediation-applying jobs and sum the count of hosts in all of them.

definitions/reports/advisor_on_prem_remediations.rb
adamlazik1: It doesn't work without it, `report generate` does not return a value and trying to replicate the query on db console throws an error.
```
foreman=# SELECT COUNT(*) FROM (SELECT DISTINCT tinv.host_id, jobs.id
        FROM job_invocations AS jobs
        INNER JOIN remote_execution_features AS rexf ON jobs.remote_execution_feature_id = rexf.id
        INNER JOIN template_invocations AS tinv ON jobs.id = tinv.job_invocation_id
        WHERE rexf.label = 'rh_cloud_remediate_hosts'
        AND tinv.host_id IS NOT NULL);
ERROR:  subquery in FROM must have an alias
LINE 1: SELECT COUNT(*) FROM (SELECT DISTINCT tinv.host_id, jobs.id
                             ^
HINT:  For example, FROM (SELECT ...) [AS] foo.
foreman=# 
```

definitions/reports/advisor_on_prem_remediations.rb
adamlazik1: Applied.

adamlazik1:
> After talking with @ares we landed on counting the hosts - find all the relevant remediation-applying jobs and sum the count of hosts in all of them.

Applied. Should I count the host in if at least one task on the host from the related job invocation succeeded, or should I disregard task result completely?

adamruzicka:
Disregard the result altogether. We want to count the attempts, not necessarily if it succeeded or not.

definitions/reports/advisor_on_prem_remediations.rb
adamruzicka: Eh, right, but we don't really need a distinct in there, do we? job invocation has many template invocations, template invocation belongs to a single host, but a single host can't be in a single job invocation multiple times.

adamruzicka: > I understood this as counting each host in a single job invocation only once

Well, yes. But a host either is or is not in a job, a single host can't be in a single job multiple times. We don't need to do anything to remove duplicates as there are no duplicates.

definitions/reports/advisor_on_prem_remediations.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def advisor_on_prem_remediations_count
+      sql_count('(
+        SELECT tinv.host_id, jobs.id
+        FROM job_invocations AS jobs
+        INNER JOIN remote_execution_features AS rexf ON jobs.remote_execution_feature_id = rexf.id
+        INNER JOIN template_invocations AS tinv ON jobs.id = tinv.job_invocation_id
+        WHERE rexf.label = \'rh_cloud_remediate_hosts\'
+        AND tinv.host_id IS NOT NULL
+      ) AS subquery')
```
adamruzicka, line 22: This should give the same results
```suggestion
      sql_count('
        job_invocations AS jobs
        INNER JOIN remote_execution_features AS rexf ON jobs.remote_execution_feature_id = rexf.id
        INNER JOIN template_invocations AS tinv ON jobs.id = tinv.job_invocation_id
        WHERE rexf.label = \'rh_cloud_remediate_hosts\'
        AND tinv.host_id IS NOT NULL
      ')
```

adamlazik1: I confirm that it gives the same result in my testing case. Applied.

definitions/reports/advisor_on_prem_remediations.rb
```diff
@@ -0,0 +1,24 @@
+module Reports
+  class AdvisorOnPremRemediations < ForemanMaintain::Report
+    metadata do
+      description 'Report number of remediations based on advisor rules from advisor on premise'
+      confine do
+        feature(:installer)&.answers&.dig('foreman::plugin::rh_cloud', 'enable_iop_advisor_engine')
```
adamruzicka, line 6: By having this here, the key-value pair is omitted from the resulting report completely if iop advisor is disabled. Personally I'd prefer to have the key there with nil value if it is not enabled at all. @ShimShtein opinions?

ShimShtein: Let's have two fields: 
boolean `advisor_on_prem_remediations_enabled` and `advisor_on_prem_remediations_count` that will constantly be present in the report. 
I don't think we should make fields optional. It will make the summary queries much more complex.

adamlazik1: Applied.

ShimShtein:
LGTM, @adamruzicka can you make sure the joins do not multiply the counts (as in all joins are 1:1 relations)

adamruzicka:
> can you make sure the joins do not multiply the counts (as in all joins are 1:1 relations)

They're not, but in this case that's [what we want](https://github.com/theforeman/foreman_maintain/pull/994#pullrequestreview-2760171208)

adamlazik1:
> I can't say I'm exactly happy that we rely on calling the methods in the right order so that the instance variable is set. Could the `_enabled` memoize the result and have `_count` call the `_enabled` method instead of relying on its side effects?

Applied.

adamruzicka:
That would do, but the linter is unhappy

adamlazik1:
Oops, I somehow missed that. Linter happy now.

adamruzicka (approved):

PR #995, opened 2025-04-08
Add check on whether foreman runs as a container

adamlazik1:


ehelms:
This is premature as we have no deployments yet for foreman as a container. Do you want to move this to draft or close and revisit when the time comes?

adamlazik1:
I can move to draft for now.

ekohl:
We won't have foreman_maintain in containerized installations so I think this needs to be revisited in a foremanctl implementation.

PR #996, opened 2025-04-11
Condense the generated report

adamruzicka:
I admit this is a bit rough, but I wanted to put together something quick that we can talk about something a bit more specific.

TODO:
- [x] polish this
- [x] make it less prone to crashing if some of the source data is missing
- [ ] ? move the code around
- [ ] ???
- [x] have all the prerequisites

# Prerequisites
- https://github.com/theforeman/foreman_maintain/pull/990
- https://github.com/theforeman/foreman_maintain/pull/994
- https://github.com/theforeman/foreman_maintain/pull/993
- https://github.com/theforeman/foreman_maintain/pull/992

# Rationale

`foreman-maintain report generate` collects _a lot_ of metrics. This high number may be fine in some cases, but it might be problematic in others. The goal here is to be able to derive a condensed report containing only the data points that we absolutely need.

# How it's meant to be used
The goal is to have something that is able to generate the condensed report either on the fly or from an pre-existing full report. It can also take into consideration the modification time of the full report so that the condensed report is reasonably fresh.

# Example data
```
# foreman-maintain report condense --output /tmp/condensed-report.json
-----B<-----SNIP-----B<-----

# json_reformat </tmp/condensed-report.json
{
    "foreman.advisor_on_prem_remediations: 12,
    "foreman.rhel_ai_workload_host_count: 1,
    "foreman.host_rhel_count": 203,
    "foreman.host_redhat_count": 20,
    "foreman.host_other_count": 8,
    "foreman.image_mode_host_count": 4,
    "foreman.use_dualstack": false,
    "foreman.use_ipv6": true
}
```

### DIFF
--- a/lib/foreman_maintain/cli/report_command.rb
+++ b/lib/foreman_maintain/cli/report_command.rb
@@ -37,10 +37,10 @@ module ForemanMaintain
       subcommand 'condense', 'Condense the report' do
         def execute
           data = if fresh_enough?(@input, @max_age)
-            YAML.load_file(@input)
-          else
-            generate_report
-          end
+                   YAML.load_file(@input)
+                 else
+                   generate_report
+                 end
 
           report = condense_report(data)
           report = prefix_keys(report)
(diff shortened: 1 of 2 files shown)


lib/foreman_maintain/cli/report_command.rb
```diff
@@ -3,23 +3,125 @@ module Cli
...
+                 end
+
+          report = condense_report(data)
+          report = prefix_keys(report)
+          save_report(JSON.dump(report), @output)
+        end
+
+        def condense_report(data)
+          result = {}
+          %w[advisor_on_prem_remediations_count rhel_ai_workload_host_count].each do |key|
```
ShimShtein, line 53: Why is it needed? Both fields are using `sql_count`, which means the fields will always be in the report

adamruzicka: Back then when I wrote this it wasn't the case. I can change it, but on the other hand, does it hurt to be extra defensive and not assume anything?

ShimShtein (approved):

evgeni (approved):

PR #999, opened 2025-05-12
Add advisor engine container update procedure

ehelms:


definitions/features/iop.rb
```diff
@@ -42,4 +58,26 @@ def services
...
+      'quay.io/iop'
+    end
+  end
+
+  def container_names
+    if feature(:instance).downstream
+      CONTAINER_NAMES.map { |container_name| "#{container_name}-rhel9" }
+    else
+      CONTAINER_NAMES
+    end
```
evgeni, line 75: @ehelms is there any chance we could not have this weird naming schema? :)

evgeni: The rhel9 in the satellite name, and no rhel9 in the "upstream" one

definitions/features/iop.rb
```diff
@@ -42,4 +58,26 @@ def services
     ]
   end
   # rubocop:enable Metrics/MethodLength
+
+  def container_base
+    if feature(:instance).downstream
+      'registry.redhat.io/satellite'
```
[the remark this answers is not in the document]
evgeni: We could allow providing/overriding the base via command line param or env var or config option?

I think we have something similar for the maintenance repo to allow self updating in the disconnected case

PR #1002, opened 2025-06-30
add iop database backups and restore

evgeni:


definitions/procedures/backup/online/iop_inventory_db.rb
```diff
@@ -0,0 +1,20 @@
+module Procedures::Backup
+  module Online
+    class IopInventoryDB < ForemanMaintain::Procedure
+      metadata do
+        description 'Backup IoP Inventory database'
+        tags :backup
+        label :backup_online_iop_inventory_db
+        for_feature :iop_inventory_database
+        preparation_steps { Checks::Candlepin::DBUp.new }
```
evgeni, line 9: We need to define checks still

evgeni:
restores are missing ;-)

definitions/features/iop_vmaas_database.rb
```diff
@@ -0,0 +1,45 @@
...
+    [
+      system_service('postgresql', 10, :component => 'iop',
+        :db_feature => feature(:iop_vmaas_database)),
+    ]
+  end
+
+  private
+
+  def load_configuration
+    podman_command = "podman exec iop-service-vmaas-reposcan bash -c 'env |grep POSTGRESQL_'"
```
ehelms, line 27: I was trying to think if there was a cleaner way to provide this information.

evgeni: Make the installer store it in a yaml file on the host.

ehelms: We can go this route for now, I will add a re-factoring task to add secrets.

PR #1005, opened 2025-07-14
add a check whether the system is correctly logged in to the registry

evgeni:


lib/foreman_maintain/concerns/downstream.rb
```diff
@@ -57,6 +58,10 @@ def satellite_upgrade_allowed?
           ForemanMaintain.upgrade_in_progress == satellite_maintain_target_version
       end
 
+      def connected?
+        File.exist?(REDHAT_REPO_FILE) && File.new(REDHAT_REPO_FILE).read.include?('https://cdn.redhat.com')
+      end
```
evgeni, line 63: The alternative would be to check the `subscription_connection_enabled` setting.
No idea what's better (and if that setting truly represents "it's disconnected")

PR #1010, opened 2025-07-17
fix in os count calculations

pondrejk:
before:
```
~]# cat sample.yml 
---
hosts_by_os_count|RedHat: 20
hosts_by_os_count|CentOS: 10
hosts_by_family_count|Redhat: 30

~]# foreman-maintain report condense --input sample.yml --output smallreport

~]# cat smallreport 
{...
"foreman.host_rhel_count":20,"foreman.host_redhat_count":10,"foreman.host_other_count":-20,
...}
```

Assuming other means "other than RH os", not "other than RH family"

with this change:
```
~]# cat smallreport 
{...
"foreman.host_rhel_count":20,"foreman.host_redhat_count":30,"foreman.host_other_count":10,
...}
```

### DIFF
--- a/lib/foreman_maintain/cli/report_command.rb
+++ b/lib/foreman_maintain/cli/report_command.rb
@@ -70,7 +70,7 @@ module ForemanMaintain
           rhel_count = data['hosts_by_os_count|RedHat'] || 0
           rh_count = data['hosts_by_family_count|Redhat'] || 0
           result['host_rhel_count'] = rhel_count
-          result['host_redhat_count_without_rhel'] = rh_count - rhel_count
+          result['host_redhat_without_rhel_count'] = rh_count - rhel_count
           result['host_other_count'] = data.select do |k, _|
             k.start_with?('hosts_by_os_count')
           end.values.sum - rh_count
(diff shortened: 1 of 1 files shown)


adamruzicka:
The original intent was to divide the hosts into three groups - proper RHELs, RHEL-like (centoses, fedoras, rockys and so on) and everything else. With the example you provided, I'd expect host_rhel_count 20, host_redhat_count 10 and host_other_count 0

pondrejk:
so with current changes the output for my example looks like
```
"foreman.host_rhel_count":20,"foreman.host_redhat_count_without_rhel":10,"foreman.host_other_count":0
```

lib/foreman_maintain/cli/report_command.rb
```diff
@@ -70,10 +70,10 @@ def aggregate_host_count(data)
           rhel_count = data['hosts_by_os_count|RedHat'] || 0
           rh_count = data['hosts_by_family_count|Redhat'] || 0
           result['host_rhel_count'] = rhel_count
-          result['host_redhat_count'] = rh_count - rhel_count
+          result['host_redhat_count_without_rhel'] = rh_count - rhel_count
```
adamruzicka, line 73: I'd like to keep the `_count` suffix, would this do?
```suggestion
          result['host_redhat_without_rhel_count'] = rh_count - rhel_count
```

adamruzicka (approved):

PR #1016, opened 2025-07-21
add support for timer management

evgeni:


lib/foreman_maintain/feature.rb
```diff
@@ -25,6 +25,15 @@ def services
       []
     end
 
+    # Override method with list of applicable timers for feature.
+    # Services have a number for priority in order to ensure
+    # they are started and stopped in the correct order.
+    # example:
+    # [ system_service('foo', 10), system_service('bar', 20) ]
+    def timers
```
stejskalleos, line 33: When I changed  this to:
```
    def timers
      [system_service('foreman-timer', 10)]
    end
```

and run `bundle exec bin/foreman-maintain maintenance-mode start` it failed with:
```
...
Stop systemd timers: 
- Stopping systemd timers                                             [FAIL]    
No timers found matching your parameters
```

My timer and the service are running and enabled:
```
systemctl status foreman-timer.timer
● foreman-timer.timer - Run foreman-timer.service every 5 minutes
     Loaded: loaded (/etc/systemd/system/foreman-timer.timer; enabled; preset: disabled)
     Active: active (waiting) since Fri 2025-07-25 08:03:13 UTC; 19min ago

systemctl status foreman-timer.service
● foreman-timer.service - Log current date and time for Foreman
     Loaded: loaded (/etc/systemd/system/foreman-timer.service; static)
     Active: activating (start) since Fri 2025-07-25 08:23:12 UTC; 1ms ago
TriggeredBy: ● foreman-timer.timer
```

stejskalleos (changes requested):
1. Is there a documentation PR documenting the timer features? If not, should we document it?
2. I see that the `backup` scenario doesn't have steps for the configured timers. It's because that's taken care of by the installer, so we don't have to back them up. Is that correct?
3. There is a command for managing service, `foreman-maintain service <action>`. Would there be a use case where Foreman users would want to manage timers in the same way they manage services?
4. The `Features::Timer` shares a lot of code with `Features::Services`. Maybe some refactoring of common code to some service objects might reduce the duplications, wdyt?

lib/foreman_maintain/feature.rb
evgeni: it needs to be `foreman-timer.timer`, otherwise it looks at the service, not the timer.

definitions/scenarios/satellite_upgrade.rb
```diff
@@ -74,6 +74,7 @@ def compose
       add_steps(
         Procedures::MaintenanceMode::EnableMaintenanceMode,
         Procedures::Crond::Stop,
+        Procedures::Timer::Stop,
```
ehelms, line 77: In theory this shouldn't be needed for the current upgrade, but if there are no timers then I assume this will be a noop.

ehelms (approved):

definitions/scenarios/satellite_upgrade.rb
evgeni: Correct.

PR #1018, opened 2025-07-21
backup kafka volume

evgeni:
In https://github.com/theforeman/puppet-iop/pull/40 we switched to a podman volume, so we should backup that instead of /var/lib/kafka


adamruzicka (approved):
Untested, but this looks right

PR #1021, opened 2025-07-22
mark steps that were filtered due to confine as skipped

evgeni:
this is especially interesting when running checks, as sometimes a check is not applicable to a system (and thus confined) but we'd still report the result as "OK" which can confuse the user


jameerpathan111:
> draft, as I don't like the way it's currently rendered (visually too close to a warning), but didn't want to leave that unmentioned after a discussion with @jameerpathan111 

@evgeni How about using Orange color for `WARNING`? and keeping yellow for `SKIPPED`?

PR #1031, opened 2025-08-13
extract pulp data *after* the DB has been restored

evgeni:
this has the benefit that if the DB restore fails for some reason, this is detected faster


evgeni:
includes #1030 to avoid conflicts

PR #1032, opened 2025-08-14
execute amcheck before doing backups

evgeni:
this ensures that the database is restorable or if it isn't (when there are index issues) the user knows at the time of the backup

### DIFF
--- a/definitions/checks/candlepin/db_index.rb
+++ b/definitions/checks/candlepin/db_index.rb
@@ -1,0 +1,25 @@
+module Checks
+  module Candlepin
+    class DBIndex < ForemanMaintain::Check
+      metadata do
+        description 'Make sure Candlepin DB indexes are OK'
+        label :candlepin_db_index
+        tags :db_index
+        for_feature :candlepin_database
+        confine do
+          feature(:candlepin_database)&.local?
+        end
+      end
+
+      def run
+        status, output = feature(:candlepin_database).amcheck
+
+        if !status.nil?
+          assert(status == 0, "Candlepin DB indexes have issues:\n#{output}")
+        else
+          skip 'amcheck is not available in this setup'
+        end
+      end
+    end
+  end
+end
--- a/lib/foreman_maintain/concerns/base_database.rb
+++ b/lib/foreman_maintain/concerns/base_database.rb
@@ -140,6 +140,40 @@ module ForemanMaintain
                 ' Make sure system has psql utility installed.'
       end
 
+      def amcheck
+        # executing the check requires superuser privileges
+        return unless local?
+
+        return unless amcheck_installed?
+
+        psqlcmd = "runuser - postgres -c 'psql --set=ON_ERROR_STOP=on #{configuration['database']}'"
+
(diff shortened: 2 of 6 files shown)


evgeni:
```
# foreman-maintain health check --tags db-index
Make sure Candlepin DB indexes are OK:                                [OK]
Make sure Foreman DB indexes are OK:                                  [FAIL]
Foreman DB indexes have issues:
ERROR:  item order invariant violated for index "index_fact_names_on_name_and_type"
DETAIL:  Lower index tid=(1,7) (points to heap tid=(0,5)) higher index tid=(1,8) (points to heap tid=(0,4)) page lsn=0/2ED6F40.
Make sure Pulpcore DB indexes are OK:                                 [OK]

  [foreman-db-index]

--whitelist="foreman-db-index"
```

ehelms:
Should we just create/enable amcheck as a default extension in our installations?

lib/foreman_maintain/concerns/base_database.rb
```diff
@@ -140,6 +140,38 @@ def raise_psql_missing_error
...
+      # rubocop:disable Metrics/MethodLength
+      def amcheck
+        # creating the extension and executing the check requires superuser privileges
+        return unless local?
+
+        psqlcmd = "runuser - postgres -c 'psql --set=ON_ERROR_STOP=on #{configuration['database']}'"
+
+        amcheck_create = 'CREATE EXTENSION IF NOT EXISTS amcheck;'
+
+        execute!(psqlcmd, :stdin => amcheck_create)
```
evgeni, line 152: @ehelms would you prefer if this would be a `return unless amcheck_installed?` and I add "install amcheck" to foreman/pulpcore/candlepin puppet?

definitions/checks/candlepin/db_index.rb
```diff
@@ -0,0 +1,21 @@
...
+        for_feature :candlepin_database
+        confine do
+          feature(:candlepin_database)&.local?
+        end
+      end
+
+      def run
+        status, output = feature(:candlepin_database).amcheck
+
+        assert(status == 0 || status.nil?, "Candlepin DB indexes have issues:\n#{output}")
```
evgeni, line 17: `status` is `nil` when `db.amcheck` returned early, without actually running amcheck. I considered returning `[0, '']`, but this felt wrong, as the status is really not `0` (`OK`).
I could also write this as
```suggestion
        assert(status == 0 , "Candlepin DB indexes have issues:\n#{output}") unless status.nil?
```

what do you prefer?

evgeni: We could also do smth like
```
        unless status.nil?
          assert(status == 0, "Candlepin DB indexes have issues:\n#{output}")
        else
          skip "amcheck not possible"
        end
```

evgeni:
- [ ] https://github.com/theforeman/puppet-pulpcore/pull/387
- [ ] https://github.com/theforeman/puppet-foreman/pull/1232
- [ ] https://github.com/theforeman/puppet-candlepin/pull/271

PR #1037, opened 2025-08-28
Run duplicate cleanup as part of pre-upgrade tasks

sjha4:
We want to run the task https://github.com/Katello/katello/blob/master/lib/katello/tasks/repository.rake#L82 to clean up duplicate erratum packages before the migration runs on the table as part of this migration:
https://github.com/Katello/katello/blob/master/db/migrate/20250613210050_use_big_int_for_erratum_packages_id.rb

The migration on the ID field re-runs indexes on the table apparently and that causes failures in case of duplicates existing on the table.


definitions/procedures/repositories/cleanup_duplicate_erratum_packages.rb
```diff
@@ -0,0 +1,16 @@
...
+  class CleanupDuplicateErratumPackages < ForemanMaintain::Procedure
+    metadata do
+      description 'Cleanup duplicate erratum packages'
+      confine do
+        feature(:katello)
+      end
+    end
+
+    def run
+      puts "This may take a while to complete on large databases..."
```
ehelms, line 11: Do we have estimates or ranges of how long this can take?

definitions/scenarios/satellite_upgrade.rb
```diff
@@ -72,6 +72,7 @@ class PreMigrations < Abstract
 
     def compose
       add_steps(
+        Procedures::Repositories::CleanupDuplicateErratumPackages,
```
evgeni, line 75: This is in Katello master (not even in 4.18 rc2), so I think it can't run as a pre-migration before the upgrade?

sjha4: We have the PR in the cherrypicks for 4.18 GA.. https://github.com/Katello/katello/pull/11494

evgeni: Ok, but then it still needs to be post-migrations to have 4.18 :)

sjha4: Oh..Are the pre-migrations run before the yum update? We want this step to run before the db:migrate on upgrades where the dup data causes errors..

evgeni: Yeah. You'd need to edit Migrations to do what you want, but it sounds dangerous to me, as at that moment the services will already be down and you'd have to restart them with the new code but old db.

sjha4:
After discussions offline, we decided against having this as an upgrade step. The rake task itself can live independently for support purposes while not being an upgrade blocker. Closing this for reasons above.

PR #1046, opened 2025-09-22
sort report keys alphabetically

pondrejk:
for readability and ease of comparison


adamruzicka (approved):
The output was meant to be primarily machine-readable, but that doesn't mean it has to be difficult for people to read. Let's get this in

adamruzicka:
Thank you @pondrejk !

PR #1052, opened 2025-10-02
confine installer feature until installer run once

evgeni:
otherwise last_scenario.yml doesn't exist and things break


evgeni:
alternative aproach in https://github.com/theforeman/foreman_maintain/pull/1053, but I don't really like either :crying_cat_face:

ehelms (approved):

PR #1083, opened 2026-04-10
self-upgrade next target_version

stejskalleos:
For the self-upgrade scenario:
* set the target version to X.Y+1
* disable the repository after the upgrade

### DIFF
--- a/definitions/scenarios/self_upgrade.rb
+++ b/definitions/scenarios/self_upgrade.rb
@@ -4,7 +4,10 @@ module ForemanMaintain::Scenarios
     include ForemanMaintain::Concerns::Versions
 
     def target_version
-      feature(:instance).target_version
+      @target_version ||= begin
+        v = Gem::Version.new(feature(:instance).target_version)
+        "#{v.segments[0]}.#{v.segments[1] + 1}"
+      end
     end
 
     def current_version
@@ -39,19 +42,30 @@ module ForemanMaintain::Scenarios
       true
     end
 
+    def maintain_version
+      @maintain_version ||= feature(:instance).target_version
+    end
+
+    def current_major
+      Gem::Version.new(current_version).segments[0..1].join('.')
+    end
+
+    def upgrade_repo_version
+      Gem::Version.new(current_version).bump.segments[0..1].join('.')
+    end
+
     def req_repos_to_update_pkgs
+      version = upgrade_repo_version
       if use_rhsm?
-        main_rh_repos + [maintenance_repo_id(target_version)]
+        main_rh_repos + [maintenance_repo_id(version)]
       else
-        [maintenance_repo_id(target_version)]
+        [maintenance_repo_id(version)]
       end
     end
 
     def self_upgrade_allowed?
-      target = Gem::Version.new(target_version).version
-
-      Gem::Version.new(current_version).segments[0..1].join('.') == target ||
-        Gem::Version.new(current_version).bump.segments[0..1].join('.') == target
+      current_major == maintain_version ||
+        Gem::Version.new(current_version).bump.segments[0..1].join('.') == maintain_version
     end
   end
 
@@ -67,10 +81,10 @@ module ForemanMaintain::Scenarios
       unless self_upgrade_allowed?
         raise(
           ForemanMaintain::Error::Warn,
-          "foreman-maintain is too many versions ahead. The target " \
-          "version is #{target_version} while the currently installed " \
-          "version is #{current_version}. Please rollback " \
-          "foreman-maintain to the proper version."
+          "foreman-maintain does not support the installed version of Satellite." \
+          "The currently installed version in #{current_version}," \
+          "while foreman-maintain only supports #{maintain_version}." \
+          "Please install the right version of foreman-maintain."
         )
       end
 
--- a/test/definitions/scenarios/self_upgrade_test.rb
+++ b/test/definitions/scenarios/self_upgrade_test.rb
@@ -16,40 +16,29 @@ describe ForemanMaintain::Scenarios::SelfUpgradeBase do
 
   it 'computes the target version correctly for downstream' do
     assume_satellite_present
+    ForemanMaintain.package_manager.
+      stubs(:find_installed_package).
+      with('satellite', '%{VERSION}').returns('6.16.0')
 
-    assert_equal '6.16', scenario.target_version
+    assert_equal '6.17', scenario.upgrade_repo_version
   end
 
   it 'returns target version correctly for upstream' do
     assume_feature_present(:foreman_install)
 
-    assert_equal '3.11', scenario.target_version
-  end
-
-  it 'allow upgrade when versions are equal' do
-    assume_satellite_present
-    scenario.expects(:current_version).returns('6.16.0')
-
-    assert scenario.self_upgrade_allowed?
-  end
-
-  it 'allow upgrade when versions are equal for z-stream' do
-    assume_satellite_present
-    scenario.expects(:current_version).returns('6.16.4')
-
-    assert scenario.self_upgrade_allowed?
+    assert_equal '3.12', scenario.target_version
   end
 
   it 'allow upgrade when versions are off by 1' do
     assume_satellite_present
-    scenario.expects(:current_version).twice.returns('6.15.8')
+    scenario.stubs(:current_version).returns('6.15.8')
 
     assert scenario.self_upgrade_allowed?
   end
 
   it 'does not allow upgrade when versions are off by 2 or more' do
     assume_satellite_present
-    scenario.expects(:current_version).twice.returns('6.14.4')
+    scenario.stubs(:current_version).returns('6.14.4')
 
     refute scenario.self_upgrade_allowed?
   end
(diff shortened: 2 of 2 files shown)


definitions/scenarios/self_upgrade.rb
```diff
@@ -74,6 +74,10 @@ def downstream_self_upgrade(pkgs_to_update)
         )
       end
 
+      add_step(Procedures::Repositories::Enable.new(
+        repos: [maintenance_repo_id(target_version)],
+        use_rhsm: use_rhsm?
+      ))
```
evgeni: `enabled_repos: req_repos_to_update_pkgs` two lines below should already be doing this, minus the fact that using `Procedures::Repositories::Enable` will leave the repository enabled after the scenario is finished.

stejskalleos: Then I need to investigate why it's not enabling the repo without this step ...

Q: Should we disable the repo after the self-upgrade? Is that a problem when it stays open?

evgeni: Yes, "self-upgrade" should alter the system as little as possible.

evgeni: If my memory serves me right, the issue is not that no repo enablement is happening, but that we're enabling the repo of the _current_ release, not the _next_ one.

stejskalleos: Yeah, but when I was trying to reproduce the issue on 6.16, next_release was correctly set to 6.17, so the issue wasn't in it.
I need to dig deeper.

stejskalleos: So the `target_version` seems to be hardcoded in 1.7.13:
```ruby
class Features::Satellite < ForemanMaintain::Feature
  def target_version
    '6.16'
  end
end
```

And in the master branch, it just points to the [current version](https://github.com/theforeman/foreman_maintain/blob/master/lib/foreman_maintain/concerns/downstream.rb#L60-L62).

So I guess, for the self-upgrade task, we should decide on the next version and enable the repos for it.

stejskalleos:
@evgeni, ready for review, fixed the bump version & updated tests.

definitions/scenarios/self_upgrade.rb
```diff
@@ -76,6 +80,10 @@ def downstream_self_upgrade(pkgs_to_update)
 
       add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true,
         enabled_repos: req_repos_to_update_pkgs))
+      add_step(Procedures::Repositories::Disable.new(
+        repos: [maintenance_repo_id(target_version)],
+        use_rhsm: use_rhsm?
+      ))
```
evgeni: Why is this needed? `Procedures::Packages::Update` will already do all the required actions for us to only enable that repo during the upgrade.

https://github.com/theforeman/foreman_maintain/blob/75ebdda30981fbcf9dea4e59dcbdf368b6fc0f92/definitions/procedures/packages/update.rb#L20
https://github.com/theforeman/foreman_maintain/blob/75ebdda30981fbcf9dea4e59dcbdf368b6fc0f92/lib/foreman_maintain/package_manager/dnf.rb#L144-L146

definitions/procedures/repositories/disable.rb
evgeni, line 1: Why is this needed for the self-upgrade?

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -17,32 +17,32 @@
...
     assume_feature_present(:foreman_install)
 
-    assert_equal '3.11', scenario.target_version
+    assert_equal '3.12', scenario.target_version
   end
 
   it 'allow upgrade when versions are equal' do
     assume_satellite_present
-    scenario.expects(:current_version).returns('6.16.0')
+    scenario.expects(:current_version).returns('6.17.0')
```
evgeni: Shouldn't *current* version remain 6.16?!

test/definitions/scenarios/self_upgrade_test.rb
evgeni, line 1: I think the fact that you had to update the tests, but none of the tests actually asserts the list of repositories to enable indicates that there is something off with out logic in this scenario (given the bug we're trying to fix, probably already before this PR).

The idea is, assuming the following releases exist: 3.16, 3.17, 3.18 and the user currently on 3.16, you get:
- with f-m from 3.16: "self-upgrade" updates to f-m from 3.17 by temp-enabling 3.17 repos
- with f-m from 3.17: "self-upgrade" ensures you have the latest f-m from 3.17 by temp-enabling 3.17 repos
- with f-m from any other release (e.g. 3.18): error "foreman-maintain is too many versions ahead."

stejskalleos:
I'm not able to make the tests work, so pushing without green runs.
@evgeni, is this the approach we should take to fix it?
I don't handle the 6.y -> 7.0 update, do I need to care about it?

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -94,7 +94,7 @@
       package_manager.
       expects(:find_installed_package).
       with('satellite', "%{VERSION}").
-      returns('6.15.3')
+      returns('6.16.3')
```
evgeni, line 97: the test says "off by 1", but this now tests "equals" as the baseline is 6.16:
https://github.com/theforeman/foreman_maintain/blob/75ebdda30981fbcf9dea4e59dcbdf368b6fc0f92/test/definitions/test_helper.rb#L217-L218

definitions/scenarios/self_upgrade.rb
```diff
@@ -4,7 +4,11 @@ class SelfUpgradeBase < ForemanMaintain::Scenario
     include ForemanMaintain::Concerns::Versions
 
     def target_version
-      feature(:instance).target_version
+      @target_version ||= begin
+        current_target = feature(:instance).target_version
+        v = Gem::Version.new(current_target)
+        "#{v.segments[0]}.#{v.segments[1] + 1}"
+      end
     end
```
evgeni, line 12: So I have been starring at this patch for way too long now, but I think I had a realization that should be shared.

In the (downstream) metadata, we define `current_satellite_version` and `previous_satellite_version`, which in itself should indicate "I am the maintain tool for current version of *Satellite*, and I know how to upgrade *Satellite* from previous version".

When it then comes to the Satellite "feature", we expose `current_satellite_version` as `.target_version` as the primary user is the *Satellite* upgrade scenario, not the *Self* (=Maintain) upgrade scenario (we don't expose previous as a value, it's just used in one method raw). We then (confusingly) add a `.current_version` value, which resolves to the actually installed (package) version.

However, the target version in the Self Upgrade scenario is totally unrelated to that. In the self upgrade scenario we need to upgrade to the Maintain tooling for the "next" version of Satellite, so `feature.current_version+1`, not `feature.target_version+1` (this will avoid the case where we already upgraded the maintain tooling and accidentally try to upgrade it again).

stejskalleos:
@evgeni update, added tests & disabling the repos.

definitions/scenarios/self_upgrade.rb
```diff
@@ -76,6 +79,7 @@ def downstream_self_upgrade(pkgs_to_update)
 
       add_step(Procedures::Packages::Update.new(packages: pkgs_to_update, assumeyes: true,
         enabled_repos: req_repos_to_update_pkgs))
+      add_step(Procedures::Repositories::Disable.new(repos: req_repos_to_update_pkgs))
```
stejskalleos, line 82: Should we add the repo disable to the upstream step as well? 🤔

evgeni: you don't need to disable the repo at all as it only gets enabled for *that one* dnf call via `--enablerepo`

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -63,60 +49,112 @@
...
+        'rhel-9-for-x86_64-baseos-rpms',
+        'rhel-9-for-x86_64-appstream-rpms',
+        'satellite-maintenance-6.17-for-rhel-9-x86_64-rpms',
+      ]
+    end
+
+    it 'with f-m from 6.16 updates by temp-enabling 6.17 repos' do
+      ForemanMaintain.package_manager.
+        expects(:find_installed_package).
+        with('satellite', "%{VERSION}").returns('6.16.3')
```
evgeni, line 68: The code is using `feature(:instance).target_version` which does not come from `package_manager` but from the foreman/satellite-maintain configuration (see below). So this (while probably called by some code) has no effect on the result of the test.

https://github.com/theforeman/foreman_maintain/blob/344d337491072e6bbed1b7987284d8f41889cbf9/definitions/features/satellite.rb#L13-L15
https://github.com/theforeman/foreman_maintain/blob/344d337491072e6bbed1b7987284d8f41889cbf9/test/definitions/test_helper.rb#L215-L223

stejskalleos: Without it the test fails with:

```
Error:
ForemanMaintain::Scenarios::SelfUpgrade::downstream#test_0001_with f-m from 6.16 updates by temp-enabling 6.17 repos:
NoMethodError: undefined method `succ' for nil:NilClass
    /home/lstejska/.rbenv/versions/3.0.7/lib/ruby/3.0.0/rubygems/version.rb:236:in `bump'
    definitions/scenarios/self_upgrade.rb:57:in `self_upgrade_allowed?'
    definitions/scenarios/self_upgrade.rb:70:in `downstream_self_upgrade'
    definitions/scenarios/self_upgrade.rb:115:in `compose'
    lib/foreman_maintain/scenario.rb:93:in `initialize'
    test/definitions/scenarios/self_upgrade_test.rb:67:in `new'
    test/definitions/scenarios/self_upgrade_test.rb:67:in `block (3 levels) in <top (required)>'
```

evgeni: Correct, that's why I said "probably called by some code".

The test is called "with f-m from 6.16 updates by temp-enabling 6.17 repos" but you're not modifying the f-m version metadata, but the satellite package. Which is used by the self-upgrade code, but does not indicate anything about f-m.

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -63,60 +49,100 @@
...
+
+      update_step = scenario.steps.find { |s| s.is_a?(Procedures::Packages::Update) }
+      assert update_step, 'Expected a Packages::Update step'
+      assert_equal expected_repos, update_step.instance_variable_get(:@enabled_repos)
+    end
+
+    it 'with f-m from 6.15 raises error as too many versions ahead' do
+      ForemanMaintain.package_manager.
+        expects(:find_installed_package).
+        with('satellite', "%{VERSION}").returns('6.15.1')
```
evgeni: This test uses f-m from 6.16 on a 6.15 satellite:
https://github.com/theforeman/foreman_maintain/blob/344d337491072e6bbed1b7987284d8f41889cbf9/test/definitions/test_helper.rb#L215-L223
(which imho should not fail but upgrade to the latest 6.16-originated f-m to perform the 6.15→6.16 upgrade)

evgeni: As we don't have a way to alter what `mock_satellite_maintain_config` returns (today), I think we could write the tests like this (`mock_satellite_maintain_config` is implied, as it's in a `before` block above):

```ruby
def mock_satellite_package_version(satver)
  ForemanMaintain.package_manager.
        expects(:find_installed_package).
        with('satellite', "%{VERSION}").returns(satver)
end

it 'with maintain from 6.16 on satellite 6.14.1 raises error' do
  mock_satellite_package_version('6.14.1')
  assert error
end

it 'with maintain from 6.16 on satellite 6.15.1 upgrades maintain to latest from 6.16' do
  mock_satellite_package_version('6.15.1')
  assert "dnf called with --enablerepo 6.16"
end

it 'with maintain from 6.16 on satellite 6.16.1 upgrades maintain to latest from 6.17' do
  mock_satellite_package_version('6.16.1')
  assert "dnf called with --enablerepo 6.17"
end

it 'with maintain from 6.16 on satellite 6.17.1 raises error' do
  mock_satellite_package_version('6.17.1')
  assert error (the error message might be wrong, but I don't think we care)
end
```

Edit: updated the title of the 6.17 test, it was wrong

stejskalleos:
@evgeni self_test updated

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -63,60 +49,109 @@
...
+    end
+
+    it 'with maintain from 6.16 on satellite 6.15.1 raises error' do
+      mock_satellite_package_version('6.15.1')
+      assume_feature_present(:satellite)
+
+      assert_raises(ForemanMaintain::Error::Warn) do
+        ForemanMaintain::Scenarios::SelfUpgrade.new
+      end
+    end
```
evgeni: it should use 6.16 repos, not produce an error.

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -63,60 +49,109 @@
...
+    it 'with maintain from 6.16 on satellite 6.17.1 upgrades maintain to latest from 6.17' do
+      mock_satellite_package_version('6.17.1')
+      assume_feature_present(:satellite)
+      scenario = ForemanMaintain::Scenarios::SelfUpgrade.new
+      scenario.compose
+
+      update_step = scenario.steps.find { |s| s.is_a?(Procedures::Packages::Update) }
+      assert update_step, 'Expected a Packages::Update step'
+      assert_equal expected_repos, update_step.instance_variable_get(:@enabled_repos)
+    end
```
evgeni: the title was wrong in my comment, but the code said "assert error"

why would a 6.16 maintain on a 6.17 box produce any reasonable result?

definitions/scenarios/self_upgrade.rb
evgeni: I think now that we updated the definition of `self_upgrade_allowed?`, we also should update this message a bit.
How about:
```
foreman-maintain does not support the installed version of Satellite.
The currently installed version in #{current_version},
while foreman-maintain only supports #{maintain_version}.
Please install the right version of foreman-maintain.
```

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -17,39 +17,25 @@
   it 'computes the target version correctly for downstream' do
```
evgeni, line 17: `target_version` is not (directly) used in downstream repo calculation anymore, so we probably should test `upgrade_repo_version` instead?

test/definitions/scenarios/self_upgrade_test.rb
```diff
@@ -17,39 +17,25 @@
...
-    scenario.expects(:current_version).returns('6.16.4')
-
-    assert scenario.self_upgrade_allowed?
+    assert_equal '3.12', scenario.target_version
   end
 
   it 'allow upgrade when versions are off by 1' do
     assume_satellite_present
-    scenario.expects(:current_version).twice.returns('6.15.8')
+    scenario.stubs(:current_version).returns('6.16.8')
```
evgeni:
```suggestion
    scenario.stubs(:current_version).returns('6.15.8')
```

otherwise it's not "off by one"?

definitions/scenarios/self_upgrade.rb
```diff
@@ -39,19 +42,37 @@ def use_rhsm?
...
+    def upgrade_repo_version
+      current_bump = Gem::Version.new(current_version).bump.segments[0..1].join('.')
+      if current_major == maintain_version
+        target_version
+      elsif current_bump == maintain_version
+        maintain_version
+      else
+        target_version
+      end
+    end
```
evgeni, line 55: is this the same as?

```suggestion
    def upgrade_repo_version
      Gem::Version.new(current_version).bump.segments[0..1].join('.')
    end
```

the test do pass and I'd argue that:
- if current major and maintain target match, so will the bumps of them
- if current bump and maintain target match, you can use the current bump directly
- the else branch is never hit

evgeni (commented):
overall 👍

test/definitions/scenarios/self_upgrade_test.rb
stejskalleos: But the next version is 6.17, so the test will fail with 6.15

evgeni: it passes for me locally

stejskalleos:
@evgeni 
I applied the changes and updated the tests. If you ack it, I'd like to rebase the PR into a single commit before merge.

evgeni:
@stejskalleos you decided not to update the error text?

otherwise the diff looks fine to me. will do a few more tests, but shouldn't prevent you from squashing.

stejskalleos:
> @stejskalleos you decided not to update the error text?

I did, but then probably accidentally reverted it when editing other code. Fixed again.

evgeni (approved):
tested on 6.16, 6.17 and 6.18 boxes.

6.16 → raises an error, as expected.
6.17 (with f-m built for 6.18) → tries to upgrade via 6.18 repos
6.18 (with f-m built for 6.18) → tries to upgrade via 6.19 repos

Also tested `--maintenance-repo-label` -- that still works and is unaffected by the changes here.

PR #1085, opened 2026-04-16
Speed up csv compression on SMP machines

sbernhard:
The output of pbzip2 is fully compatible with bzip2. pbzip2 is optimized on SMP machines as it splits the input in multiple chunks and uses threads to run on multiple cpu cores to speed up the compression.

This change requires to add pbzip2 as a requirements in the RPM spec for foreman-maintain.

### DIFF
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -177,7 +177,7 @@ class Features::ForemanTasks < ForemanMaintain::Feature
       f.write(csv_output)
       f.close
     end
-    execute("bzip2 #{filepath} -c -9 > #{filepath}.bz2")
+    execute("pbzip2 #{filepath} -c -9 > #{filepath}.bz2")
     FileUtils.rm_rf(filepath)
   end
 
--- a/definitions/features/foreman_tasks.rb
+++ b/definitions/features/foreman_tasks.rb
@@ -1,3 +1,5 @@
+require "etc"
+
 class Features::ForemanTasks < ForemanMaintain::Feature
   MIN_AGE = 30
   TIMEOUT_FOR_TASKS_STATUS = 300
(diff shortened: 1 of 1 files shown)


sbernhard:
See https://community.theforeman.org/t/backup-dynflow-actions-taking-a-long-time/42206

sbernhard:
Thoughs on this @evgeni ?

sbernhard:
@stejskalleos can you maybe have a look at this?

definitions/features/foreman_tasks.rb
```diff
@@ -177,7 +177,7 @@ def export_csv(sql, file_name, state)
       f.write(csv_output)
       f.close
     end
-    execute("bzip2 #{filepath} -c -9 > #{filepath}.bz2")
+    execute("pbzip2 #{filepath} -c -9 > #{filepath}.bz2")
```
stejskalleos: What if the pbzip2 is not available? I could not find information about where it is installed by default, so maybe some checks in place would be a good idea:

```ruby
compressor = system("which pbzip2 > /dev/null 2>&1") ? "pbzip2" : "bzip2"
execute("#{compressor} #{filepath} -c -9 > #{filepath}.bz2")
```

Second question:
By default, `pbzip2` will use all available cores. This can make the Foreman instance unresponsive and can cause some access issues. Should we lower the process's priority or (somehow) limit the number of cores used for execution? So we don't exhaust the server's resources?

maximiliankolb:
ping @sbernhard

definitions/features/foreman_tasks.rb
sbernhard: thank you very much for your suggestions. very good. I adjust the PR.

sbernhard:
Some measurements on a system with 16 cores. bigfile has size of 1.9GB.

bzip2:
```
$ time nice -n 10 bzip2 -9 bigfile
real	2m12,857s
user	2m11,971s
sys	0m0,763s
```

pbzip2:
```
$ time nice -n 10 pbzip2 -p8 -9 bigfile
real	0m43,990s
user	5m41,660s
sys	0m8,612s
```

stejskalleos (approved):

stejskalleos:
**LGTM**

🍏 check for pbzip2 is present
🍏 check for number of threads is there
:green_apple: `nice` for prioritization is used

stejskalleos:
Thanks @sbernhard !

